PR #60, opened 2022-09-14
removing fixme and adding a few routines

maxhant:
- In `qibolab/multiplex`, the IF of the RO is no longer hard coded. Therefore, the `#FIXME` followed by changing the LO is no longer required. However, the runcard needs to be updated accordingly. I made a small function in the `utils.py` to take care of that. 
- Adding to very useful routines which are `qubit_spectroscopy_track`. It avoids using a large frequency span and it corrects for the change in the resonator's spectroscopy. 
- Adding `rabi_attenuation` which takes the rabi for different attenuations. This is important because to avoid noise one needs to maximize the attenuation.

### DIFF
--- a/runcards/actions_qq_5q.yml
+++ b/runcards/actions_qq_5q.yml
@@ -51,6 +51,7 @@ actions:
   #   precision_start: -500_000
   #   precision_end: 500_000
   #   precision_step: 100_000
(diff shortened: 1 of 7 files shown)


andrea-pasquale:

PR #71, opened 2022-09-19
Maxime/ro optimization

maxhant:
* Adding rotation of i/q bins to optimize discrimination
* Adding read-out fidelity calculations
* Adding optimization of LO frequency for read-out
* Segregating calibrations utils into files because it was becoming cumbersome


andrea-pasquale:

PR #72, opened 2022-09-21
Monitor QPUs' status

MatteoRobbiati:
In this pull request me and @Edoardo-Pedicillo have implemented a monitoring tool to check the current status of the chips and the job's queue. We are keeping track of the following:

- QPUs status;
- Jobs in queue for each QPU;
- platform runcards (from `qibolab`'s main branch);

It should look something like this:
monitor

If everyone is happy with this layout we will make this run periodically on the cluster.


MatteoRobbiati (commented):
Thank you @Edoardo-Pedicillo for ending the `*.sh` file. It sounds great to me!

scarrazza:
Just to let you know that I have installed the accounting module for slurm, so if you want to show statistics about runs you can use sacct and sacctmgr.

Edoardo-Pedicillo:
@MatteoRobbiati, could you run the bash script and let me know if everything is ok. We have to remember to remove the lines with the #TODO tag.

MatteoRobbiati:
> @MatteoRobbiati, could you run the bash script and let me know if everything is ok? We have to remember to remove the lines with the #TODO tag. As of Saturday, the names of the platforms are `tii1q` and `tii5q`, instead of `qpu1q` and `qpu5q`, for this reason, I have changed the values of the variable `platform_name` in **qiboqueue.py**.

Thank you @Edoardo-Pedicillo for this. I have tested it and it properly works.
I have also modified some `css` feature in order to display better with small screens. The work is not ended yet. 
I have opened #160 related to the `echo` commands that should be removed.

MatteoRobbiati:
With the last commit I add this line to the head part of the `index.html`:

```html
  <meta http-equiv="refresh" content="5" />
```

As discussed with @Edoardo-Pedicillo, this solution seems to work well (I tested it). The `content="5"` means a refresh rate of 5 seconds. It should be changed according to the frequence of run of `qiboqueue.py`

scarrazza:
Given this PR is related to a specific lab configuration, let me propose to close it now, and as soon as you have time, port this webapplication to the qrc cluster repo.

AleCandido:
Actually, we could even put directly on the website, and just add some data files (e.g. JSON) in `qrccluster`. This should make it easier for other people to update them, and the web page can parse them on the client (downloading from an endpoint).

PR #74, opened 2022-09-22
Merging Flipping to #70 and adding phase fitting to flipping and ramsey

maxhant:
* Adding phase fitting to Ramsey and flipping because it is typically a less noisy measure

This points both to #71 and #8.


andrea-pasquale:

PR #82, opened 2022-09-26
Update platform runcard after every calibration

stavros11:
Small changes to the platform runcard loading/saving to allow passing parameters from every routine to the next one. This should be useful for autocalibration.

When a qq job is launched, the `ActionBuilder` creates a copy (named new_platform.yml) of the current qibolab runcard to the qq folder and defines the platorm using this copy as a custom runcard. The new_platform.yml is updated after every routine using the results of fitting and reloaded in the next routine using `platform.reload_settings()`.This should allow to automatically pass the latest parameters to each calibration.

If one executes an action runcard containing all the required calibrations the final version of new_platform.yml will contain all the calibrated parameters.

### DIFF
--- a/src/qcvv/cli/builders.py
+++ b/src/qcvv/cli/builders.py
@@ -33,13 +33,13 @@ class ActionBuilder:
         backend_name = self.runcard.get("backend", "qibolab")
         platform_name = self.runcard.get("platform", "dummy")
         self.backend, self.platform = self._allocate_backend(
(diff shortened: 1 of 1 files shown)


andrea-pasquale:
@stavros11 can we merge this PR?
I've fixed the issue with the resonator spectroscopy in #89.
If there are no problems other than the resonator spectroscopy I will merge #89 here, and then this one to #8.

PR #88, opened 2022-10-03
Making QCVV platform independant

maxhant:
Currently, there are very few routines which are platform-dependent. For instance, `platform.xport[qubit].attenuation` or `platform.xport[qubit].lo_frequency` are only for QBlox. On top of that, there is some redundancy in the platform runcard. The function `check_frequency` solves the second problem. 

I am working on this with related PR in Qibolab. I just needed to make that PR before the #83 because it will require rewriting all routines, so let's do it in a platform independent way.

### DIFF
--- a/src/qcvv/calibrations/resonator_spectroscopy.py
+++ b/src/qcvv/calibrations/resonator_spectroscopy.py
@@ -23,24 +23,23 @@ def resonator_spectroscopy(
     software_averages,
     points=10,
 ):
-    platform.reload_settings()
-    check_frequency(platform, write=False)
 
     platform.reload_settings()
     sequence = PulseSequence()
     ro_pulse = platform.create_qubit_readout_pulse(qubit, start=0)
     sequence.add(ro_pulse)
 
-    resonator_frequency = platform.characterization["single_qubit"][qubit][
-        "resonator_freq"
-    ]
-
     frequency_range = (
         variable_resolution_scanrange(
             lowres_width, lowres_step, highres_width, highres_step
         )
-        + resonator_frequency
+        + ro_pulse.frequency
     )
+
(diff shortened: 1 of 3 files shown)


src/qcvv/calibrations/resonator_spectroscopy.py
```diff
@@ -23,24 +23,23 @@ def resonator_spectroscopy(
...
-
     frequency_range = (
         variable_resolution_scanrange(
             lowres_width, lowres_step, highres_width, highres_step
         )
-        + resonator_frequency
+        + ro_pulse.frequency
     )
+
+    if platform.mixer_based:
```
aorgazf, line 39: I do not advise doing this. Basically it is taking decisions that the platform should take based on its knowledge of the instruments connected to it. The user at qibocal should not have to know/care about these details

src/qcvv/calibrations/utils/runcard.py
```diff
@@ -117,3 +118,57 @@ def check_frequency(platform, write=False):
         log.info(f"WARNING: Writting YAML")
         with open(path, "w") as f:
             yaml.dump(settings, f, sort_keys=False, indent=4, default_flow_style=None)
+
+
+def _frequency_allocation(freq, bandwidth=600e6, weights=[1, 1, 1, 1]):
```
aorgazf, line 123: I see a similar problem here. Deciding on the best LO is something that the platform, if not the instrument itself, should do. This logic should be kept encapsulated there, and not brought to qcvv/qibocal.

aorgazf (changes requested):
I do not support this refactor. I believe it extracts logic form the platform that should remain in there. I would support incorporating the function to calculate the LO frequency there, in qibolab.

PR #106, opened 2022-10-12
adding attenuation

maxhant:
adding attenuation important for characterization and removing fixme which is unnecessary

### DIFF
--- a/runcards/actions_qq_5q.yml
+++ b/runcards/actions_qq_5q.yml
@@ -57,6 +57,7 @@ actions:
   #   precision_start: -500_000
   #   precision_end: 500_000
   #   precision_step: 100_000
+  #   attenuation: 46
   #   software_averages: 1
   #   points: 10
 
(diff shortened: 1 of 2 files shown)


andrea-pasquale (approved):
Thanks @maxhant. 
Looks good to me.

DavidSarlle:
@maxhant from the point the view of the code I have no objection regarding the changes. It works as expected. 

From the point of view of the platform design. In my opinion, only the parameters wich the experimentalist wants to play with in the routine in order to  find the best value after run and fit the results should be passed to the script. 
If the parameter is something pre decided in advance and somehow "static", it should be modified directly in the platform runcard. 

Let's see the opinion of the rest.

DavidSarlle:
Also @maxhant , just a comment/curiosity. If we increase the attenuation too much, the results are goin to be quite weak. If you use HW demodulation, the arithmetic precision of the qblox fpga is going to be affected, loosing precission. Is the 46db att an example or is it the correct value for the qubit 3?

maxhant:
> Let's see the opinion of the rest.

No need to get everyone's opinions on tiny details. I will close the PR.

PR #107, opened 2022-10-12
Maxime/qubit spec track

maxhant:
Adding the qubit frequency spectroscopy while varying the flux. It requires re-adjustments of the resonator's frequency, and it tracks the peak to adjust the centre frequency. 
Note that it needs extra parameters that can be found in `qibolab/maxime/runcard_binning`

PR #109, opened 2022-10-13
Qubit attenuation spectroscopies 2d

maxhant:
2D scan with attenuation in spectroscopies to get thinner qubit peaks

### DIFF
--- a/runcards/actions_qq_1q.yml
+++ b/runcards/actions_qq_1q.yml
@@ -34,6 +34,14 @@ actions:
     software_averages: 1
     points: 10
 
+  # qubit_attenuation:
+  #   freq_start: -100.e+6
+  #   freq_end: +100.e+6
+  #   freq_step: 2.e+6
+  #   attenuation_list: range(38,60,2)
+  #   software_averages: 1
+  #   points: 1
+
   # rabi_pulse_amplitude:
   #   pulse_amplitude_start: 0 # 0<=a<=1
   #   pulse_amplitude_end: 1
--- a/runcards/actions_qq_5q.yml
+++ b/runcards/actions_qq_5q.yml
@@ -71,6 +71,15 @@ actions:
   #   software_averages: 2
   #   points: 10
 
+  # qubit_spectroscopy_flux_track:
+  #   freq_width: 50_000_000
+  #   freq_step: 1_000_000
+  #   current_offset: 10.e-3
+  #   current_step: 4.e-3 # 0.0001
+  #   attenuation: 46
+  #   software_averages: 1
+  #   points: 10
+
   # rabi_pulse_length:
   #   pulse_duration_start: 4 # minimum 4ns
   #   pulse_duration_end: 200
--- a/runcards/actions_qq_5q.yml
+++ b/runcards/actions_qq_5q.yml
@@ -80,6 +80,14 @@ actions:
   #   software_averages: 1
   #   points: 10
 
+  # qubit_attenuation:
+  #   freq_start: -100.e+6
+  #   freq_end: +100.e+6
+  #   freq_step: 2.e+6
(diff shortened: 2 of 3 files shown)


DavidSarlle:
@maxhant looks good to me. It works with the actual runcard in qibolab/transpilers5q

Just a comment. We should start thinking how to extract data automatically from heatmap routines. Is not a priority but, it would be nice for the future automatization of the chracaterizations.

maxhant:
> Just a comment. We should start thinking how to extract data automatically from heatmap routines. Is not a priority but, it would be nice for the future automatization of the chracaterizations.

Yes, it is indeed. I would say that it is more critical for the resonator's punchout.

runcards/actions_qq_1q.yml
```diff
@@ -34,6 +34,14 @@ actions:
     software_averages: 1
     points: 10
 
+  # qubit_attenuation:
+  #   freq_start: -100.e+6
+  #   freq_end: +100.e+6
+  #   freq_step: 2.e+6
+  #   attenuation_list: range(38,60,2)
```
andrea-pasquale, line 41: I think that we should keep 3 arguments like we do for the frequency, otherwise we create an asymmetry in the code. 
Something like: `att_start`, `att_end` and `att_step`.

runcards/actions_qq_5q.yml
```diff
@@ -78,6 +78,14 @@ actions:
   #   software_averages: 1
   #   points: 10
 
+  # qubit_attenuation:
+  #   freq_start: -100.e+6
+  #   freq_end: +100.e+6
+  #   freq_step: 2.e+6
+  #   attenuation_list: range(38,60,2)
```
andrea-pasquale, line 85: Same here

src/qibocal/calibrations/characterization/qubit_spectroscopy.py
```diff
@@ -263,3 +263,60 @@ def qubit_spectroscopy_flux_track(
...
+
+    data = Dataset(
+        name=f"data_q{qubit}", quantities={"frequency": "Hz", "attenuation": "dB"}
+    )
+
+    lo_qcm_frequency = platform.characterization["single_qubit"][qubit]["qubit_freq"]
+    freqrange = np.arange(freq_start, freq_end, freq_step) + lo_qcm_frequency
+
+    if isinstance(attenuation_list, str):
+        attenuation_list = eval(attenuation_list)
```
andrea-pasquale, line 296: If you perform the above changes here you can do something like:
```
attrange = np.arange(att_start, att_end, att_step)
```

andrea-pasquale (commented):
Thanks @maxhant.
You can find below some comments. I will add a `do not merge` tag for the same reason as #107.

runcards/actions_qq_1q.yml
maxhant: I disagree because sometimes I need array like [10,12, 20,22] and sometimes [0,2,4,6,8,10]. I am aware that it is not the most elegant way but it is the only practical way to do it.

andrea-pasquale: I see, but I still don't like that this will generate an asymmetry in the code.
I would prefer to have three different values as arguments. Lets see also what others think @aorgazf @DavidSarlle.

src/qibocal/calibrations/characterization/qubit_spectroscopy.py
stavros11: Just adding that many people dislike `eval` for various reasons. For example, see [here](https://stackoverflow.com/a/9384005) for a potential way to screw up your machine by using this with a malicious runcard. I did not test this but our runcard acts `input` so this allows the user to execute arbitrary code.

PR #116, opened 2022-10-14
add Dataset set()

aorgazf:
Hi, In this PR I add a `set()` method to Dataset so that the entire dataframe can be loaded at once from a dictionary, instead of row by row.
I believe this feature had long back been requested by @maxhant .
Row by row is very time consuming, it takes ~5min for adding 1024 rows in a calibrate_qubit_states_binning routine.

example of use:
```python
    exc_sequence = PulseSequence()
    RX_pulse = platform.create_RX_pulse(qubit, start=0)
    ro_pulse = platform.create_qubit_readout_pulse(qubit, start=RX_pulse.duration)
    exc_sequence.add(RX_pulse)
    exc_sequence.add(ro_pulse)
    data_exc = Dataset(name=f"data_exc_q{qubit}", quantities={"iteration": "dimensionless"})
    msr, phase, i, q = platform.execute_pulse_sequence(exc_sequence, nshots)['binned_integrated'][ro_pulse.serial]
    
    iq_exc = i + 1j * q
    results = {
        "MSR[V]": msr,
        "i[V]": i,
        "q[V]": q,
        "phase[rad]": phase,
        "iteration[dimensionless]": np.arange(nshots),
    }
    data_exc.set(results)
    yield data_exc
```

### DIFF
--- a/src/qibocal/calibrations/characterization/allXY.py
+++ b/src/qibocal/calibrations/characterization/allXY.py
@@ -4,7 +4,7 @@ from qibolab.platforms.abstract import AbstractPlatform
 from qibolab.pulses import PulseSequence
 
 from qibocal import plots
-from qibocal.data import Dataset
+from qibocal.data import DataUnits
 from qibocal.decorators import plot
 from qibocal.fitting.methods import drag_tunning_fit
 
@@ -50,7 +50,7 @@ def allXY(
         platform.characterization["single_qubit"][qubit]["state1_voltage"]
     )
 
-    data = Dataset(
+    data = DataUnits(
         name=f"data_q{qubit}",
         quantities={"probability": "dimensionless", "gateNumber": "dimensionless"},
     )
@@ -132,7 +132,7 @@ def allXY_iteration(
         platform.characterization["single_qubit"][qubit]["state1_voltage"]
     )
 
-    data = Dataset(
+    data = DataUnits(
         name=f"data_q{qubit}",
         quantities={
             "probability": "dimensionless",
@@ -201,7 +201,7 @@ def drag_pulse_tunning(
         - qd_pulse_test.frequency
     )
 
-    data = Dataset(name=f"data_q{qubit}", quantities={"beta_param": "dimensionless"})
+    data = DataUnits(name=f"data_q{qubit}", quantities={"beta_param": "dimensionless"})
 
     count = 0
     for beta_param in np.arange(beta_start, beta_end, beta_step).round(4):
--- a/src/qibocal/calibrations/characterization/calibrate_qubit_states.py
+++ b/src/qibocal/calibrations/characterization/calibrate_qubit_states.py
@@ -4,7 +4,7 @@ from qibolab.platforms.abstract import AbstractPlatform
 from qibolab.pulses import PulseSequence
 
 from qibocal import plots
-from qibocal.data import Dataset
+from qibocal.data import DataUnits
 from qibocal.decorators import plot
 
 
@@ -33,7 +33,7 @@ def calibrate_qubit_states(
         - RX_pulse.frequency
     )
 
-    data_exc = Dataset(name=f"data_exc_q{qubit}", quantities={"iteration": "s"})
+    data_exc = DataUnits(name=f"data_exc_q{qubit}", quantities={"iteration": "s"})
 
     count = 0
     for n in np.arange(niter):
@@ -57,7 +57,7 @@ def calibrate_qubit_states(
     ro_pulse = platform.create_qubit_readout_pulse(qubit, start=0)
     gnd_sequence.add(ro_pulse)
 
-    data_gnd = Dataset(name=f"data_gnd_q{qubit}", quantities={"iteration": "s"})
+    data_gnd = DataUnits(name=f"data_gnd_q{qubit}", quantities={"iteration": "s"})
     count = 0
     for n in np.arange(niter):
         if count % points == 0:
--- a/src/qibocal/calibrations/characterization/flipping.py
+++ b/src/qibocal/calibrations/characterization/flipping.py
@@ -4,7 +4,7 @@ from qibolab.platforms.abstract import AbstractPlatform
 from qibolab.pulses import PulseSequence
 
 from qibocal import plots
-from qibocal.data import Dataset
+from qibocal.data import DataUnits
 from qibocal.decorators import plot
 from qibocal.fitting.methods import flipping_fit
 
@@ -22,7 +22,7 @@ def flipping(
         "amplitude"
     ]
 
-    data = Dataset(name=f"data_q{qubit}", quantities={"flips": "dimensionless"})
+    data = DataUnits(name=f"data_q{qubit}", quantities={"flips": "dimensionless"})
 
     sequence = PulseSequence()
     RX90_pulse = platform.create_RX90_pulse(qubit, start=0)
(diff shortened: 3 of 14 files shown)


andrea-pasquale (commented):
Thanks for the proposal @aorgazf.
Sounds good to me.
If you agree I will perform a few changes directly here:
- add same functionality for `Data` class
- add some tests to avoid losing coverage

andrea-pasquale:
@aorgazf I've implemented the set method in more pythonic way using `@property`.
Now instead of using set you can do something like:
```python
data_exc.data = result
```

aorgazf:
> @aorgazf I've implemented the set method in more pythonic way using `@property`. Now instead of using set you can do something like:
> 
> ```python
> data_exc.data = result
> ```

thanks @andrea-pasquale for taking care of the tests.
I noticed that you have refactored the dataframe from `df` to `data` and you have modified the code in the calibration routines that used it. Doing so has resulted in some code that looks like this: 
```python
        qubit_frequency = data.data.frequency[
            data.data.MSR.index[data.data.MSR.argmin()]  # pylint: disable=E1101
        ].magnitude
```
in my opinion this `data.data` doesn't read well. Before `data.df` would tell that you were accessing the dataframe of the `Dataset` object. Now it is no longer evident the difference between the two.
How do you suggest dealing with that?

Also, instead of `Dataset` being a class that contains a dataframe attribute, (considering that the functionality added by `Dataset` is rather small), have you considered the option of `Dataset` being a subclass of pandas `Dataframe` with extended functionality? This would give users access to all the capabilities of pandas Dataframe. Is there any reason why you think it should be hidden?

If you prefer we can discuss these questions during our next meeting.

andrea-pasquale:
> I noticed that you have refactored the dataframe from `df` to `data` and you have modified the code in the calibration routines that used it. Doing so has resulted in some code that looks like this:
> 
> ```python
>         qubit_frequency = data.data.frequency[
>             data.data.MSR.index[data.data.MSR.argmin()]  # pylint: disable=E1101
>         ].magnitude
> ```
> 
> in my opinion this `data.data` doesn't read well. Before `data.df` would tell that you were accessing the dataframe of the `Dataset` object. Now it is no longer evident the difference between the two. How do you suggest dealing with that?

I agree, maybe `data` is not the best choice. I did it because I wanted to avoid using abbreviations in the attributes.
I'm also fine with keeping the name `df`. Maybe there is also a better way to do this, @stavros11 when you have time could you please have a look?

Regarding the previous lines I think that the correct way to do it is by using `get_values`:
```python
qubit_frequency = data.get_values("frequency", "Hz")[np.argmin(data.get_values("MSR", "V"))]
```
In this way we avoid `data.data` and we are also using the methods of `Dataset` instead of directly manipulating the dataframe.





> Also, instead of `Dataset` being a class that contains a dataframe attribute, (considering that the functionality added by `Dataset` is rather small), have you considered the option of `Dataset` being a subclass of pandas `Dataframe` with extended functionality? This would give users access to all the capabilities of pandas Dataframe. Is there any reason why you think it should be hidden?

I think that the main reason for which it is better to have a `Dataset` class that contains a `DataFrame` attribute, instead of having directly a subclass of pandas `DataFrame`, is that we want to create a specific object that interacts with all the other components of qibocal (qq, qq-live, etc..) in a specific way. If we code `Dataset` as a subclass of `DataFrame` we should theoretically check that all the functionalities provided by pandas will not break qibocal, in my opinion.

Secondly, the class was originally created since we are using `pint` to store the units of measure. Therefore, having a class with a df attribute is useful mainly for loading/storing data in a specific way. Moreover, as we already saw multiples times, using pint can create some issues (possibility of saving strings, array conversion, ...) so by creating our own class we force the user to use a specific set of methods that we know that should work.

Lets see also what the others think @scarrazza @DavidSarlle @maxhant .

scarrazza:
Getters and setters are usually the best option. 

Please have a look at the C++ core guidelines which are usually a good reference for OOP languages:
- http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-philosophy
- http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-interfaces

stavros11:
> I think that the main reason for which it is better to have a Dataset class that contains a DataFrame attribute, instead of having directly a subclass of pandas DataFrame, is that we want to create a specific object that interacts with all the other components of qibocal (qq, qq-live, etc..) in a specific way. If we code Dataset as a subclass of DataFrame we should theoretically check that all the functionalities provided by pandas will not break qibocal, in my opinion.

I am also not a big fan of subclassing classes from other libraries, even though I have done it in a few cases in qibo. It is a dangerous practice because your object may have attributes that you are not aware of and you may even be unintentionally overriding methods. Then it will be is not clear if the user has to look at our docs or pandas docs in order to properly use this object. It could work if we really know what we are doing but as Andrea says it could easily break something in the future. 

Keeping the dataframe as an attribute may require more lines of code but sounds safer and also allow us to control the methods we expose to the user. It also provides an abstraction of what methods are required for the dataset in order to be compatible with the rest of qibocal. However, I agree that the `data.data` does not look nice and I would also prefer `data.df`. I believe there is no issue with defining it as `self._df` and exposing it as a `df` property with a `df.setter`.

Another way to make `Dataset` behave more like a pandas dataframe without being a dataframe is to define some of the magic methods, such as `__getitem__` which enables indexing and also opting for methods like `.get_values` instead of directly manipulating `.df` in the code.

Regarding naming, another thing that is not very clear is that there is a `Dataset` and a `Data` class. From what I understand these are almost identical but `Dataset` is using `pint`. If this is the only difference we could have better names, such as `Data` and `DataUnits` (for pint), or something similar. The class using pint could also inherit from the more general class. These are not really urgent features though.

aorgazf:
Thank you both @andrea-pasquale and @stavros11 for your comments and explanations, they make a lot of sense.
Just bear in mind that the reason for wanting to access the dataframe directly is beacuse it offers a lot of functionality that obviously isn't available in our Dataset. 

> Keeping the dataframe as an attribute may require more lines of code but sounds safer and also allow us to control the methods we expose to the user. It also provides an abstraction of what methods are required for the dataset in order to be compatible with the rest of qibocal. However, I agree that the `data.data` does not look nice and I would also prefer `data.df`. I believe there is no issue with defining it as `self._df` and exposing it as a `df` property with a `df.setter`.

If you implement the above, please provide a getter so that users still have access to the dataframe.

> @aorgazf I've implemented the set method in more pythonic way using `@property`. Now instead of using set you can do something like:
> 
> ```python
> data_exc.data = result
> ```

And finally, after thinking about it, I think the above could be a little confusing, since `result` is a dictionary and `data_exc.data` (or df) a pandas dataframe. It may lead people to believe that data is a dictionary. I would use a method instead that informed about what is being done: perhaps `load_data_from_dict()`

andrea-pasquale:
Thanks for the feedback @aorgazf @stavros11 @scarrazza.
In order to address all the points I've implemented the following changes:

- Renaming the DataFrame object stored by the class in `df` instead of `data`
- Making `df` a property so it can easily be accessed using `.df`
- Implementing a setter for `df` which accepts only a pd.DataFrame, and a separate method pass a dictionary called
  `load_data_from_dict()` which corresponds to the `set` method initially introduced by Alvaro.
- Renaming `Dataset` in `DataUnits` as Stavros suggested
- Using `get_values` in `resonator_spectroscopy` and `qubit_spectroscopy`, instead of manipulating the dataframe directly.

Let me know if there is something else that we need to implement/fix otherwise we can merge this PR to main.

andrea-pasquale:
I think that we can merge this PR to main.

andrea-pasquale (approved):

Issue #94, opened 2022-10-07
Refactor plotting methods


andrea-pasquale:
Now in qibocal (`add_routines` branch), all the plotting methods are saved in scatters.py and heatmaps.py. These file are becoming quite long. It could be useful to re-organize them, as noted my @maxhant  in #78.
Given that we almost have  1:1 correspondence between calibration routines and plotting methods:

- frequency_msr_phase__fast_precision:
     - resonator_spectroscopy
     - qubit_spectroscopy
- frequency_flux_msr_phase:
      - resonator_spectroscopy_flux
      - qubit_spectroscopy_flux
- frequency_attenuation_msr_phase:
    - resonator_punchout
- frequency_attenuation_msr_phase__cut
    - resonator_punchout
- frequency_flux_msr_phase__matrix
    - resonator_spectroscopy_flux_matrix
- time_msr
    - ramsey_frequency_detuned
    - ramsey
-  t1_time_msr_phase
    -  t1
- time_msr_phase
    - rabi_pulse_length
-  gain_msr_phase
    -  rabi_pulse_gain
-   amplitude_msr_phase
    - rabi_pulse_amplitude
-  duration_gain_msr_phase
    - rabi_pulse_length_and_gain
- duration_amplitude_msr_phase
    - rabi_pulse_length_and_amplitude
- flips_msr_phase
    - flipping
- exc_gnd
    - calibrate_qubit_states_binning 
- prob_gate
   - allXY 
- prob_gate_iteration
    - allXY_iteration
- msr_beta
    - drag_pulse_tunning

A possible solution is to have a structure similar to what we have for the calibration routines.
```
plots/
* spectroscopy.py (with all the plots that referred to the resonator and the qubit spectroscopy)
* ramsey.py (with all the plots regarding ramsey)
* rabis.py (with all the plots regarding rabi oscillations)
```
and so on.
For some of the plotting methods we can also try to simplify the code where it is possible.
I don't think that this is a priority right now, since in `add_routines` all the plots are already working.
However, at some point we should see if we an simplify this structure.

andrea-pasquale:
This is the issue that @maxhant was talking about @DavidSarlle @aorgazf.

andrea-pasquale:
Fixed in #122

maxhant:
Hi @andrea-pasquale, 
To be consistent, what do you think about doing the same for the fitting?

andrea-pasquale:
Hi @maxhant,

[...]

Issue #80, opened 2022-09-23
Removing directory prints


maxhant:
Hi @andrea-pasquale 

Could we remove this print please?

```
[Qcvv 0.0.1.dev0|WARNING|2022-09-23 19:17:35]: Directory 2022-09-23-000-maxime-hantute already exists.
[Qcvv 0.0.1.dev0|WARNING|2022-09-23 19:17:35]: Trying to create directory 2022-09-23-001-maxime-hantute
[Qcvv 0.0.1.dev0|WARNING|2022-09-23 19:17:35]: Directory 2022-09-23-001-maxime-hantute already exists.
[Qcvv 0.0.1.dev0|WARNING|2022-09-23 19:17:35]: Trying to create directory 2022-09-23-002-maxime-hantute
[Qcvv 0.0.1.dev0|WARNING|2022-09-23 19:17:35]: Directory 2022-09-23-002-maxime-hantute already exists.
[Qcvv 0.0.1.dev0|WARNING|2022-09-23 19:17:35]: Trying to create directory 2022-09-23-003-maxime-hantute
[Qcvv 0.0.1.dev0|WARNING|2022-09-23 19:17:35]: Directory 2022-09-23-003-maxime-hantute already exists.
[Qcvv 0.0.1.dev0|WARNING|2022-09-23 19:17:35]: Trying to create directory 2022-09-23-004-maxime-hantute
[Qcvv 0.0.1.dev0|WARNING|2022-09-23 19:17:35]: Directory 2022-09-23-004-maxime-hantute already exists.
[Qcvv 0.0.1.dev0|WARNING|2022-09-23 19:17:35]: Trying to create directory 2022-09-23-005-maxime-hantute
[Qcvv 0.0.1.dev0|WARNING|2022-09-23 19:17:35]: Directory 2022-09-23-005-maxime-hantute already exists.
[Qcvv 0.0.1.dev0|WARNING|2022-09-23 19:17:35]: Trying to create directory 2022-09-23-006-maxime-hantute
[Qcvv 0.0.1.dev0|WARNING|2022-09-23 19:17:35]: Directory 2022-09-23-006-maxime-hantute already exists.
[Qcvv 0.0.1.dev0|WARNING|2022-09-23 19:17:35]: Trying to create directory 2022-09-23-007-maxime-hantute
[Qcvv 0.0.1.dev0|WARNING|2022-09-23 19:17:35]: Directory 2022-09-23-007-maxime-hantute already exists.
[...]

PR #122, opened 2022-10-20
Minor fixes

andrea-pasquale:
In this PR I've fixed a few minor problems raised in #94 and #80.
In particular:

- the plotting methods have been splitted into files as already described in #94
- I've downgraded the logs regarding folder creation from warnings to info, so they will not appear anymore in
  the output file. It is possible to change the log level using the environment variable `QIBOCAL_LOG_LEVEL`.
- now the logs are colored, you can see the colors using for example `less -R output_file`

I've also updated the `pyproject.toml` by adding a `lmfit` dependency since it is no longer a qibolab dependency and it is needed for all the fitting methods that we have in qibocal. 
Checklist:
- [x] Tests are passing.


andrea-pasquale:
@scarrazza as discussed yersterday I will merge this now. Tests are not passing only because of coverage we will fix this in another PR.

PR #129, opened 2022-10-24
added spin echo 3 pulses and live plotting

DavidSarlle:
This PR implements the calculation of T2 using spin echos.

Checklist:
- [ ] Tests are passing.


andrea-pasquale (approved):
Thanks @DavidSarlle.

@aorgazf @maxhant could you please review, so we can merge this to #128?

PR #155, opened 2022-11-16
Add average to plots + qubit list argument

### LINKED ISSUE #147: Software average behavior
Software average plots the new data points that are obtained during the following scans. It would be greatly helpful to plot the average of new and old data points on the same graph.

Cheers and thanks.

andrea-pasquale:
Closes #147

Here is an example on how the average will look
Screenshot from 2022-11-16 17-00-48

This PR also implements the following:
- passing list of qubits instead of a single qubit for each calibration routine
- all the improvements added by @aorgazf in PR #192

Checklist:
- [x] Tests are passing.


andrea-pasquale:
For now I'm converting this to draft since I still need to check if everything is working correctly with the qpus.

andrea-pasquale:
I've tested all the routines on `qw5q_gold` and they all seem to work as expected.
@DavidSarlle @aorgazf @maxhant could you please give it a try?

andrea-pasquale:
@scarrazza @stavros11 this PR is ready to be merged. 
Feel free to have a look otherwise we can merge it.

scarrazza:
@stavros11 could you please have a quick look/test?

runcards/actions_qq_1q.yml
```diff
@@ -1,63 +1,121 @@
-platform: tii1q_b1
+platform: qblox
 
 runcard: tii1q_b1.yml
```
stavros11, line 3: This action runcard assumes that there is a platform runcard called tii1q_b1.yml in the directory that we are running qq, right? I was doing some quick tests and I had to change this to
```yaml
platform: tii1q_b1
```
in order to use the platform runcard from the qibolab repository. This behavior is fine for me, it is just as if one clones the repository and tries to run this example runcard it will not work out of the box.

src/qibocal/web/app.py
```diff
@@ -41,6 +41,11 @@ def get_graph(n, current_figure, url):
         url = f"/data{url}"
 
     method, folder, routine, qubit, format = url.split(os.sep)[2:]
+    try:
+        qubit = int(qubit)
+    except:
+        pass
```
stavros11, line 47: This qubit is passed to the plotting method. Right now I believe the expected types for the qubit are string or int, right? If yes, then maybe 
```suggestion
    if qubit.isdigit():
        qubit = int(qubit)
```
would also work and looks slightly better (avoids the global exception).

stavros11 (approved):
I tried some resonator spectroscopy and rabi-length on tii1q_b1 and it works for me, including the average. It should be fine to merge, thanks.

runcards/actions_qq_1q.yml
andrea-pasquale: You are right. I was doing some tests with a specific runcard. I've commented that line in the runcard.

src/qibocal/web/app.py
andrea-pasquale: I agree. Thanks for the suggestion!

scarrazza:
Thank you.

PR #157, opened 2022-11-18
Getting the anharmonicity

maxhant:
@Boulos-Alfakes this will allow you to get the anharmonicity of your qubits. I am working on the fitting, but it should be available soon. It will extract the Ec and Ej hopefully

Checklist:
- [ ] Tests are passing. 

### DIFF
--- a/src/qibocal/calibrations/characterization/higher_transition.py
+++ b/src/qibocal/calibrations/characterization/higher_transition.py
@@ -1,0 +1,101 @@
+import numpy as np
+from qibolab.platforms.abstract import AbstractPlatform
+from qibolab.pulses import PulseSequence
+
+from qibocal import plots
+from qibocal.data import Dataset
+from qibocal.decorators import plot
+
+
+@plot("MSR vs length and amplitude", plots.offset_amplitude_msr_phase)
+def rabi_ef(
+    platform: AbstractPlatform,
+    qubit,
+    pulse_offset_frequency_start,
+    pulse_offset_frequency_end,
+    pulse_offset_frequency_step,
+    pulse_amplitude_start,
+    pulse_amplitude_end,
+    pulse_amplitude_step,
+    software_averages,
+    points=10,
+):
+    """
+    Calibration routine to excite the |1> to |2>
+    Sequence run is: RX - Pulse - RX - M
+    The Pulse is the pulse that is being changed to excite the |1> - |2> transition.
+    One needs to be mindfull of the IF used for the RX pulse. The offset frequency is removed from it.
+    Rabi oscillations should be observed if the |1> - |2> frequency is hit because the amplitude is varied.
+    Parameters
+    ----------
+    platform : AbstractPlatform
+        Platform containing instrument and runcard being used in the routine
+    qubit :
+        Qubit name
+    pulse_offset_frequency_start : float
+        Pulse frequency offset start in the np.arange method from the |0> and |1> frequency
+    pulse_offset_frequency_end : float
+        Pulse frequency offset end in the np.arange method from the |0> and |1> frequency
+    pulse_offset_frequency_step : float
+        Pulse frequency offset step in the np.arange method from the |0> and |1> frequency
+    pulse_amplitude_start : float
+        Pulse amplitude start in the np.arange method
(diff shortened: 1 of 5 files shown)


maxhant:
Hi @AlejandroSopena or @igres26, 
Could you check the part where I calculate the anharmonicity and let me know if you think it is correct, please? 
Thanks
M

AlejandroSopena:
Hi @maxhant, I took a look at your code and it seems to me that the calculation of the anharmonicity $\alpha = \omega_{12} - \omega_{01}$ from the experimental data is correct. I just have a small suggestion. You are trying to calculate $E_C$ and $E_J$ by diagonalising the hamiltonian to obtain $\omega_{12}$ and $\omega_{01}$ and then minimising a cost function. This seems fine to me but it would be useful for the optimisation to give some initial values of $E_C$ and $E_J$. In the transmon approximation $\omega_{01} \simeq \sqrt{8E_JE_C}-E_C$ and $\omega_{12} \simeq  \sqrt{8E_JE_C}-2E_C$ so $E_C \simeq -\alpha$ and $E_J \simeq -(\omega_{01}-\alpha)^2/8\alpha$. You can use these values as a starting point and then check how much they change after optimisation.

maxhant:
> Hi @maxhant, I took a look at your code and it seems to me that the calculation of the anharmonicity α=ω12−ω01 from the experimental data is correct. I just have a small suggestion. You are trying to calculate EC and EJ by diagonalising the hamiltonian to obtain ω12 and ω01 and then minimising a cost function. This seems fine to me but it would be useful for the optimisation to give some initial values of EC and EJ. In the transmon approximation ω01≃8EJEC−EC and ω12≃8EJEC−2EC so EC≃−α and EJ≃−(ω01−α)2/8α. You can use these values as a starting point and then check how much they change after optimisation.

Thanks @AlejandroSopena! That's a good idea, but it requires the runcard to be updated with the relevant parameters. As it is not necessary, I will update this when issue #224 is dealt with (https://github.com/qiboteam/qibolab/issues/224). In the meantime, I would say that it is ready to merge.

maxhant:
image

This is a typical result of this script

src/qibocal/plots/higher_transition.py
```diff
@@ -0,0 +1,49 @@
+import numpy as np
+import plotly.graph_objects as go
+from plotly.subplots import make_subplots
+
+from qibocal.data import DataUnits
+
+
+def offset_amplitude_msr_phase(folder, routine, qubit, format):
+    data = DataUnits.load_data(folder, routine, format, f"data_q{qubit}")
```
andrea-pasquale, line 9: I would put this line in a `try-except` block as we do for the other plotting functions.

src/qibocal/calibrations/characterization/higher_transition.py
```diff
@@ -0,0 +1,178 @@
...
+    idx = np.argmax(abs(data.get_values("MSR", "V") - data.get_values("MSR", "V")[0]))
+    qubit_12frequency = qubit_frequency - data.get_values("offset", "Hz")[idx]
+    ec_ej_res = minimize(
+        lambda x: fit_score(
+            x[0], x[1], qubit_frequency * 1e-9, qubit_12frequency * 1e-9
+        ),
+        [0.4, 4.8],
+    )
+    print(
+        f"EC = {ec_ej_res['x'][0]}, Ej = {ec_ej_res['x'][1]} for an unharmonicity of {qubit_frequency - qubit_12frequency}"
```
andrea-pasquale, line 113: I think that it would be nice to show this line directly in the plot.
To do this you should save the result of the fitting into a file and then load this data in the plotting function, then you can use [fig.annotation](https://plotly.com/python/text-and-annotations/) to put a string under the plot.
The easiest way to do this is to code a function for the fitting that should yield to file.
You can find an example here:
https://github.com/qiboteam/qibocal/blob/84353e955c80fad7d1162b13cc4140203ab8119d/src/qibocal/calibrations/characterization/resonator_spectroscopy.py#L49 

Let me know if you agree.

andrea-pasquale:
You should also have a look at why tests are failing.

PR #159, opened 2022-11-23
Standardrb

wilkensJ:
Implemented new structure with three classes, set in ```abstract.py```:

- ```Circuitfactory``` class, an iterator producing the desired random circuits
- ```Experiment``` class, holding the methods to execute circuits coming from an iterator, storing the desired data and pre-/postprocess methods 
- ```Result``` class, being able to perform the last steps of post processing and producing reports

Every module can inherit the basic functionality of the above described classes and with that build new protocols with different groups of circuits, different filter functions and so on. Every module should be able to perform experiments by itself but also have an API to use for qq.

Here, the standardrb.py module was build using these classes from ```abstract.py```. More protocols will follow.


wilkensJ:
Again, closing in favor of pull request #166 (jadwiga/rb_architecture) where the commits are held to a minimum.

scarrazza:
@wilkensJ thanks. Just a small comment, you could also use git squash to reduce or group commits together.

wilkensJ:
Oh well, that I did not know and it sounds very useful for the future.
Thank you!

Am Di., 29. Nov. 2022 um 10:30 Uhr schrieb Stefano Carrazza <
***@***.***>:

PR #178, opened 2022-12-09
Fixing tests failing on MacOS

### LINKED ISSUE #177: Tests failing on mac os
As title says

andrea-pasquale:
Closes #177 by removing `pre-commit` and `black` from dev deps.

Currently tests are failing because there is an issue with the latest release of  [black](https://github.com/psf/black/issues/3312) on MacOS. 

After talking to @AleCandido we agreed that actually both `black` and `pre-commit` are not needed inside `pyproject.toml`. In fact  `pre-commit` is not needed inside the environment given that the user could commit outside the environment generated by poetry. The same thing is true for `black` given that we are using it only with `pre-commit`.

@scarrazza let me know if you agree.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -68,14 +68,6 @@ optional = false
 python-versions = ">=3.6"
 
 [[package]]
-name = "cfgv"
-version = "3.3.1"
-description = "Validate configuration and produce human readable error messages."
-category = "dev"
-optional = false
-python-versions = ">=3.6.1"
-
-[[package]]
 name = "charset-normalizer"
 version = "2.1.1"
 description = "The Real First Universal Charset Detector. Open, modern and actively maintained alternative to Chardet."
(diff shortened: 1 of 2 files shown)


AleCandido:
Just to add a few details for a future reader:
- concerning `pre-commit`, you want to commit outside the environment: any IDE will be outside the environment, but also in a terminal you should try to use `poetry shell` as little as possible, prefer `poetry run` (that of course only works for single commands, but at least you always know what you are using), and you do not want to commit inside the Poetry shell (also because it is by design an isolated environment, and you can not use any Python CLI installed system or user-wide) - for the CI it is just a useless dependency
- concerning `black`, instead it's even harmful: if `pre-commit` and Poetry happen to have two different versions of Black, if you run inside Poetry you will also enter a loop, in which one modify the code one way and the other revert it

PR #189, opened 2022-12-28
Refactor routines in order to work `AbstractPlatform` methods

andrea-pasquale:
This PR is closely related to qiboteam/qibolab#262.
The following routines have been updated and they currently work with `tii1q_b1` and with the `dummy` platform in qibolab.

- [x] resonator_spectroscopy
- [x] resonator_punchout

If we like this approach I will continue to modify the other calibration routines.
Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.

PR #219, opened 2023-01-24
Minor fix useful for multiplex

andrea-pasquale:
Fixing a minor typo following PR #189.
The error is not triggered by single qubit platforms but only for multi qubit platforms.

Checklist:
- [x] Tests are passing.


stavros11:
At this point I believe also this can be merged, right @andrea-pasquale? Most routines seem to work well with QM too. Don't forget to change the `qubits` to `dict` in the typing and docstring. This is not a breaking issue but may be useful for documentation when writing new routines.

PR #221, opened 2023-01-25
Fix allXY plot label and plot

andrea-pasquale:
After merging #189 I changed the allXY routines because I thought that in the y axis we were supposed to plot the probability of the ground state (I got confused by the label in the plot). This PR fixes this issue.
Moreover, I changed the y label of the plot to "Expectation value of Z" to avoid future misunderstandings.

image

Checklist:
- [x] Tests are passing.


stavros11 (approved):
I merged this in my branch and it works, thanks.

PR #214, opened 2023-01-20
Live plotting refresh rate selector

DavidSarlle:
This PR implements a selector for customizing the refresh interval time in every plot of qq-live. 
Auto = time neeeded for each plotting function to retreive the data a create the figure
No refresh = stop refreshing graphs

The setting is independent for each action/graph showed in a report.

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.

PR #222, opened 2023-01-27
Adding routine to optimize Read-out

maxhant:
This is NOT a PR. It is just to share my work. 

Here is a list of 5 routines that each vary one parameter with respect to fidelity being read-out pulse amplitude, duration, frequency, and TWPA power and frequency. I will open a PR on Qibolab name QM runcard for you to have the parameters. 

I had to change how the fidelity was calculated because it was too slow otherwise (1 minute per point before compared to seconds). I haven't extensively reviewed how accurate this new method is, but as I am plotting the IQ blobs, I was referring to that as well. However, something is clearly wrong with it and the previous one because we don't get the correct rotation parameter as QM does in its script (see below a result with QM scripts)

Here are the fidelities that I got: 
calibrate_iq_blobs_qubit2
calibrate_iq_blobs_qubit1
calibrate_iq_blobs_qubit4
calibrate_iq_blobs_qubit3

As you can see, I was not able to reach the 90% fidelity that @igres26 requested. However, I found that using software classifier, I was able to reach 93% so I am working now on that. Almost done

Note that qq-live does not work with this large amount of data (sliders are heavy but really useful), so one can use a script like this to plot the data. I've put the directory where my data is so you can have a look at it this way directly! 

```python
from qibocal import plots
import pathlib
import os, glob
from qibocal.cli.builders import load_yaml

def get_latest_datafolder():
    cwd = pathlib.Path(__file__).parent / "data"
    list_dir = sorted(glob.glob(os.path.join(cwd, "*/")), key=os.path.getmtime)
    for i in range(len(list_dir)):
        if os.path.isdir(cwd / list_dir[-i] / "data"):
            return cwd / list_dir[-i]

qubits = [1, 2, 3, 4]
folder = "/home/users/maxime.hantute/measurements/16-qm_qw5q_gold/data" / "calibrate_qubit_state" #name of the routines
if __name__ == "__main__":
    if folder is None:
        folder=get_latest_datafolder()
    print(f"Plotting data from {folder}...")
    actions = load_yaml(folder / "runcard.yml")
    for qubit in qubits:
        for action in actions["actions"]:
            figures = getattr(plots, action)(folder, action, qubit, "csv")
            for figure in figures:
                figure.show()
                figure.write_html(folder / f"{action}_qubit{qubit}.png")
```

Checklist:
- [ ] Tests are passing.


aorgazf:
Hi @maxhant ,
Thanks for sharing these scripts.
Could you please merge sweepers into your branch so that we can review only the changes related to readout optimisation.
Currently it shows 35 files with changes

maxhant:
> Hi @maxhant ,
Thanks for sharing these scripts.
Could you please merge sweepers into your branch so that we can review only the changes related to readout optimization.
Currently, it shows 35 files with changes

Done, but it does not change a lot because this branch also has PR #214 merged into it. I wanted to use that for the `cz_ramsey` script that I've tried adapting for the sweepers as it would allow me to do the analysis in the plotting as a temporary solution.

It will look small once both previous PRs are merged. However, note that this PR will require the Qibolab refractors to be merged before, so not now yet. @aorgazf

andrea-pasquale:
@maxhant shall we close this in favor of #310 which is almost ready to merge?

Issue #229, opened 2023-02-01
Intermediate saving of HTML reports 


aorgazf:
Hi,
Currently, the html reports are only generated and saved when the qibocal routine finishes. Sometimes the routine does not complete or is stopped by the user. Ideally intermediate html reports are saved.

stavros11:
@andrea-pasquale @scarrazza this would be very useful to have, particularly after merging the sweepers, as the report can serve as a better alternative to qq-live and it is much faster to load too. Now we dump the report only once in the end, ideally we should dump it after every routine.

This is linked to #232, however as pointed out in https://github.com/qiboteam/qibocal/pull/232#issuecomment-1430884114 I do not think it is addressed there yet.

Also, there are many cases where the data acquisition works, but the report generation fails, for example due to a bug in a plot or a fitting fail, so we end up with a folder that has all data but no report. Do you think it would be useful to have a `qq-*` command to regenerate a report using existing data? This would also be useful when one wants to change something in a plot only, but not necessarily rerun the whole data acquisition.

scarrazza:
> Do you think it would be useful to have a qq-* command to regenerate a report using existing data?

Yes, I think is a good idea to wrap the report generator function in another program. In fact, if there is agreement about a future deprecation of qq-live, we could introduce operational modes for qq, e.g. data-acquisition, post-processing, following a similar pattern already used by RB.

andrea-pasquale:
I also agree. We definitively need to dump the report after every routine.


> In fact, if there is agreement about a future deprecation of qq-live, we could introduce operational modes for qq, e.g. data-acquisition, post-processing, following a similar pattern already used by RB.

Yes, once we merge the #239 we can think about changing all the routines in an `Experiment` object (or something similar) to be able to define different sub parts of each routine as @scarrazza.

For now I can work on a PR just to implement #232.

PR #232, opened 2023-02-02
David/pdfreports

DavidSarlle:
This PR implements:

- Automatic resizing of plotting area when multiple figures are returned in a plotting function
- Added fitting report in a styled html table instead of using hover property of plotly link
- Restyled fitting area and refresh rate area in plots 
- Added funcionality for exporting into pdf file or print any report
- Minor fixes intemplate html styling  

Checklist:
- [x] Tests are passing.


DavidSarlle:
> @DavidSarlle I was looking into this, may I ask what is the motivation to generate PDF reports given that qibocal provides tools to store reports and share URL links?

@scarrazza Some of the advantages are:
1. It gives the users the possibilty of exporting/saving/print reports directly from the qibocal GUI
2. Adds extra funcionlity for sharing to qibocal
3. The funcionality gives a user-frinedly option for sharing reports, without using qq-export
4. With this funcionality, the user can generate reports even if the calibration routine has no finished (useful if the routine crasesh before ending)
5. Report integration into larger reports can be done more easily with pdf documents than with URLs

These are the reasons that have motivated the implementation of this functionality. Talking with fabrication team and experimentalist at TII, they were happy with this possibility.

src/qibocal/web/app.py
```diff
@@ -10,6 +11,9 @@
 from qibocal.data import DataUnits
 from qibocal.web.server import server
 
+# import pandas as pd
+# from collections import OrderedDict
+
```
andrea-pasquale, line 16: We can remove this if they are no longer needed.

src/qibocal/web/app.py
```diff
@@ -117,16 +113,60 @@ def get_graph(interval, url, value):
...
+                                html.Th(className="th_styles", children="Value"),
+                            ]
+                        )
+                    ]
+                )
+            ]
+            table_rows = []
+
+            for i in range(0, len(fitting_params), 3):
+                # print(fitting_params[i])
```
andrea-pasquale, line 140: This line can also be removed.

andrea-pasquale (commented):
Thanks @DavidSarlle.
I like the table that you implemented to display the results from the fitting.
I think that the pdf report is not working as it should:
image
Can you reproduce this error?

andrea-pasquale:
I see also that this PR is linked to #229, however I don't think that this issue is addressed in this PR, correct?

src/qibocal/web/app.py
DavidSarlle: Done

src/qibocal/web/app.py
DavidSarlle: Done

DavidSarlle:
> I see also that this PR is linked to #229, however I don't think that this issue is addressed in this PR, correct?

It was somehow related because the possibility to export or print at any point of the execution of a calibration routine could be exported at that point. So, the user has the possibility to print/save an intermediate report. But, for sure is not automatic. the implementation in #245 is the best option.

DavidSarlle:
@andrea-pasquale thanks for the comment. I've been able to reproduce the problem that you pointed out. I have modified the css @media print, so all the elements in the report fit in a A4 portarit orientation (and, of course in a A4 landscape). Fitting the iframes that allocates the content from the dask app (the figures and the fitting data) for all resolutions is not easy. I'have done my best, and now the reports are good enought for almost all resolutions. 
I have also elimnated some info from the pdf report to be able to show more than one figures per page. basically I have removed the summary. See below examples for typical resolutions:

width: 1920px
image

width: 1366px
image

DavidSarlle:
@andrea-pasquale @scarrazza the PR can also be included in the first release if you are happy with the implementation. The most importat thing included in this PR is the auto resizing of the plotting area and the way of showing the fitting data. I think that is interersting if we include the PR in the first release.

scarrazza:
Looks good to me, @andrea-pasquale let me know.

andrea-pasquale:
Thanks @DavidSarlle. LGTM now.

andrea-pasquale (approved):

andrea-pasquale:
@scarrazza when can merge it now. Tests are failing only for coverage of the plottng methods which currently we don't test.

PR #195, opened 2023-01-13
Routines using hardware sweepers

stavros11:
This is a prototype for using hardware sweepers to speed-up characterization scans, following the idea described in #187. It is a breaking change in the API so I would suggest to not merge it until the abstract platform and sweeper interface is decided. In the meantime I am using it to test the Quantum Machines driver from qiboteam/qibolab#259. 

This branch only works with qiboteam/qibolab#259. Here are some reports for the qw5q_gold using the QM instruments (needs access to the qq-upload server):

* [resonator punchout](http://login.qrccluster.com:9000/tYFmq18RRTSTajsXqKZpbQ==/)
* [resonator flux](http://login.qrccluster.com:9000/p3Uo2HnSS52eTON2MFussA==/)
* [qubit spectroscopy](http://login.qrccluster.com:9000/Yhn2vEkuRxaTF-WA3eHOyA==/)

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.

PR #234, opened 2023-02-03
Fix routines for QM and Qblox

andrea-pasquale:
In this PR I will adapt all the routines that we currently have in qibocal in order to have them working sweepers.
In this way we should be able to have common routines that will run both on QM and Qblox.
In order to achieve this most of routines will need changes.

Checklist:
- [ ] Tests are passing.


andrea-pasquale:
@stavros11 I fixed conflicts. I've also tried to run all the routines (beside routines involving two qubits gate) in this branch with the `Dummy` platform. Everything seems to work apart from the routines involving flux ( I think that this is related to the fact the both qblox and dummy are not using channels).
As discussed last friday feel free to merge this in #195 so that we can work on a single PR.
Let me know if I missed something when fixing conflicts.

src/qibocal/calibrations/characterization/rabi.py
```diff
@@ -511,6 +496,8 @@ def rabi_pulse_length_and_amplitude(
     pulse_amplitude_start,
     pulse_amplitude_end,
     pulse_amplitude_step,
+    relaxation_time=None,
+    nshots=None,
```
stavros11, line 500: I have never used this routine with QM, that's why it is not updated, but since you added this it reminded me that we can already use a `Sweeper` for the amplitude here. This is not super urgent, so if you want you can ignore this comment and maybe I can do it in my PR.

By the way, I am not sure how useful all the different Rabis are at the end. The strategy we were using was to fix the pi-pulse length to 40ns and find the correct amplitude. This requires only the rabi-amplitude routine, the rest are just for verification. But there may be uses that I am not aware of, so I would not suggest to remove anything at this stage.

src/qibocal/calibrations/characterization/t1.py
```diff
@@ -108,7 +108,7 @@ def t1(
 
             for ro_pulse in ro_pulses.values():
                 # average msr, phase, i and q over the number of shots defined in the runcard
-                r = results[ro_pulse.serial].to_dict()
+                r = results[ro_pulse.serial].to_dict(average=True)
```
stavros11, line 111: This seems right, however when I run T1 I get the following error:
```
builders.py 162 _execute_single_action
for data in results:

t1.py 122 t1
yield t1_fit(

methods.py 513 t1_fit
min(voltages.values),

ValueError:
min() arg is an empty sequence
```
Do you have any idea how to fix this? I don't think it is related to this PR but you have a better of the code so maybe you can identify easier.

src/qibocal/calibrations/characterization/resonator_spectroscopy.py
```diff
@@ -463,39 +542,35 @@ def resonator_spectroscopy_flux(
...
+            biases = (
+                np.repeat(delta_bias_range, len(delta_frequency_range))
+                + platform.qubits[fluxline].flux.offset
+            )  # TODO: this will not work with dummy or qblox
+            freqs = np.array(
+                len(delta_bias_range)
+                * list(delta_frequency_range + ro_pulses[qubit].frequency)
+            ).flatten()
+            # store the results
+            r = result.to_dict()
```
stavros11, line 560: Same here
```suggestion
            r = {k: v.ravel() for k, v in result.to_dict().items()}
```

stavros11 (changes requested):
Thanks for fixing the conflicts. If you can fix the comments below, mainly the ones that are causing errors, then I agree that we should merge and have a single PR.

> I've also tried to run all the routines (beside routines involving two qubits gate) in this branch with the `Dummy` platform. 

I tried with the QM platform and most routines work except from the comments below. We can still test without having a qubit, we can just not verify if the results make sense, but if the code runs without errors that's already something.

> Everything seems to work apart from the routines involving flux ( I think that this is related to the fact the both qblox and dummy are not using channels).

Do you mean the `Channel` object? Ideally they should work even without it, as I don't think I am using it anywhere apart from some 2q-gate routines with `FluxPulse`. Is these what you mean?

runcards/actions_qm.yml
```diff
@@ -15,7 +17,7 @@ actions:
   #  relaxation_time: 50
   #  software_averages: 1
 
-  #resonator_punchout:
+  # resonator_punchout_amplitude:
```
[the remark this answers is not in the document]
andrea-pasquale: Yeah, sorry I was testing the routines using `qblox_sweeper.yml` which was originally a copy of `actions_qm.yml` where I changed some values. I forgot to update `actions_qm.yml`...
At this point I think that it should be alright if we have two different runcards (one for qblox and one for qm).
I'll rename the second runcard `sweeper_qblox.yml` in `actions_qblox` to keep the same naming convention.
Let me know what you think.

src/qibocal/calibrations/characterization/qubit_spectroscopy.py
```diff
@@ -357,36 +332,30 @@ def qubit_spectroscopy_flux(
...
+            )  # TODO: this will not work for dummy or qblox
+            freqs = np.array(
+                len(delta_bias_range)
+                * list(delta_frequency_range + qd_pulses[qubit].frequency)
+            ).flatten()
+            r = result.to_dict()
+            r.update(
+                {
+                    "frequency[Hz]": freqs,
+                    "bias[A]": biases,
```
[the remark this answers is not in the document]
andrea-pasquale: Thanks for catching this.
`bias[dimensionless]` sounds good to me.

src/qibocal/calibrations/characterization/rabi.py
andrea-pasquale: I was also discussing with @igres26 about this. I also know that the majority of this rabi routines will not be needed to perform the (auto)calibration. For the time being I suggest to keep them. Once the autocalibration is completed we can decide which calibration routines we should drop in order to simplify the code.

src/qibocal/calibrations/characterization/t1.py
```diff
@@ -108,15 +108,16 @@ def t1(
...
                 r.update(
                     {
                         "wait[ns]": wait,
-                        "qubit": qubit,
+                        "qubit": ro_pulse.qubit,
```
andrea-pasquale, line 115: This is why the T1 was not working with multiplex. Now it is working.

andrea-pasquale:
Thanks for the review @stavros11. I should have addressed all the comments that you raised. I've also tested directly the routines on QM. Let me know if something else is not working.

src/qibocal/calibrations/characterization/t1.py
stavros11: I think this is right, however now I am getting a different error:
```
builders.py 149 execute
self.update_platform_runcard(qubit, action)

builders.py 172 update_platform_runcard
params = data_fit.df.to_dict("index")[qubit if qubit == 0 else qubit - 1]
```
I think we have seen this in the past, maybe it appears because I am doing multiplex on qubits [1, 2, 3, 4] and qubit 0 is not included?

src/qibocal/calibrations/characterization/qubit_spectroscopy.py
```diff
@@ -198,49 +186,31 @@ def qubit_spectroscopy(
...
+    # calculate and save fit
     yield lorentzian_fit(
         precision_sweep_data,
         x="frequency[Hz]",
         y="MSR[uV]",
         qubits=qubits,
         resonator_type=platform.resonator_type,
-        labels=["drive_frequency", "peak_voltage"],
+        labels=["qubit_freq", "peak_voltage"],
     )
```
stavros11, line 214: Here I believe the indent was correct before your last commit. This needs to be outside the loop over `qubits` but can still be inside the loop over software average. This still works as it is, it is just that now you are only dumping at the end, not after every software average. If you are using software_average=1 like me of course there is no difference.

src/qibocal/calibrations/characterization/t1.py
```diff
@@ -108,15 +108,16 @@ def t1(
...
                         "iteration": iteration,
                     }
                 )
                 data.add(r)
+                print(data.df)
```
stavros11, line 120:
```suggestion
```

stavros11 (commented):
Thanks @andrea-pasquale, I confirm that the issues raised above are fixed, apart from T1 which now raises a different error (posted above) and also you have forgotten a `print` statement.

Other than that, one last issue that can also be addressed in my branch is the CI failing (particularly pylint). I see that you are using qibolab-main in poetry, so this should be fixed (or easier to fix) now that qiboteam/qibolab#293 is merged, but let's see.

andrea-pasquale:
> Other than that, one last issue that can also be addressed in my branch is the CI failing (particularly pylint). I see that you are using qibolab-main in poetry, so this should be fixed (or easier to fix) now that [qiboteam/qibolab#293](https://github.com/qiboteam/qibolab/pull/293) is merged, but let's see.

I can run `poetry update` directly in this PR. I think that this should fix at least pylint as you said.

andrea-pasquale:
Pylint is still failing because we are still missing qiboteam/qibolab#308 (CI is complaining about the `relative_phase` sweeper that you introduced in that PR). Once we merge that to main all tests should pass.

andrea-pasquale:
> Do you mean the `Channel` object? Ideally they should work even without it, as I don't think I am using it anywhere apart from some 2q-gate routines with `FluxPulse`. Is these what you mean?

Yes, I mean the `Channel` object.

If I try for example to run a `qubit_spectroscopy_flux` I get the following error:
```
    + platform.qubits[fluxline].flux.offset
AttributeError: 'NoneType' object has no attribute 'offset'
```
I think that we have this error because of this line in qibolab:
https://github.com/qiboteam/qibolab/blob/771687ca4be2664ff8ae38306f1165f31a5a37f1/src/qibolab/platforms/abstract.py#L69

Any ideas on how to fix this @stavros11?

andrea-pasquale:
I was able to reproduce the error for T1, now it should be fixed. I was not able to detect it the first time since I was using qq-live...
Let me know if this fix works also for you @stavros11.

stavros11:
> I was able to reproduce the error for T1, now it should be fixed. I was not able to detect it the first time since I was using qq-live... Let me know if this fix works also for you @stavros11.

Yes, the fix works, thanks.

> Any ideas on how to fix this @stavros11?

I see, thanks for explaining. One way to fix that is to implement a `get_bias` method in the `AbstractPlatform` and use it to get the `offset` value (which I think I should also rename to `bias` to avoid the confusion). I see that in the `MultiqubitPlatform` you already have `get_bias` and `set_bias` so this solution should work for both platforms. If you agree, I will open a PR for this in qibolab.

Other than that, if you agree we can merge this to my PR and do the remaining changes there. These should only be the above fix and a poetry update after merging qiboteam/qibolab#308, which hopefully will happen tomorrow.

andrea-pasquale:
> I see, thanks for explaining. One way to fix that is to implement a `get_bias` method in the `AbstractPlatform` and use it to get the `offset` value (which I think I should also rename to `bias` to avoid the confusion). I see that in the `MultiqubitPlatform` you already have `get_bias` and `set_bias` so this solution should work for both platforms. If you agree, I will open a PR for this in qibolab.

I see. Feel free to open a PR for this qibolab. Remember to modify also the `DummyPlatform` if necessary so that we don't break it for these particular routines.

> Other than that, if you agree we can merge this to my PR and do the remaining changes there. These should only be the above fix and a poetry update after merging [qiboteam/qibolab#308](https://github.com/qiboteam/qibolab/pull/308), which hopefully will happen tomorrow.

Also for me it is ok to merge this to your PR. Can you approve it before I merge it? It feels weird to merge a PR where I still see `Changes requested`.

stavros11 (approved):

Issue #233, opened 2023-02-02
Tree Executor


AleCandido:
Here the basic layout (as pseudo-Python-code, don't take too seriously), for a tree runcard executor.
Sequential, single-threaded, but we might scale later on (but we won't if it is not needed).

The new runcard is going to be very similar to the current one, and it will resemble the following:
```yaml
global-par1: ...
global-par2: ...
global-par3: ...

actions:
  - id: resonator_punchout-1
    action: resonator_spectroscopy
    par1: ...
    par2: ...
  - id: resonator_spectroscopy-1
    action: resonator_spectroscopy
    par1: ...
    par2: ...
    requires: [resonator_punchout-1]
  ...
```
(all the `*par*` are of course placeholders)

The key properties are the `id` and `requires`, which define the tree itself.

Then:
```py
# --- operation.py ---
import resonator_spectroscopy, resonator_punchout

class Operation(enum.Enum):
    """available operations"""
    resonator_spectroscopy = resonator_spectroscopy
    resonator_punchout = resonator_punchout

# --- task.py ---

Id = str

class TaskParameter(Generic[T]):
    value: Optional[T]
    source: Optional[Id]

@abstract
class Parameters:
    pass

@abstract
class Update
    pass

@abstract
class Output
    status: Status
    results: Results
    update: Update

class Task:
    operation: Operation
    parameters: Parameters
    requirements: dict[Id, bool]

    @classmethod
    def load(cls, card):
         name = ...
         parameters = ...
         requirements = ...
         return cls(operation=Operation[name], parameters=parameters, requirements=requirements) 

    def run(self) -> Output:
        self.operation.value(self.parameters)

# --- pending.py ---

class Queue:
    queue: list[Task]

    def free(self, completed=None):
        ready = []
        for task in queue:
            task.complete(completed)
            if task.ready:
                 queue.drop(task)
                 ready.append(task)
             
        return ready

# --- history.py ---

class Completed:
    task: Task
    output: task.Output

class History:
    steps: list[Completed]

    def record(self, task: Task, output: task.Output):
        completed = Completed(task, output)
        self.steps.push(completed)

    def time_travel(self, initial: Platform, time: int):
        plat = initial.copy()
        for step in steps[:time]:
            plat.update(step.output.update)

        return plat

# --- execute.py ---

class ExecutionQueue:
    queue: list[Task]

    def __iter__(self):
        wip = self.queue.pop(0)
        yield wip, wip.run()

class Executor:
[...]

PR #237, opened 2023-02-07
Automated runner: normal flow

Edoardo-Pedicillo:
This PR implements the tree executor described by #233 

- [x] design structures
- [x] graph loading
- [x] main branch execution
- [x] linear branch execution
- [x] waiters for branching
- [x] documentation

Checklist:
- [x] Tests are passing.


src/qibocal/tree/operation.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def command_3():
+    print("command 3")
+
+
+class Operation(Enum):
+    command_1 = command_1  # TODO: this is not working
+    command_2 = command_2
+    command_3 = command_3
```
AleCandido, line 19:
```suggestion
from dataclasses import dataclass
from typing import Callable

@dataclass
class Routine:
    name: str
    routine: Callable

def _command_1():
    print("command 1")

command_1 = Routine("1", routine=_command_1)

def _command_2():
    print("command 2")

command_2 = Routine("2", routine=_command_2)

def _command_3():
    print("command 3")

command_3 = Routine("3", routine=_command_3)

class Operation(Enum):
    command_1 = command_1
    command_2 = command_2
    command_3 = command_3
```

Let's start like this, test if it's working :)

AleCandido: In principle, we could use one element tuples. But let's use immediately something that looks closer to what I have in mind for the final product.

AleCandido: `Operation.command_1.routine()` should invoke your function.

If you have a string for the name of the element (the LHS inside the class, not the attribute) you can use it as `Operation['command_1']`, obtaining the same object as `Operation.command_1`.
However, whenever you can avoid the string.

src/qibocal/tree/test.py
```diff
@@ -1,7 +1,33 @@
...
+    list = [Task.load(i) for i in tasks]
+    print(type(list[0]))
+    q = Queue(list)
+    print("prova", q.queue)
+    ready = q.free()  # TODO: fix AttributeError: 'Queue' object has no attribute 'free
+    assert ready[0].requirements == {"start": True}
+    assert ready[0].ready == True
+
+
+def test_Task():
```
AleCandido, line 26:
```suggestion
def test_task():
```
Please, respect Python conventions on identifiers case
https://peps.python.org/pep-0008/

src/qibocal/tree/task.py
```diff
@@ -21,16 +21,32 @@ class Output(ABC):
...
+    _requirements: dict
+    _ready: bool
+
+    @property
+    def ready(self):
+        return self._ready
+
+    @property
+    def requirements(self):
+        return self._requirements
```
AleCandido, line 33: These properties are actually aliases, why do you need them?

src/qibocal/tree/task.py
```diff
@@ -21,16 +21,32 @@ class Output(ABC):
 class Task:
     operation: Operation
     parameters: Parameters
-    requirements: dict
+    _requirements: dict
+    _ready: bool
+
+    @property
+    def ready(self):
+        return self._ready
```
AleCandido, line 29: Actually, this is fine to be a property, but it should be:
```suggestion
    @property
    def ready(self):
        return all(self.requirements.values())
```

src/qibocal/tree/task.py
```diff
@@ -21,16 +21,32 @@ class Output(ABC):
 class Task:
     operation: Operation
     parameters: Parameters
-    requirements: dict
+    _requirements: dict
```
AleCandido, line 24:
```suggestion
    _requirements: Dict[str, bool]
```

src/qibocal/tree/task.py
```diff
@@ -21,16 +21,32 @@ class Output(ABC):
...
+    _ready: bool
```
AleCandido, line 25: Drop this attribute in favor of the property

AleCandido (changes requested):

src/qibocal/tree/task.py
```diff
@@ -0,0 +1,63 @@
...
+class Parameters(ABC):
+    pass
+
+
+class Update(ABC):
+    pass
+
+
+@dataclass
+class Output:  # TODO: write Output as abstract class
```
AleCandido, line 16: https://stackoverflow.com/a/60669138/8653979

But maybe we should do something simpler, let me think about it

Edoardo-Pedicillo:
About the method `time_travel', it will be deployed in the next step, right ?

AleCandido:
> About the method `time_travel', it will be deployed in the next step, right ?

The name is almost a joke (not completely: [time travel debugging](https://en.wikipedia.org/wiki/Time_travel_debugging) is the technical name for a similar concept, but we are not writing a debugger).

However, all the ingredients we'll be implemented here, that are:
- the history
- the outputs in the history

But in order to actually implement the time travel feature with need to know which object we want to make travelling, and at the moment we're not adding any ingredient related to the platform/runcard.
What we can do, it's to define a protocol: we assume that we can "step" any object that is capable to consume an `Output` object, with an `.update()` or `.forward()` method. At that point we could also implement `History.time_travel(traveler)`, by repeated call to `traveler.forward()`.

AleCandido:
Actually, the "normal flow" should be completed, apart from the docstrings and Sphinx description (e.g. it is nowhere written what "normal" and "exceptional flow" mean).
I will complete those, and split the implementation of the exceptional flow to a further PR.

Hooking the actual autocalibration routines can be done before or after the exceptional flow implementation, but without it, they would not be incredibly useful. However, in the end all elements mentioned will be required, even for a minimal product.

In any case, I'd merge after #262

AleCandido:
I would suggest you to have a look at the example runcards https://github.com/qiboteam/qibocal/tree/tree/tests/runcards and docs https://github.com/qiboteam/qibocal/blob/tree/doc/source/getting-started/autoruncard.rst

In particular, it should be possible to reproduce both @igres26 diagram for single qubit, but also with all the black and purple lines (no pink, grey, and green).
Dynamic conditions are not implemented (this is the *Exceptional flow*, it will be implemented last with #264), and routines are not yet accessible from the runcards (they will be implemented in a separate PR, following #265).

If you are fine with the current layout, I'd propose merging, and then directly starting with #265 to make an actual automated Normal Flow.

src/qibocal/auto/execute.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        candidates = self.successors(self.current)
+
+        if len(candidates) == 0:
+            candidates.extend([])
+
+        candidates = list(filter(lambda t: self.available(t), candidates))
+
+        # sort accord to priority
+        candidates.sort(key=lambda t: t.priority)
```
andrea-pasquale, line 77: Here we are sorting in ascending order of priority.
Maybe it is just me but I was expecting that the nodes should be ordered in descending order of priority.
If a node has a higher priority value compared to another one I would expect that it should be executed first.

andrea-pasquale (commented):
Thanks @AleCandido.
It looks very good to me.

src/qibocal/auto/draw.py
```diff
@@ -0,0 +1,35 @@
...
+
+import networkx as nx
+from networkx.drawing.nx_pydot import graphviz_layout
+
+from .graph import Graph
+from .runcard import Id
+
+
+def draw(graph: Graph, ax=None):
+    """Draw a graph according to its priority."""
```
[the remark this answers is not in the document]
AleCandido: This was a nice-to-have but still rather incomplete...

I implemented mainly for debugging, but I know that eventually we want to use it.
I was considering to make it graphviz independent, but it might be a certain amount of work and I'm not sure it's worth.

For the time being, I just made sure that if you don't use it nothing breaks

src/qibocal/auto/execute.py
AleCandido: I have not been able to find a nice solution, so I kept one of the least bad ones...

Suggestions are welcome.

andrea-pasquale: Perhaps we can use something like `wait_time`?
So that the nodes with the with the lower `wait_time` are executed first.

AleCandido: Let's try to involve other people and get more proposals: @igres26 @stavros11 @Edoardo-Pedicillo  @Jacfomg @rodolfocarobene we need some brainstorming!

In the end, it is just a UI problem, but definitely a relevant one :)

igres26 (approved):
All looks good on my end. 
As for the "priority" naming issue, I agree that it can be a bit misleading but I don't think it's that strange. I will think of alternatives. Is this something we want to consolidate in this pull request?

AleCandido:
> Is this something we want to consolidate in this pull request?

Unless right after this PR will have tons of runcards with that name, I'd say we can consider the naming issue not urgent (but better not to forget, otherwise we will have to motivate the weird name over and over).

andrea-pasquale (approved):

PR #243, opened 2023-02-14
optimizing data agroupation with numpy arrays

DavidSarlle:
This PR optimizes the plotting functions reducing the computational cost of generating calibration plots whith big sets of data. 
The function "groupby()" and the application of any math over the whole data set (like mean()) over a pandas data frame was the bottleneck. When the number of agrupations and the number of stored data increases, the computational cost increases exponentially.

The new function calculates the same output using numpy arrays. Also the data sets has been reduced to the minimal expression needed en each case before making any calculation with them, reducing the execution time.

The new method is 50% faster for big datasets and agruppations made by 1 or 2 columns.

Checklist:
- [x] Tests are passing.


DavidSarlle:
@scarrazza @andrea-pasquale can you take a look into this PR? From my side is ready to be merged into main. As described, it implements a way of groupping data frames and afterthat, calculate the mean, avoiding the use of "dataFrame.groupby" method that was causing a bottleneck when loading large sets of data. 

It would be interesting on merging it before the first release of qibocal because, the user will not have live plotting and, when the dataSset is large and we operate over it, the computational cost is to high with the current version of the plotting functions, and the user impression is that the plotting is not working.

Also, when the computational cost is too high, the dash callback enters in an infinite loop where the callback is called before it finish. For that, I have set an initial refreseh time quite hight that gives the callback time to calculate the optimal refresh rate avoiding the eternal loop. 

Now also the dropdown menu for setting the refresh rate has only "auto" and "no refresh" options to avoid the user setting a wrong refresh rate that may cause a eternal loop.

src/qibocal/plots/rabi.py
```diff
@@ -266,13 +268,13 @@ def gain_msr_phase(folder, routine, qubit, format):
...
+
             fig.add_trace(
                 go.Scatter(
-                    x=gains,
-                    y=data.df.groupby("gain")["phase"]  # pylint: disable=E1101
-                    .mean()
-                    .pint.to("rad")
-                    .pint.magnitude,
+                    x=unique_gains,
+                    y=mean_phases,  # pylint: disable=E1101
```
andrea-pasquale, line 277: I think that pylint should not complain so probably you can remove it from this line.

src/qibocal/plots/spectroscopies.py
```diff
@@ -342,13 +349,12 @@ def frequency_attenuation_msr_phase_cut(folder, routine, qubit, format):
...
+            unique_frequencies, mean_measurements = grouped_by_mean(data.df, 2, 0)
             fig.add_trace(
                 go.Scatter(
-                    x=frequencies,
-                    y=data.df.groupby("frequency")["MSR"]  # pylint: disable=E1101
-                    .mean()
-                    .pint.to("uV")
-                    .pint.magnitude,  # pylint: disable=E1101
+                    x=unique_frequencies,
+                    y=mean_measurements * 1e6,  # pylint: disable=E1101
```
andrea-pasquale, line 357: Remove pylint

src/qibocal/plots/spectroscopies.py
```diff
@@ -357,13 +363,11 @@ def frequency_attenuation_msr_phase_cut(folder, routine, qubit, format):
...
+            unique_frequencies, mean_phases = grouped_by_mean(data.df, 2, 1)
             fig.add_trace(
                 go.Scatter(
-                    x=frequencies,
-                    y=data.df.groupby("frequency")["phase"]  # pylint: disable=E1101
-                    .mean()
-                    .pint.to("rad")
-                    .pint.magnitude,  # pylint: disable=E1101
+                    x=unique_frequencies,
+                    y=mean_phases,  # pylint: disable=E1101
```
andrea-pasquale, line 370: Remove pylint

src/qibocal/plots/spectroscopies.py
```diff
@@ -411,7 +415,7 @@ def frequency_flux_msr_phase(folder, routine, qubit, format):
         iterations = data.df["iteration"].unique()
         fluxlines = data.df["fluxline"].unique()
         frequencies = data.df["frequency"].pint.to("Hz").pint.magnitude.unique()
-        biass = data.df["bias"].pint.to("V").pint.magnitude.unique()
+        # biass = data.df["bias"].pint.to("V").pint.magnitude.unique()
```
andrea-pasquale, line 418: Remove commented line

src/qibocal/plots/spectroscopies.py
```diff
@@ -803,6 +831,7 @@ def frequency_attenuation(folder, routine, qubit, format):
     return figures, fitting_report
 
 
+# Not modified or checked
```
andrea-pasquale, line 834: I understand that you didn't modify this particular plotting method.
I'm guessing that we can remove the comment, right?
Otherwise, you can open an issue to remember to do this in a later stage if needed.

src/qibocal/plots/utils.py
```diff
@@ -23,3 +25,73 @@ def get_color_state0(number):
 
 def get_color_state1(number):
     return "rgb" + str(hls_to_rgb((-0.02 - number * 9 / 20) % 1, 0.6, 0.75))
+
+
+def grouped_by_mean(df, g_column1, m_column1, g_column2=None, m_column2=None):
+    r"""
+    Given a pandas data frame, the function generates groups of unique pairs of g_column1 and g_column2
+    and calculates the mean of the given columns m_column1 and m_column2.
```
andrea-pasquale, line 33: This function looks good, however I would suggest to change the arguments from numbers to the column names to increase code readability and maintenance. If I understood correctly in the plotting methods you progressively remove the columns that are no longer needed and then you apply this function passing specific indices which can vary from 0 to 4.  If you use directly column names I think that it will be easier to avoid side effects. Let me know what you think.

andrea-pasquale (changes requested):
Thanks for implementing this @DavidSarlle.
I have a few suggestions in the comments, in particular regarding the `grouped_by_mean` function. Let me know if you agree.

andrea-pasquale:
I forgot to mention that also `pylint` is failing. Please have a look when you have time.

src/qibocal/web/app.py
```diff
@@ -55,8 +51,7 @@
         dcc.Location(id="url", refresh=False),
         dcc.Interval(
             id="interval",
-            # TODO: Perhaps the user should be allowed to change the refresh rate
-            interval=5 * 1000,
+            interval=160 * 1000,
```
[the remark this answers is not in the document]
DavidSarlle: @andrea-pasquale. The initial value has been increased to a large enough time, to avoid calling the callback method for the second time before the firts call has finished. Right after the end of the callback, in the first call, the refresh rate is calculate, modied and adjusted automatically to the time needed by each function, so the code never waits that long

src/qibocal/plots/utils.py
```diff
@@ -1,6 +1,8 @@
 import os
 from colorsys import hls_to_rgb
 
+import numpy as np
```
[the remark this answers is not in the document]
DavidSarlle: Done

src/qibocal/plots/rabi.py
DavidSarlle: Removed

src/qibocal/plots/spectroscopies.py
DavidSarlle: Removed

src/qibocal/plots/spectroscopies.py
DavidSarlle: Removed

src/qibocal/plots/spectroscopies.py
DavidSarlle: Sorry. It was a personal note. I was adding notes to each routine when I tested the optimization. 
Comment removed.

src/qibocal/web/app.py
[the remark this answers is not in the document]
andrea-pasquale: Oh I see. Thank you very much for the explanation.

src/qibocal/plots/utils.py
DavidSarlle: @andrea-pasquale change suggested implemented. I totally agree with your comment. Please take a look

andrea-pasquale: Thanks @DavidSarlle. Looks good to me.

andrea-pasquale:
I had a look, this issue has been already reported to pylint PyCQA/pylint#4577.
These errors are false positives generated by the fact that when reading from file it is assume that the object is a `TextFileReader` and not a pandas dataframe.
A possible workaround could to introduce a configuration file for pylint, see for example [here](https://github.com/PyCQA/pylint/issues/4577#issuecomment-1000245962_), otherwise we can manually suppress these error as we were doing before in the code.
What do you suggest @scarrazza?

DavidSarlle:
@andrea-pasquale @scarrazza we have detected that the computational bottleneck in plotting functions was the agroupation of data where the data in each column of the data frames were quantified using pint. Using pint, all the calculations with a pandas data frame columns were extremely slow. 

After some tests, we have decided to remove the custom "group_by" function using numpy arrays and use the pandas function removing pint. That give us a more flaxible method for agroupations without increasing too much the computational time (compared with the custom fuction using numpy arrays).

For that, the data load uses an auxiliar funcion "load_data" located at plots/utils.py where the df of DataUnits is not quantified, instead of using the data.py --> load_data() function.

I have modified all the plotting methods accordinly, in order to show the correct magnituds that was taking care "pint"

It can be consedered as a first step to remove completly pint from qibocal.

I have tested all the methods and results. From my side is ready to merge as soon as we fix the tests.

DavidSarlle:
@andrea-pasquale @scarrazza we have detected that the computational bottleneck in plotting functions was the agroupation of data where the data in each column of the data frames were quantified using pint. Using pint, all the calculations with a pandas data frame columns were extremely slow. 

After some tests, we have decided to remove the custom "group_by" function using numpy arrays and use the pandas function removing pint. That give us a more flaxible method for agroupations without increasing too much the computational time (compared with the custom fuction using numpy arrays).

For that, the data load uses an auxiliar funcion "load_data" located at plots/utils.py where the df of DataUnits is not quantified, instead of using the data.py --> load_data() function.

I have modified all the plotting methods accordinly, in order to show the correct magnituds that was taking care "pint"

It can be consedered as a first step to remove completly pint from qibocal.

I have tested all the methods and results. From my side is ready to merge as soon as we fix the tests.

DavidSarlle:
> I had a look, this issue has been already reported to pylint [PyCQA/pylint#4577](https://github.com/PyCQA/pylint/issues/4577). These errors are false positives generated by the fact that when reading from file it is assume that the object is a `TextFileReader` and not a pandas dataframe. A possible workaround could to introduce a configuration file for pylint, see for example [here](https://github.com/PyCQA/pylint/issues/4577#issuecomment-1000245962_), otherwise we can manually suppress these error as we were doing before in the code. What do you suggest @scarrazza?

@andrea-pasquale  I have been googleling the error "Instance of 'TextFileReader' has no 'xxxx' member" and it indicates that the instance used does not have a 'xxxx' method. In our case, even though you're using a Pandas DataFrame, PyLint is interpreting it as an object of the TextFileReader class. 

Seems that this error is common in PyLint, as it can sometimes have trouble inferring the type of object being used. 

We are sure that the object used is a Pandas DataFrame and not a TextFileReader, so I think we can ignore this PyLint error by using the comment # pylint: disable=E1101 before the line that's giving the error (as we did it before), and in parallel, explore the possibilty of introducing a configuration file for pylint.

andrea-pasquale:
> @andrea-pasquale @scarrazza we have detected that the computational bottleneck in plotting functions was the agroupation of data where the data in each column of the data frames were quantified using pint. Using pint, all the calculations with a pandas data frame columns were extremely slow.
> 
> After some tests, we have decided to remove the custom "group_by" function using numpy arrays and use the pandas function removing pint. That give us a more flaxible method for agroupations without increasing too much the computational time (compared with the custom fuction using numpy arrays).
> 
> For that, the data load uses an auxiliar funcion "load_data" located at plots/utils.py where the df of DataUnits is not quantified, instead of using the data.py --> load_data() function.
> 
> I have modified all the plotting methods accordingly, in order to show the correct magnituds that was taking care "pint"
> 
> It can be consedered as a first step to remove completly pint from qibocal.
> 
> I have tested all the methods and results. From my side is ready to merge as soon as we fix the tests.

Thanks for the update @DavidSarlle! I'm happy to hear that you were able to found the bottleneck. I'm not surprised that it was `pint`. I will have a closer look later at this PR, but, as you said, I think that this is a first step to remove pint from qibocal.

DavidSarlle:
@andrea-pasquale @scarrazza have you decided how to fix the false positive errors of pylint? Temporary, I have fixed inserting comments to ignore the errors in the lines causing the errors (as we didi it some time ago).

The last commit also fixes the issue #247.

scarrazza:
Please disable for those lines.

andrea-pasquale (commented):
Thanks @DavidSarlle.
I'll review the fix for #247 and then we can merge this.

andrea-pasquale:
Generally the fix for #247 looks good to me.
I've noticed that this branch introduces a bug for `drag_pulse_tuning`. The report is not generated.
Could you please have a look @DavidSarlle ?

DavidSarlle:
> Generally the fix for #247 looks good to me. I've noticed that this branch introduces a bug for `drag_pulse_tuning`. The report is not generated. Could you please have a look @DavidSarlle ?

thanks @andrea-pasquale for pointing this out. I have fixed the problem. There was a mistake constructing the fitting_report. Now it works.

andrea-pasquale:
Thanks @DavidSarlle. I've tested it and I can confirm that it works.

andrea-pasquale (approved):

src/qibocal/web/app.py
```diff
@@ -101,9 +96,9 @@ def get_graph(interval, url, value):
         # # return getattr(getattr(plots, routine), method)(data)
         figs, fitting_report = getattr(plots, method)(folder, routine, qubit, format)
         et = time.time()
-
+        print(int(et - st))
```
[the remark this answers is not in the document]
DavidSarlle: nope! Sorry. I was printing the elapsed time to check the improvements during the plotting in qq-live.
Print removed

andrea-pasquale:
I think that we can merge this now.
Tests are not passing only for coverage which is fine.

PR #248, opened 2023-02-17
Minor fixes before release

andrea-pasquale:
This PR fixes a few issues and cleans up some parts of the code before the release.

@scarrazza I'm also planning to run `poetry update` before making the first release once we publish qibolab.

Checklist:
- [ ] Tests are passing.


andrea-pasquale:
> Thanks, please go ahead with poetry, the qibolab release 0.0.1 is now online.

Done. Tests are not passing because of an issue with coverage. I have no clue on how to fix it.
@stavros11 @AleCandido could you please have a look?

scarrazza:
Looks like windows is not happy with some RB tests, I believe this issue was already fixed somewhere.

src/qibocal/calibrations/characterization/qubit_spectroscopy.py
```diff
@@ -307,35 +307,31 @@ def qubit_spectroscopy_flux(
...
+    bias_min = {}
+    bias_max = {}
 
     if fluxlines == "qubits":
         fluxlines = qubits
 
     for fluxline in fluxlines:
         # TODO: check if this is correct
-        sweetspot_currents[fluxline] = qubits[fluxline].sweetspot
+        sweetspot_biass[fluxline] = qubits[fluxline].sweetspot
```
AleCandido, line 321: I see you just ran a "find/replace" ^^

AleCandido (approved):
In principle, this is isomorphic to current `main`, apart from the updated dependency on `qibolab` and one neglected return.

I would say it should be good to go.

Concerning the tests, I can't see any specific change responsible for the bug, so I guess the fault has to be attributed to the updated dependencies.
I would suggest inspecting more carefully the logs, if you don't understand the error, paste a link to it in the logs, and I will have a look.

stavros11:
Also, I think the resonator_spectroscopy_flux does not work with the `DummyPlatform` we were discussing before. When I try using this branch and qibolab main I get
```
 line 437, in resonator_spectroscopy_flux
    sweetspot_biass[fluxline] = qubits[fluxline].sweetspot
AttributeError: 'int' object has no attribute 'sweetspot'
```

> Concerning the tests, I can't see any specific change responsible for the bug, so I guess the fault has to be attributed to the updated dependencies.

Another possibility is that this test does something random and fails sometimes and just happened to fail on windows. I am not very familiar with the RB so I may be completely wrong, but we had situations like this in qibo.

andrea-pasquale:
> Also, I think the resonator_spectroscopy_flux does not work with the `DummyPlatform` we were discussing before. When I try using this branch and qibolab main I get
> 
> ```
>  line 437, in resonator_spectroscopy_flux
>     sweetspot_biass[fluxline] = qubits[fluxline].sweetspot
> AttributeError: 'int' object has no attribute 'sweetspot'
> ```

Thank you @stavros11 for catching this. This issue was due to a side effect that the RB PR introduced on the action builder. Now it should be fine. Let me know.

src/qibocal/tests/test_niGSC_circuitfactory.py
```diff
@@ -37,27 +37,6 @@ def abstract_factorytest(gfactory):
...
-    same_count = 0
-    count = 0
-    for circuit1, circuit2 in zip(factory1, factory2):
-        same_circuit = True
-        for gate1, gate2 in zip(circuit1.queue[:-1], circuit2.queue[:-1]):
-            same_circuit *= np.array_equal(gate1.matrix, gate2.matrix)
-        same_count += same_circuit
-        count += 1
-    # Half of the runs should not give the same
-    assert same_count <= count * 0.5
```
andrea-pasquale, line 58: This was the test that was causing issue. I think that it fails randomly.
@scarrazza since you want to make a release today I've removed it for now. I think that coverage should not decrease but let's have a look once the CI is done.

scarrazza: Thanks, this seems to work.

andrea-pasquale: I think that this is ready to merge.
Let me know if there is something else to do.

Issue #249, opened 2023-02-17
Release issues


scarrazza:
Let me collect some issues which are blocking the first release and should be addressed as soon as possible:
- [ ] Tests are still failing in the main branch, [see here](https://github.com/qiboteam/qibocal/actions/runs/4207153945/jobs/7301743135#step:10:165)
- [ ] There is no workflow for poetry deployment in this repository and in the workflows.
- [ ] We cannot have a pypi package which includes pandas-pint master as dependency:
   ```
   Invalid value for requires_dist. Error: Can't have direct dependency: 
   'Pint-Pandas @ git+https://github.com/hgrecco/pint-pandas.git@master'
   ```

The first and last points are critical for the moment.

AleCandido:
>  We cannot have a pypi package which includes pandas-pint master as dependency: 

The latest commit has been on November 24th, and the latest release, v0.3, on November 15th, and it is also on PyPI

https://github.com/hgrecco/pint-pandas/commits/master
https://pypi.org/project/Pint-Pandas/

After the tag the source has not been touched (judging from the commit messages), so v0.3 should be perfectly equivalent

https://github.com/hgrecco/pint-pandas/commits/master/pint_pandas/pint_array.py

AleCandido:
However, given the popularity and maintenance of that project (https://github.com/hgrecco/pint-pandas/graphs/contributors), I would trim this dependency later on. 

Software reuse is a great thing, but for a long-term project it is worth to keep only very reliable dependencies.

Concerning pandas itself: this is a very reliable project, but I would keep in the user facing part, since I don't believe here you gain much from `Dataframe` manipulations, e.g. inside the `calibrations` subpackage.
I would just replace with dedicated objects containing Numpy arrays. 

However, all of this will naturally come as part of #233 and the auto-calibration implementation anyhow. So, I would not do it immediately, and just limit to switch Pint-pandas v0.3 dependency for this release.

AleCandido:
> Tests are still failing in the main branch, [see here](https://github.com/qiboteam/qibocal/actions/runs/4207153945/jobs/7301743135#step:10:165)

Regarding the failing test: maybe there is an easier solution for that test in particular, but a general one when Numpy arrays are involved is to get the binary dump of an array, and encode in a base64 string (all doable with Numpy and standard library).

https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tobytes.html
https://docs.python.org/3/library/base64.html#base64.b64encode

[...]

PR #252, opened 2023-02-20
Add workflow for poetry deployment

andrea-pasquale:
This PR is related to qiboteam/workflows#36 and #249 (second point).
I'm not an expert in workflows so I'm open to suggestions (especially from @AleCandido :) ). 

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,22 +1,10 @@
 name: Deploy
 
 on:
-
-  workflow_call:
-    inputs:
-        os:
-          required: true
-          type: string
-        python-version:
-          required: true
-          type: string
-        publish:
-          required: false
-          type: boolean
-          default: false
-    secrets:
-      PYPI-TOKEN:
-        required: false
+  push:
+  release:
+    types:
+      - published
 
 jobs:
   build:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -19,6 +19,3 @@ jobs:
       python-version: ${{ matrix.python-version }}
       publish: ${{ github.event_name == 'release' && github.event.action == 'published' && matrix.os == 'ubuntu-latest' && matrix.python-version == '3.9' }}
     secrets: inherit
-    with:
-      os: ${{ matrix.os }}
-      python-version: ${{ matrix.python-version }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -2,6 +2,7 @@ name: Deploy
 
 on:
   push:
+  merge_group:
   release:
     types:
       - published
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,8 +1,22 @@
 name: Deploy
 
 on:
-  workflow_dispatch:
-  push:
+
+  workflow_call:
+    inputs:
+        os:
+          required: true
+          type: string
+        python-version:
+          required: true
+          type: string
+        publish:
+          required: false
+          type: boolean
+          default: false
+    secrets:
+      PYPI-TOKEN:
+        required: false
 
 jobs:
   build:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -17,3 +17,8 @@ jobs:
     with:
       os: ${{ matrix.os }}
       python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,30 @@
+name: Deploy
+
+on:
+
+  workflow_call:
```
Edoardo-Pedicillo, line 5: It does not seem a reusable workflow (otherwise I suggest you move it in the `workflow` repository). You can choose the right trigger [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows).

Edoardo-Pedicillo (changes requested):

.github/workflows/deploy.yml
andrea-pasquale: I was not so sure about this first part. Btw, the workflow is already in `workflows` repository.
Maybe the trigger should be `workflow_dispatch`?

andrea-pasquale: I think that something like this should work:
```yaml
on:
  push:
  release:
    types:
      - published
```
which is currently what we have in qibo.

AleCandido:
https://github.com/NNPDF/eko/blob/master/.github/workflows/pypi.yml
https://github.com/NNPDF/workflows/blob/v2/.github/workflows/python-poetry-pypi.yml

AleCandido (approved):
Wait to merge until https://github.com/qiboteam/workflows/pull/36 is merged. Fine for the rest :)

andrea-pasquale:
Thanks. I've already took inspiration from those :) 

> https://github.com/NNPDF/eko/blob/master/.github/workflows/pypi.yml https://github.com/NNPDF/workflows/blob/v2/.github/workflows/python-poetry-pypi.yml

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,19 @@
...
+  release:
+    types:
+      - published
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: [3.8, 3.9, '3.10']
```
scarrazza, line 15:
```suggestion
        python-version: [3.8, 3.9, '3.10']
        publish: ${{ github.event_name == 'release' && github.event.action == 'published' && matrix.os == 'ubuntu-latest' && matrix.python-version == '3.9' }}
```

@AleCandido @andrea-pasquale, are we missing this, right?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,19 @@
...
+    uses: qiboteam/workflows/.github/workflows/deploy-pip-poetry.yml@main
```
andrea-pasquale, line 16:
```suggestion
    uses: qiboteam/workflows/.github/workflows/deploy-pip-poetry.yml@main
    with:
      os: ${{ matrix.os }}
      python-version: ${{ matrix.python-version }}
      publish: ${{ github.event_name == 'release' && github.event.action == 'published' && matrix.os == 'ubuntu-latest' && matrix.python-version == '3.9' }}
    secrets: inherit
```
This is how it is currently done in Qibo.

.github/workflows/deploy.yml
andrea-pasquale: Yes, I think so. Although maybe this line should be added below line 16 (see my comment) (?)

AleCandido: Ah, right: for me, every time this workflow is run is just to actually deploy, so I usually do not put such a key.

But if you want to build wheels on `push:` as well you definitely need to filter out this run (you do not want to deploy a version at each push).
And since it is a pure Python package, you'd be building OS independent wheels, so deploying from more than one OS would cause a clash and a workflow failure (even though the wheel would be actually deployed).

.github/workflows/deploy.yml
andrea-pasquale: @scarrazza maybe you should add those lines here (if I understood correctly).

scarrazza: Yes, indeed, but we are missing the filtering key in the workflow, @Edoardo-Pedicillo could you please have a look?

Edoardo-Pedicillo: After merging qiboteam/workflow#37 we can submit @andrea-pasquale suggested change.

andrea-pasquale:
My suggestion was not fully correct. Now it should be fine.

Issue #231, opened 2023-02-01
Support for 2q gates


aorgazf:
- [ ] characterize CZ gates
- [ ] be able to run qibo circuits with arbitrary 2q gates

andrea-pasquale:
Closing this as we have already separated issues dedicated to the bullet points in the first comment.

PR #256, opened 2023-02-23
0->2 transition and qubit spectroscopies flux fits

aorgazf:
This PR includes a new routine to find 0->2 transition and a basic fitting function for qubit spectroscopies flux.
It addresses some of the tasks included in issue #231 

Checklist:
- [ ] Tests are passing.


src/qibocal/calibrations/characterization/qubit_spectroscopy.py
```diff
@@ -215,13 +215,12 @@ def qubit_spectroscopy(
...
 
 @plot(
-    "MSR and Phase vs Qubit Drive Frequency and Flux Current",
+    "MSR and Phase vs Qubit Drive Frequency and Flux Current - Transition 0->1",
     plots.frequency_flux_msr_phase,
 )
 def qubit_spectroscopy_flux(
     platform: AbstractPlatform,
     qubits: dict,
-    drive_amplitude,
```
stavros11, line 224: Here we introduced this argument because we observed that using a lower drive amplitude than the default loaded from the runcard usually leads to better signal to noise ratio. This is because the drive pulse amplitude in the runcard is usually the pi-pulse amplitude, but here we are sending a pulse that is much longer, so the plot looks better with smaller amplitude.

stavros11 (commented):
The fitting method looks okay, even though I did not test it yet. 
Regarding the new routine, if I am right it is exactly the same as the existing `qubit_spectroscopy_flux` with the only differences being the drive amplitude that is doubled and the plot title. Personally, I disagree with duplicating the whole routine only for these two changes but I would also like to hear the opinion of @andrea-pasquale and @scarrazza.

Not tested but we may be able to get the new routine without duplicating 150 lines of code.
```py
@plot(
    "MSR and Phase vs Qubit Drive Frequency and Flux Current - 2-photon Transition 0->2",
    plots.frequency_flux_msr_phase,
)
def qubit_spectroscopy_flux_02(
    platform: AbstractPlatform,
    qubits: dict,
    drive_amplitude,
    freq_width,
    freq_step,
    bias_width,
    bias_step,
    fluxlines,
    nshots=1024,
    relaxation_time=50,
    software_averages=1,
):
    yield qubit_spectroscopy_flux(platform, qubits, 2 * drive_amplitude, freq_width, freq_step, bias_width, bias_step, fluxlines, nshots, relaxation_time, software_averages)
```

src/qibocal/fitting/methods.py
```diff
@@ -1082,3 +1082,63 @@ def calibrate_qubit_states_fit(data, x, y, nshots, qubits):
         }
         parameters.add(results)
     return parameters
+
+
+def qubit_spectroscopy_flux_fit(data, x, y, qubits, resonator_type):
+    r""" """
```
andrea-pasquale, line 1088: Missing docstring

src/qibocal/fitting/methods.py
```diff
@@ -1082,3 +1082,63 @@ def calibrate_qubit_states_fit(data, x, y, nshots, qubits):
...
+
+    data_fit = Data(
+        name=f"fits",
```
andrea-pasquale, line 1091:
```suggestion
        name="fits",
```

andrea-pasquale (commented):
Thanks for implementing this @aorgazf.
I left a few comments.


> Regarding the new routine, if I am right it is exactly the same as the existing `qubit_spectroscopy_flux` with the only differences being the drive amplitude that is doubled and the plot title. Personally, I disagree with duplicating the whole routine only for these two changes but I would also like to hear the opinion of @andrea-pasquale and @scarrazza.

I agree with @stavros11 since I would also prefer to avoid code duplication whenever this is possible. I think that having an argument `drive_amplitude` could be extremely useful especially if during the autocalibration we need only small adjustments compare to the original value.

PR #261, opened 2023-03-01
Classify qubit states

Edoardo-Pedicillo:
This PR introduces the folder `qibocal/extras`, where there are the classifiers benchmarks.
Checklist:
- [x] Tests are passing.


src/qibocal/fitting/classifier/qubit_fit.py
```diff
@@ -0,0 +1,79 @@
...
+    options = ["qubit", "state"]
+    data = DataUnits(options=options)
+    data_dict = {
+        "MSR[V]": [0] * len(y),
+        "i[V]": x[:, 0].tolist(),
+        "q[V]": x[:, 1].tolist(),
+        "phase[rad]": [0] * len(y),
+        "state": y.tolist(),
+        "qubit": [1] * len(y),
+    }
```
AleCandido, line 64: Soon we will drop Pint, so all these names are not going to be that useful...

However, we should use some `QubitData` object, either already defined somewhere else, or defined in `qibocal.fitting.classifier.data`, rather than relying on dictionaries.

src/qibocal/fitting/classifier/qubit_fit.py
```diff
@@ -0,0 +1,79 @@
+from dataclasses import dataclass
+from math import cos, sin
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+from qibocal.data import DataUnits
+from qibocal.fitting.methods import calibrate_qubit_states_fit
```
AleCandido, line 8: Use relative imports:
```py
from ...data import DataUnits
```

It is definitely more robust.

AleCandido, line 8: The linear method is this function, so, after checking that is not used anywhere else, I'd move it inside here.

src/qibocal/fitting/classifier/run.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+class Classifiers(enum.Enum):
+    # linear_svm = linear_svm
+    # naive_bayes = naive_bayes
+    # rbf_svm = rbf_svm
+    # ada_boost = ada_boost
+    # random_forest = random_forest
+    # gaussian_process = gaussian_process
+    # nn = nn
```
AleCandido, line 39: You already implemented a `filter` argument, stop commenting the code.

src/qibocal/fitting/classifier/run.py
```diff
@@ -0,0 +1,235 @@
...
+    data,
+    gaussian_process,
+    linear_svm,
+    naive_bayes,
+    nn,
+    plots,
+    qubit_fit,
+    random_forest,
+    rbf_svm,
+)
```
AleCandido, line 24: Since this makes Qibocal depending on Tensorflow, Keras, and Scikit-Learn in any case (even when only the linear "Ramiro"'s classifier is used) I'd transform this imports in dynamic imports, in such a way that we can make those dependencies optional without breaking anything.

https://docs.python.org/3/library/importlib.html#module-importlib

AleCandido:
@Edoardo-Pedicillo @scarrazza from what I understood "Ramiro"'s method is a [linear SVM](https://en.wikipedia.org/wiki/Support_vector_machine#Linear_SVM), simply *manually* fitted.

Is there any value in using that, instead of an actual SVM? (or just the Naive Bayes, that is orders of magnitude faster)

Of course, the current implementation is definitely useful for benchmarking, and to prove the effectiveness and performances of the other methods.

scarrazza:
I would like to see this for performance benchmarking.

src/qibocal/fitting/classifier/plots.py
```diff
@@ -59,43 +59,50 @@ def plot_conf_matr(y_test, base_dir: pathlib.Path, classifiers=None):
...
-            ax=ax,
-            color="darkorange",
-        )
+
+        print(model)
+        try:
+            RocCurveDisplay.from_estimator(
+                model, x_test, y_test, ax=ax, color="darkorange"
+            )
+        except:
```
AleCandido, line 92: Never catch any kind of exceptions, be more specific (such that if it is an unxpected exception, still bubbles up)

src/qibocal/fitting/classifier/plots.py
```diff
@@ -59,43 +59,50 @@ def plot_conf_matr(y_test, base_dir: pathlib.Path, classifiers=None):
...
-            y_pred,
-            ax=ax,
-            color="darkorange",
-        )
+
+        print(model)
+        try:
+            RocCurveDisplay.from_estimator(
+                model, x_test, y_test, ax=ax, color="darkorange"
+            )
```
AleCandido, line 91: What if you try to do:
```suggestion
			y_pred = model.decision_function(x_test)
            RocCurveDisplay.from_predictions(
                y_test, y_pred, ax=ax, color="darkorange"
            )
```

like in here?
https://scikit-learn.org/stable/modules/generated/sklearn.metrics.RocCurveDisplay.html#sklearn.metrics.RocCurveDisplay.from_predictions

Edoardo-Pedicillo: I tried, but the neural network and `qubit fit` have not the `decision_function` method.

Edoardo-Pedicillo: Yes, I obtained the same results, but it works only with the scikit-learn classifiers.

AleCandido: That is fine and expected, but then how did you obtain `y_pred` before for the scikit ones?

Edoardo-Pedicillo: I used `predict` method. I understand your point, I will change it soon.

AleCandido: At this point, it would also be interesting to understand the difference between `.predict()` and `.decision_function()`. I would have also expected to have the same effect for a classifier, but maybe `.predict()` is returning something else...

AleCandido: It looks like they're methods of different objects:
- `.predict()` is a method of a calibrated classifier (that wraps an estimator)
- `.decision_function()` of an estimator

Why they give back different results, I still don't know
https://scikit-learn.org/stable//modules/generated/sklearn.calibration.CalibratedClassifierCV.html

AleCandido: > Predict the target of new samples.
> The predicted class is the class that has the highest probability, and can thus be different from the prediction of the uncalibrated classifier.

https://scikit-learn.org/stable//modules/generated/sklearn.calibration.CalibratedClassifierCV.html#sklearn.calibration.CalibratedClassifierCV.predict

Not sure if we need this calibration, and how are we doing it, but the plots definitely seemed weird with `.predict()`...

src/qibocal/fitting/classifier/run.py
```diff
@@ -47,6 +32,14 @@ class Classifiers(enum.Enum):
 base_dir = pathlib.Path()
 
 
+def import_classifiers(cls_names: list[str]):
+    importing_func = lambda mod: importlib.import_module(
+        ".." + mod, "qibocal.fitting.classifier.*"
+    )
+    Classifiers = list(map(importing_func, cls_names))
+    return Classifiers
```
AleCandido, line 40:
```suggestion
    classifiers = list(map(importing_func, cls_names))
    return classifiers
```

Only classes should be named in Pascal case

AleCandido: However, many kudos for the `list(map(...))` instead of the list comprehension `[importing_func(c) for c in cls_names]`.

I like both of them (both directly coming from FP), but I had the feeling no one would have ever used the first one xD

src/qibocal/fitting/classifier/run.py
```diff
@@ -0,0 +1,337 @@
...
+        classifiers = CLS_MODULES  # [i.value for i in Classifiers]
+
+    classifiers = import_classifiers(classifiers)
+
+    for mod in classifiers:
+        classifier = Classifier(mod, qubit_dir)
+        try:
+            classifier.savedir.mkdir()
+        except:
+            pass
```
AleCandido, line 246:
```suggestion
      	classifier.savedir.mkdir(exist_ok=True)
```

src/qibocal/fitting/classifier/run.py
```diff
@@ -0,0 +1,337 @@
...
+        print(mod, classifier)
+        print(classifier.name)
```
AleCandido, line 248: Remember to use [`logging`](https://docs.python.org/3/library/logging.html) eventually

extras/classification/NN-benchmark.py
```diff
@@ -0,0 +1,562 @@
+import argparse
```
AleCandido, line 1: Most likely, we can drop this file at this point (I'm pretty sure you're not using it, since it still contains merge conflicts)

extras/classification/play.py
```diff
@@ -0,0 +1,27 @@
+from pathlib import Path
+
+from qibocal.fitting.classifier import (
+    linear_svm,
+    naive_bayes,
+    plots,
+    random_forest,
+    rbf_svm,
+    run,
```
AleCandido, line 9: Since you converted to strings, in principle now you should be able to avoid the imports here
```suggestion
    plots,
    run,
```

AleCandido:
No complaint for this time, just for the future: docstrings are good, but duplicated lines are not great.
If your docstrings are all equal, because there is not much to comment, keep them short: even a single line is just fine, without all the parameters and return description, if that is the same for a bunch of similar functions. Just describe them once somewhere, and that's it :)

src/qibocal/fitting/classifier/data.py
```diff
@@ -3,6 +3,7 @@
 import matplotlib.pyplot as plt
 import pandas as pd
 import seaborn as sns
+from matplotlib.colors import ListedColormap
```
AleCandido, line 6: You're importing this, but not using

src/qibocal/fitting/classifier/qubit_fit.py
Edoardo-Pedicillo: The function is used [here](https://github.com/qiboteam/qibocal/blob/0577e91a67eb7117b7eb7ef7e3381f74c16be335/src/qibocal/calibrations/characterization/calibrate_qubit_states.py#L100), I prefer to leave it where it is.

src/qibocal/fitting/classifier/run.py
```diff
@@ -90,28 +90,6 @@ def hyperfile(self):
...
-        """
-        inst = cls(Classifiers[name], base_dir)
-        hyperpars = inst.load_hyper()
-        return inst.create_model(hyperpars)
-
-    @classmethod
-    def model_from_dir(cls, folder: pathlib.Path):
-        name = folder.name
-        base_dir = folder.parent
-        return cls.load_model(name, base_dir)
```
AleCandido: Actually why?

Edoardo-Pedicillo: I deleted this part as it was not used and not compatible with the later changes, now I have restored it with the necessary corrections.

Edoardo-Pedicillo:
I think that the code is complete, @AleCandido I would ask your advice regarding the dictionary in `qubit_fit `
https://github.com/qiboteam/qibocal/blob/01d0a4bf387378cbbf7418fa19f2aedeb2a280f0/src/qibocal/fitting/classifier/qubit_fit.py#L80-L87
for the time being, we could implement a simple solution like
```python
@dataclass
class QubitData:
    i:float
    q:float
    ...
```
or a more complex one like
``` python

@dataclass
class UnitElemet:
    value:float
    unit: str
    # implement the arithmetic 
    # operations  between UnitElements


@dataclass
class QubitData:
    i: UnitElement
    q: UnitElement
    ...
```
What do you think ?

AleCandido:
> I think that the code is complete

Then it should not be draft any longer :)

AleCandido:
> for the time being, we could implement a simple solution like

Simpler is better.

It's not clear to me that we really need to keep track of units. Most likely, a good enough solution is to standardize units inside the code (e.g. using always `V` instead of `mV`, and things like that).
Different quantities should not be mixed based on their types, so, if ever, we can do something like:

```py
from typing import NewType

Tension = NewType("Tension", float)
Angle = NewType("Angle", float)

@dataclass
class QubitData:
    i: Tension
    q: Tension
    ...
```

Types are only checked by the static analyzer, and they disappear at runtime (i.e. at runtime they are fully equivalent to floats).

src/qibocal/fitting/classifier/ada_boost.py
```diff
@@ -0,0 +1,40 @@
...
+        x_train: Training inputs.
+        y_train: Training outputs.
+        _path (path): Model save path.
+
+    Returns:
+        Dictionary with model's hyperparameters.
+    """
+    clf = AdaBoostClassifier()
+    cv = RepeatedStratifiedKFold(n_splits=10, n_repeats=3, random_state=1)
+    space = dict()
```
AleCandido, line 21:
```suggestion
    space = {}
```

The literal is considered more Pythonic.

AleCandido (commented):
I reviewed scikit classifiers, and they are simple and nice, and mostly fine :D

The qubit classifier (Ramiro's) is also mostly fine, but a bit messier, so you can improve a bit.

I'm still missing plots (which are not that relevant, only the result is) and the neural network.

tests/test_classifier.py
```diff
@@ -0,0 +1,76 @@
...
+        for file in files:
+            assert os.path.exists(qubit_dir / file)
+
+
+def test_accuracy(initialization):
+    table, _, y_test = initialization
+    real_accuracies = table["accuracy"].tolist()
+    # The model is evaluated good if it
+    # misclassifys less than two points
+    min_accuracy = (len(y_test) - 2.0) / len(y_test)
```
AleCandido, line 73:
```suggestion
    min_accuracy = 1 - 2.0 / len(y_test)
```

It's irrelevant here, but just a good practice in general: algebraic "simplification" save computations (improving performances), and reduce the numerical error.

> [...] just applies all the major simplification operations in SymPy, and uses heuristics to determine the simplest result. ***But “simplest” is not a well-defined term.***
> <p align="right"><a href="https://docs.sympy.org/dev/tutorials/intro-tutorial/simplification.html#simplify">Sympy docs</a></p>

(in computation it is more well-defined: fewer operations performed; but then there are better and worse operations for floats...)

tests/test_classifier.py
```diff
@@ -0,0 +1,76 @@
...
+
+    for acc in real_accuracies:
+        assert acc > min_accuracy
```
AleCandido, line 76:
```suggestion
    assert all(acc > min_accuracy for acc in real_accuracies)
```

This is just a matter of taste, so it is only a suggestion, since in practice they are doing the same thing (at the first `False` encountered, `all` return `False`, since it is short-circuit).
But I believe to be quite Pythonic, and, in the spirit of Python, it reads more like English.
I.e., it is closer to the natural language, "`assert` that `all` `acc`uracies are *greater than* `min`imum `accuracy`" (but even the one above is rather close, simply it would only sound more mechanic).

AleCandido: Of course it's a personal choice, so you have to make your own choice, but take into account that:

> Readability counts.
> *The Zen of Python*, https://peps.python.org/pep-0020/

AleCandido (commented):
I usually don't look too much at tests, and the worse they are the least time I spend (since life is short...).

But in this case they were short enough and already meaningful, so I thought it would have been *educative* to comment a bit.

In any case, you already did a great job :)
The only major point is that they are not really "unit tests", since they do not test individual units in isolation (but this is really difficult and time-consuming). But they are simple and clear enough, so it's already pretty good!

Edoardo-Pedicillo:
Because the tests failed due to [TensorFlow](https://github.com/qiboteam/qibocal/actions/runs/4489457620/jobs/7895304018#step:8:101), I ran the command `poetry update tensorflow`, @AleCandido could you confirm this is the right way to solve this problem?

AleCandido:
> Because the tests failed due to [TensorFlow](https://github.com/qiboteam/qibocal/actions/runs/4489457620/jobs/7895304018#step:8:101), I ran the command `poetry update tensorflow`, @AleCandido could you confirm this is the right way to solve this problem?

The reason for the failure was that you had a very old version of TensorFlow (2020), that had wheels only for py3.5-7, so completely out of our current supported range.
https://pypi.org/project/tensorflow/2.0.2/#files

You can run `poetry update tensorflow` (that is a subset of `poetry update`, that will try to update all of them), but that version might be required by some of the other dependencies, since for you it was an indirect dependency.
So, most likely Poetry will also add it explicitly to your `pyproject.toml`, since the previously resolved one should already be the most recent compatible with all the other dependencies.

I suspect that `keras` might be a bit outdated, but I haven't even looked into it. If possible, it would be better to keep only one explicitly dependency between Keras and TensorFlow, and the best candidate would be to use `tf.keras`, if possible.

Can you try to perform this investigation?

pyproject.toml
```diff
@@ -28,9 +28,14 @@ furo = { version = "^2023.3.27", optional = true }
...
 recommonmark = { version = "^0.7.1", optional = true }
 sphinx_markdown_tables = { version = "^0.0.17", optional = true }
+scikit-learn = {version = "^1.2.1", optional = true}
+keras-tuner = {version = "^1.3.0,<1.3.1", optional = true}
+matplotlib = "^3.7.0"
+seaborn = "^0.12.2"
 networkx = "^3.0"
 pydantic = "^1.10.5"
 pydot = { version = "^1.4.2", optional = true }
+tensorflow-io-gcs-filesystem = "0.31"
```
AleCandido, line 38: Why do you need extra IO support?

src/qibocal/fitting/classifier/nn.py
```diff
@@ -0,0 +1,140 @@
+import json
+import pathlib
+
+import keras_tuner as kt
+from keras import backend as K
+from keras import callbacks, optimizers
+from keras.layers import BatchNormalization, Dense, Layer
+from keras.models import Sequential
```
AleCandido, line 8: You removed `keras` dependency, but you're still relying on it.
So, at the moment you've an undeclared dependency, and you rely on `keras-tuner` dependency on `keras`.

As soon as you can, try to use `tf.keras` for this.

src/qibocal/fitting/classifier/qubit_fit.py
```diff
@@ -0,0 +1,141 @@
...
+def _check(value, sorted_list):
+    r"""Returns the position of the highest element in
+    `sorted_list` less than `value`.
+    """
+    for i, val in enumerate(sorted_list):
+        if value < val:
+            if i == 0:
+                return 0
+
+            return i - 1
```
AleCandido, line 124:
```suggestion
            return max(i - 1, 0)
```

pyproject.toml
Edoardo-Pedicillo: Because the [workflow](https://github.com/qiboteam/qibocal/actions/runs/4564441056/jobs/8054216806#step:8:73) has failed, I think that the problem with `tensorflow-io-gcs-filesystem` is that there was a new release on Mar 29,2023 that has not the Windows [wheels](https://pypi.org/project/tensorflow-io-gcs-filesystem/0.32.0/#files). Maybe I should put `tensorflow-io-gcs-filesystem = "^0.31`".

AleCandido: No, the problem is in a dependency of a dependency, do not add it to your own to make it passing on Windows.

We can think about more refined solutions, in particular dropping who depends on that, or lower the related version, but if you don't use it directly, don't declare it among your deps :)

src/qibocal/fitting/classifier/nn.py
```diff
@@ -126,13 +126,13 @@ def _hypermodel(hp):
...
+        optimizer = optimizers.legacy.Adam(learning_rate=learning_rate)
     elif optimizer_choice == "Adagrad":
-        optimizer = optimizers.Adagrad(learning_rate=learning_rate)
+        optimizer = optimizers.legacy.Adagrad(learning_rate=learning_rate)
     elif optimizer_choice == "SGD":
-        optimizer = optimizers.SGD(learning_rate=learning_rate)
+        optimizer = optimizers.legacy.SGD(learning_rate=learning_rate)
     elif optimizer_choice == "RMSprop":
-        optimizer = optimizers.RMSprop(learning_rate=learning_rate)
+        optimizer = optimizers.legacy.RMSprop(learning_rate=learning_rate)
```
AleCandido, line 135: No problem with it, even fine, but why `.legacy`?

Edoardo-Pedicillo: No, the issue was with `optimizers.RMSprop`

AleCandido: Which issue?

AleCandido: I didn't ask to fix it, I just wanted to understand.

Nevertheless, if you managed to make it work without, so much the better. Thank you very much @Edoardo-Pedicillo

Edoardo-Pedicillo: Now it works without 'legacy', at the moment I am not able to reproduce or to find easily the error, maybe I made some mistake manipulating the toml.

pyproject.toml
Edoardo-Pedicillo: Yes, I know, obviously tensorflow has this dependecy, I checked but downgrading it doesn't solve the problem. I can try with the other dependencies, but this could be a lot time-consuming. Do you have some hints to solve it as soon as possible?

src/qibocal/fitting/classifier/nn.py
AleCandido: No problem, so much the better :)

src/qibocal/fitting/classifier/qubit_fit.py
```diff
@@ -0,0 +1,138 @@
...
+        self.angle = -1 * atan2(iq_state1_trans[1], iq_state1_trans[0])
+
+        # rotate
+        iq_coord_rot = self.rotate(iq_coordinates_translated)
+
+        x_values_state0 = iq_coord_rot[(states == 0)][:, 0]
+        x_values_state1 = iq_coord_rot[(states == 1)][:, 0]
+
+        x_values_state0 = np.sort(x_values_state0)
+        x_values_state1 = np.sort(x_values_state1)
```
AleCandido, line 77:
```suggestion
        x_values_state0 = np.sort(iq_coord_rot[(states == 0)][:, 0])
        x_values_state1 = np.sort(iq_coord_rot[(states == 1)][:, 0])
```

Just to avoid shadowing, but it is really the same.

src/qibocal/fitting/classifier/qubit_fit.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    # data and points sorted
+    prob = np.zeros(len(input_data))
+    app = 0
+
+    for i, val in enumerate(input_data):
+        app = app + _check(val, points[app::])
+        prob[i] = app + 1
+
+    return prob / len(points)
```
AleCandido, line 138:
```suggestion
    prob = []
    app = 0

    for val in input_data:
        app += _check(val, points[app::])
        prob.append(app + 1)

    return np.array(prob) / len(points)
```

Very mild simplification.

src/qibocal/fitting/classifier/qubit_fit.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def _check(value, sorted_list):
+    r"""Returns the position of the highest element in
+    `sorted_list` less than `value`.
+    """
+    for i, val in enumerate(sorted_list):
+        if value < val:
+            return max(i - 1, 0)
+    return len(sorted_list) - 1
```
AleCandido, line 122: Can you test if this is equivalent to:
```py
np.max(np.searchsorted(sorted_list, value) - 1, 0)
```
?

The advantage of this one (apart from being rather compact) is that it works also on arrays (both `sorted_list` as an array, but also an array of `value`s).

src/qibocal/fitting/classifier/plots.py
AleCandido, line 1: If it doesn't take too much, can you lift all file names top-level?
- benchmarks.pdf
- confusion_matrices.pdf
- confusion_matrices.json
- ROC_curves.pdf
- ROC_curves.json
- results.pdf

AleCandido (commented):
Last few comments, but there is nothing blocking :)

After these we can merge (if needed, even before).

pyproject.toml
AleCandido: It is definitely a non-trivial problem. Proper solutions are:
- TensorFlow IO releasing wheels for Windows for v0.32.0 https://github.com/tensorflow/io/issues/1789#issuecomment-1491355793
- Poetry support to resolve indirect deps for multiple platforms https://github.com/python-poetry/poetry/issues/7746 (or at least invalidate the lock file for unsupported platforms)

I'm trying to consider if there are other possible workarounds, in order to avoid a direct dependency on `tensorflow-io-gcs-filesystem`, that we don't use.

pyproject.toml
```diff
@@ -28,9 +28,14 @@ furo = { version = "^2023.3.27", optional = true }
...
 recommonmark = { version = "^0.7.1", optional = true }
 sphinx_markdown_tables = { version = "^0.0.17", optional = true }
+scikit-learn = {version = "^1.2.1", optional = true}
+keras-tuner = {version = "^1.3.0,<1.3.1", optional = true}
+matplotlib = "^3.7.0"
+seaborn = "^0.12.2"
 networkx = "^3.0"
 pydantic = "^1.10.5"
 pydot = { version = "^1.4.2", optional = true }
+tensorflow = "^2.12.0"
```
AleCandido, line 38: Make this optional as the others, it should be used only by the classifiers.

pyproject.toml
```diff
@@ -28,9 +28,14 @@ furo = { version = "^2023.3.27", optional = true }
 sphinxcontrib-bibtex = { version = "^2.4.1", optional = true }
 recommonmark = { version = "^0.7.1", optional = true }
 sphinx_markdown_tables = { version = "^0.0.17", optional = true }
+scikit-learn = {version = "^1.2.1", optional = true}
+keras-tuner = {version = "^1.3.0,<1.3.1", optional = true}
+matplotlib = "^3.7.0"
+seaborn = "^0.12.2"
```
AleCandido, line 34: For the time being, even these are only used by the classifiers (even more restrictive: the classifier doesn't need them, they are required only for the comparison).

pyproject.toml
```diff
@@ -59,6 +64,8 @@ ipython = "^8.0"
 devtools = "^0.10.0"
 
 [tool.poetry.extras]
+
+classify = ["scikit-learn", "tensorflow", "keras-tuner"]
```
AleCandido, line 68: Add the other optional dependencies

pyproject.toml
AleCandido: Moreover, as a temporary solution use environment markers to skip Windows https://python-poetry.org/docs/dependency-specification/#using-environment-markers
(add a `# TODO:` explaining why).

tests/test_classifier.py
```diff
@@ -0,0 +1,76 @@
...
+    qubits = [0, 1]
+    _, base_dir, _ = initialization(data_path)
+    for qubit in qubits:
+        qubit_dir = qubit_path(base_dir, qubit)
+        assert os.path.exists(qubit_dir)
+        for file in FILES_NAME:
+            assert os.path.exists(qubit_dir / file)
+
+
+def test_accuracy(data_path):
```
AleCandido, line 69: Skip this test on Windows. Use `pytest.mark.skipif` https://docs.pytest.org/en/7.1.x/how-to/skipping.html#id1

(the section below has an example that explicitly skips Windows https://docs.pytest.org/en/7.1.x/how-to/skipping.html#skip-all-test-functions-of-a-class-or-module)

AleCandido (commented):
Let's try this way: we won't have `tf` on Windows, but we're mostly not even using it.

AleCandido (approved):

Edoardo-Pedicillo:
@scarrazza this PR is ready to be merged.

scarrazza:
Very good, thanks.

PR #263, opened 2023-03-05
Spectroscopies flux fits

AlejandroSopena:
This PR implements the fits for resonator and qubit flux spectroscopies in the transmon limit. 
The resonator frequency fit depends on the qubit frequency. For the qubit frequency, two approaches are considered depending on whether we have an initial estimate of $E_C$ and $E_J$ or not. To determine the resonator frequency, it is necessary to know the resonator frequency at high power as well as an estimate of the readout coupling. After this fit, the sweetspot, the asymmetry between the junctions ($d \approx 0$), the readout coupling, and the estimation of the qubit frequency at the sweetspot (and $E_C$ and $E_J$ if initial estimates are given) are obtained. With these parameters, the fit for the qubit flux spectroscopy can be carried out and a better value for the asymmetry can be obtained.

When qubits are available, I will think about the best way to extract the 1D curve from the 2D data to perform the fit.

Checklist:
- [ ] Tests are passing.


andrea-pasquale:
@AlejandroSopena what is the status of this PR? Is it ready for review?

AlejandroSopena:
Hi @andrea-pasquale, I just need to code the function to make the plots.
So far, I was just taking the minimums or maximums to obtain the frequency vs bias curve, but this sometimes gives bad results. I changed the method and now I use more information to extract the curve. I tested it with old data for the resonator and it seems to work well!
image

The red points are the data points that I apply the theoretically motivated fit to (blue curve). To obtain them, I'm using Ridge regression, which is basically least squares with L2 regularization.

andrea-pasquale:
Thanks @AlejandroSopena.
It looks very good!
Let me know if we need any help with the plotting functions :)

src/qibocal/fitting/methods.py
```diff
@@ -1073,3 +1076,344 @@ def calibrate_qubit_states_fit(data, x, y, nshots, qubits, degree=True):
...
+        leny = int((max_y - min_y) / step_y) + 1
+        lenx = int(len(x) / (leny))
+        max_x = np.max(x)
+        min_x = np.min(x)
+        X = np.linspace(min_x, max_x, lenx)
+        Y = np.linspace(min_y, max_y, leny)
+        X1, Y1 = np.meshgrid(Y, X)
+        M = np.zeros((lenx, leny))
+        for j in range(lenx):
+            M[j, :] = msr[j * leny : j * leny + leny]
```
Edoardo-Pedicillo, line 1179: This seems a reshape of msr in a matrix (lenx, leny), I think you could use the `np.reshape` function, for example
```suggestion
        M = np.reshape(msr, (lenx,leny))
```

Edoardo-Pedicillo (changes requested):
Thank @AlejandroSopena.
Please find below a few comments.

AlejandroSopena:
Hi, thanks for the comments. I didn't know Ridge regression was implemented in scikit. I've already changed it. I think it will be better than my implementation in terms of numerical stability.
I have also corrected the rest of the issues. The PR is not yet ready for review, so there are some things that are not well readable and there are also some prints, but I have already removed them. Only the plots for the qubit spectroscopy are missing.

igres26:
@AlejandroSopena @Edoardo-Pedicillo Seems like the comments were addressed according to the latest comment, but maybe some of this things are outdated? 

Can we check if this pull request can be merged as is or needs some updating?

andrea-pasquale (approved):
Thanks @AlejandroSopena.

PR #269, opened 2023-03-08
Fix bugs related to plotting

### LINKED ISSUE #266: Issues found while calibrating with QM
Here is a list of issues I had during the second round of calibrations using QM and the qmsweepers branch (PR #195 - synced with latest main):

* The `update_platform_runcard` method of `ActionBuilder` fails in some indexing
```py
params = data_fit.df.to_dict("index")[qubit if qubit == 0 else qubit - 1]
```
when using more than one qubits with qubit 0 included. In fact this indexing does not look right to me.
* Rabi-amplitude MSR plot fails with `KeyError: 0` in line `data_fit.df["popt0"][0]`. This is fixed by using `float(data_fit.df["popt0"])`.
* Ramsey plot fails with the same error and the same fix also works.

I do not have time to test if these issues appear on main, but since they are related to plotting/fitting and not the data acquisition, I would expect that they will appear on main if a multiqubit platform is used. An easy way to check would be to use the `DummyPlatform` with one or more qubits. Also, I have not tested all routines yet so there may be other issues not reported.

It seems to me that most of these issues are related to wrong indexing of pandas DataFrame: particularly the fact that DataFrame indeces are not affected when slicing the DataFrame with conditions. I believe that moving to numpy dtypes would simplify this a lot.

andrea-pasquale:
Closes #266.

Checklist:
- [x] Tests are passing.


stavros11 (commented):
Thanks for fixing these issues. I tested after merging this to qmsweepers branch and spectroscopies, Rabi amplitude and length and allXY work for me.

For Ramsey I am still getting an error, now in builders.py `update_platform_runcard` line 282, that `int` cannot convert 1D array. It may be related to the fact that fitting failed and fits.csv does not have data for all qubits. It may be good to make this robust to fitting failures, because Ramsey fit fails quite often but we would still like to get the report.

I am now trying T1.

andrea-pasquale:
> For Ramsey I am still getting an error, now in builders.py `update_platform_runcard` line 282, that `int` cannot convert 1D array. It may be related to the fact that fitting failed and fits.csv does not have data for all qubits. It may be good to make this robust to fitting failures, because Ramsey fit fails quite often but we would still like to get the report.

@stavros11 you mean `ramsey_frequency_detuned`, right?
The routine is pretty complicated and not optimized for multiplex I think.

stavros11:
> @stavros11 you mean `ramsey_frequency_detuned`, right? The routine is pretty complicated and not optimized for multiplex I think.

Sorry, I just saw this comment, I replied to the isse first. No, I am only using the normal `ramsey`, I have never used the detuned one. Even from before I remember that the fit for ramsey rarely works and even when it does the fitted curve is completely off the real one.

andrea-pasquale:
Thanks for the clarification.
I was able to spot the bug (see my comment).
Let me know if it works now @stavros11.

andrea-pasquale:
Coverage is not passing because we are missing some tests for fitting methods.
Given that this is out of the scope of this PR  I suggest to merge this now.
In the meantime I will open an issue for the missing tests.

PR #276, opened 2023-03-14
Two-qubit state tomography

stavros11:
Implements the data acquisition for the two-qubit state tomography scheme described in Supplementary Information B of https://arxiv.org/pdf/0903.2030.pdf and a basic histogram which compares the measuremed frequencies on each of the 15 different basis with simulation. 

Additional plots and information can be extracted and ideally the whole density matrix can be reconstructed. From some quick experiments I noticed that the minimization procedure from the paper is not very robust and can lead to non-pure states even when simulated (noiseless) data are used, but maybe I did something wrong. When I sort this out, I will add this step. Before doing so, I would like to mention two issues with the qibocal report integration:

1. The reports currently assume that each routine produces a set of plots for each qubits. This is not true here since we do two-qubit state tomography so we have a plot for each pair of qubits. I do not see any way around this, particularly if we are interested in measuring entanglement which cannot be done on a single.
This issue may also appear when we try to do 2q-gates on a >5q chip. With the 5q chip we did not have it because qubit 2 participates in all 2q-gates so we can get away by only tracking the second qubit (not the whole pair, since 2 was always there).
2. It would also be useful to add some text in the report, beyond the fitting parameters. For example here I would like to add the preperation circuit we are performing tomography on (for example using `circuit.draw()` from qibo), as well as several numbers that can be calculated from the density matrix once I reconstruct it (purity, purity of reduced states, entanglement entropy, etc.).

Checklist:
- [ ] Tests are passing.

In principle tomography also be done at the qibo level (using gates), however I add it to the characterization and I use pulses, as the main motivation for this was to use it to better understand how the flux pulses work for implementing 2q-gates.


andrea-pasquale:
@Jacfomg @stavros11 I believe that this PR will be superseded by #388.
If you want I can close it but I can keep the branch if we still need to port things from it.

PR #277, opened 2023-03-14
Swepeers (Rabi, Ramsey, T1, 3IQM coupler routines, Integration weights optimization and time of flight)

Jacfomg:
Updating to protocols in process on #383, #333, #399 and #340 

Rabi length, Ramsey and T1 with sweepers. Tested with Zurich Instruments.

Added: 3IQM coupler routines, Integration weights optimization and their plotting.

Tested:
- [x] Rabi length, Ramsey(check sweep pulse) and T1 with sweepers.
- [x] 3IQM coupler (Can't find the feature)
- [x] Integration weights optimization.

image

- [ ] Create a Integration weights optimization + RO pulse parameters variation (power, frequency) routine
- [x] Time of flight. Routine works but I'm not sure about the data I got.
- [x] Fast Reset sanity check. It works but the error amplification you get from the readout errors requires that much calibration that it should not be the default option never. Also, to avoid confusion calibrate qubit state was changed to return angles on radians.  For the plotting I think the best would be just to plot the to ro matrices from relaxation and active reset, but I need to look into that.

Checklist:
- [ ] Tests are passing.


src/qibocal/calibrations/characterization/ramsey.py
```diff
@@ -311,71 +313,46 @@ def ramsey(
...
+        #     resonator_type=platform.resonator_type,
+        #     qubit_freqs={qubit: qubits[qubit].drive_frequency for qubit in qubits},
+        #     sampling_rate=sampling_rate,
+        #     offset_freq=0,
+        #     labels=[
+        #         "delta_frequency",
+        #         "drive_frequency",
+        #         "T2",
+        #     ],
+        # )
```
andrea-pasquale, line 358: Is this commented because it is not working?

andrea-pasquale (commented):
Thanks @Jacfomg for this PR.
The code generally looks good to me.
I'm am not able to test it since Zurich are not accessible through the cluster.
As @stavros11 said, I would suggest to wait to merge this PR until we have sweepers for duration also for qm and qblox, this is why I'm adding a `do not merge` tag.

src/qibocal/calibrations/characterization/ramsey.py
Jacfomg: I was using an version from a week before, Stavros told that is fixed now so when it can't fit the data, what was happening with all my results, at least it won't crash.

Jacfomg:
Now that ZI is on the queue I will start updating these routines to meet the changes introduced

PR #279, opened 2023-03-17
Add routines to tree executor

andrea-pasquale:
Tasks to be implemented:

- [x] Introduce `AutoCalibrationBuilder` and `qq-auto` to allocate executor
- [x] Pass results of fitting from 1 routine to the other qiboteam/qibolab#341
- [x] Dump platform to runcard  qiboteam/qibolab#341
- [x] Recover html report

Routines:
- [x] resonator spectroscopy
- [x] punchout (amplitude or attenuation)
- [x] resonator flux dependence (fit missing) 
- [x] qubit spectroscopy
- [x] qubit flux dependence (fit missing)
- [x] rabi (amplitude, length) 
- [x] t1
- [x] ramsey (detuned) 
- [x] spin echo
- [x] calibrate qubit states 
- [x] allXYs 

Checklist:
- [x] Tests are passing.


src/qibocal/web/templates/autocalibration.html
AleCandido, line 1: Is there a reason why there is a new HTML file? In which way is this different from the existing one?

src/qibocal/protocols/characterization/hardware/resonator_spectroscopy.py
```diff
@@ -126,7 +126,7 @@ def resonator_spectroscopy(
                 data.add(r)
             count += 1
     # finally, save the remaining data and fits
-    yield data
+    return data
```
AleCandido, line 129: With this, the code after is dead. Is there a reason to keep `return` and the code after it?

In case this is just for debugging, add a `TODO` comment (at least reviewers are aware).

src/qibocal/auto/operation.py
```diff
@@ -79,20 +81,23 @@ def update(self) -> Dict[str, ParameterValue]:
 # contravariance on parameter type
 _ParametersT = TypeVar("_ParametersT", bound=Parameters, contravariant=True)
 _DataT = TypeVar("_DataT", bound=Data)
-_ResultsT = TypeVar("_ResultsT", bound=Results, covariant=True)
+_ResultsT = TypeVar("_ResultsT", bound=Results)
+_QubitsT = TypeVar("_QubitsT", bound=Qubits, contravariant=True)
```
AleCandido, line 85: Do you really need a generic type for this? Can you have `Qubits` subclasses?

src/qibocal/auto/operation.py
```diff
@@ -79,20 +81,23 @@ def update(self) -> Dict[str, ParameterValue]:
...
+_PlatformT = TypeVar("_PlatformT", bound=AbstractPlatform, contravariant=True)
```
AleCandido, line 86: Moreover (wrt the previous comment), I only use generics because I want the explicit type hint, and I want to read the actual type later on.

`Generic` are only a type hint feature, so they do not exist at runtime. And type hints should already understand subclasses.
So, unless you have a reason to explicit mark which `Platform` subclass you want to use in each acquisition routine, you can also skip this generic type.

src/qibocal/auto/operation.py
```diff
@@ -79,20 +81,23 @@ def update(self) -> Dict[str, ParameterValue]:
...
 
 
 @dataclass
-class Routine(Generic[_ParametersT, _DataT, _ResultsT]):
+class Routine(Generic[_PlatformT, _QubitsT, _ParametersT, _DataT, _ResultsT]):
```
AleCandido, line 90:
```suggestion
class Routine(Generic[_ParametersT, _DataT, _ResultsT]):
```

src/qibocal/auto/operation.py
```diff
@@ -79,20 +81,23 @@ def update(self) -> Dict[str, ParameterValue]:
...
     """A wrapped calibration routine."""
 
-    acquisition: Callable[[_ParametersT], _DataT]
+    acquisition: Callable[[_PlatformT, _QubitsT, _ParametersT], _DataT]
```
AleCandido, line 93:
```suggestion
    acquisition: Callable[[AbstractPlatform, Qubits, _ParametersT], _DataT]
```

src/qibocal/auto/operation.py
```diff
@@ -1,12 +1,14 @@
...
-from enum import Enum
 from typing import Callable, Dict, Generic, NewType, TypeVar, Union
 
+from qibolab.platforms.abstract import AbstractPlatform, Qubit
+
 OperationId = NewType("OperationId", str)
 """Identifier for a calibration routine."""
 ParameterValue = Union[float, int]
 """Valid value for a routine and runcard parameter."""
+Qubits = Dict[int, Qubit]
```
AleCandido, line 11: Add a brief comment, mainly to describe what do you use it for (what it is is already sufficiently clear).

src/qibocal/auto/task.py
```diff
@@ -74,28 +67,34 @@ def priority(self):
...
-        return Parameters.load(self.action.parameters)
+        return self.operation.parameters_type.load(self.action.parameters)
 
     def datapath(self, base_dir: Path):
-        return base_dir / f"{self.id}_{self.iteration}" / DATAFILE
+        path = base_dir / "data" / f"{self.id}_{self.iteration}"
+        os.makedirs(path)
+        return path
 
     def data(self, base_dir) -> Optional[Data]:
```
AleCandido, line 77:
```suggestion
    def data(self, base_dir: Path) -> Optional[Data]:
```

src/qibocal/auto/task.py
```diff
@@ -74,28 +67,34 @@ def priority(self):
...
-        if not self.datapath(base_dir).is_file():
+        if not self.datapath(Path(base_dir)).is_file():
```
AleCandido, line 78:
```suggestion
        if not self.datapath(base_dir).is_file():
```

Just ask for a `Path` in the type hint, at least we avoid calling `Path()` in tons of places (we should convert them just in the interface, and then assume they are `Path` objects).

src/qibocal/auto/task.py
```diff
@@ -74,28 +67,34 @@ def priority(self):
...
             return None
-
         Data = self.operation.data_type
+        # print(self.datapath(Path(base_dir)))
+        return Data.load_data(
+            base_dir, "data", f"{self.id}_{self.iteration}", "csv", "data"
```
AleCandido, line 83: `f"{self.id}_{self.iteration}"` you can also make it a property, it is sufficiently meaningful :)

src/qibocal/auto/task.py
```diff
@@ -74,28 +67,34 @@ def priority(self):
...
+        )
+        # return Data(yaml.safe_load(self.datapath(base_dir).read_text(encoding="utf-8")))
 
-        return Data(yaml.safe_load(self.datapath(base_dir).read_text(encoding="utf-8")))
-
-    def run(self) -> Results:
+    def run(self, platform, qubits, folder) -> Results:
```
AleCandido, line 87:
```suggestion
    def run(self, platform: AbstractPlatform, qubits: Qubits, folder: Path) -> Results:
```

src/qibocal/auto/task.py
```diff
@@ -74,28 +67,34 @@ def priority(self):
...
             operation: Routine = self.operation
             parameters = self.parameters
         except RuntimeError:
             operation = dummy_operation
             parameters = DummyPars()
 
-        data: Data = operation.acquisition(parameters)
+        path = self.datapath(Path(folder))
+        data: Data = operation.acquisition(platform, qubits, parameters)
+        data.to_csv(path)
```
AleCandido, line 97:
```suggestion
        data: Data = operation.acquisition(platform, qubits, parameters)
        data.to_csv(self.datapath(folder))
```

As above, we can assume a `Path` to be passed. Moreover, if you use it only once you don't need a name, unless:
- your name is specific enough, so it describes better the returned object (but `path` is rather generic)
- it would result in a very long call (but this is not the case)

src/qibocal/auto/operation.py
```diff
@@ -127,64 +132,8 @@ def _dummy_fit(data: DummyData) -> DummyRes:
...
-
-command_2 = Routine(_cmd1_acq, _cmd1_fit)
-
-
-#  ---
-#  the following enum should exist, with this name, so only its content is
-#  supposed to be an example, but it should not be exported by this module, and
-#  instead should be placed inside `qibocal.calibrations.__init__.py`
-#  ---
+def _dummy_report(data: DummyData, result: DummyRes):
```
AleCandido, line 135: Not sure I'd put the report function inside the `Routine` object, since in principle it's not needed during execution (but only after that).

You can put it in the same file, fine for the time being. Maybe it makes even sense to collect the pointer together with the rest, I'm simply not sure.
Do you have an argument for putting this here?

src/qibocal/auto/execute.py
```diff
@@ -97,14 +97,14 @@ def current(self):
         assert self.head is not None
         return self.graph.task(self.head)
 
-    def run(self):
+    def run(self, qubits, platform, folder):
```
AleCandido, line 100: Do not put them as argument, use attributes: in the end, it makes sense that the instantiated runner will run on a given platform, for some qubits, and write the result in a specified folder.

If ever you need to loop over something (e.g. you have a list of `(qubits, platform)`) you can always construct multiple executors (they are not memory heavy, they do not contain data).
```suggestion
    def run(self):
```

src/qibocal/auto/execute.py
```diff
@@ -97,14 +97,14 @@ def current(self):
...
         """Actual execution."""
         self.head = self.graph.start
 
         while self.head is not None:
             task = self.current
 
-            output = task.run()
+            output = task.run(qubits, platform, folder)
```
AleCandido, line 107:
```suggestion
            output = task.run(self.qubits, self.platform, self.folder)
```

src/qibocal/web/report.py
```diff
@@ -26,3 +27,24 @@ def create_report(path, actions=None):
 
     with open(os.path.join(path, "index.html"), "w") as file:
         file.write(html)
+
+
+def create_autocalibration_report(path, history):
+    """Creates an HTML report for the data in the given path."""
+
+    filepath = pathlib.Path(__file__)
```
AleCandido, line 35: Make this a top-level constant.
```suggestion
    WEB_DIR = pathlib.Path(__file__).parent
```

src/qibocal/web/report.py
```diff
@@ -26,3 +27,24 @@ def create_report(path, actions=None):
...
+    with open(os.path.join(filepath.with_name("static"), "styles.css")) as file:
```
AleCandido, line 36:
```suggestion
	STYLES = WEB_DIR / "static" / "styles.css"
    with open(STYLES) as file:
```
Another top-level constant.

You do not need to use `os.path` if you're working with `pathlib.Path` objects.

src/qibocal/web/report.py
```diff
@@ -26,3 +27,24 @@ def create_report(path, actions=None):
...
+        css_styles = f"<style>\n{file.read()}\n</style>"
+
+    report = AutoCalibrationReportBuilder(path, history)
+    env = Environment(loader=FileSystemLoader(filepath.with_name("templates")))
```
AleCandido, line 40:
```suggestion
	TEMPLATES = WEB_DIR / "templates"
    env = Environment(loader=FileSystemLoader(TEMPLATES))
```
At this point you should know what to do ;)

AleCandido (commented):
I still have to review the actual routines... this is just about everything else :D

src/qibocal/web/templates/autocalibration.html
andrea-pasquale: The previous HTML was using jinja assuming as input type a `ReportBuilder`. See for example here:
https://github.com/qiboteam/qibocal/blob/61b036af6b633281c3ef0dffac2ec795e63932cf/src/qibocal/web/report.py#L18

`ReportBuilder` was built for the `ActionBuilder`, its main purpose is to create the report and associate each routine with the corresponding plotting functions using decorators. Given that we currently do not use decorators to introduce plotting functions I thought that it would be easier to create a separate class.

Moreover, given that the tree executor has a `History` I want to exploit it, thus modifying the python code blocks inside the html file.

Let me know what you think.

src/qibocal/auto/operation.py
andrea-pasquale: I don't think so. I wasn't familiar with some of these lines. I will fix it.

src/qibocal/auto/operation.py
andrea-pasquale: Yes, I was planning to do this soon.

src/qibocal/cli/auto_builder.py
```diff
@@ -0,0 +1,78 @@
...
+    def __init__(self, path, history):
+        self.path = path
+        self.metadata = load_yaml(os.path.join(path, "meta.yml"))
+
+        # find proper path title
+        base, self.title = os.path.join(os.getcwd(), path), ""
+        while self.title in ("", "."):
+            base, self.title = os.path.split(base)
+
+        self.runcard = load_yaml(os.path.join(path, "runcard.yml"))
```
AleCandido, line 51: Use `/` instead of `os.path.join()`, move values (like `"runcard.yaml"`) in top-level constants.

AleCandido, line 51: The runcard has an object, i.e. `qibocal.auto.runcard.Runcard`, that should work as a loader as well. Use that one, and extend if you need more parameters than the only `action` (like `format` and `qubits`).

src/qibocal/cli/auto_builder.py
```diff
@@ -0,0 +1,78 @@
...
+        self.format = self.runcard.get("format")
+        self.qubits = self.runcard.get("qubits")
```
AleCandido, line 53: Use the runcard object.

src/qibocal/cli/auto_builder.py
AleCandido, line 51: However, you do not need a dedicated function for `load_yaml`. If you want a one-liner, just use:
```suggestion
        self.runcard = yaml.safe_load((path / "runcard.yml").read_text())
```

src/qibocal/cli/auto_builder.py
```diff
@@ -0,0 +1,78 @@
...
+
+        self.history = history
+
+    def get_routine_name(self, routine, iteration):
```
AleCandido, line 57:
```suggestion
    def routine_name(self, routine, iteration):
```

Do not use `get_`, it's just redundant.

src/qibocal/cli/auto_builder.py
```diff
@@ -0,0 +1,78 @@
...
+
+        if self.platform is not None:
+            self.platform.stop()
+            self.platform.disconnect()
+
+    def dump_report(self):
+        from qibocal.web.report import create_autocalibration_report
+
+        # update end time
+        meta = load_yaml(f"{self.folder}/meta.yml")
```
AleCandido, line 32: See the other comments:
```suggestion
        meta = yaml.safe_load((self.folder / META).read_text())
```

src/qibocal/cli/auto_builder.py
```diff
@@ -0,0 +1,78 @@
...
+
+class AutoCalibrationReportBuilder:
+    def __init__(self, path, history):
+        self.path = path
+        self.metadata = load_yaml(os.path.join(path, "meta.yml"))
+
+        # find proper path title
+        base, self.title = os.path.join(os.getcwd(), path), ""
+        while self.title in ("", "."):
+            base, self.title = os.path.split(base)
```
AleCandido, line 49: Not sure I understand the logic, but I guess you want something like:
```suggestion
        self.title = path.name
```
Otherwise, just tell me which part of the path you're looking for here, and I can help you find a more explicit way of getting it :)
(one or two examples should be enough to understand)

src/qibocal/cli/auto_builder.py
```diff
@@ -0,0 +1,78 @@
...
+
+        self.runcard = load_yaml(os.path.join(path, "runcard.yml"))
+        self.format = self.runcard.get("format")
+        self.qubits = self.runcard.get("qubits")
+
+        self.history = history
+
+    def get_routine_name(self, routine, iteration):
+        """Prettify routine's name for report headers."""
+        return routine.replace("_", " ").title() + f" {iteration}"
```
AleCandido, line 59:
```suggestion
        name = routine.replace("_", " ").title()
        return f"{name} {iteration}"
```

AleCandido: I would also add a dash in the middle, usually it is a nice separator :)
```py
f"{name} - {iteration}"
```

src/qibocal/auto/operation.py
andrea-pasquale: > Not sure I'd put the report function inside the `Routine` object, since in principle it's not needed during execution (but only after that).

Theoretically we could argue that even the fitting function is not needed in certain cases (if the user is planning only to acquire data). 
The plots are needed only when building the report. In particular I was taking advantage of the fact that in the `History` we have access to each `Routine` object to be able to recover the associated plotting function.
I'm open to suggestions :)
I think that especially for development it is nice to have a single file with data_acquisition, fitting and plotting/reporting. I feel like when we were using the decorators it was less intuitive...

andrea-pasquale: In summary I think that we can put the plotting methods wherever you want as long as the creation of the HTML doesn't become too complicated.

src/qibocal/auto/execute.py
```diff
@@ -18,17 +20,36 @@ class Executor:
...
+    platform: AbstractPlatform
+    """Qubits' platform."""
+    folder: Path
+    """Output path."""
     head: Optional[Id] = None
     """The current position."""
     pending: Set[Id] = field(default_factory=set)
     """The branched off tasks, not yet executed."""
 
+    # TODO: find a more elegant way to pass everything
```
AleCandido, line 34: You can make the other arguments `Optional`, and set their default value to `None`.

Of course, if a run is attempted, and they are still `None`, you should raise an error :)

andrea-pasquale: You are right. Thank you!

andrea-pasquale:
@AleCandido @stavros11 @aorgazf @DavidSarlle @Edoardo-Pedicillo I know that I still need to address some comments from @AleCandido however I think that this PR is almost ready to review. I spent the last weeks debugging all the calibration routines and they seem to work as expected.
I may change a few things but if you have time could you please have a look at this PR?

src/qibocal/auto/operation.py
```diff
@@ -79,20 +82,21 @@ def update(self) -> Dict[str, ParameterValue]:
...
-_ResultsT = TypeVar("_ResultsT", bound=Results, covariant=True)
+_ResultsT = TypeVar("_ResultsT", bound=Results)
 
 
 @dataclass
 class Routine(Generic[_ParametersT, _DataT, _ResultsT]):
     """A wrapped calibration routine."""
 
-    acquisition: Callable[[_ParametersT], _DataT]
+    acquisition: Callable[[AbstractPlatform, Qubits, _ParametersT], _DataT]
```
andrea-pasquale, line 92: @AleCandido I would like to make `AbstractPlatform` optional here given that it will not be needed for RB.
According to [this](https://stackoverflow.com/questions/68386130/how-to-type-hint-a-callable-of-a-function-with-default-arguments) it seems that it is not possible to do. Do you have any suggestions?

andrea-pasquale: Actually by removing it the code seems to run fine. I'm guessing that it is checking only that it is a `Callable` and not the arguments(?)

runcards/actions_qm.yml
aorgazf, line 1: Hi @andrea-pasquale , minor comment here, the name of the file `actions_qm` no longer matches with the platform `qw5q_gold_qblox`

runcards/autocalibration_qblox.yml
AleCandido, line 1: Same of above, avoid commented code

src/qibocal/auto/execute.py
```diff
@@ -18,17 +20,36 @@ class Executor:
     """The graph to be executed."""
     history: History
     """The execution history, with results and exit states."""
+    folder: Path
```
AleCandido, line 23: I'd be a bit more specific: call it `output`, as you do in the comment, such that the name is already self-commenting. Try to always be brief, but as specific as possible (sometimes eating a vocabulary would help for coding xD)

Resist the temptation to call it `output_path`: the type is already telling that is a `Path`, so use the variable just to add the information missing, i.e. the information specific to the attribute (not necessarily shared by all instances of type `Path`).

src/qibocal/auto/history.py
```diff
@@ -3,7 +3,7 @@
 from dataclasses import dataclass
 from typing import Dict, Tuple
 
-from .operation import Results
+from .operation import Data, Results
```
AleCandido, line 6: Why do you import `Data` if you're not using it?

src/qibocal/auto/execute.py
```diff
@@ -18,17 +20,36 @@ class Executor:
...
         """Load execution graph and associated executor from a runcard."""
         runcard = Runcard.load(card)
 
-        return cls(graph=Graph.from_actions(runcard.actions), history=History({}))
+        return cls(
+            graph=Graph.from_actions(runcard.actions),
+            history=History({}),
+            folder=folder,
+            platform=platform,
+            qubits=qubits,
```
AleCandido, line 51: These three elements are appearing over and over together.

Consider finding a name for the group, and gather them in a single structure.

src/qibocal/auto/runcard.py
```diff
@@ -38,6 +38,8 @@ class Runcard:
     """Structure of an execution runcard."""
 
     actions: List[Action]
+    qubits: Optional[List[int]] = None
+    format: Optional[str] = None
```
AleCandido, line 42: Consider making an `enum` with available formats.

src/qibocal/auto/runcard.py
```diff
@@ -52,4 +54,8 @@ def load(cls, card: Union[dict, Path]):
             if isinstance(card, Path)
             else card
         )
-        return cls(**content)
+        return cls(
+            actions=content["actions"],
+            qubits=content["qubits"] if "qubits" in content else [],
+            format=content["format"] if "format" in content else None,
+        )
```
AleCandido, line 61: Since you're creating `content` in this scope, you could also just modify it:
```suggestion
        content.setdefault("qubits", [])
        content.setdefault("format")
        return cls(**content)
```

AleCandido: In principle, this way of doing things would fail if there is some unrecognized field in the `content`.
I consider this a feature: unknown fields should not be present in a runcard.

src/qibocal/data.py
```diff
@@ -57,6 +57,9 @@ def to_pickle(self, path):
             path (str): Path containing output folder."""
         self.df.to_pickle(f"{path}/{self.name}.pkl")
 
+    def save(self, path):
+        return self.to_csv(path)
```
AleCandido, line 61: I would drop this method here: if you want to have a generic method `.save()`, leave it abstract, such that every subclass can adopt the best format, specific to itself.

If you want to implement a generic `save` to CSV, call it `.to_csv()`, that btw it's already present, and abstract (so `.to_csv()` is not implemented, but if you `.save()` you're saving to CSV...).

AleCandido (commented):
I'm still finishing to read even the general files, definitely far from finishing reviewing all the routines.

However, better to split the review in pieces, at least you can start reading the comments, and addressing (if you agree with them).

AleCandido:
Before keep going too far: can you fix the unit tests?

andrea-pasquale:
> Before keep going too far: can you fix the unit tests?

Thanks for the review @AleCandido.
I will take a look at comments as soon as I have time.
Unit tests are failing because this branch is working with the branch [`runcard_update`](https://github.com/qiboteam/qibolab/pull/341) in qibolab. Tests locally are passing.

AleCandido:
> Unit tests are failing because this branch is working with the branch runcard_update in qibolab. Tests locally are passing.

Good to know.

In order to make it explicit, if this works with a branch, add the branch as the dependency in `pyproject.toml`, instead of the released version.
I know you already mentioned it somewhere, but it is non-trivial to keep track of the relations among all PR and issues, so better to keep the information in the code as well, when possible.

andrea-pasquale:
> In order to make it explicit, if this works with a branch, add the branch as the dependency in `pyproject.toml`, instead of the released version.
> I know you already mentioned it somewhere, but it is non-trivial to keep track of the relations among all PR and issues, so better to keep the information in the code as well, when possible.

Ok, I can do that.
You are suggesting to add the branch temporarily in the `pyproject.toml` and we will remove it before merging to main, correct?

AleCandido:
> You are suggesting to add the branch temporarily in the `pyproject.toml` and we will remove it before merging to main, correct?

Exactly :)

In principle, if we want to merge before, I'd see nothing bad in keeping it.
Actually, I wouldn't do it now, just because `main` is being treated as the working version, and releases are mainly ignored.
But it makes senses to have a working branch, in which things might be temporary, since `main` is assumed to be unstable: new commits will appear anyhow. Provided that people can work with releases.

Edoardo-Pedicillo:
We are having two problems with the dependencies:
- one is this : hgrecco/pint-pandas#168, I solved it forcing a downgrade of pint, at the moment I don't see a better solution (since we are planning to drop pint , we should not put to much effort on it) https://github.com/qiboteam/qibocal/blob/2b4697d42d94c261bcf26f1392f78b65e62fbf5e/pyproject.toml#L41
- the workflow tests are failing (https://github.com/qiboteam/qibocal/actions/runs/4944441664/jobs/8840584164#step:8:78), I don't know how to solve it

AleCandido:
> * I solved it forcing a downgrade of pint, at the moment I don't see a better solution

For this, the only better solution would be to actually drop Pint. If possible I would do in a later PR (immediately after this), but if not, we could consider doing it immediately (if the version pinning is sufficient, we'll live with it for the time being).

> * the workflow tests are failing ([qiboteam/qibocal/actions/runs/4944441664/jobs/8840584164#step:8:78](https://github.com/qiboteam/qibocal/actions/runs/4944441664/jobs/8840584164#step:8:78)), I don't know how to solve it

Actually, why didn't they run on 2b4697d?

AleCandido:
Another point: PyLint is generating more than two thousands lines of comments.

I know they are not fully related to this PR, but it would be nice to fix at least those that are.

Edoardo-Pedicillo:
> Actually, why didn't they run on https://github.com/qiboteam/qibocal/commit/2b4697d42d94c261bcf26f1392f78b65e62fbf5e?

Because yestarday Github had problems running the workflows, now it seems working.

> However, looking at the previous commit there seems to be only one error, and it is quite clear:

This is the workflow triggered by [086b61b](https://github.com/qiboteam/qibocal/pull/279/commits/086b61b7f63e678b792d9948ba3ecbad14e42fb9) and the error is fixed by https://github.com/qiboteam/qibocal/commit/2b4697d42d94c261bcf26f1392f78b65e62fbf5e

> the workflow tests are failing (https://github.com/qiboteam/qibocal/actions/runs/4944441664/jobs/8840584164#step:8:78), I don't know how to solve it

this workflow is one that I triggered manually, but now you can see the same problem here (https://github.com/qiboteam/qibocal/actions/runs/4944441664/jobs/8840584164#step:8:78)

> If possible I would do in a later PR (immediately after this)

I agree.

AleCandido:
Great, I'm glad that the green tick is finally back :D

DavidSarlle (approved):

src/qibocal/data.py
Edoardo-Pedicillo: In my opinion we can drop it, what does @andrea-pasquale think about it?

andrea-pasquale:
> For this, the only better solution would be to actually drop Pint. If possible I would do in a later PR (immediately after this), but if not, we could consider doing it immediately (if the version pinning is sufficient, we'll live with it for the time being).

I also suggest to move pint in a separate PR. I'm planning to do several things as soon as this PR is merged to main. To avoid to create a huge PR it is better to tackle each problem in a single PR.

src/qibocal/data.py
andrea-pasquale: Yeah, I think that we could drop it. I decided to add it so that users can eventually define a save method for each `Data` structure.
https://github.com/qiboteam/qibocal/blob/8e42eddb8347ab5417147806223ac0a6967459e0/src/qibocal/auto/task.py#L102
I think that we could merge by supporting only `csv`, eventually once we drop pint we could re-introduce different formats.

src/qibocal/auto/runcard.py
andrea-pasquale: As I already said below I think that for the time being we could support only a single format (csv). We could postpone the introduction of an `enum` in a different PR.

runcards/actions_qm.yml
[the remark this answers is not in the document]
andrea-pasquale: I think that we need to think about how to remove dead code in runcards. These runcards were introduced to simplify the runcard generation by simply commenting/uncommenting blocks of code. I'm completely aware that this is not the ideal approach. 

Do you have any suggestions on how to write examples which are based on runcards?

On the top of my mind I'm thinking that we could add some markdown files to thoroughly describe how each routine is executed.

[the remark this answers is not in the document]
AleCandido: A simple option can also be to cut these examples from here, and paste them in Sphinx.

You can maintain a Sphinx page with the examples, or even a section (as you prefer). Then people can copy from there, and paste on their own runcard.

scarrazza:
@andrea-pasquale @Edoardo-Pedicillo as discussed today, if the review is done (cc @AleCandido), please go ahead and merge.

andrea-pasquale:
We still need to address some comments from @AleCandido.
I propose to merge it and open issues for the things that still needs to be done.
Let me know if you agree @scarrazza.

src/qibocal/auto/operation.py
```diff
@@ -79,20 +82,28 @@ def update(self) -> Dict[str, ParameterValue]:
...
+        if self.fit == None:
+            self.fit = _dummy_fit
+        if self.report == None:
+            self.report = _dummy_report
 
     @property
     def parameters_type(self):
         sig = inspect.signature(self.acquisition)
-        param = next(iter(sig.parameters.values()))
+        param = list(sig.parameters.values())[0]
```
AleCandido, line 106: Why did you change this?

In principle the two ways give the same result, but the new one is first constructing the full list, potentially with 1B elements, and then just getting the first. The old one was only getting the first, without looking at all the other elements.

scarrazza:
Yes, please let's target next meeting as deadline for the merge (and collection of missing features in issues).

src/qibocal/cli/utils.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+    metadata["versions"]["qibo"] = (
+        metadata["versions"]["qibo"] + " , " + metadata_new["versions"]["qibo"]
+    )
+    metadata["versions"]["qibocal"] = (
+        metadata["versions"]["qibocal"] + " , " + metadata_new["versions"]["qibocal"]
+    )
+    metadata["versions"]["qibolab"] = (
+        metadata["versions"]["qibolab"] + " , " + metadata_new["versions"]["qibolab"]
+    )
```
AleCandido, line 65: Even better:
```suggestion
    def update(old, new, key):
        old[key] += " , " + new[key]
    
    for key in ("backend", "date", "end-time", "platform", "start-time", "title"):
        update(metadata, metadata_new, key)
    for key in ("numpy", "qibo", "qibocal", "qibolab"):
        update(metadata["versions"], metadata_new["versions"], key)
```

src/qibocal/cli/utils.py
```diff
@@ -0,0 +1,89 @@
...
+
+def update_runcard(rundata, rundata_new, target_compare_dir):
+    """Update runcard.yml file
+
+    Args:
+        rundata (dict): dictionary with the runcard.yml actual parameters and values
+        rundata_new (dict): dictionary with the new parameters and values to update in the actual runcard.yml
+    """
+
+    rundata["platform"] = rundata["platform"] + " , " + rundata_new["platform"]
```
AleCandido, line 78:
```suggestion
    rundata["platform"] += " , " + rundata_new["platform"]
```

AleCandido (commented):
I haven't reviewed the routines, and I'm not going to.

We will pass through them later on, while refactoring something else. Just make sure they are running.

The only other thing I'm missing is `cli.auto_builder`, but this review is already big enough, so I wanted to publish sooner.

src/qibocal/auto/task.py
```diff
@@ -74,28 +70,36 @@ def priority(self):
...
             operation: Routine = self.operation
             parameters = self.parameters
         except RuntimeError:
             operation = dummy_operation
             parameters = DummyPars()
 
-        data: Data = operation.acquisition(parameters)
+        path = self.datapath(folder)
+
+        if operation.platform_dependent and operation.qubits_dependent:
```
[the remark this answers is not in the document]
andrea-pasquale: Platform will not be necessary if you run something at the level of circuits.
The converse I don't know at the moment.
Probably we will need to adapt this mechanism once we actually include the RB here, or general routines at circuit level. Therefore I suggest to leave it as it is know given that it will probably change soon.

src/qibocal/cli/utils.py
```diff
@@ -0,0 +1,89 @@
...
+def folders_exists(folders):
+    """Check if a list of folders exists
+
+    Args:
+        folders (list): list of absolute or relative path to folders
+    """
+    foldernames = []
+    for foldername in folders:
+        expanded = list(glob(foldername))
+        if len(expanded) == 0 and "*" not in foldername:
```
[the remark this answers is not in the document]
andrea-pasquale: @AleCandido this is related to the implementation of `qq-compare` developed by @DavidSarlle, I just moved some of the functions in this file.
I feel like a refactor of this part is out of the scope of this PR. I will still commit your suggestions.
Moreover, `qq-compare` is not supported  in the current version of the autocalibration.

src/qibocal/auto/task.py
```diff
@@ -74,28 +70,36 @@ def priority(self):
...
+        return self.operation.parameters_type.load(self.action.parameters)
 
-    def data(self, base_dir) -> Optional[Data]:
-        if not self.datapath(base_dir).is_file():
-            return None
-
-        Data = self.operation.data_type
+    @property
+    def data(self):
+        return self._data
```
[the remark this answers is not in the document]
Edoardo-Pedicillo: This is a getter of a protected attribute, why should we remove it ?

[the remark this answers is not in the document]
AleCandido: Because of simplicity: there is no point in just protecting it in writing, if people are supposed not to write, they simply should not do it because they do not need, or because there is no better alternative.
If they need to do it and there is no better alternative, just denying it is quite a poor solution...

In particular, `@property` are useful when they are not trivial, and making things *private* for the sake of it is not very Pythonic (it has been said ["we are all adults here"](https://mail.python.org/pipermail/tutor/2003-October/025932.html)).

However, the proposed solution was in the property before the change: use it to load data, and use the "protected" attribute for caching, if you need it. In that case, the property would make a non-trivial task (i.e. loading).
If you only use a property to make something private, please make it public.

src/qibocal/auto/runcard.py
Edoardo-Pedicillo: For the time being, it is a placeholder for other formats, for example in the RB, that I hope we will have in autocalibration soon, they are using different formats.

src/qibocal/cli/builders.py
```diff
@@ -160,8 +160,9 @@ def __init__(self, runcard, folder=None, force=False):
                 self.qubits = self.runcard.get("qubits")
         else:
             self.qubits = []
-        self.format = self.runcard.get("format")
 
+        # Setting format. If None csv is used.
```
AleCandido, line 164: Actually, if *absent*, that is different from being `None`

AleCandido:
As discussed: create an issue for the [`self.data`](https://github.com/qiboteam/qibocal/pull/279#discussion_r1192359341) discussion, and we can merge.

src/qibocal/auto/runcard.py
```diff
@@ -38,6 +39,8 @@ class Runcard:
     """Structure of an execution runcard."""
 
     actions: List[Action]
+    qubits: Optional[List[int]] = Field(default_factory=list)
```
stavros11, line 42: I tested this branch for the 25q chip with QM and the string qubit issue is fixed just by changing this line to
```suggestion
    qubits: Optional[List[Union[int, str]]] = Field(default_factory=list)
```

So far I tested only the resonator spectroscopy and worked after making only this change and merging qiboteam/qibolab#341 to qiboteam/qibolab#349. I will try to test some more routines later today and report if I find any issues.

andrea-pasquale: That's great!
Thanks @stavros11. Let us know if see other issues.

src/qibocal/cli/builders.py
andrea-pasquale: Fixed thanks.

andrea-pasquale:
@AleCandido I've opened the issue #335, as you suggested.
As soon as we merge qiboteam/qibolab#341 I'll change the lock file here to use the latest main or latest release.

scarrazza:
Latest main please.

PR #283, opened 2023-03-20
Remove max and min bias limits from resonator and qubit spectroscopies flux

aorgazf:
Hi, in order to protect the fridge from warming up from the application of strong biassing currents, the spectroscopy routines implemented a limit on the maximum current to be applied with the SPI to 30mA. 
Now that biasing is applied in QM and qblox setting the offset (voltage), this limit no longer works. In the latest experiments with qblox we needed to apply voltages of around 0.5V.

This PR removes those protections that will now need to be implemented in the respective drivers.

Checklist:
- [x] Tests are passing.


andrea-pasquale (commented):
Thanks @aorgazf.
Looks good to me.
I would suggest to add a warning when you changed the value of the bias from the one specified by the user. It could also be useful to document this somewhere (for example in the docstrings of the calibration routine).

andrea-pasquale:
@aorgazf you may want to check qiboteam/qibolab#345 to see if this PR is still necessary

stavros11:
> @aorgazf you may want to check [qiboteam/qibolab#345](https://github.com/qiboteam/qibolab/pull/345) to see if this PR is still necessary

Indeed the idea of qiboteam/qibolab#345 was to raise an error if the bias exceeds the maximum safe value, however this was only for instruments using the `DesignPlatform`, so it will not work for qblox. Also, in order to use this feature in qibolab, you need to set the supported `max_bias` in the channels when defining the platform, eg:
```py
def create_qw5q_gold(...):
    channels = ChannelMap()
    ...
    channels["..."].max_bias = 0.2
    # then an error will be raised if the bias for this channel is set >0.2
```

andrea-pasquale:
> I would suggest to add a warning when you changed the value of the bias from the one specified by the user. It could also be useful to document this somewhere (for example in the docstrings of the calibration routine).

Actually I've misunderstood this PR. Since we are removing the max and min bias limit I think that this could merged as it is.

PR #284, opened 2023-03-20
Remove constrains on data types

aorgazf:
Hi, the current (temporary) solution implemented by @DavidSarlle to load data in plotting functions and bypassing pint, set constrains on the type of data to be loaded from the csv like so:
```
def load_data(folder, subfolder, routine, data_format, name):
    file = f"{folder}/{subfolder}/{routine}/{name}.csv"
    data = DataUnits()
    all_columns = pd.read_csv(file, nrows=1).columns.tolist()
    dtype_dict = {
        col: "float64"
        for col in all_columns
        if col not in ["Unnamed: 0", "qubit", "iteration"]
    }
    if "fit" in name or "parameters" in name:
        data.df = pd.read_csv(file, header=[0], index_col=[0])

    else:
        data.df = pd.read_csv(
            file, dtype=dtype_dict, header=[0], skiprows=[1], index_col=[0]
        )

    return data
```
forcing all column values other than qubit and iteration to be float64. 
Yet, sometimes we had other parameters with other data types, like in resonator spectroscopy flux, where fluxlines could be a string, or the Tag parameter in the cryoscope.

This PR removes these unnecessary constrains. Pandas sets the right type of data type depending on the values read in the columns.

Checklist:
- [x] Tests are passing.


stavros11 (approved):
I fell into the same bug while trying to plot a spectroscopy with flux and I confirm that this is fixing it. @andrea-pasquale if you agree we can merge this.

PR #291, opened 2023-03-27
Update Furo to hash-safe version

AleCandido:
This actually solves compatibility with Poetry modern installer, and so the vanilla 1.4.1 version (without further configurations).


andrea-pasquale (approved):
Thanks @AleCandido.

PR #308, opened 2023-04-22
Fix Classification Parameters closes #305

### LINKED ISSUE #305: Single shot classification wrong for qblox
After some debugging I found out that the `threshold` and the `rotation_angle` computed by the routine [`calibrate_qubit_states` ](https://github.com/qiboteam/qibocal/blob/598e8622c2d9419adc236dd9d28771adaefd5fd9/src/qibocal/calibrations/characterization/calibrate_qubit_states.py#L12)are not the exact parameters used by qblox. Currently in qibocal we are using Ramiro's method which consist in the following:

1. Translate 0s to the origin and 1s accordingly
2. Rotate 1s to x axis
3. Compute the threshold by looking at the cumulative distributions

According to their [documentation](https://qblox-qblox-instruments.readthedocs-hosted.com/en/master/tutorials/conditional_playback.html#Measure-qubit-histogram) qblox are doing the following:

```py
rotation = np.mod(-np.angle(np.mean(state1)-np.mean(state0)), 2*np.pi)
threshold = (np.exp(1j*rotation)*(np.mean(state1)+np.mean(state0))).real/2
```
which I don't fully understand but it is working and it doesn't involve the average of the state 0.
Here is a comparison between the two different methods using random numbers.
image

The final results is the same however they have different values for threshold and angle:
|   | ramiro  | qblox  | 
|---|---|---|
| angle  | -0.45497  | 0.45497  |  
| threshold  | 1.230  | -0.655  |   

As a test I tried to compute the angle and the rotation using qblox method and then I ran allXY which are now working. You can see the report [here](http://login.qrccluster.com:9000/qU70t2QkToSw2YU8vtpQUQ==/).

It would be nice to benchmark also this method @Edoardo-Pedicillo.
If it is faster than what we currently have in main we could use this a default since it is required by qblox.
I'm going to update the runcard with the new classification parameters in a new branch in qibolab so that everyone can use it.
Let me know what you think @stavros11 @scarrazza @Edoardo-Pedicillo @aorgazf @DavidSarlle.

aorgazf:
The PR fixes the bug introduced in the calculation of the rotation angle and threshold of the `calibrate_qubit_states_fit` function.
With this fix, AllXYs work again.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibocal/fitting/methods.py
+++ b/src/qibocal/fitting/methods.py
@@ -1027,9 +1027,7 @@ def calibrate_qubit_states_fit(data, x, y, nshots, qubits, degree=True):
         iq_mean_state0 = np.mean(iq_state0)
         origin = iq_mean_state0
 
-        iq_state0_translated = iq_state0 - origin
-        iq_state1_translated = iq_state1 - origin
-        rotation_angle = np.angle(np.mean(iq_state1_translated))
+        rotation_angle = np.angle(np.mean(iq_state1 - origin))
 
         iq_state1_rotated = iq_state1 * np.exp(-1j * rotation_angle)
         iq_state0_rotated = iq_state0 * np.exp(-1j * rotation_angle)
--- a/src/qibocal/fitting/methods.py
+++ b/src/qibocal/fitting/methods.py
@@ -1025,9 +1025,8 @@ def calibrate_qubit_states_fit(data, x, y, nshots, qubits, degree=True):
 
         iq_mean_state1 = np.mean(iq_state1)
         iq_mean_state0 = np.mean(iq_state0)
-        origin = iq_mean_state0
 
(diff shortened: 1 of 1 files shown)


Edoardo-Pedicillo:
At this point I think you can drop this two lines
https://github.com/qiboteam/qibocal/blob/159c5edcb554e7d53e986a547e2df76b1a42207a/src/qibocal/fitting/methods.py#L1030-L1031

andrea-pasquale:
> At this point I think you can drop this two lines
> 
> https://github.com/qiboteam/qibocal/blob/159c5edcb554e7d53e986a547e2df76b1a42207a/src/qibocal/fitting/methods.py#L1030-L1031

+1

andrea-pasquale (approved):

andrea-pasquale:
I'll go ahead and merge this PR.

PR #313, opened 2023-04-26
Cumulative

Edoardo-Pedicillo:
This PR implements and benchmarks a new function to evaluate the cumulative distribution.
To compare the old method (`method1`) with the new one (`method2`), you can use the following code
```python
import time

import numpy as np

REPETITIONS = 100

def cum_method1(real_values_combined, real_values_state0):
    return  [
            sum(map(lambda x: x.real >= real_value, real_values_state0))
            for real_value in real_values_combined
        ]

def cum_method2(input_data, points):
    # data and points sorted
    input_data = np.sort(input_data)
    points = np.sort(points)
 
    prob = []
    app = 0
    for val in input_data:
        app += np.max(np.searchsorted(points[app::], val), 0)
        prob.append(app)

    return np.array(prob)

def time_compare(method, input1, input2):
    start = time.time()
    for _ in range(REPETITIONS):
       cum =  method(input1, input2)
    end = time.time()
    return (end-start)/REPETITIONS, cum

state0_center = [0.0, 0.0]
state1_center = [2.0, 2.0]
cov = 0.1 * np.eye(2)
size = 100  # print(points, input_data)

data_0 = np.random.multivariate_normal(state0_center, cov, size=size)
data_1 = np.random.multivariate_normal(state1_center, cov, size=size)

all_data = np.concatenate((data_0, data_1), axis = 0)

#method1
time1, cum1 = time_compare(cum_method1,all_data[:,0],data_0[:,0])
print("Method 1: ", time1)

#method2
time2, cum2 = time_compare(cum_method2,all_data[:,0],data_0[:,0])
print("Method 2: ", time2)

#Plots
plt.figure(figsize=(14, 7), dpi=80)
plt.subplot(1,2,1)
plt.scatter(data_0[:, 0], data_0[:, 1], label = "0")
plt.scatter(data_1[:, 0], data_1[:, 1], label = "1")

plt.subplot(1,2,2)
plt.scatter(all_data[:,0],cum1, label = f"method1 ({round(time1, 4)})")
plt.scatter(np.sort(all_data[:, 0]), cum2, label = f"method2 ({round(time2,4)})")
plt.legend()
plt.show()
```
From the outputs we can see: 

1. the new method is four time faster
2. the old method has a bug ( we have to change the condition `x.real >= real_value` in `x.real <= real_value`)
```
Method 1:  0.0026085638999938966
Method 2:  0.0008060479164123535
```
image
With the change proposed in point 2 this is the result
image

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibocal/fitting/methods.py
+++ b/src/qibocal/fitting/methods.py
@@ -9,6 +9,7 @@ from qibocal.config import log
 from qibocal.data import Data
 from qibocal.fitting.utils import (
     cos,
+    cumulative,
     exp,
     flipping,
     freq_q_mathieu,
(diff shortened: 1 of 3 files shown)


andrea-pasquale (approved):

andrea-pasquale:
@Edoardo-Pedicillo could we add a quick test for the cumulative?

Edoardo-Pedicillo:
Done

PR #318, opened 2023-04-29
NIGSC changes

vodovozovaliza:
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibocal/calibrations/niGSC/XIdrb.py
+++ b/src/qibocal/calibrations/niGSC/XIdrb.py
@@ -49,13 +49,9 @@ class ModuleFactory(CircuitFactory):
     def build_circuit(self, depth: int):
         # Initiate the empty circuit from qibo with 'self.nqubits'
         # many qubits.
-        circuit = Circuit(1, density_matrix=True)
-        # There are only two gates to choose from for every qubit.
-        a = [gates.I(0), gates.X(0)]
-        # Draw sequence length many zeros and ones.
-        random_ints = np.random.randint(0, 2, size=depth)
-        # Get the Xs and Ids with random_ints as indices.
-        gate_lists = np.take(a, random_ints)
+        circuit = Circuit(1)
+        # Draw sequence length many I and X gates.
+        gate_lists = np.random.choice([gates.I(0), gates.X(0)], size=depth)
         # Add gates to circuit.
         circuit.add(gate_lists)
         circuit.add(gates.M(0))
@@ -72,18 +68,20 @@ class ModuleExperiment(Experiment):
         noise_model: NoiseModel = None,
     ) -> None:
         super().__init__(circuitfactory, data, nshots, noise_model)
+        # Make the circuitfactory a list be able to store them using save_circuits method.
+        self.prebuild()
         self.name = "XIdRB"
 
     def execute(self, circuit: Circuit, datarow: dict) -> dict:
         datadict = super().execute(circuit, datarow)
         datadict["depth"] = circuit.ngates - 1
-        # TODO change that.
-        datadict["countX"] = circuit.gate_types["x"]
+        # TODO change to circuit.gate_types["x"] for next Qibo version
+        datadict["countX"] = len(circuit.gates_of_type("x"))
         return datadict
 
 
 # Define the result class for this specific module.
-class moduleReport(Report):
+class ModuleReport(Report):
     def __init__(self) -> None:
         super().__init__()
         self.title = "X-Id Benchmarking"
@@ -137,13 +135,14 @@ def post_processing_sequential(experiment: Experiment):
 # After the row by row execution of tasks comes the aggregational task. Something like calculation
 # of means, deviations, fitting data, tasks where the whole data as to be looked at, and not just
 # one instance of circuit + other information.
-def get_aggregational_data(experiment: Experiment) -> pd.DataFrame:
+def get_aggregational_data(experiment: Experiment, ndecays: int = 2) -> pd.DataFrame:
     """Computes aggregational tasks, fits data and stores the results in a data frame.
 
     No data is manipulated in the ``experiment`` object.
 
     Args:
         experiment (Experiment): After sequential postprocessing of the experiment data.
+        ndecays (int): Number of decays to be fitted. Default is 2.
 
     Returns:
         pd.DataFrame: The summarized data.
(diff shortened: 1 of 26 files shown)


src/qibocal/calibrations/niGSC/basics/experiment.py
```diff
@@ -107,24 +107,37 @@ def load(cls, path: str) -> Experiment:
         obj = cls(circuitfactory, data=data, nshots=nshots)
         return obj
 
-    def save(self, path: str | None = None, force=False) -> str:
-        """Creates a path if None given and pickles relevant data from ``self.data``
-        and if ``self.circuitfactory`` is a list that one too.
+    def save_circuits(self, path: str | None = None, force: bool = False) -> str:
+        """Creates a path if None given and pickles ``self.circuitfactory``
+        if its a list.
```
andrea-pasquale, line 112:
```suggestion
        if it is a list.
```

tests/test_niGSC_circuitfactory.py
```diff
@@ -41,7 +39,7 @@ def test_abstract_factory():
     cfactory = CircuitFactory(1, [1, 2] * 3, qubits=[0])
     with pytest.raises(NotImplementedError):
         for circuit in cfactory:
-            print(circuit.draw())
+            circuit.draw()
```
andrea-pasquale, line 42: This could probably be removed I think

src/qibocal/calibrations/niGSC/basics/circuitfactory.py
```diff
@@ -53,7 +53,9 @@ def __next__(self) -> Circuit:
             circuit = self.build_circuit(self.depths[self.n])
             self.n += 1
             # Distribute the circuit onto the given support.
-            bigcircuit = Circuit(self.nqubits)
+            circuit_init_kwargs = circuit.init_kwargs
+            del circuit_init_kwargs["nqubits"]
```
andrea-pasquale, line 57: Is this necessary (?)

andrea-pasquale (commented):
Thanks @vodovozovaliza for implementing this.
I started with a quick a review even if this is still in draft.
I agree with the changes proposed by this PR.
You can find below a few comments.

src/qibocal/calibrations/niGSC/basics/circuitfactory.py
vodovozovaliza: This code is needed to keep the initial parameters of the circuit (such as `density_matrix`), while embedding it into a bigger circuit with new `nqubits`.

andrea-pasquale: I understand. However I don't understand if the variable `circuit_init_kwargs` is needed, since at line 58 you use directly the circuit attribute (`circuit.init_kwargs`) instead of  `circuit_init_kwargs`.
I think that if you drop lines 56-57 the code should work as expected :)

vodovozovaliza:
@andrea-pasquale @wilkensJ this is ready for review

src/qibocal/calibrations/niGSC/XIdrb.py
```diff
@@ -137,13 +134,14 @@ def post_processing_sequential(experiment: Experiment):
 # After the row by row execution of tasks comes the aggregational task. Something like calculation
 # of means, deviations, fitting data, tasks where the whole data as to be looked at, and not just
 # one instance of circuit + other information.
-def get_aggregational_data(experiment: Experiment) -> pd.DataFrame:
+def get_aggregational_data(experiment: Experiment, ndecays: int = 2) -> pd.DataFrame:
```
wilkensJ, line 137: How can this be changed through a runcard?

vodovozovaliza: This parameter is not used in the runcards for now, but can be used to redo the fitting of the same experiment

src/qibocal/calibrations/niGSC/basics/experiment.py
```diff
@@ -39,33 +39,25 @@ def __init__(
         if circuitfactory is not None and not isinstance(circuitfactory, Iterable):
             raise_error(
                 TypeError,
-                "given circuit factory has wrong type {}, must be Iterable | None.".format(
-                    type(circuitfactory)
-                ),
+                f"given circuit factory has wrong type {type(circuitfactory)}, must be Iterable | None.",
```
andrea-pasquale, line 42:
```suggestion
                f"Invalid type {type(circuitfactory)} for CircuitFactory,  must be Iterable or None.",
```

src/qibocal/calibrations/niGSC/basics/experiment.py
```diff
@@ -39,33 +39,25 @@ def __init__(
...
             )
         self.circuitfactory = circuitfactory
         if data is not None and not isinstance(data, Iterable):
             raise_error(
                 TypeError,
-                "given data has wrong type {}, must be Iterable | None ".format(
-                    type(data)
-                ),
+                f"given data has wrong type {type(data)}, must be Iterable | None ",
```
andrea-pasquale, line 48:
```suggestion
                f"Invalid data type  {type(data)}. Data must be Iterable or None.",
```

src/qibocal/calibrations/niGSC/basics/experiment.py
```diff
@@ -39,33 +39,25 @@ def __init__(
...
             )
         self.data = data
         if nshots is not None and not isinstance(nshots, int):
             raise_error(
                 TypeError,
-                "given nshots has wrong type {}, must be int | None".format(
-                    type(nshots)
-                ),
+                f"given nshots has wrong type {type(nshots)}, must be int | None",
```
andrea-pasquale, line 54:
```suggestion
                f"Invalid nshots type {type(nshots)}. Nshots must be int or None.",
```

src/qibocal/calibrations/niGSC/basics/experiment.py
```diff
@@ -39,33 +39,25 @@ def __init__(
...
             )
         self.nshots = nshots
         if noise_model is not None and not isinstance(noise_model, NoiseModel):
             raise_error(
                 TypeError,
-                "given circuit factory has wrong type {}, must be qibo NoiseModel | None .".format(
-                    type(noise_model)
-                ),
+                f"given circuit factory has wrong type {type(noise_model)}, must be qibo NoiseModel | None .",
```
andrea-pasquale, line 60:
```suggestion
                f"NoiseModel has wrong type {type(noise_model)}. NoiseModel must be qibo NoiseModel or None .",
```

src/qibocal/calibrations/niGSC/basics/plot.py
```diff
@@ -86,7 +80,12 @@ def build(self):
             width=1000,
         )
 
-        return fig
+        return fig, self.info_table()
+
+    def info_table(self):
+        return "".join(
+            [f"q{0}/r{0} | {key}: {value}<br>" for key, value in self.info_dict.items()]
```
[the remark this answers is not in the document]
andrea-pasquale: I agree. In the notation q{i}/r{j} i is the qubit number, while j is the report number (used for `qq-compare`). I don't think that the niGSC is compatible with `qq-compare` so probably it doesn't matter for now.

src/qibocal/web/static/styles.css
andrea-pasquale, line 1: This modifications also affect the report generated by the low level routines.
Moving the table to the left is not ideal for the routine which include more than one plot.
I believe this is out of the scope of this PR, we could discuss about the report generation in a separate PR/issue.

andrea-pasquale (commented):
Thanks @vodovozovaliza.
You can find below a few comments.
I think that we should be able to merge this PR today.

src/qibocal/web/static/styles.css
wilkensJ: I tried it with the routines from runcard/actions_qq.yml (a simulation) and for me it looked fine, the only change here is that it is not centered anymore and there is a bit of more whitespace between the columns.
It would be nice if you could try to generate reports with the new css settings and if something is not readable anymore I totally agree to not implement that.
Otherwise, when the table has only a 30% width our fitting parameters are in two lines and that does influence how they are read, also, it takes a lot of space, so if there are more than lets say 5 lines the breaking does make it harder to read.

andrea-pasquale: I have a problem only with the position of the table.
If you run for example t1 with dummy it looks like this:
image
which personally I don't like. As you said centered looks better.
I also understand that the current rendering may not best for the RB protocols.
For single plot you could try to stretch out the plot like we do for the T1 above.
For multiple plots I think that the table in the center should look ok.
But, again we could fix this in a separate PR and we should also ask the opinion of other people.

wilkensJ: I agree! This looks weird. I like the idea of putting it in another PR :)

wilkensJ (approved):

wilkensJ:
For me this would be fine to merge with the `simply_builder` now :)

andrea-pasquale (approved):
LGTM, thanks @vodovozovaliza.

PR #326, opened 2023-05-08
Add execution parameters and new results to autocalibration routines

Jacfomg:
- [x] execution parameters
- [x] new results
- [x] Update routines involving delays and duration to be sweeped tested on #277 (Non detuned Ramsey)

Checklist:
- [ ] Tests are passing.


src/qibocal/protocols/characterization/classification.py
```diff
@@ -95,12 +96,17 @@ def _acquisition(
 
     # execute the first pulse sequence
     state0_results = platform.execute_pulse_sequence(
-        state0_sequence, nshots=params.nshots
+        state0_sequence,
+        ExecutionParameters(
+            nshots=params.nshots,
+            relaxation_time=params.relaxation_time,
+            acquisition_type=AcquisitionType.INTEGRATION,
+        ),
```
maxhant, line 104:
```suggestion
        ExecutionParameters(
            nshots=params.nshots,
            acquisition_type=AcquisitionType.INTEGRATION,
        ),
```

src/qibocal/protocols/characterization/classification.py
```diff
@@ -111,12 +117,15 @@ def _acquisition(
 
     # execute the second pulse sequence
     state1_results = platform.execute_pulse_sequence(
-        state1_sequence, nshots=params.nshots
+        state1_sequence,
+        nshots=params.nshots,
+        relaxation_time=params.relaxation_time,
+        acquisition_type=AcquisitionType.INTEGRATION,
```
maxhant, line 123:
```suggestion
  ExecutionParameters(
              nshots=params.nshots,
              acquisition_type=AcquisitionType.INTEGRATION,
          ),
```

src/qibocal/protocols/characterization/allxy/allxy.py
```diff
@@ -88,11 +89,18 @@ def _acquisition(
...
 
         # execute the pulse sequence
-        results = platform.execute_pulse_sequence(sequence)
+        results = platform.execute_pulse_sequence(
+            sequence,
+            ExecutionParameters(
+                nshots=params.nshots,
+                relaxation_time=params.relaxation_time,
+                averaging_mode=AveragingMode.CYCLIC,
+            ),
```
maxhant, line 98:
```suggestion
            ExecutionParameters(
                averaging_mode=AveragingMode.CYCLIC,
            ),
```

maxhant (changes requested):

src/qibocal/protocols/characterization/classification.py
Jacfomg: Fixed the routine to take relaxation time

src/qibocal/protocols/characterization/allxy/allxy.py
Jacfomg: same as above, fixed ExecutionParameters missing

src/qibocal/protocols/characterization/classification.py
Jacfomg: same as above, fixed ExecutionParameters missing

andrea-pasquale:
@Jacfomg I will review this PR soon.
In the meantime could you do the following please:

1. merge main
2. update the lock file to point to the qiboteam/qibolab#400 so that we can see if tests are passing here or not.
Here is an example of what I did in #342
https://github.com/qiboteam/qibocal/blob/ef23eedbfe471c302e219fd4d2fe456a4ec5a6f6/pyproject.toml#L19
in your case you just need to replace the branch with `execute_branch` then run `poetry lock` and push both the pyproject.toml and the poetry.lock file.
If you have any problem with that I can also do it.

Jacfomg:
It seems like all the tests are failing because of pint or something

andrea-pasquale:
This is also what I saw locally. I had a chat with @Edoardo-Pedicillo   yesterday about it. We will open a PR to fix this issue.

Jacfomg:
Tests are a bit slow I may did something not right on the dummy, maybe revise it

src/qibocal/protocols/characterization/allxy/allxy.py
```diff
@@ -88,11 +93,17 @@ def _acquisition(
             )
 
         # execute the pulse sequence
-        results = platform.execute_pulse_sequence(sequence)
+        results = platform.execute_pulse_sequence(
+            sequence,
+            ExecutionParameters(
+                nshots=params.nshots,
+                averaging_mode=AveragingMode.CYCLIC,
```
rodolfocarobene, line 100:
```suggestion
                averaging_mode=AveragingMode.SINGLESHOT,
```
This should be singleshot right? Otherwise probabilities do not really make sense (also for other allxy routines)

Jacfomg: You will get a value between 0 and 1 from the averaging. I don't see a problem with that.

rodolfocarobene: Here we are averaging and classifying, maybe the problem is that I first average and than classify while you first classify and then average...
But then in the StateResult there will be something that is not 0-1 (if you average after classification) and I thought that this is not wanted

Jacfomg: Yes, I do first classification and the average. And also, the AveragedStateResults are expected not to be 0 or 1. Just a value in between so the results are the same as the old AllXY

tests/runcards/single_routines.yml
```diff
@@ -156,6 +155,8 @@ actions:
       delay_between_pulses_start: 16
       delay_between_pulses_end: 20000
       delay_between_pulses_step: 100
+      nshots: 1024
+      relaxation_time:
```
andrea-pasquale, line 159: Could we make these arguments optional?

andrea-pasquale (approved):

PR #333, opened 2023-05-16
Add lenght and delay sweepers to autocalibration protocols

Jacfomg:
Routines I need to run things properly with the sweepers I have on Zurich with updated results and ExecutionParameters

Checklist:
- [ ] Tests are passing.


rodolfocarobene:
At the moment the driver does not support time sweepers... We could merge this PR naming things like "t1_sweepers"

Jacfomg:
I agree, that was my first idea, but wasn't sure if the code duplication was right

andrea-pasquale (commented):
Thanks for implementing this @Jacfomg.
Generally it looks good to me.
I still need to test it with Zurich.
Let's try to reduce code duplication by importing from the original routine all the functions/classes that you will not modify.
In the meantime could you please merge main? Tests should pass now.

src/qibocal/protocols/characterization/ramsey_sequences.py
```diff
@@ -0,0 +1,98 @@
...
+    # DataUnits stores by default MSR, phase, i, q
+    # additionally include wait time and t_max
+    data = RamseyData(params.n_osc, params.delay_between_pulses_end, detuning_sign=+1)
+
+    # sweep the parameter
+    for wait in waits:
+        for qubit in qubits:
+            RX90_pulses2[qubit].start = RX90_pulses1[qubit].finish + wait
+            ro_pulses[qubit].start = RX90_pulses2[qubit].finish
+            if params.n_osc != 0:
```
andrea-pasquale, line 62: Why did you put this condition? I can understand that the sign of the correction is the correct one only if we apply detuning. I think that a more elegant solution could be to define this condition in the `__post_init__` of `RamseyParameters`. I would also suggest to raise an error if the user put `n_osc == 0`  telling the user to use the `T2` experiments instead of `Ramsey`. Let me know if you agree.

Jacfomg: No idea, this is the old routine I copied from main.

andrea-pasquale: I see. I would say that it makes sense to put this constraint.
If you want I can push directly here what I suggested in the comment above.

Jacfomg: My only question is does the T2 experiment update the runcard as well ?

andrea-pasquale: If you want an estimate of T2 you should run a T2 experiment directly after merging

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -119,43 +120,72 @@ def _acquisition(
...
+                        "qubit": qubit,
+                    }
                 )
+                data.add_data_from_dict(r)
+
+    else:
+        sweeper = Sweeper(
+            Parameter.delay,
+            waits,
+            [RX90_pulses1[qubit] for qubit in qubits],
```
stavros11, line 163: I believe this should be 
```suggestion
            [RX90_pulses2[qubit] for qubit in qubits],
```
We need to sweep the delay between the two drive pulses, so we sweep the start of the second pulse.

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -188,6 +218,8 @@ def _fit(data: RamseyData) -> RamseyResults:
         times = qubit_data_df["wait"].pint.to("ns").pint.magnitude
         qubit_freq = qubit_data_df["qubit_freqs"].pint.to("Hz").pint.magnitude.unique()
 
+        print("qubit_freq", qubit_freq)
```
stavros11, line 221: This print and the one below are probably not needed.

src/qibocal/protocols/characterization/ramsey_sequences.py
andrea-pasquale: The problem with the frequency correction with `n_osc = 0` is that we are not sure that the sign of the corrections is the correct one. Our implementation of the frequency corrections works only in the detuned regime. This is why I suggest to not have a `n_osc = 0` case for the ramsey experiment.

andrea-pasquale: After our discussion in the qibo meeting I think that we can resolve the conversation.
I was probably overthinking it...

andrea-pasquale (commented):
Generally it looks good to me.
As soon as you fix conflicts and address @stavros11's comments we can merge.

src/qibocal/protocols/characterization/ramsey.py
Jacfomg: I dont think so, for me this definition is way natural as the readout section can't be on its own delayed more than 2us and need to get linked to a drive or flux section with that delay. And also you will not delay the first pulse never with the other approach and that seemed odd to me.

Jacfomg: For me I just understand each pulse as two things.

1. Play the wave.
2. Delay for the next.

So the delay in this case should be part of the 1st pulse. That way when I get a sweeper I get both elements instead of having to check the last element of the sequence for the sweeper.values to apply the delay and the next to know the pulse on my loop. Maybe It's just the way I coded the driver but it seemed more straigtforward.

Jacfomg: Done

src/qibocal/protocols/characterization/rabi/length_sequences.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+
+    # create a sequence of pulses for the experiment
+    sequence = PulseSequence()
+    qd_pulses = {}
+    ro_pulses = {}
+    for qubit in qubits:
+        # TODO: made duration optional for qd pulse?
+        qd_pulses[qubit] = platform.create_qubit_drive_pulse(qubit, start=0, duration=4)
+        qd_pulses[qubit].amplitude = params.pulse_amplitude
```
stavros11, line 30: Not directly related to this PR, as it is the same in the routine on main, but personally I would prefer to be able to pass `params.pulse_amplitude` as `None` in the action runcard and use the default drive pulse amplitudes defined in the platform runcard. The main reason for this is that each qubit usually has its own pi-pulse amplitude, so it doesn't make sense to use the same value for all qubits. @andrea-pasquale what do you think?

stavros11 (commented):
Thank you for the updates, I believe now it should work for all other drivers. If you can fix the conflicts, I believe this is ready to merge.

src/qibocal/protocols/characterization/rabi/length_sequences.py
andrea-pasquale: Thanks @stavros11.
We are already implementing this mechanism in other protocols.
https://github.com/qiboteam/qibocal/blob/cae85be16cfdb92558e212b2060df253a53a4a77/src/qibocal/protocols/characterization/resonator_spectroscopy.py#L107-L108
Feel free to open a PR with your suggestion :)

stavros11: Thanks, that's the same with what I did for the rabi length. I can open a PR but I would wait for @Jacfomg to finalize the current PR first (or we could even just commit this fix directly here).

Jacfomg:
This will require https://github.com/qiboteam/qibolab/pull/484. But now that the sweepers are start and the routines were just naming delays to all the parameters do you feel like changing them as well ?

stavros11 (approved):
> This will require [qiboteam/qibolab#484](https://github.com/qiboteam/qibolab/pull/484). But now that the sweepers are start and the routines were just naming delays to all the parameters do you feel like changing them as well ?

For me this is good as it is but I am also fine if others prefer to change. I think we now need to merge qiboteam/qibolab#484 for tests to pass here.

Jacfomg:
> > This will require [qiboteam/qibolab#484](https://github.com/qiboteam/qibolab/pull/484). But now that the sweepers are start and the routines were just naming delays to all the parameters do you feel like changing them as well ?
> 
> For me this is good as it is but I am also fine if others prefer to change. I think we now need to merge [qiboteam/qibolab#484](https://github.com/qiboteam/qibolab/pull/484) for tests to pass here.

Yup, they passed before so I don't expect any problem here.

src/qibocal/protocols/characterization/coherence/t1.py
```diff
@@ -7,6 +7,7 @@
 from qibolab.platform import Platform
 from qibolab.pulses import PulseSequence
 from qibolab.qubits import QubitId
+from qibolab.sweeper import Parameter, Sweeper
```
rodolfocarobene, line 10:
```suggestion
from qibolab.sweeper import Parameter, Sweeper, SweeperType
```
We should define the sweepers with explicit type for more clarity

src/qibocal/protocols/characterization/coherence/t1.py
```diff
@@ -91,35 +92,36 @@ def _acquisition(params: T1Parameters, platform: Platform, qubits: Qubits) -> T1
         params.delay_before_readout_step,
     )
 
+    sweeper = Sweeper(
+        Parameter.start,
+        ro_wait_range,
+        [ro_pulses[qubit] for qubit in qubits],
```
rodolfocarobene, line 98:
```suggestion
        [ro_pulses[qubit] for qubit in qubits],
        type=SweeperType.ABSOLUTE
```
this more line does not change anything since absolute is already standard, but I think that helps avoiding confusion.

(I think this should be added also in the other sweepers)

src/qibocal/protocols/characterization/coherence/t1.py
Jacfomg: Yeah, but I am using the default value, right ?

src/qibocal/protocols/characterization/coherence/t1.py
Jacfomg: Ok, for clarity it may be helpful

src/qibocal/protocols/characterization/rabi/length.py
```diff
@@ -80,40 +82,42 @@ def _acquisition(
         params.pulse_duration_step,
     )
 
+    sweeper = Sweeper(
+        Parameter.duration,
+        qd_pulse_duration_range,
+        [qd_pulses[qubit] for qubit in qubits],
```
rodolfocarobene, line 88:
```suggestion
        [qd_pulses[qubit] for qubit in qubits],
        type=SweeperType.ABSOLUTE,
```

rodolfocarobene (approved):

Issue #337, opened 2023-05-17
Current Ramsey fitting bug


DavidSarlle:
@andrea-pasquale @DaniloKolesnyk @Edoardo-Pedicillo @aorgazf  I have being testing the Ramsey fitting and to me seems that the delta_phys calculated in the fitting should be subtracted in qubit_frequency instead of added. 

![imagen](https://github.com/qiboteam/qibocal/assets/39095664/bd5751bf-9ba2-4cea-962d-5ef58770b57c)

If you check the reports below, you will see that, after running multiple ramsey, if the delta_phys is subtracted, the delta_phys tends to decrease. If the delta_phys is added, it tends to increase in each iteration.

(+ delta_phys)
[multiple ramsey detuned (delta +).pdf](https://github.com/qiboteam/qibocal/files/11495292/multiple.ramsey.detuned.delta.%2B.pdf)

(-delta_phys)
[multiple ramsey detuned (delta -).pdf](https://github.com/qiboteam/qibocal/files/11495298/multiple.ramsey.detuned.delta.-.pdf)

DavidSarlle:
> 

@igres26 I did not detune the qubit frequency manually. The qubit frequency used was the one obtained from the qubit spec.
The number of oscilations was 10. Do we have to check the number of oscilations introduced by the user in order to be sure that the induced change of frequency from the oscillations is larger than the detuning due to miscalibration?

igres26:
I see what is wrong now. The `delta_phys` is computed with the right sign under the conditions I stated. But then the `corrected_qubit_frequency` should subtract the `delta_phys` to recover the real frequency.

DavidSarlle:
> I see what is wrong now. The `delta_phys` is computed with the right sign under the conditions I stated. But then the `corrected_qubit_frequency` should subtract the `delta_phys` to recover the real frequency.

That was exactly what I was suggesting, because the reports are quite clear. If we substract the delta_phys to recover the real frequency, the next Ramsey iterations are reducing the physical detunning.
I was not talking about the sign when computing the delta phys, but about the sign when reconstructing the qubit freq.

igres26:
Yep. You are right. I asked @andrea-pasquale to re-check the tests he did, since they shouldn't have worked in the first place to make sure we are not missing anything. But the found Delta should always be subtracted to recover the real frequency.

andrea-pasquale:
[...]

PR #339, opened 2023-05-18
Simplify Ramsey code

andrea-pasquale:
Simplify the Ramsey code as suggested by @igres26 following #337.
This PR will not fix the bug for qblox as it is related to a problem with the relative phase.
I tried to test it this morning with qblox but I think that something is not working.
@Jacfomg could be please rerun quickly the sanity check that we ran yesterday using iqm?

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibocal/protocols/characterization/ramsey.py
+++ b/src/qibocal/protocols/characterization/ramsey.py
@@ -45,7 +45,7 @@ class RamseyResults(Results):
 class RamseyData(DataUnits):
     """Ramsey acquisition outputs."""
 
-    def __init__(self, n_osc, t_max):
+    def __init__(self, n_osc, t_max, detuning_sign):
         super().__init__(
             name="data",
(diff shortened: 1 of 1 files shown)


Jacfomg:
Yes, I will do that. But also, we can take another look into why you could not manage to run things yesterday

andrea-pasquale:
I think that it was because I had messed things up with conda.
I will give it another try.

Jacfomg:
This is with this code, @igres26 I also substracted .5 MHz from the qubit frequency and now you see the delta being negative but the correction is still working as intended as we substract it as well

image

igres26 (approved):
Looks good. We could make it more clear that the sign in the `(-2*np.pi)` is the same as the `-int((delta_fitting...`. Maybe in the comments. But this works.

Jacfomg (approved):

andrea-pasquale:
Tested and it works both with qblox (commenting a single line) and with RFSoC

PR #343, opened 2023-05-18
Standard RB as a protocol for ExecutionParameters

Jacfomg:
Only experimental implementation for how things will look in qibolab after #400.  Thanks @vodovozovaliza for the code and help to implement. Feel free to use it as a template or change things but for now I needed some RB running on the device.

Example:

image

Checklist:
- [ ] Sequence unrolling implementation
- [ ] Tests are passing.


wilkensJ:
Hello @Jacfomg and @vodovozovaliza, thank you for implementing that! It looks like everything is working for Zurich instruments only, right? And only if `ExecutionParamters` are available?

Also, it seems that if the (compiler PR)[https://github.com/qiboteam/qibolab/pull/367] of @stavros11  is merged, the compiler rules you hard coded could be added as compiler rules.

I would suggest to make this a draft, since it will not be merged soon to main. I would love to sit together and think of a way how to incorporate this into qibocal main branch. I think there are many good ideas in the code.

Jacfomg:
ExecutionParameters are ready to be merged today as most of the drivers were updated. That means this will run on all of the updated ones.

Yes, I could take a look at the compiler rules but this was also a prototype to see if everything worked on its own as we found some problems with @vodovozovaliza on the RY, maybe they were related to the different way qblox handles phases opposed to Zurich, Qm and FPGAs. So we should find a way to address this. For now the easy solution is to ignore qblox as it the only driver that was not updated and will not be able to run on main.

I would say it could be merged as soon as the RB team approves it even today.

andrea-pasquale:
@Jacfomg @wilkensJ what should we do with this PR? It's starting to become outdated. 
Is there any change we can recover this implementation with the new rb that is currently in main.

Jacfomg:
> @Jacfomg @wilkensJ what should we do with this PR? It's starting to become outdated. Is there any change we can recover this implementation with the new rb that is currently in main.

I would leave it around until we test the other on hardware, and this approach is easier for me if I want to play with the sequence unrolling

wilkensJ:
Sounds good to me, keep it a draft until for example the unrolling is established in qibolab? Then this can happened in the background, or is this a specific qibocal feature?

Jacfomg:
> Sounds good to me, keep it a draft until for example the unrolling is established in qibolab? Then this can happened in the background, or is this a specific qibocal feature?

I expect something to change on qibocal or even qibo as well, for sure circuit.execute() can't work now with unrolling as it is now.

andrea-pasquale:
@Jacfomg shall we close this now given that we have also #398?

Jacfomg:
We can close it, I will still keep it as I want to reproduce the same RB results I got with #349. For the #398 I don't expect that I will be able to run it soon as I forgot that I was assuming the unrolling to be a long pulse sequence not a list so I need to rewrite the unrolling again for the driver as that was done before they decided to go for that.

andrea-pasquale:
Ok, fine we can keep it. Since I saw some similarities with #398 I thought that we could remove it.
Lets keep it for now.

Jacfomg:
The RB with circuits seems to work on hardware as expected so we can close this and keep #398 for the unrolling if we go for that.

PR #341, opened 2023-05-18
fixed update qubit frequency after Ramsey

DavidSarlle:
This PR fixes issue #337 returning only delta_phys in qibocal and updating the qubit freq accordingly in qibolab (https://github.com/qiboteam/qibolab/pull/426)

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.

Issue #336, opened 2023-05-16
Missing features for new layout


andrea-pasquale:
I've opened this issue to collect a list of missing features for the new layout of qibocal (autocalibration).

- [x] make parameter update mechanism optional
- [ ] improve lorentzian fit
- [x] improve flipping fit
- [x] qubits argument (see first presentation of new layout) #351
- [ ] restore qq-compare #552 
- [x] optional input parameters with default value from runcard #596 
- [x] change plotting mechanism ( I guess this could be considered completed by #546) 
- [x] Remove old routines and rename `qq-auto` in `qq`
- [x] Port NIGSC to new layout
- [x] Add documentation for new layout #485
- [ ] pass parameters of previous steps to later ones (already mentioned in #265)
- [ ] add optimization loop #497 
- [x] enable protocols that only have a fitting/execute without data acquisition and fitting separated
- [x] generate report after each routine #438 
- [x] fix type hints #345
- [x] implement mechanism to switch between voltages and probabilities easily (duplicating only acquisition for now)

@aorgazf @DavidSarlle @maxhant @ingoroth @wilkensJ @vodovozovaliza @igres26 @Edoardo-Pedicillo @AleCandido @stavros11 @scarrazza feel free to drop any suggestions here.

wilkensJ:
Thank you Andrea, that list looks great. I have one suggestion, but it is more like an overall suggestion. 
I think the result objects you introduced are a very powerful tool to implement many features, for example: 
- Make post processing repeatable (meaning repeat a fit with other initial values, with another fitting function, even use a different post processing scheme, ...)
- Display the result differently, depending on the user (for example for qq an html string is returned, but if I want to look at the result from a different angle I can just import the result in a script/jupyter notebook and play around with it)

Also, I don't know what the status concerning the optimization loops is, but that also would be great, @vodovozovaliza already coded something in that direction.

andrea-pasquale:
Thanks @wilkensJ.

> * Display the result differently, depending on the user (for example for qq an html string is returned, but if I want to look at the result from a different angle I can just import the result in a script/jupyter notebook and play around with it)

I had something similar in mind when I put the entry `change plotting mechanism`. I will expand it a bit more.


> * Make post processing repeatable (meaning repeat a fit with other initial values, with another fitting function, even use a different post processing scheme, ...)

[...]

PR #345, opened 2023-05-19
Fixing type hints

andrea-pasquale:
This PR addressing this point from #336.
> - [ ] fix type hints https://github.com/qiboteam/qibocal/pull/341#discussion_r1197895823_

At this point I could also define a type `QubitId = Union[str,int]`, I don't know whether this will be defined in qiboteam/qibolab#387 or if in the future we will support only strings.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibocal/protocols/characterization/classification.py
+++ b/src/qibocal/protocols/characterization/classification.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass, field
-from typing import Dict, List, Tuple
+from typing import Dict, List, Tuple, Union
 
 import numpy as np
 import plotly.graph_objects as go
@@ -34,16 +34,22 @@ class SingleShotClassificationData(DataUnits):
 
 @dataclass
 class SingleShotClassificationResults(Results):
-    threshold: Dict[List[Tuple], str] = field(metadata=dict(update="threshold"))
-    rotation_angle: Dict[List[Tuple], str] = field(metadata=dict(update="iq_angle"))
-    mean_gnd_states: Dict[List[Tuple], str] = field(
+    """SingleShotClassification outputs."""
+
+    threshold: Dict[Union[str, int], float] = field(metadata=dict(update="threshold"))
+    """Threshold for classification."""
+    rotation_angle: Dict[Union[str, int], float] = field(
+        metadata=dict(update="iq_angle")
+    )
+    """Threshold for classification."""
+    mean_gnd_states: Dict[Union[str, int], complex] = field(
         metadata=dict(update="mean_gnd_states")
     )
-    mean_exc_states: Dict[List[Tuple], str] = field(
+    mean_exc_states: Dict[Union[str, int], complex] = field(
         metadata=dict(update="mean_exc_states")
     )
-    fidelity: Dict[List[Tuple], str]
-    assignment_fidelity: Dict[List[Tuple], str]
+    fidelity: Dict[Union[str, int], float]
+    assignment_fidelity: Dict[Union[str, int], float]
 
 
 def _acquisition(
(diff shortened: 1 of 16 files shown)


src/qibocal/protocols/characterization/classification.py
```diff
@@ -34,16 +34,22 @@ def nshots(self):
...
+    """Threshold for classification."""
+    rotation_angle: Dict[Union[str, int], float] = field(
+        metadata=dict(update="iq_angle")
+    )
+    """Threshold for classification."""
+    mean_gnd_states: Dict[Union[str, int], complex] = field(
         metadata=dict(update="mean_gnd_states")
     )
-    mean_exc_states: Dict[List[Tuple], str] = field(
+    mean_exc_states: Dict[Union[str, int], complex] = field(
```
stavros11, line 48: I understand that this, and all other changes here would be replaced by
```suggestion
    mean_exc_states: Dict[QubitId, complex] = field(
```
Does it make sense to have
```suggestion
    mean_exc_states: Dict[Qubit, complex] = field(
```
instead? This way we would avoid creating a new type altogether. I undrestand this may not work as expected with the `MultiqubitPlatform` (but maybe it will), but it would be good to design these things with the future in mind.

stavros11 (commented):
> At this point I could also define a type `QubitId = Union[str,int]`, I don't know whether this will be defined in [qiboteam/qibolab#387](https://github.com/qiboteam/qibolab/pull/387) or if in the future we will support only strings.

Yes, I agree that defining the `QubitId` type in qibolab makes sense. It will probably make maintenance easier too, if in the future we support only strings (or only ints) we just change `QubitId = str` in one place in qibolab instead of having to update all routines here. If you want I can do it in [qiboteam/qibolab#387](https://github.com/qiboteam/qibolab/pull/387).

src/qibocal/protocols/characterization/classification.py
andrea-pasquale: > Yes, I agree that defining the `QubitId` type in qibolab makes sense. It will probably make maintenance easier too, if in the future we support only strings (or only ints) we just change `QubitId = str` in one place in qibolab instead of having to update all routines here. If you want I can do it in [qiboteam/qibolab#387](https://github.com/qiboteam/qibolab/pull/387).

Sure, if you agree I believe that we can implement it.

I think that the second approach will not work.
If you try something like this:
```py
from qibolab.platforms.abstract import Qubit
qubit = Qubit(0)
test_dict = {qubit : "qubit"} 
```
It will raise the following error
```sh
TypeError: unhashable type: 'Qubit'
```
we could define `__hash__` for Qubit but I don't what is better between this and the QubitId type.

src/qibocal/protocols/characterization/resonator_punchout_attenuation.py
```diff
@@ -230,9 +230,7 @@ def _fit(
         lp_min_att_dict[qubit] = lp_min_att
         hp_max_att_dict[qubit] = hp_max_att
         hp_min_att_dict[qubit] = hp_min_att
-        log.warning(
-            f"max att: {lp_max_att} -  min att: {lp_min_att} -  readout_attenuation: {ro_att}"
-        )
+    print(hp_min_att_dict)
```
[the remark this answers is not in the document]
andrea-pasquale: Thanks for catching this.
We should probably import this test also here :)

src/qibocal/protocols/characterization/classification.py
stavros11: Right, I guess `Qubit` is not hashable since it is mutable. We need it to be mutable as we typically change its parameters (frequencies, etc.) during calibration, otherwise we would need to create a new `Qubit` everytime. It is probably possible to make a mutable object hashable, but I am not sure if it is a good idea.

Maybe just defining the type `QubitId` is the best solution. I will do it there.

PR #346, opened 2023-05-22
Fix punchout platform update

andrea-pasquale:
Fixing a bug concerning the amplitude computed with `resonator_punchout`.
I want to take a look also the `resonator_punchout_attenuation`.

Checklist:
- [ ] Tests are passing.


andrea-pasquale:
The punchout over attenuation updates the platform correctly, the problem was related only the resonator punchout over amplitude given that we sweep amplitude by multiplying the value of the amplitude present in the runcard.

andrea-pasquale:
@stavros11 @rodolfocarobene it could be nice if you can check that after running punchout the correct amplitude is written in the runcard.

rodolfocarobene (approved):

PR #349, opened 2023-05-23
Standard rb auto

wilkensJ:
Routine for auto mode of standard rb 

Checklist:
- [ ] Tests are passing.


andrea-pasquale:
@wilkensJ before proceeding with the review could you please fix tests and conflicts?

Jacfomg:
Were you able to test on hardware ? At least see the RB running even if the results were crappy ?

src/qibocal/calibrations/niGSC/basics/fitting.py
andrea-pasquale, line 1: Since we are going to remove the old routines, it would be nice if you can move to the `RB` folder all the things that you need in order to run protocols.

src/qibocal/calibrations/niGSC/basics/fitting.py
```diff
@@ -116,19 +116,21 @@ def fit_exp1B_func(
...
                 xdata,
                 ydata,
-                p0=guess,
-                method="lm",
+                **kwargs,
             )
             perr = tuple(np.sqrt(np.diag(pcov)))
-        except:
+        except Exception as e:
+            log.warning(e)
```
andrea-pasquale, line 133: Here I would also add a message to explain that the error was raised during this specific fit.

src/qibocal/calibrations/niGSC/basics/fitting.py
```diff
@@ -165,7 +167,8 @@ def fit_exp1_func(
                 method="lm",
             )
             perr = tuple(np.sqrt(np.diag(pcov)))
-        except:
+        except Exception as e:
+            log.warning(e)
```
andrea-pasquale, line 171: Same here

src/qibocal/protocols/characterization/RB/result.py
```diff
@@ -0,0 +1,210 @@
...
+
+@dataclass
+class DecayResult(Results):
+    """Data being described by a single decay, Ap^x."""
+
+    # x and y data.
+    m: Union[List[numeric], np.ndarray]
+    y: Union[List[numeric], np.ndarray]
+    # Fitting parameters.
+    A: Optional[numeric] = field(default=None)
```
andrea-pasquale, line 29: Here you don't we set `None` instead of `field(default=None)`?
Same for the lines below.

src/qibocal/protocols/characterization/RB/result.py
```diff
@@ -0,0 +1,210 @@
...
+            raise ValueError(
+                "Lenght of y and m must agree. len(m)={} != len(y)={}".format(
+                    len(self.m), len(self.y)
+                )
+            )
+        if self.A is None:
+            self.A = np.max(self.y) - np.min(self.y)
+        if self.p is None:
+            self.p = 0.9
+        self.func = exp1_func
```
andrea-pasquale, line 51: This attribute is not used anywhere, you can drop it.

src/qibocal/protocols/characterization/RB/result.py
```diff
@@ -0,0 +1,210 @@
...
+        self.A, self.p = params
+        self.Aerr, self.perr = errs
+
+    def plot(self):
+        """Plots the histogram data for each point and the averges plus the fit."""
+
+        if self.hists is not None:
+            self.fig = plot_hists_result(self)
+        self.fig = plot_decay_result(self, self.fig)
+        return self.fig
```
andrea-pasquale, line 72: Here I can understand why you did it. However, we are still working on a way to display plots.
Until we find a solution I would advise against putting plotting methods inside the `Result` object.

src/qibocal/protocols/characterization/RB/result.py
```diff
@@ -0,0 +1,210 @@
...
+    """Data being described by a single decay with offset, Ap^x + B."""
+
+    B: Optional[numeric] = field(default=None)
+    Berr: Optional[numeric] = field(default=None)
+
+    def __post_init__(self):
+        super().__post_init__()
+        if self.B is None:
+            self.B = np.mean(np.array(self.y))
+        self.func = exp1B_func
```
andrea-pasquale, line 96: Not used

src/qibocal/protocols/characterization/RB/result.py
```diff
@@ -0,0 +1,210 @@
...
+class DecayWithOffsetResult(DecayResult):
+    """Data being described by a single decay with offset, Ap^x + B."""
+
+    B: Optional[numeric] = field(default=None)
+    Berr: Optional[numeric] = field(default=None)
+
+    def __post_init__(self):
+        super().__post_init__()
+        if self.B is None:
+            self.B = np.mean(np.array(self.y))
```
andrea-pasquale, line 95: I would prefer to put this directly in the `fit` function

src/qibocal/protocols/characterization/RB/result.py
```diff
@@ -0,0 +1,210 @@
...
+
+        if self.perr is not None:
+            return "({:.3f}\u00B1{:.3f})({:.3f}\u00B1{:.3f})^m + ({:.3f}\u00B1{:.3f})".format(
+                self.A, self.Aerr, self.p, self.perr, self.B, self.Berr
+            )
+        else:
+            return "DecayResult: Ap^m+B"
+
+
+def plot_decay_result(
```
andrea-pasquale, line 122: Perhaps it is better to move this plotting functions to a `plots.py`

src/qibocal/protocols/characterization/RB/standard_rb.py
```diff
@@ -0,0 +1,191 @@
...
+from qibocal.protocols.characterization.RB.result import (
+    DecayWithOffsetResult,
+    get_hists_data,
+)
+from qibocal.protocols.characterization.RB.utils import extract_from_data
+
+from .data import RBData
+from .params import RBParameters
+
+NoneType = type(None)
```
andrea-pasquale, line 19: Is this  really necessary? It seems a bit odd that you have to define a `NoneType`.

andrea-pasquale (changes requested):
Thanks @vodovozovaliza and @wilkensJ for implementing this.
You can find a some comments down below. Mainly to improve the code.
Could you please confirm that you were able to run this on hardware?
I didn't have time to do it yet.

EDIT: I did't see your comment about the hardware.
Since this is a priority I suggest to merge only when the code is working also on the hardware.

src/qibocal/protocols/characterization/RB/data.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        # FIXME this is necessary because the auto builder calls .to_csv(path_to_directory).
+        # But the DataFrame object from pandas needs a path to a file.
+        self.save_func = self.to_csv
+        self.to_csv = self.to_csv_helper
+
+    def to_csv_helper(self, path):
+        self.save_func(f"{path}/{self.__class__.__name__}.csv")
```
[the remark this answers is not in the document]
wilkensJ: The `RBData` is a child of a pandas DataFrame. If to_csv is overwritten directly, the functionality of the pandas data frame is lost. If I overwrite it and call it in the overwriting function, it calls itself which calls it self and so on. So I had to introduce the helper function.

src/qibocal/calibrations/niGSC/basics/fitting.py
```diff
@@ -116,19 +116,21 @@ def fit_exp1B_func(
         popt, perr = (ydata[0], 1.0, 0), (0, 0, 0)
     else:
         # Get a guess for the exponential function.
-        guess = kwargs.get("p0", (np.max(ydata) - np.mean(ydata), 0.9, np.mean(ydata)))
+        kwargs["p0"] = kwargs.get(
+            "p0", (np.max(ydata) - np.min(ydata), 0.9, np.min(ydata))
+        )
```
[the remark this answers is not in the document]
wilkensJ: This line creates a `p0` entry in kwargs if none is given, if the initial guess is already given it creates the guess by itself. Like this it can be passed by another function, or if it is not passed it creates it by itself. But I am not sure anymore if this belongs there, since this is happening in the result object itself.

src/qibocal/protocols/characterization/RB/result.py
```diff
@@ -0,0 +1,210 @@
...
+        if len(self.y) != len(self.m):
+            raise ValueError(
+                "Lenght of y and m must agree. len(m)={} != len(y)={}".format(
+                    len(self.m), len(self.y)
+                )
+            )
+        if self.A is None:
+            self.A = np.max(self.y) - np.min(self.y)
+        if self.p is None:
+            self.p = 0.9
```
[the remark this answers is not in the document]
wilkensJ: Since the fit function just generally fits exponential decays, the `p` does not have to be 0.9 there to keep it general. But the `DecayResult` know that is has to be around 0.9. I would keep it there, I think there is still come inconsistency though, because the other fitting function are giving a guess for themself, I will fix that. Thank you.

src/qibocal/calibrations/niGSC/basics/fitting.py
andrea-pasquale: It is not blocking the removal. I'm suggesting to start moving things around here to simplify the removal that will happen soon. If it is too complicated we can do it afterwards, but it could be useful to do it here directly.


> then the xid rb and simulfilter rb are lost in main

The old characterization routines are already not supported in main anymore. As soon as we merge some PRs (especially #363) that are currently opened I will proceed with the cleanup.

src/qibocal/protocols/characterization/RB/result.py
wilkensJ: I think because it works right now this is fine, for the development and hardware analyzed this is very useful. And the plot function is called in the plot function of the routine so there should not be any conflict.

src/qibocal/protocols/characterization/RB/result.py
wilkensJ: See above, same comment.

src/qibocal/protocols/characterization/RB/result.py
wilkensJ: See comment above.

src/qibocal/protocols/characterization/RB/standard_rb.py
wilkensJ: Yes, thank you, that is not right. Using `Optional` is much better.

src/qibocal/protocols/characterization/RB/standard_rb.py
```diff
@@ -0,0 +1,191 @@
...
+    scan = setup_scan(params)
+    # For simulations, a noise model can be added.
+    if params.noise_model:
+        from qibocal.calibrations.niGSC.basics import noisemodels
+
+        noise_model = getattr(noisemodels, params.noise_model)(*params.noise_params)
+    else:
+        noise_model = None
+    # Execute the scan.
+    data = execute(scan, params.nshots, noise_model)
```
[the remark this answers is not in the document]
wilkensJ: I strongly disagree. The more atomic approach makes it very easy to debug, run on buggy hardware or change small things when developing. The plots which were shown on Friday were only possible because of that atomic approach.

src/qibocal/calibrations/niGSC/basics/fitting.py
wilkensJ: How can I use the qibocal logging and still print the exception and the function that happened?:) I would like to display the error occurred during the fitting (there are multiple which can happen), then also a costume message where the error happened. Do you know how to do that?

src/qibocal/protocols/characterization/RB/standard_rb.py
```diff
@@ -0,0 +1,191 @@
...
+    Args:
+        data (RBData): Data from the data acquisition stage.
+
+    Returns:
+        StandardRBResult: Aggregated and processed data.
+    """
+
+    result = aggregate(data)
+    result.fit()
+    result.calculate_fidelities()
```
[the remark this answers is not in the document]
wilkensJ: See my reply above.

wilkensJ:
The plot when running on hardware, tii1q_b1:
Screenshot 2023-05-31 at 12 15 46 PM

src/qibocal/calibrations/niGSC/basics/fitting.py
andrea-pasquale: See for example here
https://github.com/qiboteam/qibocal/blob/d1c3fcbb6e45161ebd144dbb231382a85a8e6e6e/src/qibocal/protocols/characterization/ramsey.py#L238-L239

src/qibocal/protocols/characterization/RB/params.py
```diff
@@ -0,0 +1,30 @@
...
+
+@dataclass
+class RBParameters(Parameters):
+    """Standard Randomized Benchmarking runcard inputs."""
+
+    nqubits: int
+    """The amount of qubits on the chip """
+    qubits: list
+    """A list of indices which qubit(s) should be benchmarked """
+    depths: Union[list, dict]
```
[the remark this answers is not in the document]
andrea-pasquale: The problem with this is that it is not clear by looking only at the `RBParameters` which kind of dict I should pass (the keys of the dictionary). If you really want this flexibility you could add optional arguments such as `depths_start`, `depths_end` and `depths_step` so that if the user doesn't provide `depths` he can specify them using those arguments. In order to avoid side effects you should also raise an error if the user specify boths (depths and start, step and stop).

As you can see this can get tricky... The best solution is just to choose of the two approaches (generic list or start,step,stop params), either one is fine by me, but not both.
-

src/qibocal/protocols/characterization/RB/result.py
andrea-pasquale: I probably missed it, sorry.
Anyway it is not a good practice to introduce a new attribute in the `__post_init__`. It should be better to add this directly as an optional dataclass attribute with default value `exp1_func`.

wilkensJ:
> Thanks for the plot @wilkensJ. You said that there were some issues in qibolab, correct?

Yes, qiblob tried to load `nshots` from the setting section of the runcard, there is none specified. Also, loading the topology gives an error. I commented out that part in qibolab and added the `nshots` to the runcard just to make it work. This is of course no long term solution.

andrea-pasquale:
> > Thanks for the plot @wilkensJ. You said that there were some issues in qibolab, correct?
> 
> Yes, qiblob tried to load `nshots` from the setting section of the runcard, there is none specified. Also, loading the topology gives an error. I commented out that part in qibolab and added the `nshots` to the runcard just to make it work. This is of course no long term solution.

Are you using `qibolab_qrc_platforms`, right?

Jacfomg:
> > Thanks for the plot @wilkensJ. You said that there were some issues in qibolab, correct?
> 
> Yes, qiblob tried to load `nshots` from the setting section of the runcard, there is none specified. Also, loading the topology gives an error. I commented out that part in qibolab and added the `nshots` to the runcard just to make it work. This is of course no long term solution.

Can it be because the topology should be the adjacency matrix now @stavros11 ?

wilkensJ:
Thanks @andrea-pasquale and @Jacfomg, the protocol works now on hardware, I tested it again :)

src/qibocal/protocols/characterization/RB/standard_rb.py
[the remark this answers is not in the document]
andrea-pasquale: The `setup_scan` function is just a single line, therefore I would say that it is not a function. It's more like an alias.
Execute is also very minimal as a function. I would say that you can benefit from this approach if you are dealing with huge functions but this is not the case. 

If in the future we will develop more complicated protocols we can think about adding this extra layer consistently for all the protocols. For the time being I believe that it is not necessary.

andrea-pasquale:
> Thanks @andrea-pasquale and @Jacfomg, the protocol works now on hardware, I tested it again :)

Just to elaborate on this. The problem was that in `qibolab_platforms_qrc` only the `connectivity` which is now on main had the correct runcards.

src/qibocal/protocols/characterization/RB/standard_rb.py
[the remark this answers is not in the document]
Jacfomg: It may be simple to debug but now that people that did not develop the code may start running it I would try to focus as well on having everything the most readable possible if that does not comprise this approach

src/qibocal/protocols/characterization/RB/standard_rb.py
```diff
@@ -0,0 +1,182 @@
...
+PULSES_PER_CLIFFORD = 1.875
+
+
+@dataclass
+class StandardRBResult(DecayWithOffsetResult):
+    """Inherits from `DecayWithOffsetResult`, a result class storing data and parameters
+    of a single decay with statistics.
+
+    Adds the method of calculating a fidelity out of the fitting parameters.
+    TODO calculate SPAM errors with A and B
```
Jacfomg, line 25: That would be great, is it too hard to implement for this PR as well ???

src/qibocal/protocols/characterization/RB/params.py
```diff
@@ -0,0 +1,30 @@
...
+    """A list of depths/sequence lengths. If a dictionary is given the list will be build."""
+    niter: int
+    """Sets how many iterations over the same depth value."""
+    nshots: int
```
Jacfomg, line 19: Can we make this optional as I did on the other routines ? That way if its None it take the value from the runcard

src/qibocal/protocols/characterization/RB/params.py
```diff
@@ -0,0 +1,30 @@
...
+from typing import Union
+
+from qibocal.auto.operation import Parameters
+
+
+@dataclass
+class RBParameters(Parameters):
+    """Standard Randomized Benchmarking runcard inputs."""
+
+    nqubits: int
```
Jacfomg, line 11: Is there a way of you getting this from the platform ?

Jacfomg: Also, I see you use to create a circuit that may be bigger than the one needed for the qubits you are actually using. Is there a reason for this ?

Jacfomg:
It runs on Zurich, can comment if everything worked well as I did that on an uncalibrated qubit but it runs

src/qibocal/protocols/characterization/RB/standard_rb.py
```diff
@@ -0,0 +1,182 @@
...
+    data_list = []
+    # Iterate through the scan and execute each circuit.
+    for c in scan:
+        # The inverse and measurement gate don't count for the depth.
+        depth = (c.depth - 2) if c.depth > 1 else 0
+        if noise_model is not None:
+            c = noise_model.apply(c)
+        samples = c.execute(nshots=nshots).samples()
+        # Every executed circuit gets a row where the data is stored.
+        data_list.append({"depth": depth, "samples": samples})
```
Jacfomg, line 83: Is it here were the unrolling should be coded ? I am a bit lost on how that can be done using your functions or if it can be done with circuits instead of pulses. You may leave this for when this gets merged but for sure we need a way to code the sequence unrolling to accelerate this. The only example I can provide is the RB_sequence_unrolling branch I have.

Jacfomg: Do you know if our idea, at least for the single qubit gates, it's to have all of them converted to U3s and implemented them with two pulses ? This may allow me to check if waveform replacement is useful for Zurich and make RB much faster in theory.

src/qibocal/calibrations/niGSC/basics/fitting.py
andrea-pasquale: The new protocols don't share a common fitting folder where all the fitting methods.
Usually we create a folder to group protocols that have similar fitting/plotting methods.
See for example the [coherence](https://github.com/qiboteam/qibocal/tree/main/src/qibocal/protocols/characterization/coherence) folder where all the protocols are calling https://github.com/qiboteam/qibocal/blob/29e3c0e1e7ab1eeac2fac381eaabbacf3d6ddbc2/src/qibocal/protocols/characterization/coherence/utils.py#L11 which is in the file `coherence/utils.py`.

For the RB / niGSC I suggest to move everything that is necessary to run protocols in the[ `RB` ](https://github.com/qiboteam/qibocal/tree/standard_rb_auto/src/qibocal/protocols/characterization/RB) folder. If you want to maintain temporarily the old routines I would suggest to call any fitting/plotting methods directly from the RB folder. 

For the sake of this PR this is not strictly necessary. I would make it here just to facilitate the removal of the all routines.

ingoroth: Ok. then let's do this.  We port the function to RB and load in `niGCS` from `RB` . Then it is easier to delete the legacy files eventually.

src/qibocal/protocols/characterization/RB/standard_rb.py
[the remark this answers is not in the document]
ingoroth: I agree that the setup function is a bit funky here, the reason being that the core functionality for defining the RB circuits is still imported from the `niGSC` module.   This should be resolved already in this PR, .e.g. by hardcoding it.  As a matter of principle, the modularity however is very good, especially, since a lot of characterization protocol will follow the same blue print.

src/qibocal/calibrations/niGSC/basics/fitting.py
andrea-pasquale: That would be perfect. Thank you!

src/qibocal/protocols/characterization/RB/data.py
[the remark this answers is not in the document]
andrea-pasquale: Something like this should work I think then
```py
def to_csv(self, path):
     super().to_csv(f"{path}/{self.__class__.__name__}.csv")
```
See here

src/qibocal/protocols/characterization/RB/params.py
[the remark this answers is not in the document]
andrea-pasquale: > And it can be found in the post_init of the parameter object.

The user is not supposed to look the post__init to understand which parameters should be passed. So I'm still not convinced :)
If you want to document it properly you can give it a go, but it should be a priority given the content of this PR.

src/qibocal/protocols/characterization/RB/data.py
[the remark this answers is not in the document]
wilkensJ: That's great! Thank you. This looks much cleaner.

src/qibocal/protocols/characterization/RB/params.py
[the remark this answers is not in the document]
wilkensJ: Would it be enough to write a small comment in the runcard?

src/qibocal/protocols/characterization/RB/result.py
andrea-pasquale: With the new API we distinguish between structure data and operations.
There are three data structure:

- `Parameters`
- `Data`
- `Results`

And three operations:

- `acquisition` which generates `Data` starting from `Parameters`
- `fit/post-processing` which generates `Results` starting from `Data`
- `plot` which generates a list of figures + table starting from `Data` and `Results`

This is the current model that we are adopting for all protocols: inserting fitting or plotting results directly in the `Results` object generates confusion because the `Results` should not know anything about plotting or fitting, it's just a container. Since we are refactoring everything, it might be better to have the same approach for all protocols.

src/qibocal/protocols/characterization/RB/params.py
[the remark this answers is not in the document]
andrea-pasquale: The runcard is not the right place to document the code.
Everything should be done through docstrings.
Again I think that a custom list should be more than enough.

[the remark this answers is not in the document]
wilkensJ: I would like to keep the feature, I can take it out of the runcard so users who know how to use it can use it but users who not know just pass lists.

src/qibocal/protocols/characterization/RB/result.py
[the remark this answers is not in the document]
wilkensJ: I added the guess to the fitting function, but kept it in the result object, so before calling the fitting this guess already is there.

src/qibocal/protocols/characterization/RB/standard_rb.py
wilkensJ: Do you mean use the architecture on how to define gates and how the compiler is encoded from #343? I think this should be very easy to implement in the given architecture of this PR. Just the scan and execute function have to be changed, better would be to add the compiler (by hand if necessary) to the qibolab compiler and then use all the functionality from this PR.

Jacfomg: I was thinking of another way of doing some kind pulse replacement to make RB faster on my device. But it the end I think the better would be the thing I mentioned in Slack. So forget about this.

src/qibocal/protocols/characterization/RB/params.py
andrea-pasquale: There is a relative simple way to do this.
You can get `nqubits` from the `Platform` object `platform.nqubits`.
If you pass the platform as an argument like we do for other protocols this should be easy to do.

Another option is to obtain the `platform` from the `backend` object.

src/qibocal/protocols/characterization/RB/params.py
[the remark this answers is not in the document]
andrea-pasquale: @Jacfomg @Edoardo-Pedicillo what do you think?

src/qibocal/protocols/characterization/RB/standard_rb.py
```diff
@@ -0,0 +1,212 @@
...
+    Returns:
+        RBData: _description_
+    """
+
+    # 1. Set up the scan (here an iterator of circuits of random clifford gates with an inverse).
+    scan = setup_scan(params)
+    # For simulations, a noise model can be added.
+    if params.noise_model:
+        from qibocal.calibrations.niGSC.basics import noisemodels
+
```
andrea-pasquale, line 153: Usually it is better to avoid this type of import, unless they are particularly heavy.
You can move the import directly at the top of the file.

andrea-pasquale (commented):
Thanks @wilkensJ for all the updates.
Please find below a few suggestions.
Also can you rename the folder something different other than `RB` using we use lower case letters for folders and files.

src/qibocal/protocols/characterization/RB/result.py
andrea-pasquale: This will break the initial idea of having `Results` only as a container.


> I would also say it is actually a bit confusing that the plotting function in `plots.py` actually only require `Results` type objects as input but would actually break if one does not provide the correct subclass of the specific routine

This should not be the case because the every protocols should have only a single plotting method consisting of multiple figures if necessary.

andrea-pasquale: For the time being it should be fine. However just keep in mind that with this (having a plot method inside the results class) you are defining another API on top of the agreed one which is outside of the purpose of this PR that was meant to port the RB to the new  layout.

src/qibocal/protocols/characterization/RB/standard_rb.py
```diff
@@ -0,0 +1,212 @@
...
+        Returns:
+            Iterable: The iterator of circuits.
+    """
+
+    def make_circuit(depth):
+        """Returns a random Clifford circuit with inverse of `depth`."""
+
+        def layer_gen():
+            """Returns a circuit with a random single-qubit clifford unitary."""
+            return random_clifford(len(params.qubits), return_circuit=True)
```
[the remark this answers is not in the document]
wilkensJ: You mean with the lambda syntax? Or do you have something else in mind?:)

```python
lambda x: random_clifford(len(params.qubits), return_circuit=True)
```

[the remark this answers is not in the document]
andrea-pasquale:
```py
circuit = layer_circuit(random_clifford(len(params.qubits), return_circuit=True), depth)
```
at line 70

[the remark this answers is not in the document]
wilkensJ: It has to be a callable though, random_clifford returns a gate, a gate is not callable.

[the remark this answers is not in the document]
andrea-pasquale: Does it need to be callable? With a few simple changes in `layer_circuits` everything works as expected.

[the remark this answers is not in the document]
wilkensJ: This functions are build to make more complicated circuits possible. I see that for now this is not very useful (because there will only be one scan, random_cliffords with an inverse scan), but by only changing the layer_gen() function the new layers of Xs and 1s builds the xid scan (which is just one line), same for the Z3, Z4, interleaved rb, ... The only change is then to remove the inverse layer and rewrite the layer_gen() function, everything else can stay then :)
And I think that it is not overcomplicating things for now.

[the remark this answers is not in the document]
andrea-pasquale: I understand. I'm not saying to drop the `layer_gen` for each protocol.
If was just suggesting to not use a callable. In this way you can avoid to generate the layer directly in the `layer_circuit` method.
https://github.com/qiboteam/qibocal/blob/e479473ca58400141c899ccf639efdebe6b49ce7/src/qibocal/protocols/characterization/rb/circuit_tools.py#L51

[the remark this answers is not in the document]
wilkensJ: I think why a callable is needed is that each call returns a new layer, that's for the randomness.
What do you have to say to this @ingoroth (since this is your code)?:)

src/qibocal/protocols/characterization/RB/params.py
wilkensJ: Would that work without specifying a platform? When the protocol is simulated? :)

src/qibocal/protocols/characterization/RB/standard_rb.py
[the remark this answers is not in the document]
ingoroth: I do not really understand, how and why you want to avoid to make it callable, @andrea-pasquale.  Could you explain what you would like to change?  My approach was that `layer_gen` is supposed to simply wrap the `random_clifford` generator.  The latter you actually have to call *every time* you want a different layer and the number of layers you need depends on the other parameters of the `layer_circuit` method.  I find the  `layer_circuit` taking a callable generator actually a very neat syntax. If it is about streamlining the code and avoiding the local definition, I am also ok with something like
```
layer_gen = partial(random_clifford, len(params.qubits), return_circuit=True)
```

src/qibocal/protocols/characterization/RB/result.py
wilkensJ: Every plotting function has to take a specific type of data and, if needed, a specific type of Result object (I just skipped through the routines in qibocal auto, this is what I saw). The plotting functions break, if another type of object is passed. So the plotting function is tightly bound to the passed data object and the result object.

The only difference in the rb section is now, that it is made explicit. In the qibocal auto parts this is the case, but implicit. Both works on the top level.

For developers it is very convenient to have this API. The result can just be stored and later fitted and plotted. Even with different parameters. Even with another plotting function. Without having to know how the data looks like or from which module the results came.

src/qibocal/protocols/characterization/RB/params.py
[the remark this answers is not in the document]
AleCandido: @wilkensJ in order to avoid accepting a whatever `dict`, you could use in your type:

- a [`dataclass`](https://docs.python.org/3/library/dataclasses.html) or
- a [`TypedDict`](https://docs.python.org/3/library/typing.html#typing.TypedDict)

Essentially the two things are similar, but a `dataclass` creates an actual object, so you will be able to access its content with `instance.attribute` instead of `instance["attribute"]`, and you get some more runtime methods defined, while a `TypedDict` is fully a `dict` at runtime, but the fields are checked statically (so attempting to access `depths["ciao"]` would be flagged as an error by the analyzer).

In this case, I have to a have a second look about how we deserialize the runcards parameters, but I might suggest a `TypedDict` (because I don't believe I added any supported for nested deserialization, so you would get a `dict` anyhow, and it would clash with the `dataclass` type).

In any case, runcards are not validated using types at the moment, so the type check is only happening in the code (if the type declared in parameters is consistent with its usage in the routine).
The best solution for this, would be to add a specific loader supporting both the inputs, deserializing and validating them.

However, at the moment increase the support for complex parameters specification is not a priority, so I would suggest limiting to `Union[list, MyTypedDictForDepth]`, or to only use `list`. But the type itself might be useful in general (support both `list` and `start:,stop:,step:`, we could call it `RangeList`), simply is more on the UI side, and first there are more substantial features missing.
If you wish, you could make an issue for this.

src/qibocal/protocols/characterization/RB/result.py
AleCandido: About the methods of `Results` objects: in principle it is true that it has been designed as just a container, and all the other routines are using in this way. In practice, this is irrelevant what each single routine is doing.

The idea is that `Results` has to be a container, as `Parameters` and `Data`, and a `plot` function has to exist for the report, as `acquisition` and `fit` for the calibration process.
But in practice, everything that is done on top of this is just routine specific, so it won't be used by the overall infrastructure.

The reason to follow a common layout is only to make it easier for developers themselves: if you're familiar with what is happening with `Results` of one routine, you might already know what to expect from the others. Standardization helps to reduce the complexity.

@ingoroth the difference you're discussing is just to call `results.plot()` for a specific routine, or `MyRoutine.plot(results)`. Both are fine, and pretty much equivalent. We should only decide on a single one ([PEP 20, 13th aphorism](https://peps.python.org/pep-0020/)), to improve consistency simplifying the code. But this can be done even after this PR (smaller PRs would help reviewers).

src/qibocal/protocols/characterization/RB/result.py
AleCandido: > Every plotting function has to take a specific type of data and, if needed, a specific type of Result object (I just skipped through the routines in qibocal auto, this is what I saw). The plotting functions break, if another type of object is passed. So the plotting function is tightly bound to the passed data object and the result object.

Perfectly agree, that's why I did not complain when the `plot()` functions were being defined in the routines (they were absent in the initial `qibocal.auto` design).

And I agree to keep them in the same module of the respective result, this is fine.

> For developers it is very convenient to have this API. The result can just be stored and later fitted and plotted. Even with different parameters. Even with another plotting function. Without having to know how the data looks like or from which module the results came.

And I would also agree on this, but you do not need to make them `classmethod` for this reason: a method is a function that receives the instance as input, and the `classmethod` receives the class. But you're not using the class at all, so it should not be a `classmethod` (at most a `staticmethod`).

> The only difference in the rb section is now, that it is made explicit. In the qibocal auto parts this is the case, but implicit. Both works on the top level.

About this I fully disagree:
https://github.com/qiboteam/qibocal/blob/29e3c0e1e7ab1eeac2fac381eaabbacf3d6ddbc2/src/qibocal/auto/operation.py#L94-L99
Every routine has its own `Results` type, and the `report` function (that's the actual name of `plot`) is definitely specific for that results type.
If you ignore the type, that's not a problem of the function. But it would be the exact same with a method (methods are also normal functions, they are simply *bound* to objects, so it simply means that you have a special syntax to call them, `object.method()` vs `method(object)`).

src/qibocal/protocols/characterization/RB/standard_rb.py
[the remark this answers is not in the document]
AleCandido: > I strongly disagree. The more atomic approach makes it very easy to debug, run on buggy hardware or change small things when developing. The plots which were shown on Friday were only possible because of that atomic approach.

> As a matter of principle, the modularity however is very good, especially, since a lot of characterization protocol will follow the same blue print.

In this case I fully agree with @ingoroth and @wilkensJ. The more atomic, the better.
This would be beneficial for all the code written, not just for the protocol. And writing atomic functions avoid a lot of repetitions (I've asked plenty of times to split things in separate functions).

About the one line function: even one-line functions in general are worth, if they are conceptual units.
In this case, it is true that `setup_scan` is practically an alias, but that's not bad. In particular, there is an idiomatic way of defining this *alternative constructors*: `classmethod`s.

```py
class StandardRBScan:
    @classmethod
    def setup_scan(cls, params: RBParameters):
        return cls(params.nqubits, params.depths * params.niter, params.qubits)
```
(btw, the original hint, `-> Iterable`, is either wrong or too general, since it should have been `-> StandardRBScan`)

src/qibocal/protocols/characterization/RB/params.py
AleCandido: In principle, a `Platform` should be always available, at least if you're using `qq-auto` (since it is expected to have a runcard to be calibrated, and routines might directly request access to the driver, not just gates).

If simulation is supposed to be used to actually simulate the calibration process, I wonder if we should make a platform for it...
Would `dummy` be enough for simulation?

andrea-pasquale: The problem with dummy is that it produces random numbers...
We already talked in the past about implementing some sort of platform for simulation but it is not trivial to do.

src/qibocal/protocols/characterization/RB/result.py
wilkensJ: > And I would also agree on this, but you do not need to make them `classmethod` for this reason: a method is a function that receives the instance as input, and the `classmethod` receives the class. But you're not using the class at all, so it should not be a `classmethod` (at most a `staticmethod`).

What I have in mind here:

```python
class MyResult:
   ...
   def plot(self, *args, **kwargs):
      # Do something class specific, maybe dependent on args and kwargs.
      self.fig = specific_plot1(self) # The figure is stored, it can be displayed later/changed later/another figure can be added
      return self.fig # Just for convenience

def specific_plot1(result: MyResult):
   # Plot and 
   # then return a fig
```
Now, it can be called as a class method *and* a function. And, when the result is loaded from disc, I don't have to know which plotting function I need in order to display the result correctly. I can even define a new function which takes a result and plots it differently, when in development mode.


> If you ignore the type, that's not a problem of the function. But it would be the exact same with a method (methods are also normal functions, they are simply _bound_ to objects, so it simply means that you have a special syntax to call them, `object.method()` vs `method(object)`).

I think we have something else in mind here. The plotting functions I saw (in `utils` of diverse folders but also the module specific `rerport` function), only work with their specific type of data and result object, because attributes are called and a certain structure needed. Meaning that `method(object)` would break, if `object` cannot deliver the wanted attributes or keys. It is tightly bound to what object it gets. Which is totally fine. But then why not *actually* bind it to the object by making it a class method? This is what I mean with implicit and explicit.

src/qibocal/protocols/characterization/RB/params.py
AleCandido: > We already talked in the past about implementing some sort of platform for simulation but it is not trivial to do.

Agreed, but maybe we should discuss and actually schedule it, if it's going to be used.

wilkensJ: The RB modules (and in general the high level characterization routines) need a way to be simulated, but they only need to simulate circuits, not pulses (sometimes it is convenient, because the transpiler and compiler is hard coded, to define them on a pulse  level though). That is why in general (not always though) the high level characterizations (or gate level characterizations) don't need a platform directly. Just a backend. That is why the parameters `nqubits` and `nshots` (maybe even more could be loaded directly from the platform) have to be specified.

src/qibocal/protocols/characterization/RB/result.py
ingoroth: @AleCandido the explicit vs. implicit point here was IMO only a matter of annotation not of code structure. In the `plot.py` the methods are actually not *annotated* with the required subtype.   Other than this I fully agree that is just two APIs.  The main reason for us to introduce this (and let's discuss if we keep it after the PR),  is that in the future (=next weeks) with more complicated characterization routines, like simultaneous RB, the one-to-one relation between a single plot function and one routine is actually limiting. Actually many routines will have exactly the same type of result etc.  But again let's keep this discussion to a point when everyone is more what is required.  

Regarding, your point of it being actually `staticmethod` and not a `classmethod`, I fully agree.

src/qibocal/protocols/characterization/RB/params.py
AleCandido: I see your need, but avoiding the simulation platform would require a separate execution path to be maintained.
While saving for sure the time to develop the additional platform, there will always be the additional track to be maintained.

Moreover, it might be beneficial to be able to simulate in the exact same context in which the hardware execution happens. Creating differences at a higher level would make it more difficult to trace where the differences in the output are coming from.

src/qibocal/protocols/characterization/RB/standard_rb.py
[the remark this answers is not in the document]
ingoroth: We already kind of resolved this by actually re-implementing the functionality from the legacy `niGSC` module here.  So now it is not an alias for class instantiation anymore but actually doing something ;)

src/qibocal/protocols/characterization/RB/params.py
wilkensJ: Right now everything works by specifying a backend in the runcard, or loading the functionality directly into a script and make a plain circuit simulation, both times the circuits are just simulated with qibo.

Or do you have a change in the RB module in mind to make more transparent what is simulated where?

src/qibocal/protocols/characterization/RB/standard_rb.py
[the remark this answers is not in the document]
andrea-pasquale: > I think why a callable is needed is that each call returns a new layer, that's for the randomness. 

You are right. Now I understand.
Instead of a wrapper we could also use a **proper** python generator but I would say that this is fine for the time being.

src/qibocal/protocols/characterization/RB/standard_rb.py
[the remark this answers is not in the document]
andrea-pasquale: It is a matter of personal preference. Personally I prefer a  more readable approach like @Jacfomg 's implementation in #343. @AleCandido 's proposal seems like a good compromise.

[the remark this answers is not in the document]
andrea-pasquale: Sorry I also got lost in this conversation.
Lets keep it as it is.

src/qibocal/protocols/characterization/RB/params.py
andrea-pasquale: Maybe lets add this point to #374

src/qibocal/protocols/characterization/RB/result.py
andrea-pasquale: > The main underlying reason for our preference of introducing more API, is that we are also developing for different user and workflow in characterization that is distinct from what is happening in the low-level auto-calibration.

I see.
For the future I would suggest to tackle only a single problem for one PR.
In this particular case, even though I understand that you may need a different API, it would have been better to first have the RB ported to the new layout and only in a different PR introduce a different API. In this way we could have discussed more thoroughly about the problems in current API and find a proper way to solve it, instead of introducing partial changes targeted only for a specific part of the code.

andrea-pasquale (approved):
As soon as conflicts are fixed we should be ready to merge.
Thanks for all the updates and the explanations @wilkensJ @vodovozovaliza @ingoroth.

wilkensJ:
Thank you everybody, @vodovozovaliza, @andrea-pasquale, @ingoroth, @AleCandido and @Jacfomg!

Jacfomg:
> Thank you everybody, @vodovozovaliza, @andrea-pasquale, @ingoroth, @AleCandido, @Edoardo-Pedicillo and @Jacfomg!

Good, now can we get interleaved working ?

ingoroth:
Thanks you ***@***.*** … First, filtered RB, small groups, simultaneous/correlated, two-qubit Clifford, cycle benchmarking, Pauli noise learning, gate-set shadows for robust QPT, GST integration, overlapping detector tomography … On 6. Jun 2023, at 10:26, Juan Cereijo ***@***.***> wrote:﻿

Thank you everybody, @vodovozovaliza, @andrea-pasquale, @ingoroth, @AleCandido, @Edoardo-Pedicillo and @Jacfomg!

Good, now can we get interleaved working ?

Jacfomg:
@ingoroth It would be nice to have it working so we can reproduce the IQM spreadsheet before they take the chip away at the beggining on august.

PR #351, opened 2023-05-23
Add optional local qubits parameter

### LINKED ISSUE #301: Global qubits parameter in runcard to action specific parameter
Would it be possible to specify the `qubits` parameter in the qibocal runcards in each action instead of globally? When using `niGSC` protocols, it could be very useful to specify target qubits in the same runcard.
But I can see that, when characterizing only one qubit with many actions, the global parameter is very useful.
So instead of
https://github.com/qiboteam/qibocal/blob/9a27c90074150cec3106010a16b2c104011ca451/runcards/niGSC.yml#L1-L28

Have something like

```yaml
backend: numpy
format: pickle

actions:
  standardrb:
    qubits: [0]
    nqubits: 5
    depths: [1,3,5,7,10]
    runs: 2
    nshots: 1024
    noise_model: PauliErrorOnUnitary
    noise_params: [0.01, 0.01, 0.01]
  XIdrb:
    qubits: [2]
    nqubits: 5
    depths: [1,2,3,4,5,6,7,8,9,10]
    runs: 5
    nshots: 10
    noise_model: PauliErrorOnX
    noise_params: [0.05, 0.01, 0.01]
  simulfilteredrb:
    qubits: [0,1,2,3,4]
    nqubits: 5
    depths: [1,3,5,7,10]
    runs: 2
    nshots: 1024
    noise_model: PauliErrorOnUnitary
    noise_params: [0.01, 0.01, 0.01]
```

andrea-pasquale:
Closes #301 and this point
> - [x] qubits argument (see first presentation of new layout)

from #336.

This PR adds a local `qubits` parameter for each routine.
If this parameter is not provided it will be used the global parameter `qubits`, if provided it will overwrite the global parameter.
Probably this is not enough for supporting two qubit gates routines.
We could address this in another PR once we port two qubit gates to the new layout.

Checklist:
- [x] Tests are passing.


stavros11 (approved):
It looks good to me, thanks. I did a quick test using the resonator spectroscopy routines and the `qubits` parameter seems to work as expected. A real test for this would be to port the two-qubit gate routines and try to use the local qubits parameter for them but I guess this can be done in a separate PR (ideally using this).

src/qibocal/protocols/characterization/allxy/drag_pulse_tuning.py
```diff
@@ -28,6 +28,10 @@ class DragPulseTuningParameters(allxy_drag_pulse_tuning.AllXYDragParameters):
     """DRAG pulse beta end sweep parameter."""
     beta_step: float
     """DRAG pulse beta sweep step parameter."""
+    qubits: Optional[list] = field(default_factory=list)
```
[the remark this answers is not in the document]
andrea-pasquale: Thanks @stavros11.
Yes, you are right. I believe that it can be dropped.

src/qibocal/cli/auto_builder.py
```diff
@@ -72,6 +75,11 @@ def routine_name(self, routine, iteration):
         name = routine.replace("_", " ").title()
         return f"{name} - {iteration}"
 
+    def routine_qubits(self, routine_name, iteration):
+        """Get local qubits parameter from Task if available otherwise use global one."""
+        local_qubits = self.history[(routine_name, iteration)].task.parameters.qubits
+        return local_qubits if local_qubits else self.qubits
```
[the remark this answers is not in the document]
andrea-pasquale: I think that this should be `>0`, right?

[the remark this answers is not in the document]
stavros11: Yes, sorry, you are right, it should be `>0`.

src/qibocal/cli/utils.py
```diff
@@ -117,3 +118,11 @@ def generate_output_folder(folder, force):
...
     os.makedirs(path)
     return folder
+
+
+def allocate_qubits(platform: Platform, qubit_ids: list):
+    """Load platform' qubits."""
+    if platform is not None:
+        return {q: platform.qubits[q] for q in qubit_ids if q in platform.qubits}
+    else:
+        return qubit_ids
```
AleCandido, line 128:
```suggestion
    return qubit_ids
```

src/qibocal/protocols/characterization/allxy/allxy.py
```diff
@@ -18,6 +18,12 @@ class AllXYParameters(Parameters):
 
     beta_param: float = None
     """Beta parameter for drag pulse."""
+    qubits: list = field(default_factory=list)
+    """Local qubits (optional)."""
+    update: Optional[
+        bool
+    ] = False  # False because this routine does not update the runcard
```
AleCandido, line 25: Is this really an `Optional[bool]`?

An `Optional[bool]` is a variable that can be `False`, `True`, or `None`.
If this is *optional* in the sense that has a default, the type should just be `bool`.

src/qibocal/protocols/characterization/allxy/allxy_drag_pulse_tuning.py
```diff
@@ -24,6 +24,12 @@ class AllXYDragParameters(Parameters):
     """Final beta parameter for Drag pulse."""
     beta_step: float
     """Step beta parameter for Drag pulse."""
+    qubits: list = field(default_factory=list)
+    """Local qubits (optional)."""
+    update: Optional[
+        bool
+    ] = False  # False because this routine does not update the runcard
```
AleCandido, line 31: Same as above.

src/qibocal/cli/utils.py
AleCandido: And since you're actually defining now, definitely place somewhere else...

src/qibocal/protocols/characterization/allxy/drag_pulse_tuning.py
```diff
@@ -28,6 +28,10 @@ class DragPulseTuningParameters(allxy_drag_pulse_tuning.AllXYDragParameters):
     """DRAG pulse beta end sweep parameter."""
     beta_step: float
     """DRAG pulse beta sweep step parameter."""
+    qubits: list = field(default_factory=list)
+    """Local qubits (optional)."""
+    update: Optional[bool] = None
```
AleCandido, line 33: Are you sure that it shouldn't be:
```suggestion
    update: bool = False
```
as for the others?

src/qibocal/protocols/characterization/classification.py
```diff
@@ -16,6 +16,12 @@
 
 @dataclass
 class SingleShotClassificationParameters(Parameters):
+    nshots: int
+    """Number of shots."""
+    qubits: list = field(default_factory=list)
+    """Local qubits (optional)."""
+    update: Optional[bool] = None
```
AleCandido, line 23: As above

src/qibocal/protocols/characterization/resonator_spectroscopy.py
```diff
@@ -27,6 +27,10 @@ class ResonatorSpectroscopyParameters(Parameters):
     amplitude: Optional[float] = None
     """Readout amplitude (optional). If defined, same amplitude will be used in all qubits.
     Otherwise the default amplitude defined on the platform runcard will be used"""
+    qubits: list = field(default_factory=list)
+    """Local qubits (optional)."""
+    update: Optional[bool] = None
```
AleCandido, line 32: If you are actually adding to all parameters, then we need a better mechanism them parameters.

Let's pull the `update` key out of them.

tests/runcards/test_autocalibration.yml
```diff
@@ -119,6 +119,32 @@ actions:
       nshots: 10
       relaxation_time:
 
+  - id: rabi amplitude
+    priority: 0
+    operation: rabi_amplitude
+    next: rabi length
+    parameters:
+      update: False
```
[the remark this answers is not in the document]
AleCandido: If you want to add a routine default for some routines, define an attribute (e.g. just `update`) for the routine class itself, and check it during execution. If the key is missing, use the local default (per-routine), otherwise the global default.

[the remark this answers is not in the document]
andrea-pasquale: Thanks @AleCandido.
Indeed this is a good idea.
I think that we can do the same for the `qubits` argument, right?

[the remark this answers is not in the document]
AleCandido: Definitely.

Unfortunately, I've never reviewed the routines themselves, and I had not realized it was there.
Moreover, there was already a plan for `qubits`, so it is very relevant to have it outside (even though I thought it was something that should have been global, but if you need per routine add it, at most we can have in both places, or turn it into something like `exclude-qubits` per routine, according to the most common use case).

src/qibocal/auto/task.py
```diff
@@ -6,6 +6,8 @@
 
 from qibolab.platform import Platform
 
+from qibocal.cli.utils import allocate_qubits
```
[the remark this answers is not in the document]
andrea-pasquale: You are right. Since this function is utils by both `cli` and `auto` module I decided to put in a generic utils.py.
This should avoid circular imports. Let me know if you agree.

tests/runcards/test_autocalibration.yml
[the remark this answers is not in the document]
andrea-pasquale: Actually this PR originally implemented this additional argument `qubits`, this is why you never saw it.
People have asked for more flexibility when choosing on which qubits to run protocols see #301.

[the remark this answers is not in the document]
andrea-pasquale: I've just pushed the implementation that you suggested.
We could even go a step further and do the same for `nshots` and `relaxation_time` given that they are common to all the routines. I'm not too sure given that they are parameters of the routine itself. Thoughts @stavros11 @rodolfocarobene @Jacfomg?

src/qibocal/auto/execute.py
```diff
@@ -26,6 +26,8 @@ class Executor:
     """Qubits to be calibrated."""
     platform: Optional[Platform] = None
     """Qubits' platform."""
+    update: Optional[bool] = True
```
AleCandido, line 29: Second iteration: are you sure you need an `Optional[bool]` and not just a `bool` with a default?

src/qibocal/auto/execute.py
```diff
@@ -128,5 +132,10 @@ def run(self):
             completed = Completed(task, output, Normal())
             self.history.push(completed)
             self.head = self.next()
+
             if self.platform is not None:
-                self.platform.update(completed.res.update)
+                # update platform if:
+                # - global update is True and local not specified
+                # - local update is True
```
AleCandido, line 139: Great, but better lifted in the docstring ;)

src/qibocal/auto/runcard.py
```diff
@@ -39,7 +43,7 @@ class Runcard:
     """Structure of an execution runcard."""
 
     actions: List[Action]
-    qubits: Optional[List[Union[int, str]]] = Field(default_factory=list)
+    qubits: Optional[list] = Field(default_factory=list)
```
AleCandido, line 46:
```suggestion
    qubits: Optional[list[QubitID]] = Field(default_factory=list)
```
maybe?

src/qibocal/auto/task.py
```diff
@@ -70,6 +72,14 @@ def priority(self):
     def parameters(self):
         return self.operation.parameters_type.load(self.action.parameters)
 
+    @property
+    def qubits(self):
+        return self.action.qubits
+
+    @property
+    def update(self):
+        return self.action.update
```
AleCandido, line 81: Since they start to be a lot of properties following this pattern, I wonder if it would be less redundant to have a `from_action` constructor, and put all of these as regular fields.

The only advantage of the current layout is that they are *read-only* (if you do not access the internal object, of course).

It can (and probably should) be tackled in a separate PR, we can make an issue for this.

src/qibocal/cli/utils.py
```diff
@@ -8,6 +8,7 @@
 
 import click
 import yaml
+from qibolab.platform import Platform
```
AleCandido, line 11: This new import seems to be unused... (it is new, and there is no more new code)

AleCandido: Since this pattern appears also in next files, I added myself #385, and I'm going to run on this PR before proceeding

AleCandido:
I changed base just to reduce the number of files changed. #385 is a very tiny PR, so (I hope) it should become soon enough `main` again.

src/qibocal/auto/runcard.py
andrea-pasquale: You are right

src/qibocal/auto/execute.py
andrea-pasquale: Yep,`Optional` is not needed

src/qibocal/auto/task.py
andrea-pasquale: Also to me it feels like we are redundant. I've opened an issue for this (#386)

andrea-pasquale:
Thanks for the second review @AleCandido.
I've implemented your suggestions.

src/qibocal/auto/execute.py
AleCandido: Since you lifted, you can also remove from here: it's just written a few lines above ;)

src/qibocal/web/templates/autocalibration.html
```diff
@@ -152,7 +152,7 @@ <h1>{{ report.title }}</h1>
           class="d-flex justify-content-between flex-wrap flex-md-nowrap align-items-center pt-3 pb-2 mb-3 border-bottom">
           {{ report.routine_name(routine, iteration) }}</h4>
         {% if report.qubits != [] %}
-        {% for qubit in report.qubits %}
+        {% for qubit in report.routine_qubits(routine, iteration) %}
```
AleCandido, line 155: Even here: `(routine, iteration)` is a `task.uid`.

src/qibocal/utils.py
```diff
@@ -0,0 +1,13 @@
...
+from qibolab.qubits import Qubit, QubitId
+
+
+def allocate_qubits(
+    platform: Platform, qubit_ids: List[QubitId]
+) -> Union[List[QubitId], Dict[QubitId, Qubit]]:
+    """Convert List[QubitId] -> Dict[QubitId, Qubit] for non-trivial platform."""
+    if platform is not None:
+        return {q: platform.qubits[q] for q in qubit_ids if q in platform.qubits}
+    return qubit_ids
```
AleCandido, line 13:
```suggestion
def allocate_qubits(platform: Platform) -> Dict[QubitId, Qubit]:
    """Convert List[QubitId] -> Dict[QubitId, Qubit] for non-trivial platform."""
    return {q: platform.qubits[q] for q in qubit_ids if q in platform.qubits}
```
Make it simpler: the `None` check you can do before calling the function, and in this way the function have one less argument and a well-defined return type.

Moreover, with the current implementation the `platform` type would be even *slightly wrong*: it should be `Optional[Platform]`.

src/qibocal/cli/auto_builder.py
```diff
@@ -71,6 +74,11 @@ def routine_name(self, routine, iteration):
         name = routine.replace("_", " ").title()
         return f"{name} - {iteration}"
 
+    def routine_qubits(self, routine_name, iteration):
```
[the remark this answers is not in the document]
andrea-pasquale: Implemented

src/qibocal/auto/runcard.py
andrea-pasquale: Actually with this modification the code doesn't work.
I think that it might be related to the fact that we are reading a list directly from a yaml. I need to investigate more about this.

src/qibocal/auto/task.py
```diff
@@ -22,6 +21,9 @@
 DATAFILE = "data.csv"
 """Name of the file where data acquired by calibration are dumped."""
 
+UId = tuple[Id, int]
```
AleCandido, line 24: At a second thought: maybe better `TaskId`, because it is completely unclear what it is outside this scope (and it is used outside).

src/qibocal/auto/task.py
```diff
@@ -22,6 +21,9 @@
...
+"""Useful type for Task."""
```
AleCandido, line 25:
```suggestion
"""Unique identifier for executed tasks."""
```

src/qibocal/auto/runcard.py
andrea-pasquale: To reproduce the error you can just run one of the protocols in this [runcard](https://github.com/qiboteam/qibocal/blob/qubit_local/runcards/actions_qq_auto.yml)

AleCandido: Try with this: https://docs.pydantic.dev/latest/usage/model_config/#smart-union

andrea-pasquale: Thanks, now it works.

andrea-pasquale: I guess that it will be interesting to add it everywhere to have more control in a  separate PR.

src/qibocal/utils.py
```diff
@@ -0,0 +1,11 @@
...
+
+from qibolab.platform import Platform
+from qibolab.qubits import Qubit, QubitId
+
+
+def allocate_qubits(
+    platform: Optional[Platform], qubit_ids: list
+) -> Dict[QubitId, Qubit]:
+    """Convert List[QubitId] -> Dict[QubitId, Qubit] for non-trivial platform."""
+    return {q: platform.qubits[q] for q in qubit_ids if q in platform.qubits}
```
maxhant, line 11:
```suggestion
def allocate_qubits(
    platform: Optional[Platform], qubit_ids: List[QubitId]
) -> Union[List[QubitId], Dict[QubitId, Qubit]]:
    """Convert List[QubitId] -> Dict[QubitId, Qubit] for non-trivial platform."""
    qubits = {}
    for q in qubit_ids:
        if not isinstance(q, list) and q in platform.qubits:
            qubits[q] = platform.qubits[q]
        else:
            qubits[tuple(sorted(q))] = platform.pairs[tuple(sorted(q))]
    return qubits
```

maxhant (changes requested):

src/qibocal/cli/auto_builder.py
```diff
@@ -74,13 +76,13 @@ def routine_name(self, routine, iteration):
         name = routine.replace("_", " ").title()
         return f"{name} - {iteration}"
 
-    def routine_qubits(self, routine_name, iteration):
+    def routine_qubits(self, task_uid: TaskId):
```
AleCandido, line 79: Tiny simplification: in this context (same below), you can also just call it `task`. The type is already telling you that is a `TaskId`, so that's enough :)
```suggestion
    def routine_qubits(self, task: TaskId):
```

src/qibocal/utils.py
```diff
@@ -1,13 +1,11 @@
...
+from typing import Dict, Optional
 
 from qibolab.platform import Platform
 from qibolab.qubits import Qubit, QubitId
 
 
 def allocate_qubits(
-    platform: Platform, qubit_ids: List[QubitId]
-) -> Union[List[QubitId], Dict[QubitId, Qubit]]:
+    platform: Optional[Platform], qubit_ids: list
```
AleCandido, line 8:
```suggestion
    platform: Optional[Platform], qubit_ids: List[QubitId]
```
Keep the type as specific as possible (unless you really want to allow for a whatever list to be passed).

src/qibocal/utils.py
AleCandido: @maxhant I believe that the hint is not exact, what you meant is not that you have a dictionary with either a single qubit or tuples, but rather:
```py
dict[Union[QubitId, Tuple[QubitId, QubitId]], Union[Qubit, QubitPair]]
```
i.e. you have a dict with both type of keys and values allowed.

I know it is unpleasant, because you would like to specify that a certain type of key is only associated to a certain type of value, but this is not possible with dictionaries (it is not something that you can guarantee at compile time).

However, exactly because of this, I'd rather propose to have two dictionaries, so to have something closer to your type:
```py
Tuple[dict[QubitId, Qubit], dict[Tuple[QubitId,QubitId], QubitPair]]
```
Even better if this could be done in two separate functions ^^

maxhant:
@andrea-pasquale 

I've re-merged this branch to #384 and it is crashing again:

```
  File "pydantic/dataclasses.py", line 325, in pydantic.dataclasses._add_pydantic_validation_attributes.new_init
    # had __set_name__ called on it at the right time.
  File "pydantic/dataclasses.py", line 425, in pydantic.dataclasses._dataclass_validate_values
    
pydantic.error_wrappers.ValidationError: 2 validation errors for Runcard
actions -> 0 -> qubits -> 0
  str type expected (type=type_error.str)
actions -> 0 -> qubits -> 0
  value is not a valid integer (type=type_error.integer)
```

AleCandido:
> @andrea-pasquale
> 
> I've re-merged this branch to #384 and it is crashing again:

Thanks @maxhant to report this.

We definitely need `qubits` in the autocalibration tests, not only to remove `Optional`[*], but actually to check that the loading is properly working.

[*]: that of course I like

AleCandido (commented):
For me, this PR is only missing `qubits:` key tests
https://github.com/qiboteam/qibocal/pull/351#issuecomment-1580563872

andrea-pasquale:
> For me, this PR is only missing `qubits:` key tests
> https://github.com/qiboteam/qibocal/pull/351#issuecomment-1580563872

Ok, I can do that.

src/qibocal/auto/task.py
```diff
@@ -76,7 +78,11 @@ def parameters(self):
 
     @property
     def qubits(self):
-        return self.action.qubits
+        return self._qubits
+
+    @qubits.setter
+    def qubits(self, qubits):
+        self._qubits = qubits
```
AleCandido, line 85: Nice: this is a fully useless property (`self.qubits` is exactly following `self._qubits`) and the underlying attribute is even undeclared (not appearing in `__init__`, i.e. the class attributes from which `__init__` is generated)

AleCandido: Can't we just have `qubits` as a class attribute?

src/qibocal/auto/task.py
```diff
@@ -100,13 +106,19 @@ def run(self, folder: Path, platform: Platform, qubits: Qubits) -> Results:
...
-            )
+            if platform is not None:
+                if self.action.qubits:
+                    qubits = allocate_qubits(platform, self.action.qubits)
+            if qubits:
+                self._data: Data = operation.acquisition(
+                    parameters, platform=platform, qubits=qubits
+                )
+                # after acquisition we update the qubit parameter
+                self.qubits = list(qubits.keys())
```
AleCandido, line 118:
```suggestion
                self.qubits = list(qubits)
```
`list()` is creating a list out of an iterable. The default iteration on `dict`s return keys :)

src/qibocal/auto/task.py
```diff
@@ -100,13 +106,19 @@ def run(self, folder: Path, platform: Platform, qubits: Qubits) -> Results:
...
         if operation.platform_dependent and operation.qubits_dependent:
-            if self.qubits:
-                qubits = allocate_qubits(platform, self.qubits)
-            self._data: Data = operation.acquisition(
-                parameters, platform=platform, qubits=qubits
-            )
+            if platform is not None:
+                if self.action.qubits:
+                    qubits = allocate_qubits(platform, self.action.qubits)
+            if qubits:
```
AleCandido, line 113: Is `qubits` a boolean?

Otherwise, formulate a condition with it.
(when is `qubits` falsy?)

src/qibocal/auto/task.py
andrea-pasquale: Yeah, I know. I pushed this version but I'm currently working on a different approach. Thanks!

src/qibocal/auto/task.py
andrea-pasquale: It was just a shortcut to check if the list/dict was empty.

AleCandido:
```suggestion
            if len(qubits) > 0:
```

src/qibocal/auto/runcard.py
```diff
@@ -29,7 +29,7 @@ class Action:
     """Priority level, determining the execution order."""
     qubits: list[QubitId] = Field(default_factory=list)
     """Local qubits (optional)."""
-    update: bool = True
+    update: bool = None
```
AleCandido, line 32: I guess you wanted:
```suggestion
    update: bool = False
```

src/qibocal/auto/task.py
```diff
@@ -31,6 +32,11 @@
 class Task:
     action: Action
     iteration: int = 0
+    qubits: List[QubitId] = field(default_factory=list)
+
+    def __post_init__(self):
+        if len(self.qubits) == 0:
+            self.qubits = self.action.qubits
```
AleCandido, line 39: Actually, why do we need two mechanisms for `qubits`?

I'm now wondering in which cases the `qubits` of the task are different from the `qubits` in the underlying action...

src/qibocal/auto/task.py
```diff
@@ -108,14 +106,14 @@ def run(self, folder: Path, platform: Platform, qubits: Qubits) -> Results:
         path = self.datapath(folder)
         if operation.platform_dependent and operation.qubits_dependent:
             if platform is not None:
-                if self.action.qubits:
-                    qubits = allocate_qubits(platform, self.action.qubits)
-            if qubits:
+                if len(self.qubits) != 0:
+                    qubits = allocate_qubits(platform, self.qubits)
+            if len(qubits) != 0:
```
AleCandido:
```suggestion
                if len(self.qubits) > 0:
                    qubits = allocate_qubits(platform, self.qubits)
            if len(qubits) > 0:
```

As specific as possible :)

src/qibocal/auto/runcard.py
andrea-pasquale: No actually `None` is correct, since this is the local update relative to a protocol, which can be absent.

src/qibocal/auto/task.py
andrea-pasquale: They are not, but since action is also an attribute of the task this is a possible solution.
The reason why I'm adding a qubit attribute (or before with the property) is that I want to achieve the following:
if a local argument qubit is not passed I want it to be overwritten by the global parameter qubits. In this when I perform tests 
https://github.com/qiboteam/qibocal/blob/3b263246bd8f2d2a8e00a4e89eb9a5cb933b5fc2/tests/test_task_options.py#L44
I can check whether the task run with the global qubits or with the local qubits.
I'm sure that there is a better way to do it. This is what I came up with.

src/qibocal/auto/task.py
andrea-pasquale: Already fixed :)

src/qibocal/auto/runcard.py
AleCandido: But then we are back to the wrong type hint: it is not a `bool`, that can only be `True` or `False`, not `None`.

If `update`:
- is a toggle, keep it a `bool`, and if it can be absent, just choose what is the default value (out of the two allowed)
- if it's an object, then it is not a bool, and if it can be absent, it should be `Optional[MyUpdate]`, and you can set to `None`

But `None` can't be a `bool`. And definitely, do not use an `Optional[bool]`, because this is just 3-state logic (and I believe you will end up acting as if it were one of the other two values).

andrea-pasquale: I feel like we are over complicating things. It is an optional value that the user can provide.
I cannot put a default value (`True`/`False`) since by providing a value to a local parameter, the local parameter will overwrite the global option which is something that I don't want. If no local option is provided the program should choose the global option.


 

> And definitely, do not use an `Optional[bool]`, because this is just 3-states logic (and I believe you will end up acting as if it were one of the other two values).

I feel like in this case this is exactly what I want. Since my scenario should be the following:

1. if local is provided it will overwrite the global option
2. if local is not provided (`local == None`) the program should use the global option


> But `None` can't be a `bool`.

I agree with this.

src/qibocal/auto/task.py
AleCandido: Having considered the alternatives, this is a good idea. Sorry if I've been slow :)

src/qibocal/auto/runcard.py
AleCandido: You're creating more cases than needed.

I understand the logic of locally prevailing definition (i.e. scopes), but here I believe it could be much simpler.

- by default, you run the autocalibration to update the runcard, so you can make the assumptions you're always updating
- if you want to disable this option for some routines, you can set the local `update` key to `False`, for each of them that you want to set to `False`
- the only reason to set the global one to `False` is to never update

I believe that you had in mind something like "if I have more `False`, I set the global to `False` and switch the few different ones explicitly, otherwise I do the opposite". And that's why you'd like to set an overwritable default.
But this makes all the logic more complicate, because you're allowing every time for two complementary ways of setting the same configuration (global `False` and set only local `True`s, or the opposite).

Instead, only the local key is required to have the full flexibility, so in principle you could avoid having the global one at all.
But for the special (and arguably relevant [*]) case of disabling update for all, you can provide the global key.

[*]: I actually don't know how much this is going to be used; if no one is using it, we can really avoid defining the global key

andrea-pasquale: To me this is fine. Since we were discussing about this in  #366 with @vodovozovaliza and @wilkensJ. I want to make sure that they also agree.

AleCandido (approved):
There would still be bits to be cleaned up. But part of them are already addressed by #387.

I'm becoming mad with all the changes here and in the other PR. If it works, better to merge both, and at most keep optimizing in a further one (after merging both of them)

tests/test_niGSC_qq.py
```diff
@@ -10,7 +10,7 @@
 def test_command_niGSC():
     path_to_runcard = here / "niGSC.yml"
     test_folder = tempfile.mkdtemp()
-    builder = ActionBuilder(path_to_runcard, test_folder, force=True)
+    builder = ActionBuilder(str(path_to_runcard), test_folder, force=True, update=False)
```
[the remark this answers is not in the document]
andrea-pasquale: Yes, exactly this will be dropped by #387.

src/qibocal/cli/utils.py
```diff
@@ -79,11 +81,12 @@ def update_runcard(rundata, rundata_new, target_compare_dir):
         )
 
 
-def load_yaml(path):
-    """Load yaml file from disk."""
-    with open(path) as file:
-        data = yaml.safe_load(file)
-    return data
+def load_card(card: Union[dict, str]):
+    if isinstance(card, str):
```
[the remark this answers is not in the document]
andrea-pasquale: Also this should be fixed by #387.

src/qibocal/cli/builders.py
```diff
@@ -153,39 +154,36 @@ class ActionBuilder:
...
 
         # Setting format. If absent csv is used.
         self.format = self.runcard.get("format", "csv")
         # Saving runcard
-        shutil.copy(runcard, f"{self.folder}/runcard.yml")
+        if isinstance(runcard, dict):
+            with open(f"{self.folder}/runcard.yml", "w") as file:
+                yaml.dump(runcard, file)
+        else:
+            shutil.copy(runcard, f"{self.folder}/runcard.yml")
```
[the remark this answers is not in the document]
andrea-pasquale: Already addressed in #387. Let me merge it here.

PR #358, opened 2023-05-24
adding tests for python 3.11

scarrazza:
Checklist:
- [ ] Tests are passing.


Edoardo-Pedicillo:
Now tests are passing. I have updated the `pylint` version and forced poetry to install `setuptools` that is removed by poetry during installation (this fix https://github.com/qiboteam/qibocal/actions/runs/5114509833/jobs/9194818064#step:10:78, I am not sure if this is the best way to solve it)

andrea-pasquale (approved):
Thanks @Edoardo-Pedicillo.
Personally I don't see a problem with the current implementation.
If there is a better solution we can implement it later.

PR #369, opened 2023-05-30
Add sweeper types

rodolfocarobene:
Sweepers type (factor, offset, absolute) added as per [qibolab #439](https://github.com/qiboteam/qibolab/pull/459).
For now, I tried to not change the expected behaviour of the routines, in particular:

| Sweeper Parameter     | Sweeper Type |
| ----------- | ----------- |
| Frequency    | OFFSET        |
| Amplitude   | FACTOR        |
| Bias    | ABSOLUTE        |
| Attenuation    | ABSOLUTE        |

For attenuation sweepers I'm not sure if absolute is the expected behaviour, I never used them

Checklist:
- [ ] Tests are passing.


andrea-pasquale (dismissed):
Thanks @rodolfocarobene, LGTM.

I can confirm you that attenuation is swept using absolute values.
For the future we could also think about passing this sweeper type directly as an argument of the protocol to give more freedom to the user as we were discussing yesterday.

Btw, if you want to see if tests are passing here you can change the qibolab branch in the pyproject.toml and then regenerate the new lock file using `poetry lock`.

Edoardo-Pedicillo (dismissed):

rodolfocarobene:
Locally everything is passing, but before merging qibolab needs to be updated so this can wait a bit more.
It's not in draft mode since it should not require modifcations (other then maybe the updated poetry lock)

rodolfocarobene:
@andrea-pasquale @Edoardo-Pedicillo, this PR should be know ready for review and mergable (since we merged the qibolab PR).


[Here](https://github.com/qiboteam/qibocal/actions/runs/5176752174/jobs/9325967196#step:11:27) I broke almost all the niGSC tests when updating the lock file (in particular using latest version of qibo 0.0.14)...
The problem is fixed when fixing in the lock file qibo==0.0.13, but it should be addressed in some other way.

rodolfocarobene:
> The issue with niGSC tests is solved in #380, that was merged in main. Merging the `main` in your branch should solve the problem.

You are right, I thought to have merged main but I forgot to pull... sorry!

andrea-pasquale (approved):

Issue #254, opened 2023-02-22
Dropping pint(-pandas)


andrea-pasquale:
As already discussed in #249 we are planning to drop the `pint-pandas` dependency.
The main motivation is that, although it is interesting to have a mechanism to store units of measure and change them directly on the fly, this comes at the cost of having a lot of complications inside the code especially regarding data saving/loading.
By not using `pint` we will be able to simplify several parts of the code including the calibration routines.
As far as I understood we are planning to use custom `dtypes` following the same approach that we used in qibolab for the [`ExecutionResults`](https://github.com/qiboteam/qibolab/blob/1ae75a8a21aef417f589cf95325edcac0b9d135a/src/qibolab/result.py#L41) object.

We can use this issue to further elaborate on which is the best solution to adopt for qibocal.

aorgazf:
Hi,
I would like to add another reason for dropping `pint-pandas`, which is that it dramatically slows the performance of pandas.
The problems we have with the plotting functions would go away just by `de-quantifying` the dataframe before working with it:
```python
data.df[quantities] = data.df[data.quantities].pint.dequantify()
```

andrea-pasquale:
I've a proposal for a new data structure when storing the data acquired from a calibration routines.
Looking at the calibration routines I think that we need a structure that should be able to store

1. Qubit(Pair) specific quantities which may include voltages, frequencies, amplitudes...
2. Non Qubit(Pair) specific quantities which may include initial parameters (to be able to pass them to the results) and platform attributes which will be again processed in the results such as the resonator type.

The current approach involving pint is the following:

```py
class ResonatorSpectroscopyData(DataUnits):
    """ResonatorSpectroscopy acquisition outputs."""

    def __init__(self, resonator_type, power_level=None, amplitude=None):
        super().__init__(
            "data",
            {"frequency": "Hz"},
            options=["qubit"],
        )
        self._resonator_type = resonator_type
        self._power_level = power_level
        self._amplitude = amplitude

    @property
    def resonator_type(self):
        """Type of resonator"""
        return self._resonator_type

    @property
    def power_level(self):
        """Resonator spectroscopy power level"""
        return self._power_level

    @property
    def amplitude(self):
        """Readout pulse amplitude common for all qubits"""
        return self._amplitude
```
 in which 1 are stored through a `DataUnits` object which is basically a pandas dataframe + pint and 2 are currently not stored but they are used in runtime by the results object.

[...]

PR #372, opened 2023-06-01
Removing pint

andrea-pasquale:
A possible approach for removing pint using dataclasses, record arrays and custom dtypes.
For more info check https://github.com/qiboteam/qibocal/issues/254#issuecomment-1571503394.

Checklist:
- [x] Tests are passing.


src/qibocal/protocols/characterization/resonator_spectroscopy.py
```diff
@@ -45,53 +47,65 @@ def __post_init__(self):
...
+    power_level: PowerLevel
+    """Power regime of the resonator."""
+    resonator_type: str
+    """Resonator type."""
+    amplitudes: list
+    """Amplitudes provided by the user."""
+    data: Dict[QubitId, npt.NDArray[ResSpecType]] = field(default_factory=dict)
+    """Raw data acquired."""
+
+    def load_qubit(self, qubit, freq, msr, phase):
```
stavros11, line 83: I think this function name is a bit misleading, it says `load` while it is actually saving (creating the data array that will be saved). Maybe something like `add_qubit_data`/`add_qubit_output` would be better?

Alternatively we could have a seperate function (say `create_output`) that constructs the array given `[freq, msr, phase]` and implement the `__setitem__` for `ResonatorSpectroscopyData`. In the end in the acquisition you would do
```py
for qubit in qubits:
        data[qubit] = create_output(
            msr=result.magnitude,
            phase=result.phase,
            freq=delta_frequency_range + ro_pulses[qubit].frequency,
        )
```

src/qibocal/protocols/characterization/resonator_spectroscopy.py
```diff
@@ -45,53 +47,65 @@ def __post_init__(self):
...
-        """Readout pulse amplitude common for all qubits"""
-        return self._amplitude
+    def qubits(self):
+        """Access qubits from data structure."""
+        return [q for q in self.data]
+
+    def __getitem__(self, qubit):
+        return self.data[qubit]
+
+    def to_csv(self, path):
```
stavros11, line 99: Is this storing to csv? I think you are just storing to `npy` (binary format). 
I am not sure if there is a numpy method to dump the array with the custom dtype directly to csv (readable) or we need to do some kind of wrapping ourselves.

stavros11 (commented):
Thanks for this proposal, it looks good to me, certainly much better than the current solution. I would wait for one or two other people to confirm they like this before you propagate this to other routines, but from my side I agree, other than some small comments below.

A general suggestion since this is a [POC], maybe it will be easier for others to review if you only changed one file in this branch. If I understand correctly the only relevant file regarding the [POC] is the resonator_spectroscopy.py, the rest is just replacing `Union[str, int]` -> `QubitId`. Not very important if we know where to look, but in the future it may be better to seperate renames (which cause large file changes) from actual implementations which need to be reviewed in more detail.

AleCandido:
Thanks @andrea-pasquale for the proposal, I'm going to take a look.

> A general suggestion since this is a [POC], maybe it will be easier for others to review if you only changed one file in this branch. If I understand correctly the only relevant file regarding the [POC] is the resonator_spectroscopy.py, the rest is just replacing `Union[str, int]` -> `QubitId`. Not very important if we know where to look, but in the future it may be better to seperate renames (which cause large file changes) from actual implementations which need to be reviewed in more detail.

And thanks @stavros11 for the comment.
In general, this kind of tiny tasks that are completely decoupled are a perfect fit for stand-alone PRs. You make them separate, it's simple, non-controversial, and thus easy to merge. And it's not going to hide the important changes in some other PRs :)

andrea-pasquale:
> A general suggestion since this is a [POC], maybe it will be easier for others to review if you only changed one file in this branch. If I understand correctly the only relevant file regarding the [POC] is the resonator_spectroscopy.py, the rest is just replacing `Union[str, int]` -> `QubitId`. Not very important if we know where to look, but in the future it may be better to seperate renames (which cause large file changes) from actual implementations which need to be reviewed in more detail.

You are totally @stavros11, sorry. I will be more careful :)





> Since it was a good idea, and it was already a separate commit, I made the PR myself :P

Thanks @AleCandido.
Now it is much better indeed.

src/qibocal/protocols/characterization/resonator_spectroscopy.py
andrea-pasquale: This only a prototype. Here I had to put `to_csv` since in the executor currently I had hardcoded a `to_csv` function.
https://github.com/qiboteam/qibocal/blob/9f800393cfda17e8b79d0bd7c6c6e0c1da87f668/src/qibocal/auto/task.py#L104



> I think you are just storing to `npy` (binary format).

Indeed this is the case.

src/qibocal/protocols/characterization/resonator_spectroscopy.py
andrea-pasquale: > I think this function name is a bit misleading, it says `load` while it is actually saving (creating the data array that will be saved). Maybe something like `add_qubit_data`/`add_qubit_output` would be better?

Yes, I also was not happy with `load_qubit`. It was just a placeholder for time being. `add_qubit_data` sounds good to me.



> Alternatively we could have a seperate function (say `create_output`) that constructs the array given `[freq, msr, phase]` and implement the `__setitem__` for `ResonatorSpectroscopyData`


I've also thought about this. It could an option. I don't have a strong preference between this and the option above.

src/qibocal/protocols/characterization/resonator_spectroscopy.py
AleCandido: I would just drop the idea of supporting multiple formats, it is an over-head and not very useful. There are already other libraries that do this (and that you would use to implement the various formats), let's delegate this to the user.

I'd propose to go back to just `save()`.

andrea-pasquale:
> I would also keep a general data structure like `DataUnits`

`DataUnits` without pint is just a `pandas.DataFrame`. At this point I would suggest to inherit directly the `DataFrame` class.

AleCandido:
> At this point I would suggest to inherit directly the `DataFrame` class.

Or even just use `DataFrame` directly, without inheriting. But I would leave the pleasure to the user to transform a `np.ndarray` in a `pd.DataFrame` (that is a rather simple task, since it's just one call), and only keep arrays inside `qibocal`.

Once that data are saved in one format, and you can simply load them, converting it is definitely a simple task (since Numpy and Pandas are giving you the tools to do it anyhow).

I'm not against other data formats, but maintaining many options inside a single library/application is expensive. A package should focus on its core task, that can only be done by it. Then, it can always be composed with other ones to tackle specific use cases.

andrea-pasquale:
If there are no complains I would just change `load_qubit` to `add_qubit_data` and I will propagate those changes for all the routines.
I think that I would convert the global file into a json to be consistent with #395.

src/qibocal/protocols/characterization/resonator_spectroscopy.py
AleCandido: I believe `_data` to be redundant, but unfortunately `add_` is a bit generic (that's why you need then `_data`).

`record_`/`register_`/`import_`/...?

src/qibocal/protocols/characterization/resonator_spectroscopy.py
```diff
@@ -60,33 +61,51 @@ class ResonatorSpectroscopyResults(Results):
...
+    def __getitem__(self, qubit):
+        return self.data[qubit]
+
+    def save(self, path):
+        """Store results."""
+        self.power_level = self.power_level.name
+        my_dict = asdict(self)
+        my_dict.pop("data")
+        with open(path / "conf.yml", "w") as outfile:
+            yaml.dump(my_dict, outfile)
```
AleCandido, line 105: I know I said JSON, but it will be just the same, so I'll write for YAML because it's the current one, and the proposal will be slightly closer:
```suggestion
        (path / "conf.yml").write_text(yaml.dump(my_dict))
```
https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_text

src/qibocal/protocols/characterization/resonator_spectroscopy.py
andrea-pasquale: `register_qubit` sounds good to me.

src/qibocal/protocols/characterization/resonator_spectroscopy.py
```diff
@@ -60,33 +61,51 @@ class ResonatorSpectroscopyResults(Results):
...
+
+        for q in self.data:
+            np.save(path / f"data_q{q}.npy", self[q])
```
AleCandido, line 108: I would suggest writing all of them in a single file.

You could concatenate and make a single array, but it is slightly more annoying if you then are using them as separate arrays.
Otherwise, you just have to use a `.npz` file (despite the name, it is not compressed, it's just a single file containing multiple independent arrays).

https://numpy.org/doc/stable/reference/generated/numpy.savez.html#numpy.savez

andrea-pasquale: Thanks @AleCandido. In https://github.com/qiboteam/qibocal/pull/372/commits/cd0e236a9df12dc7aa1ab3b3d2e1193112dcd62b I tried to implement the dumping to both json and npz. Let me know if see something wrong :)

src/qibocal/auto/operation.py
```diff
@@ -45,6 +52,16 @@ def load(cls, parameters):
 class Data:
     """Data resulting from acquisition routine."""
 
+    @staticmethod
+    def to_npz(path, data_dict: dict):
+        """Helper function to use np.savez while converting keys into strings."""
+        np.savez(path / DATAFILE, **{str(i): data_dict[i] for i in data_dict})
+
+    @staticmethod
+    def to_json(path, data_dict: dict):
```
AleCandido, line 61:
```suggestion
    def to_npz(path, data: dict):
        """Helper function to use np.savez while converting keys into strings."""
        np.savez(path / DATAFILE, **{str(i): data_dict[i] for i in data_dict})

    @staticmethod
    def to_json(path, data: dict):
```

`_dict` is redundant, it's already in the type.

Why do you want to have them as `staticmethod`? They could be just regular functions...

andrea-pasquale: > `_dict` is redundant, it's already in the type.

Sure





> Why do you want to have them as `staticmethod`? They could be just regular functions...

I agree. I wanted to do it this way just to avoid to import for all the protocols these two functions.
I don't if I should call it laziness or just a strategy to avoid some imports.

src/qibocal/protocols/characterization/resonator_spectroscopy.py
AleCandido: It seems great :)

Just test it ^^

src/qibocal/protocols/characterization/allxy/allxy.py
```diff
@@ -28,14 +29,39 @@ class AllXYResults(Results):
...
+    def qubits(self):
+        """Access qubits from data structure."""
+        return [q for q in self.data]
+
+    def __getitem__(self, qubit):
+        return self.data[qubit]
+
+    def save(self, path):
+        """Store results."""
+        self.to_npz(path, self.data)
```
AleCandido, line 64: Maybe it's worth to make `.data` part of the interface, to avoid repeating these methods over and over for all the routines.

Is it possible?

andrea-pasquale: I was thinking about it, also I don't like to repeat these methods every time.
The only problem in making them part of the interface is that we then force the user to use this interface (?)
In theory he can overwrite all the methods or simply not inherit from `Data`.
I was planning to clean up these methods.

AleCandido: > The only problem in making them part of the interface is that we then force the user to use this interface (?)
> In theory he can overwrite all the methods or simply not inherit from `Data`.

Well, enforcing its usage is the whole point of an interface. If you can make it difference, it's not an interface any longer ^^
But the advantage is that, if you have a bunch of different objects with the same interface, you could standardize more operations on them, despite being different.

About inheriting or not, that's a separate discussion, since both nominal and structural subtyping are both possible.
However, if you do not inherit, you might have to repeat some code. But if it might be worth in certain cases (but at the moment we really rely on inheritance, even in type hints, using bounds, if we wanted to go structural, we should start defining protocols).

AleCandido: > Btw, `qubits` and `__getitem__` are different for some routines. See for example

Well, this is not really different:
```py
    def __getitem__(self, qubit_state: tuple):
        qubit, state = qubit_state
        return self.data[qubit, state]
```
is equivalent to:
```py
    def __getitem__(self, qubit_state: tuple):
        return self.data[qubit_state]
```
Names do not matter, and if you do not specify the type (i.e. make the type a union) then the two things are still compatible.

AleCandido: > Finally, also `save` is not completely standardized given that in some routines a json is not needed since all the data are just arrays.

Well, if not needed we can leave it empty :)

tests/runcards/single_routines.yml
```diff
@@ -118,8 +118,8 @@ actions:
       max_amp_factor: 4.0
       step_amp_factor: 0.1
       pulse_length: 30
-      relaxation_time: 0
-      nshots: 10
+      nshots: 1024
+      relaxation_time:
```
AleCandido: Without a value is an empty string, not 0

src/qibocal/protocols/characterization/allxy/allxy.py
andrea-pasquale: > Well, if not needed, we can leave it empty :)

Another option would be to check if the dict that it going to be dumped to json is empty.

andrea-pasquale: Also about the `save` method is using the `data` attribute which is defined in the child class (which will make pylint complain).
I could put in the `Data` class but then I will lose the specific type hint that I define in each child class...

andrea-pasquale: Or I could just put something like `data: dict` and then it will be overwritten in the child class actually.

src/qibocal/auto/operation.py
```diff
@@ -45,6 +52,16 @@ def load(cls, parameters):
...
+        """Helper function to dump to json in JSONFILE path."""
+        (path / JSONFILE).write_text(json.dumps(data_dict, indent=4))
```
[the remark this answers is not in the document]
andrea-pasquale: If I understood correctly here, your suggestion is just to remove the indentation and if the user wants a readable format you suggest to use a json formatter.
At least for now this json are very short (just a few keys), I think that the indentation should not hurt as much for now.
If these files will increase we always turn off the indentation.

andrea-pasquale:
I've added a `deploy on hardware` label given that this PR has changed a lot of things in the protocols.
Apart from that it is ready to review, currently I'm testing everything with the dummy and tests seem to pass.

src/qibocal/auto/operation.py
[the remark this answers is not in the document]
AleCandido: Not immediately, it was not even my proposal, just keep that in mind, since it could save a relevant bunch of bytes (every whitespace is one ASCII character, so 1 byte in UTF-8, and indentation uses a bunch of them).

src/qibocal/protocols/characterization/allxy/allxy.py
AleCandido: > The `save` method is using the `data` attribute which is defined in the child class (which will make pylint complain). I could put in the `Data` class but then I will lose the specific type hint that I define in each child class...

In case of doubt, make it `Any` (if you have too, otherwise leave it empty, that is the same), you could also overwrite in the child (for as long as it is compatible). I believe there might be a suitable (and not too complex) type hint that is more specific, but we can see later on.

AleCandido: > Another option would be to check if the dict that it going to be dumped to json is empty.

This is possible, but it's a more special case wrt to the empty file. Every special case is more code to handle, every time you have to.
However, the two of them are simple enough, so they might be both good (in this case the trade-off is not so clear, even because it's not so crucial), you decide your favorite.

andrea-pasquale: > Another option would be to check if the dict that it going to be dumped to json is empty.

Lets stick with this for time being.

src/qibocal/protocols/characterization/allxy/drag_pulse_tuning.py
```diff
@@ -173,30 +188,29 @@ def _fit(data: DragPulseTuningData) -> DragPulseTuningResults:
...
-
-        pguess = [
-            0,  # Offset:    p[0]
-            beta_params.values[np.argmax(voltages)]
-            - beta_params.values[np.argmin(voltages)],  # Amplitude: p[1]
-            4,  # Period:    p[2]
-            0.3,  # Phase:     p[3]
-        ]
+        qubit_data = data[qubit]
+        voltages = qubit_data.msr * 1e6
```
rodolfocarobene, line 197: I would define a constant instead of using `1e6`  so that is a bit clearer what you are doing (`UV_TO_V` or whatever)

src/qibocal/protocols/characterization/coherence/t1.py
```diff
@@ -146,39 +163,38 @@ def _plot(data: T1Data, fit: T1Results, qubit):
...
-    qubit_data = data.df[data.df["qubit"] == qubit]
+    qubit_data = data[qubit]
+    waits = qubit_data.wait
 
     fig.add_trace(
         go.Scatter(
-            x=qubit_data["wait"].pint.to("ns").pint.magnitude,
-            y=qubit_data["MSR"].pint.to("uV").pint.magnitude,
+            x=waits,
+            y=qubit_data.msr * 1e6,
```
rodolfocarobene, line 172: Same constant than before... I don't know how many other times it appears, but maybe could be interesting to have it stored in a common place?

src/qibocal/protocols/characterization/dispersive_shift.py
```diff
@@ -252,12 +248,12 @@ def _plot(data: DispersiveShiftData, fit: DispersiveShiftResults, qubit):
         )
     ):
         opacity = 1
-        frequencies = q_data.df["frequency"].pint.to("GHz").pint.magnitude.unique()
+        frequencies = q_data.freq / 1e9
```
rodolfocarobene, line 251: better a constant

src/qibocal/protocols/characterization/dispersive_shift.py
```diff
@@ -312,23 +308,23 @@ def _plot(data: DispersiveShiftData, fit: DispersiveShiftResults, qubit):
     )
 
     fig.add_vline(
-        x=fit.best_freq[qubit],
+        x=fit.best_freq[qubit] / 1e9,
         line=dict(color="orange", width=3, dash="dash"),
         row=1,
-        col=2,
+        col=1,
```
rodolfocarobene, line 314: I didn't try this routine, but this change seems a bit weird...
Was it wrong before?

src/qibocal/protocols/characterization/flux_depedence/utils.py
```diff
@@ -18,12 +18,12 @@ def flux_dependence_plot(data, fit, qubit):
...
-
+    frequencies = qubit_data.freq / 1e9
     fig.add_trace(
         go.Heatmap(
-            x=qubit_data["frequency"].pint.to("GHz").pint.magnitude,
-            y=qubit_data["bias"].pint.to("V").pint.magnitude,
-            z=qubit_data["MSR"].pint.to("uV").pint.magnitude,
+            x=frequencies,
+            y=qubit_data.bias,
+            z=qubit_data.msr * 1e4,
```
rodolfocarobene, line 26: what unit of measure is this? Elsewhere we had 1e6

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -48,36 +48,38 @@ class RamseyResults(Results):
...
+    detuning_sign: int
+    """Sign for induced detuning."""
+    qubit_freqs: Dict[QubitId, float] = field(default_factory=dict)
+    """Qubit freqs for each qubit."""
+    data: Dict[QubitId, npt.NDArray[RamseyType]] = field(default_factory=dict)
+    """Raw data acquired."""
+
+    def register_qubit(self, qubit, wait, msr, phase):
+        """Store output for single qubit."""
+        shape = (1,) if np.isscalar(wait) else wait.shape
```
rodolfocarobene, line 74: Not sure what's happening here, maybe a comment would be nice

Edoardo-Pedicillo:
Thank @andrea-pasquale, I have only one doubt concerning the dumping: with your implementation the global data are saved in separate files, since most of them are parameters that can be found in the runcards, there is a duplication of information in the report folder. From my point of view, the users should not move or touch directly the files in the report and when they want to load a `Data` class, Qibocal should be able to take what it needs from the runcards and the .npz. 

About the attribute `data`, I am not sure that we need a dictionary, it could be a suitable ndarray where the first axis runs on the qubits. If we want the qubit ID, we could evaluate it from the runcards.

rodolfocarobene (changes requested):
Thanks for the nice PR @andrea-pasquale! 

I have added a couple of minor comments. In particular, I didn't write it every time, but:
* would be nice to have the conversion constants in a shared file
* would be nice to use the built-in types when possible

In any case, these are minor things, I'll try to run some routine to check everything works :-)

pyproject.toml
```diff
@@ -19,7 +19,6 @@ python = ">=3.9,<3.12"
 qibolab = { git = "https://github.com/qiboteam/qibolab.git", branch = "main" }
 qibo = "^0.1.9"
 pandas = "^1.4.3"
-Pint-Pandas = "^0.3"
```
AleCandido, line 22: :tada:

src/qibocal/auto/task.py
```diff
@@ -19,9 +19,6 @@
 
 """
 
-DATAFILE = "data.csv"
-"""Name of the file where data acquired by calibration are dumped."""
-
 TaskId = tuple[Id, int]
```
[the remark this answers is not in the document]
andrea-pasquale: I think that it will be more clear once we have also a validation step. Basically the `int` should be the amount of time that you repeat a routine (if the fit `fails`).

src/qibocal/protocols/characterization/allxy/allxy.py
```diff
@@ -102,16 +118,10 @@ def _acquisition(
         )
 
         # retrieve the results for every qubit
-        for ro_pulse in ro_pulses.values():
-            z_proj = 2 * results[ro_pulse.serial].probability(0) - 1
+        for qubit in qubits:
+            z_proj = 2 * results[ro_pulses[qubit].serial].probability(0) - 1
```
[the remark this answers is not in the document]
andrea-pasquale: In the literature for the allXY plot usually on the y axis it is shown the expectation value of Z.

src/qibocal/protocols/characterization/allxy/drag_pulse_tuning.py
```diff
@@ -139,19 +153,20 @@ def _acquisition(
...
+            data.register_qubit(qubit, r1.magnitude - r2.magnitude, beta_param)
+            # r = {
+            #     "MSR[V]": r1.magnitude - r2.magnitude,
+            #     "i[V]": r1.voltage_i - r2.voltage_i,
+            #     "q[V]": r1.voltage_q - r2.voltage_q,
+            #     "phase[rad]": r1.phase - r2.phase,
+            #     "beta_param[dimensionless]": beta_param,
+            #     "qubit": ro_pulse.qubit,
+            # }
+            # data.add_data_from_dict(r)
```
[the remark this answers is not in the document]
andrea-pasquale: Done, thanks.

src/qibocal/protocols/characterization/allxy/drag_pulse_tuning.py
andrea-pasquale: Done

src/qibocal/protocols/characterization/ramsey.py
andrea-pasquale: Done

src/qibocal/auto/task.py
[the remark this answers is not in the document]
AleCandido: As @andrea-pasquale said, it is part of the *exceptional flow*, in which you may have loops (even when you do not fail, you might iterate to improve, or even go back to a previous step, if you remember Sergi's diagram).

src/qibocal/protocols/characterization/dispersive_shift.py
rodolfocarobene: Indeed it was wrong before :-)

rodolfocarobene (approved):

src/qibocal/protocols/characterization/dispersive_shift.py
andrea-pasquale: Yes, sorry I forgot to reply to this point. Indeed it was wrong before.

src/qibocal/auto/operation.py
stavros11: I am also not sure why these need to be `@staticmethod`. Is there any routine where you use them with a `data_dict` that is not `self.data` and `self.global_params_dict`?

stavros11 (approved):
Thanks @andrea-pasquale, this is certainly an important simplifcation.

andrea-pasquale:
Thanks @stavros11.
You are right. Before, I had to do something like this because `global_params_dict` and `data` were defined only in the child classes. Now it doesn't make sense to have these methods as `staticmethods`.

PR #373, opened 2023-06-01
Error bars and uncertainties for Randomized Benchmarking

vodovozovaliza:
This PR implements semi-parametric bootstrap resampling to get uncertainties of the Randomised Benchmarking fit as described in Section 2.6.3 of https://arxiv.org/pdf/1811.10040.pdf.

2 optional variables are added to the `RBParameters` that can be specified in the runcard:
- `n_bootstrap`: number of bootstrap iterations, can be set to `0`. Default is `100`.
- `uncertainties`: method of computing the uncertainties. If `'std'`, the errors are calculated as the standard deviation. If a number from 0 to 100, computes the confidence interval. If `None`, does not compute the errors. Default is `95`. 

An example of the report with the data from the hardware, `n_bootstrap=1000` and `uncertainties=95`:
srb_error

Overview of the resampling method:

1. Get the original data (probabilities of random circuits) for each `x` (sequence length) and compute the average `y(x)`.
2. For each `x`, resample the probabilities with replacement and with the same number of points.
3. For each resampled probability `p`, generate "corrected" samples (0s and 1s) of size `nshots` from the binomial distribution and get the new probability `p'`. Then compute the new average `y'(x)`.
4. Repeat `n_bootstrap` times. 
5. Repeat for each `x`, perform the fitting and record the fitting parameters.
6. The error bars of `y` and the uncertainties of fitting parameters can be computed from the bootstrap distribution of their estimates according to the specified `uncertainties` method.

The diagram below displays steps 1-4 for `x=m` to get the error bar of `y(m)`:

Screenshot 2023-06-02 at 15 06 42

Checklist:
- [ ] Tests are passing.


vodovozovaliza:
The code in this PR will be changed to match the implementation in #391.

vodovozovaliza:
@wilkensJ @Jacfomg @andrea-pasquale @Edoardo-Pedicillo this is ready for review

src/qibocal/protocols/characterization/randomized_benchmarking/standard_rb.py
```diff
@@ -43,6 +44,17 @@ class StandardRBParameters(Parameters):
...
     nshots: int
     """For each sequence how many shots for statistics should be performed."""
+    uncertainties: Union[str, float] = 95
+    """Method of computing the error bars and uncertainties of the data. If ``None``, does not
+    compute the errors. If ``"std"``, computes the standard deviation. If ``float`` or ``int``
+    between 0 and 100, computes the corresponding confidence interval. Defaults to ``95``."""
+    n_bootstrap: int = 100
+    """Number of bootstrap iterations for the fit uncertainties and error bars.
+    If ``0``, gets the fit uncertainties from the fitting function and the error bars
+    from the distribution of the measurements. Defaults to ``1``."""
```
Jacfomg, line 54: defaults to 1 ?

Jacfomg (commented):
Maybe it's just me but can we avoid calling uncertainties errors in a couple places. Apart from that  it looks good from my side. I would like to test in hardware first at the end of this week if able.

src/qibocal/protocols/characterization/randomized_benchmarking/utils.py
```diff
@@ -86,6 +87,105 @@ def random_clifford(qubits, seed=None):
...
+        if isinstance(precision, int) is False:
+            raise_error(
+                TypeError,
+                f"`precision` must be of type int. Got {type(precision)} instead.",
+            )
+        if precision < 0:
+            raise_error(
+                ValueError,
+                f"``precision`` cannot be negative. Got {precision}.",
+            )
```
Edoardo-Pedicillo, line 141:
```suggestion
```
The doc is enough.

src/qibocal/protocols/characterization/randomized_benchmarking/standard_rb.py
```diff
@@ -198,13 +260,61 @@ def _fit(data: RBData) -> StandardRBResult:
...
+        )
+
+    # Fit the initial data and compute error bars
+    y = [np.mean(y_row) for y_row in y_scatter]
+    error_y = data_errors(y_estimates, uncertainties, symmetric=False, data_median=y)
+    sigma = (
+        np.max(error_y, axis=0)
+        if error_y is not None and len(error_y.shape) == 2
+        else error_y
+    )
```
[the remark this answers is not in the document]
vodovozovaliza: I replaced `error_y` with `error_bars`, do you think this name would be clear or should I change it to `uncertainty_y`?

Edoardo-Pedicillo (commented):
Thank @vodovozovaliza

Edoardo-Pedicillo (commented):
I missed a comment

src/qibocal/protocols/characterization/randomized_benchmarking/utils.py
```diff
@@ -86,6 +87,102 @@ def random_clifford(qubits, seed=None):
...
+    Returns:
+        int: position of the first significant digit or ``2`` if the given number
+        is integer, ``inf`` or ``0``. Returns ``-1`` if ``number`` is ``None``.
+    """
+
+    position = (
+        ceil(-log10(abs(np.real(number))))
+        if not isinf(np.real(number)) and np.real(number) != 0
+        else 2
+    )
```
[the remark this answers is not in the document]
vodovozovaliza: I only use this for numbers ≤ 1 for now, but feel free to suggest how the other numbers should be handled in your opinion

src/qibocal/protocols/characterization/randomized_benchmarking/standard_rb.py
Jacfomg: Thanks !

src/qibocal/protocols/characterization/randomized_benchmarking/utils.py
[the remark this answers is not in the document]
Jacfomg: It would be still similar. You first convert to scientific notation so you may have something like 3.2145 \pm 0.5426e7 that could be displayed as 3.21 \pm 0.54 e7. But I would not worry about it for this PR.

tests/test_task_options.py
```diff
@@ -27,7 +27,17 @@
...
+        {
+            "id": "standard rb",
+            "priority": 0,
+            "operation": "standard_rb",
+            "parameters": {
+                "depths": [1, 5, 10],
+                "niter": 3,
+                "nshots": 10,
+            },
+        },
```
andrea-pasquale, line 40: Here I've added only the first routine since the test should be the same for each routine since we are only testing things at the level of `Task`. Is there a particular reason on why you decided to add it here?

src/qibocal/auto/runcard.py
```diff
@@ -45,7 +45,7 @@ class Runcard:
     """Structure of an execution runcard."""
 
     actions: List[Action]
-    qubits: List[QubitId] = Field(default_facotry=list)
+    qubits: List[QubitId]
```
andrea-pasquale, line 48: I remember that we discussed about this. Why did you decide to remove the default value?

andrea-pasquale (approved):
Thanks @vodovozovaliza. It looks good to me. I have only two small comments.
There are some conflicts which were generated after merging #372. Let me know if you have any problem fixing them.

Concerning #372, now there is no `pint` dep anymore. For the other protocols we are now using `dataclasses` with an attribute `data` which a custom `np.array` to store qubit-dependent features. See for example here https://github.com/qiboteam/qibocal/blob/2222ce76f50b21a073807280fd79916354620e44/src/qibocal/protocols/characterization/resonator_spectroscopy.py#L72-L82

There are also some default methods for saving the new data structure
https://github.com/qiboteam/qibocal/blob/2222ce76f50b21a073807280fd79916354620e44/src/qibocal/auto/operation.py#L54-L90

For the standard rb you are using pandas so everything should work as expected. If you feel like switching to this approach with `dataclasses` and `np.array` you can follow what we are doing in other protocols.

src/qibocal/auto/runcard.py
vodovozovaliza: The previous implementation did not make `qubits` optional and led to this error:
```python
pydantic.error_wrappers.ValidationError: 1 validation error for Runcard
qubits
  field required (type=value_error.missing)
``` 
This had 2 solutions:
1. Add `Optional[List[QubitId]]` to the `qubits` type so that they can be optional in the runcard.
2. Remove the default value so that `qubits` are non-optional.

After the discussion with @andrea-pasquale, we chose the latter, so that a user always has to specify global `qubits` in the runcard, but can overwrite them with local `qubits` for a specific action. Now the error raised for not specifying global `qubits` will be
```python
TypeError: Runcard.__init__() missing 1 required positional argument: 'qubits'
```

vodovozovaliza: Update: making global qubits optional also leads to this error now
```python
File "/qibocal/src/qibocal/web/templates/template.html", line 195, in top-level template code
    {% set figures, fitting_report = report.plot(routine, iteration) %}
^^^^^^^^^^^^^^^^^
TypeError: ReportBuilder.plot() takes 2 positional arguments but 3 were given
```

tests/test_task_options.py
vodovozovaliza: This test was added for the coverage, to check the case of `platfrom=None` in `Task.run()` (lines 109-110):
https://github.com/qiboteam/qibocal/blob/046798bf41d0a0a54f0263f00846c2102813ea9a/src/qibocal/auto/task.py#L106-L110
However, it seems like only `standard_rb` works with `platfrom=None` at the moment, so I had to add it here.

andrea-pasquale: I see. It makes sense.

vodovozovaliza:
> For the standard rb you are using pandas so everything should work as expected. If you feel like switching to this approach with dataclasses and np.array you can follow what we are doing in other protocols.

This is a good idea, can we leave this for a separate PR?

andrea-pasquale:
> > For the standard rb you are using pandas so everything should work as expected. If you feel like switching to this approach with dataclasses and np.array you can follow what we are doing in other protocols.
> 
> This is a good idea, can we leave this for a separate PR?

Sure, it is just a suggestion.

Jacfomg:
I know, it looks good to me but I want to test on the hardware as well.

vodovozovaliza:
Result from a non-calibrated qubit:
tii1q_b1_strb

andrea-pasquale:
Thank you very much @vodovozovaliza.
From my side we are ready to merge, do you agree @Jacfomg?

Jacfomg (approved):

PR #375, opened 2023-06-02
Improve punchout routine

andrea-pasquale:
This PR introduces some modifications to `resonator_spectroscopy` and `resonator_punchout`.

Currently in resonator spectroscopy there is the option to specify an amplitude value which will be automatically set in the platform runcard. I decided to do this given that the punchout fitting is not reliable yet.
After talking with @rodolfocarobene we noticed that this is not the expected behavior when running the resonator spectroscopy at high power, i.e. in the high power regime the amplitude provided should be written in the runcard. This is true especially when _recalibrating_.

This was done at the beginning because we were expecting that the user would start with a completely empty runcard. In this way the punchout should be done starting from high power and moving to low power. However, given that the scan is performed with equally spaced points it become difficult to see the low power regime. 

This PR fixes this issue by:

1. not writing to the runcard the amplitude when running resonator spectroscopy at high power.
2. specifying an optional argument `amplitude` in resonator punchout which corresponds to the initial guess for the readout amplitude 

@DavidSarlle if you agree I can propagate these changes also for the routines involving attenuation.

Checklist:
- [x] Tests are passing.


rodolfocarobene:
Thanks @andrea-pasquale , I'll have to try it but it LGTM :-)

igres26 (approved):
Looks good to me!

PR #381, opened 2023-06-05
Two qubits gates routines

rodolfocarobene:
Converting Maxime routines for two qubits gates (chevron and landscape/virtual z) to the new layout.

This is still a work in progress and I just started with the chevron, however I already have a question @andrea-pasquale, @Edoardo-Pedicillo: to the `_acquisition` function by default gets passed a `qubits` object and, for `_plot` I suppose that the function gets called separately for every qubit. 
This is a bit weird for two qubits routine since the important things are the pairs (parameter of the routine), not the single qubits. 
I don't know what to do, for example, in the plotting function where for every row I need to draw plots for two qubits.

Checklist:
- [x] Tests are passing.


andrea-pasquale:
> This is still a work in progress and I just started with the chevron, however I already have a question @andrea-pasquale, @Edoardo-Pedicillo: to the `_acquisition` function by default gets passed a `qubits` object and, for `_plot` I suppose that the function gets called separately for every qubit.

Thanks @rodolfocarobene.
Yes, this is something that we should need to address. Here the plotting function takes a pair of qubits, right?
Actually after we merge #351 the report will look at each element of the arguments qubits.
https://github.com/qiboteam/qibocal/blob/85e03d981368fcb5a3c20aec03354fe23abb4982/src/qibocal/web/templates/autocalibration.html#L155
If for the two qubit gates you specify that the qubits inputs is a list of pairs it could work I think...
Maybe we just need to rename qubit to something else.

andrea-pasquale:
You can give it a try by merging #351 which should be ready to merge here.

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
```diff
@@ -0,0 +1,251 @@
...
+    amplitude_factor_step: float
+    """Amplitude step."""
+    duration_min: float
+    """Duration minimum."""
+    duration_max: float
+    """Duration maximum."""
+    duration_step: float
+    """Duration step."""
+    pairs: Union[list[Tuple[int], Tuple[int]]]
+    """Pair(s) of qubit to probe."""
```
andrea-pasquale, line 33: Following my suggestion this should be:

```suggestion
    qubits: List[List[QubitId,QubitId]]
    """Pair(s) of qubit to probe."""
```
where you can import `QubitId` from qibolab.qubits

rodolfocarobene:
Thanks @andrea-pasquale, I'll proceed merging #351. 
The only problem is that I will not be able to test it with actual hardware, but ok

andrea-pasquale:
If it works with dummy you can try with that.

rodolfocarobene:
@andrea-pasquale, @Edoardo-Pedicillo What should I do to add tests for the routine? I saw that there are several "runcards" in the tests directory, but I am not sure which I should modify

andrea-pasquale:
A simple way to test this routine is to add it in [here](https://github.com/qiboteam/qibocal/blob/two-qubit-gates/tests/runcards/single_routines.yml) and [here](https://github.com/qiboteam/qibocal/blob/two-qubit-gates/tests/runcards/test_autocalibration.yml).

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
```diff
@@ -0,0 +1,273 @@
...
+    sequence = PulseSequence()
+
+    ro_pulses = {}
+    qd_pulses = {}
+
+    qd_pulses[ord_pair[0]] = platform.create_RX_pulse(ord_pair[0], start=0)
+    qd_pulses[ord_pair[1]] = platform.create_RX_pulse(ord_pair[1], start=0)
+    fx_pulse = FluxPulse(
+        start=max([qd_pulses[ord_pair[0]].se_finish, qd_pulses[ord_pair[1]].se_finish])
+        + 8,
```
maxhant, line 88: The +8 here was to avoid any timing problems when using the bias tee. I have kept it because it does not hurt. However, as it is a bit arbitrary, I would remove it, or create a parameter 
Now that we are not using bias-tee pulses should arrive at the scheduled time exactly.

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
```diff
@@ -0,0 +1,273 @@
...
+        channel=platform.qubits[ord_pair[0]].flux.name,
+        qubit=ord_pair[0],
+    )
+
+    ro_pulses[ord_pair[0]] = platform.create_MZ_pulse(
+        ord_pair[0], start=fx_pulse.se_finish + 8
+    )
+    ro_pulses[ord_pair[1]] = platform.create_MZ_pulse(
+        ord_pair[1], start=fx_pulse.se_finish + 8
+    )
```
maxhant, line 101: same comment for 8 here

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
```diff
@@ -0,0 +1,273 @@
...
+        platform: Platform to use.
+        params: Experiment parameters.
+        qubits: Qubits to use.
+
+    Returns:
+        DataUnits: Acquisition data.
+    """
+    if params.qubits is None:
+        raise ValueError("You have to specifiy the pairs. Es: [[0, 1], [2, 3]]")
+        params.qubits = platform.settings["topology"]
```
maxhant, line 174: Why that line?

maxhant (changes requested):
I personally don't like the functions but don't worry.  
I would also use the `iq_distance` like I've done in the SNZ PR to be able to discriminate the second state. 
Also, you can have a look at the fitting done by @aorgazf. I tried to find it in the branches that I know but couldn't find it...
It is pretty simple. Take the FFT for every amplitude, get the main frequency of oscillation, from all the smallest one (center chevron) is the one you want, fit it with a sin/cos and the period is your duration. 
Anyway, it is not strickly necessary for now. Just the few comments, and it's good to me

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
rodolfocarobene: I would like that, if I don't define the qubit parameter, the topology is used... But in any case I have to wait for the local qubit to fully work

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
```diff
@@ -0,0 +1,333 @@
...
+    results: dict[Union[QubitId, str], AveragedSampleResults],
+    sweep_data: ChevronData,
+    delta_amplitude_range: np.ndarray,
+    delta_duration_range: np.ndarray,
+):
+    """Save results of the experiment in the Data object."""
+    for state, qubit in zip(["low", "high"], pair):
+        ro_pulse = ro_pulses[qubit]
+        result = results[ro_pulse.serial]
+        prob = result.statistical_frequency
```
Jacfomg, line 140:
```suggestion
        result = results[measure_target.serial].serialize
        prob = result["0"]
```
Same here I might take a look into the results.py as well

src/qibocal/protocols/characterization/two_qubit_interaction/cz_virtualz.py
```diff
@@ -0,0 +1,416 @@
...
+                    ExecutionParameters(
+                        nshots=params.nshots,
+                        acquisition_type=AcquisitionType.DISCRIMINATION,
+                        averaging_mode=AveragingMode.CYCLIC,
+                    ),
+                    sweeper,
+                )
+
+                result = results[measure_target.serial]
+                prob = result.statistical_frequency
```
Jacfomg, line 208:
```suggestion
                result = results[measure_target.serial].serialize
                prob = result["0"]
```
This should save probabiliies not as complex numbers

Jacfomg:
They run properly on the devices

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
```diff
@@ -0,0 +1,296 @@
...
+                z=data[ordered_index].prob,
+                name=f"Qubit {qubit} |{state}>",
+                coloraxis=colouraxis[states.index(state)],
+            ),
+            row=1,
+            col=states.index(state) + 1,
+        )
+
+        fit_report += f"q{qubit} - {state} frequency| "
+        fit_report += f"Period of oscillation: {fit.period[str(ordered_index)]} ns<br>"
```
Jacfomg, line 242:
```suggestion
       #  fit_report += f"Period of oscillation: {fit.period[str(ordered_index)]} ns<br>"
```
Comment it for now until we get the fitting

src/qibocal/protocols/characterization/two_qubit_interaction/cz_virtualz.py
```diff
@@ -0,0 +1,411 @@
...
+    if parking:
+        # if parking is true, create a cz pulse from the runcard and
+        # add to the sequence all parking pulses
+        cz_sequence, _ = platform.pairs[
+            tuple(sorted([target_qubit, control_qubit]))
+        ].native_gates.CZ.sequence(start=0)
+        for pulse in cz_sequence:
+            if pulse.qubit not in {target_qubit, control_qubit}:
+                pulse.start = flux_sequence[setup].start
+                pulse.duration = flux_sequence[setup].duration
```
Jacfomg, line 140:
```suggestion
                pulse.start = flux_sequence.start
                pulse.duration = flux_sequence.duration
```

Jacfomg (commented):
Small changes to make them run on the devices.

andrea-pasquale:
This is ready to review and it has been tested on hardware.
A few things to be changed:
- fititng on chevron
- change from MSR to probabilities

src/qibocal/auto/operation.py
```diff
@@ -67,6 +67,11 @@ def qubits(self):
             return list({q[0] for q in self.data})
         return [q for q in self.data]
 
+    @property
+    def pairs(self):
+        """Access qubit pairs from data structure."""
+        return list({tuple(sorted(q[:2])) for q in self.data})
```
rodolfocarobene, line 73: In which way are the qubit sorted? I would write it in the docstring for clarity

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+
+def order_pairs(
+    pair: list[QubitId, QubitId], qubits: dict[QubitId, Qubit]
+) -> list[QubitId, QubitId]:
+    """Order a pair of qubits by drive frequency."""
+    if qubits[pair[0]].drive_frequency > qubits[pair[1]].drive_frequency:
+        return pair[::-1]
+    return pair
```
rodolfocarobene, line 87: I know that is basically my code so my bad, but maybe here it would be better to use a named tuple so that it's clear (here and elsewhere) which is the high frequency qubit and which the low. Something like:
```python
from collections import namedtuple

Pair = namedtuple('Pair', 'high_freq low_freq')
mypair = Pair(qubit0, qubit1)

mypair.low_freq
> qubit0
```

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
```diff
@@ -0,0 +1,270 @@
...
+    """Amplitude step."""
+    duration_min: float
+    """Duration minimum."""
+    duration_max: float
+    """Duration maximum."""
+    duration_step: float
+    """Duration step."""
+    nshots: Optional[int] = None
+    """Number of shots per point."""
+
```
rodolfocarobene, line 38: I see that you removed the parking parameter, may I ask you how you manage it?

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
```diff
@@ -0,0 +1,270 @@
...
+        sequence.add(initialize_highfreq)
+        sequence.add(initialize_lowfreq)
+        cz, _ = platform.create_CZ_pulse_sequence(
+            qubits=(ordered_pair[1], ordered_pair[0]), start=initialize_highfreq.finish
+        )
+        sequence.add(cz)
+
+        # add readout
+        measure_lowfreq = platform.create_qubit_readout_pulse(
+            ordered_pair[0], start=initialize_lowfreq.finish + params.duration_max + 20
```
rodolfocarobene, line 131: These 20s could be a routine parameter, something like dt set to 20 by default

src/qibocal/protocols/characterization/two_qubit_interaction/cz_virtualz.py
```diff
@@ -0,0 +1,384 @@
...
+
+
+@dataclass
+class CZVirtualZResults(Results):
+    """CzVirtualZ outputs when fitting will be done."""
+
+    fitted_parameters: dict[tuple[str, QubitId],]
+    """Fitted parameters"""
+    virtual_phase: dict[tuple[QubitId, QubitId], float]
+    cz_phase: dict[tuple[QubitId, QubitId], float]
```
rodolfocarobene, line 48: Are these different?

src/qibocal/protocols/characterization/two_qubit_interaction/cz_virtualz.py
```diff
@@ -0,0 +1,384 @@
...
+    RX_pulse_start = platform.create_RX_pulse(control_qubit, start=0, relative_phase=0)
+
+    flux_sequence, virtual_z_phase = platform.create_CZ_pulse_sequence(
+        (highfreq, lowfreq),
+        start=max(Y90_pulse.finish, RX_pulse_start.finish),
+    )
+
+    theta_pulse = platform.create_RX90_pulse(
+        target_qubit,
+        start=flux_sequence.finish + 20,
```
rodolfocarobene, line 115: I think that here the dt could be used (and it's already a parameter)

src/qibocal/auto/operation.py
andrea-pasquale: Good point, here they are just sorted alphanumerically.

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
andrea-pasquale: You are right, at this point we could add also here a `dt` parameter or maybe something more meaningful than `dt`.

src/qibocal/protocols/characterization/two_qubit_interaction/cz_virtualz.py
andrea-pasquale: Yes, CZ phase should be pi, while virtual_phase is the virtual_phase computed to correct the additional phase acquired during the 02 - 11 interaction. Now that you make think about it perhaps here the calculation of the virtual phase is wrong... I will double check :)

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
andrea-pasquale: At this point I was thinking that we could move this to qibolab...
Or to already order the qubit pair in qibolab according to the drive frequency.
What do you think @rodolfocarobene @Jacfomg @stavros11 ?

src/qibocal/protocols/characterization/two_qubit_interaction/cz_virtualz.py
andrea-pasquale: After discussing with @igres26, I modified the result object.
Currently there is also one parameter which is the CZ angle. There is no way to detect the virtual phase through fitting.

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
rodolfocarobene: Imho it makes sense to have them here, in qibolab they could be ordered in respect to the topology (like, for a linear chip it could become confusing)

Jacfomg (commented):
Apart from the parking seems good to me. I think we should wait until they get tested on hardware, I am not aware of the situation right now.

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
andrea-pasquale: I implemented the `namedtuple` in https://github.com/qiboteam/qibocal/pull/381/commits/35fb20383a7b4f3548420c373879a86f42f52a75, it looks much more readable now. Thanks!

andrea-pasquale:
I should have implemented all the comments. Let me know if there is something else to do, otherwise I think that we can merge @rodolfocarobene @Jacfomg

andrea-pasquale:
> Apart from the parking seems good to me. I think we should wait until they get tested on hardware, I am not aware of the situation right now.

I'm testing on hardware with qblox, the protocols seem to be working. I need to test only the parking.

rodolfocarobene:
> I'm testing on hardware with qblox, the protocols seem to be working. I need to test only the parking.

It looks good to me! (approved)

andrea-pasquale (approved):

andrea-pasquale:
> > I'm testing on hardware with qblox, the protocols seem to be working. I need to test only the parking.
> 
> It looks good to me! (approved)

I approved on your behalf ahahah

Jacfomg (approved):

PR #383, opened 2023-06-06
Add optimal integration weights routine

Jacfomg:
Add integration weights. I would say it would need a way to save the fitting, #395, so I can get the integration weights and pass them to the devices.

Need to read a bit to understand what I get from the hardware.

https://arxiv.org/pdf/1711.05336.pdf and its 29 reference. 

Checklist:
- [ ] Tests are passing.


src/qibocal/protocols/characterization/signal_experiments/optimal_integration_weights.py
```diff
@@ -0,0 +1,197 @@
...
+    signal acquired which maximally distinguish between state 1 and 0.
+    """
+
+    def save(self, path):
+        for q in self.optimal_integration_weights.keys():
+            # """Helper function to use np.savez while converting keys into strings."""
+            np.savez(
+                path / f"optimal_integration_weights_q{q}.npz",
+                self.optimal_integration_weights[q],
+            )
```
andrea-pasquale, line 42: The advantage of `npz` is that you can store [multiple arrays](https://numpy.org/doc/stable/reference/generated/numpy.savez.html). Therefore you don't need a specific file for each qubit. Unless you are doing this on purpose.

src/qibocal/protocols/characterization/signal_experiments/optimal_integration_weights.py
```diff
@@ -0,0 +1,197 @@
...
+    def register_qubit(self, qubit, samples, state):
+        """Store output for single qubit."""
+        ar = np.empty(samples.shape, dtype=OptimalIntegrationWeightsType)
+        ar["samples"] = samples
+        if qubit in self.data:
+            self.data[qubit, state] = np.rec.array(
+                np.concatenate((self.data[qubit], ar))
+            )
+        else:
+            self.data[qubit, state] = np.rec.array(ar)
```
andrea-pasquale, line 66: If you append only once for the same combination of `qubit` and `state` you can keep only the line in the `else` statement.

andrea-pasquale (commented):
Thanks @Jacfomg.
LGTM, just a few minor comments.

src/qibocal/protocols/characterization/signal_experiments/optimal_integration_weights.py
Jacfomg: Yeah, if I have them is separate files I think it will be easier to handle them, at least now. So if you were recalibrating some qubits you can let the files for the others untouched. If not you will need to calibrate them all to have the file, right ?

andrea-pasquale:
I'm closing this as outdated but I will keep the branch, I will also open an issue to keep track of it.

andrea-pasquale:
@Jacfomg this branch can be deleted since you have already #582 right?

PR #384, opened 2023-06-06
SNZ two qubit gate routine

maxhant:
I have edited the Qibocal file here to have the qubit pairs being passed on [properly](https://github.com/qiboteam/qibocal/blob/b33db4f11e5dd46e7e234a67d2868086e532c56c/src/qibocal/utils.py#L11).

You can use the following action for that:
Actions to use:
```
  - id: snz
    priority: 0
    operation: snz_tuning
    parameters:
      pairs: [[0,3]]
      amplitude_factor_start: 0.0
      amplitude_factor_end: 1.0
      amplitude_factor_step: 0.1
      b_amplitude_factor_start: 0.0
      b_amplitude_factor_end: 1.0
      b_amplitude_factor_step: 0.1
      detuning_start: 0
      detuning_end: 180
      detuning_step: 20
```

I have commented on the update field in the results due to a fundamental problem in the routine that needs non-trivial changes in qibolab. 

To sweep the amplitude, I need to overlap two flux pulses, the small one that has the B amplitude and the overall one that has the A amplitude. This is why I am defining it as followed in the runcard:

```python
            "0-3": {
                "CZ": [
                    {
                        "duration": 34,
                        "amplitude": 0.055,
                        "shape": SNZ(1),
                        "qubit": 3,
                        "relative_start": 0,
                        "type": "qf",
                    },
                    {
                        "duration": 4,
                        "amplitude": 0.055,
                        "shape": SNZ(1),
                        "qubit": 3,
                        "relative_start": 14,
                        "type": "qf",
                    },          
                    {
                        "duration": 34,
                        "amplitude": 0.1,
                        "shape": "Rectangular()",
                        "qubit": 3,
                        "relative_start": 0,
                        "type": "qf",
                    },                              
                    {"type": "virtual_z", "phase": -1.5707963267948966, "qubit": 3},
                    {"type": "virtual_z", "phase": -1.5707963267948966, "qubit": 0},
                ]
           }
```
Once the pulse is calibrated, maybe it is best to have the SNZ as one pulse only as it takes the B amplitude as an optional parameter. 
This is why I cannot use the update field. 

I was suggesting @stavros11 that we could have pulses' ID and that routines could update `CZ_SNZ` instead of `CZ_square` or `CZ_SNZ_calibration`. For the compiler, we would have a place holder somewhere to associate the CZ to a specific ID.

This routine was tested on the `dummy` only. I don't know if it will run well. I had tested it on the old format a long time back, and I had sensible results but not conclusive. Improving this routine will take about a week of work I believe. This is why I think that it is worth merging it and opening an issue to test on hardware. This is at the discretion of the reviewers. 

Checklist:
- [ ] Tests are passing.


andrea-pasquale:
Can we close also this one and keep the branch open @Jacfomg ?

PR #385, opened 2023-06-06
Add pycln to pre-commit hooks

AleCandido:
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,3 +23,8 @@ repos:
     rev: v3.4.0
     hooks:
       - id: pyupgrade
+  - repo: https://github.com/hadialqattan/pycln
+    rev: v2.1.5
+    hooks:
+      - id: pycln
+        args: [--config=pyproject.toml]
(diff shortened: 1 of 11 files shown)


AleCandido:
Actually, I've found a couple of export files that are not following the best possible pattern:
https://github.com/qiboteam/qibocal/blob/c7739b85f0c5e9ba4a481f9e35e29a498ae05c0b/src/qibocal/plots/__init__.py#L1-L12
https://github.com/qiboteam/qibocal/blob/c7739b85f0c5e9ba4a481f9e35e29a498ae05c0b/src/qibocal/calibrations/__init__.py#L1-L14

At the moment they are ignored (the `# nopycln: file` comment), but they should be either dropped (the content, not the file, of course) or should define the `__all__` variable, declaring what they export.

However, I suspect that these files to be outdated, and they might be fully dropped soon, so for the time being I haven't taken action.
Should I make this an issue?

andrea-pasquale:
> Should I make this an issue?

Thanks @AleCandido for the fix.
Actually those files belong to the old layout. 
I'm planning to open a PR today to remove everything that it is no longer used.

stavros11 (approved):
Thanks @AleCandido. Looks useful, shall we also add this to qibo and qibolab?

PR #390, opened 2023-06-07
Add supports for qubit pairs

andrea-pasquale:
As title says

Checklist:
- [x] Tests are passing.


andrea-pasquale:
@rodolfocarobene can you retry now?
I tested it with a dummy routine and it seems to work.

rodolfocarobene:
> @rodolfocarobene can you retry now? I tested it with a dummy routine and it seems to work.

I now get a different error at the end of the routine:
```
  File "/home/rodolfoc/venv/qibo/qibocal/src/qibocal/cli/auto_builder.py", line 55, in dump_report
    create_autocalibration_report(self.folder, self.executor.history)
  File "/home/rodolfoc/venv/qibo/qibocal/src/qibocal/web/report.py", line 43, in create_autocalibration_report
    html = template.render(
  File "/home/rodolfoc/venv/qibo/lib/python3.10/site-packages/jinja2/environment.py", line 1301, in render
    self.environment.handle_exception()
  File "/home/rodolfoc/venv/qibo/lib/python3.10/site-packages/jinja2/environment.py", line 936, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "/home/rodolfoc/venv/qibo/qibocal/src/qibocal/web/templates/autocalibration.html", line 195, in top-level template code
    {% set figures, fitting_report = report.plot(routine, iteration) %}
TypeError: AutoCalibrationReportBuilder.plot() takes 2 positional arguments but 3 were given
```

andrea-pasquale:
Uhm... it seems weird because the report should call `report.single_qubit_plot` and not `report.plot`.
I will put here the routines that I'm using to debug.

```py
import plotly.graph_objects as go


import numpy as np


from qibocal.auto.operation import Parameters, QubitsPairs, Results, Routine



@dataclass
class TwoParams(Parameters):
    """Two qubit gates dummy routine."""
    ok : int



@dataclass
class TwoResults(Results):
    """Two results."""




class TwoData:
    """ResonatorSpectroscopy acquisition outputs."""
    def to_csv(self, path):
        pass

def _acquisition(
    params: TwoParams, platform: Platform, qubits: QubitsPairs
) -> TwoData:
    """Data acquisition for resonator spectroscopy."""
    print(qubits)
    for qubit in qubits:
        print(qubit)
    return TwoData()



def _fit(data: TwoData) -> TwoResults:
    return TwoResults()

def _plot(data: TwoData, fit: TwoResults, qubits):
    """Plotting function for ResonatorSpectroscopy."""
    figures = []
    fig = go.Figure()

    fig.add_trace(
        go.Scatter(
            x=np.random.rand(10),
            y=np.random.rand(10),
        )
    )
    figures.append(fig)
    return figures, ""


two = Routine(_acquisition, _fit, _plot)
"""ResonatorSpectroscopy Routine object."""
```


and this is the runcard

```yml
platform: dummy

qubits: [0,1]

format: csv

actions:

  - id: two
    priority: 0
    operation: two
    qubits: [[0, 1]]
    # main: resonator punchout
    parameters:
      ok: 2

```
we can have a look also tomorrow don't worry

andrea-pasquale:
Apparently the issue was in the html template.
It was failing if you didn't providing a global qubits option.
I've also removed a generic plot function from the builder which was not used.

rodolfocarobene:
Thank you @andrea-pasquale, now it works fine. The only thing, eventually not necessary, is that if I do not specify the qubits parameter then it tries to use the dictionary of single qubits

src/qibocal/cli/builders.py
```diff
@@ -58,9 +58,13 @@ def backend(self):
...
-            return allocate_qubits(self.platform, self.runcard.qubits)
-
-        return self.runcard.qubits
+            if any(isinstance(i, list) for i in self.runcard.qubits):
+                return allocate_qubits_pairs(self.platform, self.runcard.qubits)
+            else:
+                return allocate_single_qubits(self.platform, self.runcard.qubits)
+        # TODO: check if this is needed
+        else:
+            return self.runcard.qubits
```
rodolfocarobene, line 67:
```suggestion
            if any(isinstance(i, list) for i in self.runcard.qubits):
                return allocate_qubits_pairs(self.platform, self.runcard.qubits)
            return allocate_single_qubits(self.platform, self.runcard.qubits)
        # TODO: check if this is needed
        return self.runcard.qubits
```
These else were not needed. Also, what is this TODO?

rodolfocarobene (changes requested):
Thanks @andrea-pasquale, I think this is almost ready right? To me looks good in general, I will approve as soon as the conflict are solved :-)

src/qibocal/cli/builders.py
andrea-pasquale: Indeed @rodolfocarobene the `else` can be removed.
The `TODO` is not necessary anymore. I believe that line is necessary when simulating the rb.

rodolfocarobene (approved):

PR #396, opened 2023-06-12
pycln-fix

rodolfocarobene:
Adding `--all` parameter to `pycln`. 
See that I had to add an exception for pint-pandas in data.py, one of the unused/used imports...

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibocal/data.py
+++ b/src/qibocal/data.py
@@ -5,6 +5,7 @@ from abc import abstractmethod
 
 import numpy as np
 import pandas as pd
+import pint_pandas  # nopycln: import
 
 from qibocal.config import raise_error
 


src/qibocal/data.py
```diff
@@ -5,7 +5,7 @@
 
 import numpy as np
 import pandas as pd
-import pint_pandas
+import pint_pandas  # nopycln: import
```
AleCandido, line 8: ?

rodolfocarobene: If I remove that import, several tests fail (I suppose every tests that uses this file).
I am not sure why, but I think that it's implicitly using it with
https://github.com/qiboteam/qibocal/blob/0b72989e53f66be52d0ac21c37e2357025880082/src/qibocal/data.py#L76-L82

AleCandido: Here our first side effect. Most likely `pint_pandas` is patching `pandas` (i.e. replacing some content of the module) upon import.

Luckily, we have #372 :)

andrea-pasquale: Indeed this is just how pint works for some reason...

andrea-pasquale (approved):
Thanks @rodolfocarobene

src/qibocal/data.py
AleCandido: It wants to be a *drop-in* replacement of Pandas, that's the reason. But, yeah, I don't like it so much...

PR #402, opened 2023-06-16
add time of flight for readout

Jacfomg:
Add time of flight for readout. It did not seem trivial to fit this. We should get more time of flight plots from different devices before attemping it. I would say it's better to do it after.

Test should fail until we merge #410. But We should merge that soon so I will not point this there.

Checklist:
- [x] Tests are passing.


src/qibocal/protocols/characterization/signal_experiments/tof_readout.py
```diff
@@ -0,0 +1,131 @@
...
+    """Number of shots."""
+    relaxation_time: Optional[int] = None
+    """Relaxation time (ns)."""
+
+
+@dataclass
+class ToFResults(Results):
+    """ToF outputs."""
+
+    ToF: Dict[Union[str, int], float] = field(metadata=dict(update="time_of_flight"))
```
andrea-pasquale, line 29:
```suggestion
    ToF: Dict[QubitId, float] = field(metadata=dict(update="time_of_flight"))
```
You can do this by importing `QubitId` from `qibolab.qubits`

src/qibocal/protocols/characterization/signal_experiments/tof_readout.py
andrea-pasquale, line 1: I would prefer to have `TimeOfFlight` instead of `ToF`

andrea-pasquale (commented):
Thanks @Jacfomg for implementing this protocol.
You can find below some suggestions.

src/qibocal/protocols/characterization/signal_experiments/tof_readout.py
Jacfomg: It should also be something like `TimeOfFlightReadout` but I feel like it's too long too

andrea-pasquale: Don't worry if it is too long. I would rather have a long name which is more understandable than a short name.
Choose whatever you prefer between `TimeOfFlight` and `TimeOfFlightReadout`.

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
```diff
@@ -0,0 +1,185 @@
...
+
+    data: dict[QubitId, npt.NDArray[TimeOfFlightReadoutType]] = field(
+        default_factory=dict
+    )
+    """Raw data acquired."""
+
+    def register_qubit(self, qubit, samples, window_size):
+        """Store output for single qubit."""
+        shape = (1,) if np.isscalar(samples) else samples.shape
+        ar = np.empty(shape, dtype=TimeOfFlightReadoutType)
```
andrea-pasquale, line 50:
```suggestion
        ar = np.empty(samples.shape, dtype=TimeOfFlightReadoutType)
```
Since you are appending data only once at the end you will never have samples as a scalar, right?

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
```diff
@@ -0,0 +1,185 @@
...
+        ar["samples"] = samples
+        ar["window_size"] = window_size
+        if qubit in self.data:
+            self.data[qubit] = np.rec.array(np.concatenate((self.data[qubit], ar)))
+        else:
+            self.data[qubit] = np.rec.array(ar)
```
andrea-pasquale, line 56:
```suggestion
        self.data[qubit] = np.rec.array(ar)
```
See comment above.

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
```diff
@@ -0,0 +1,185 @@
...
+            )
+            if last is None:
+                last = window_average
+            moving_average_deltas.append(window_average - last)
+            last = window_average
+            i += 1
+
+        max_average_change = moving_average_deltas.index(max(moving_average_deltas))
+        # FIXME: Where do I get the sampling rate from ???
+        # time_of_flight_readout = max_average_change * sampling_rate
```
andrea-pasquale, line 124: For this you can add an attribute `sampling_rate` to `TimeOfFlightReadoutData`. See for example https://github.com/qiboteam/qibocal/blob/6f567c1804463a32b9ddcd6cad21a01331fab70e/src/qibocal/protocols/characterization/resonator_spectroscopy.py#L75

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
```diff
@@ -0,0 +1,185 @@
...
+    )
+
+    fig.update_layout(
+        showlegend=True,
+        uirevision="0",  # ``uirevision`` allows zooming while live plotting
+        xaxis_title="Sample",
+        yaxis_title="MSR (uV)",
+    )
+
+    # FIXME: Need Sampling Rate again
```
andrea-pasquale, line 159: You can access it through the data class if you follow my previous comment.

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
```diff
@@ -0,0 +1,185 @@
...
+    figures = []
+    fitting_report = ""
+    fig = go.Figure()
+
+    qubit_data = data[qubit]
+    y = qubit_data.samples
+
+    fig.add_trace(
+        go.Scatter(
+            # x=data.df["sample"],
```
andrea-pasquale, line 143:
```suggestion
            x=qubit_data.samples
```
Right?

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+@dataclass
+class TimeOfFlightReadoutParameters(Parameters):
+    """TimeOfFlightReadout runcard inputs."""
+
+    nshots: Optional[int] = None
+    """Number of shots."""
+    relaxation_time: Optional[int] = None
+    """Relaxation time (ns)."""
```
andrea-pasquale, line 22: This two attributes can be removed, right?

andrea-pasquale (commented):
Thanks @Jacfomg.
I still need to test it on hardware, generally it looks good to me.
You can find below some comments / replies on your `#FIXME`s.

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
Jacfomg: Yup

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
Jacfomg: This one I forgot to remove, it's not even needed for the plot

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
andrea-pasquale: Yes, but they are already define here for all the protocols, right?
https://github.com/qiboteam/qibocal/blob/6f567c1804463a32b9ddcd6cad21a01331fab70e/src/qibocal/auto/operation.py#L34-L37

andrea-pasquale: I think that the code should work even if you drop them from this class.

andrea-pasquale: image

Jacfomg: But then it complains if I give it nshots as a parameter

andrea-pasquale: It is weird then. For now lets keep them and you can fix this in another PR.

Jacfomg: If I do this they will work and we could remove all the repetition of `nshots` and `relaxation_time` but as they need to be `None` to by default and are the first argument to be defined it will complain because they should be the last. Any idea on what to do with this ?

```py
@dataclass
class Parameters:
    """Generic action parameters.

    Implement parameters as Algebraic Data Types (similar to), by subclassing
    this marker in actual parameters specification for each calibration routine.

    The actual parameters structure is only used inside the routines themselves.

    """

    nshots: Optional[int] = None
    """Number of shots."""
    relaxation_time: Optional[int] = None
    """Relaxation time (ns)."""
```

andrea-pasquale: I had similar issue with both `qubits` and `update`. In that case I fix the problem by moving them outside the parameters class (see #351). We could follow the same approach here, the only problem I see is that maybe both `relaxation_time` and `nshots` are better identified as parameters...
I'll have another look.

andrea-pasquale: Apparently there is a solution but it is available only for [py3.10 ](https://docs.python.org/3/library/dataclasses.html#dataclasses.KW_ONLY) :(

Jacfomg: Thanks for looking for taking a look, should we just leave them as they are for now ?

andrea-pasquale: Yes, lets keep them as they are.

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
```diff
@@ -0,0 +1,179 @@
...
+        last = None
+        while i < len(qubit_data) - window_size + 1:
+            window_average = (
+                np.sum(qubit_data[i : i + window_size].samples) / window_size
+            )
+            if last is None:
+                last = window_average
+            moving_average_deltas.append(window_average - last)
+            last = window_average
+            i += 1
```
andrea-pasquale, line 121: Acutally we can do this in a single line.
I tested it locally with random numbers and it seems to work as expected.
```suggestion
        moving_average_deltas = np.convolve(qubit_data.samples, np.ones(window_size)/window_size, mode='valid')
```

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
```diff
@@ -0,0 +1,179 @@
...
+
+        max_average_change = moving_average_deltas.index(max(moving_average_deltas))
```
andrea-pasquale, line 123: Here I don't fully understand what is happening.
If you are interested in the max you should just use `max` or `np.max`. If you need the index you can use `np.argmax`.

andrea-pasquale (commented):
Thanks for all the updates @Jacfomg.
A few minor corrections.

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
andrea-pasquale:
```suggestion
        max_average_change = np.argmax(moving_average_deltas) + 1
```
This should do the job I think. The `+1` could be necessary if you consider the comment above.

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
```diff
@@ -0,0 +1,179 @@
...
+        default_factory=dict
+    )
+    """Raw data acquired."""
+
+    def register_qubit(self, qubit, samples, window_size, sampling_rate):
+        """Store output for single qubit."""
+        ar = np.empty(samples.shape, dtype=TimeOfFlightReadoutType)
+        ar["samples"] = samples
+        ar["window_size"] = window_size
+        ar["sampling_rate"] = sampling_rate
```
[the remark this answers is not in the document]
Jacfomg: `sampling_rate` for sure will be the same but for `window_size` I am not so sure about as qubits on the same chips can be quite different. I would still assume both will be the same for all qubits for now.

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
Jacfomg: That's a very nice way that took I yet not full inderstand why it works but I will look into it. But for this, you would get the moving averages not the change between the moving averages so I would still do something like this

```py
moving_average_deltas = []
last = None
moving_averages = np.convolve(qubit_data.samples, np.ones(window_size)/window_size, mode='valid')
for moving_average in moving_averages:
    if last is None:
        last = moving_average
    moving_average_deltas.append(moving_average - last)
    last = moving_average
```

andrea-pasquale: Oh yes sorry. I forgot to add the difference (I have done it only locally):
```suggestion
        moving_average_deltas = np.ediff1d(np.convolve(qubit_data.samples, np.ones(window_size)/window_size, mode='valid'))
```

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
Jacfomg: I converted the list into a np.array and just argmax is enough, we don't need the +1

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
Jacfomg: Yup, now it it's the same. Very fancy now on a single line.

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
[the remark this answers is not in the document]
andrea-pasquale: For the future we can also pass a list for different window sizes https://github.com/qiboteam/qibocal/issues/336#issuecomment-1571445224

andrea-pasquale (approved):

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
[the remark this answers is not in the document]
Jacfomg: Yeah, first I would like to see how this behaves for other chips or instruments as well beucase maybe a single value it's enough.

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
```diff
@@ -0,0 +1,172 @@
...
+import numpy.typing as npt
+import plotly.graph_objects as go
+from qibolab import AcquisitionType, AveragingMode, ExecutionParameters
+from qibolab.platform import Platform
+from qibolab.pulses import PulseSequence
+from qibolab.qubits import QubitId
+
+from qibocal.auto.operation import Data, Parameters, Qubits, Results, Routine
+
+SECONDS_TO_NANO = 1e9
```
rodolfocarobene, line 14: There are other constants [here](https://github.com/qiboteam/qibocal/blob/main/src/qibocal/protocols/characterization/utils.py), I would add this too

rodolfocarobene (commented):
I was just thinking , maybe it's nice also to have the readout amplitude as a parameter? Since we saw it varies a lot with it

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
andrea-pasquale: +1
Also to uniform the notation we could rename this `S_TO_NS`.

Jacfomg:
> I was just thinking , maybe it's nice also to have the readout amplitude as a parameter? Since we saw it varies a lot with it

Yeah, but the one you use for the readout should be fine. I can still add it but I am not sure if it going to be that useful. Also, what would be nice is a way to change the `time_of_flight` defined on the `create()` but it is not a `platform` parameter for now I think.

rodolfocarobene (approved):
Ok, it makes sense also to use it a bit on hardware and then eventually implement some fixes :-)

other than the consant, it looks all good to me

PR #403, opened 2023-06-16
Storing fitting results

### LINKED ISSUE #395: Save the fitting
For the #383 I need to save from the fitting part a numpy array on the report. But I would say that this can be useful in general.

andrea-pasquale:
Closes #395 by storing for each protocol the `Results` object as a json.
I had to modify a few things in some of the routines given that `np.ndarray` and `complex` are not json serializable. 
I would like to merge this PR  **after**   #372. 
I'm also planning to open a PR to remove `complex` for `mean_gnd_states` and `mean_exc_states` I think that a list should be more than enough.

Checklist:
- [x] Tests are passing.


Jacfomg:
Fitting gets saved nicely, but I am worried for the #383 that needs to save np.arrays of thousands of samples.

src/qibocal/auto/operation.py
```diff
@@ -126,6 +128,10 @@ def update(self) -> dict[str, ParameterValue]:
 
         return up
 
+    def save(self, path):
+        """Store results to json."""
+        (path / RESULTFILE).write_text(json.dumps(asdict(self), indent=4))
```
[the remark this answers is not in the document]
andrea-pasquale: In theory json can store numpy arrays. But we will have a single for each qubit.



> Maybe we could check if here one of the elements was a big array and store as a separate numpy file ?

We could do that, but if this is necessary only for one routine you could simply overwrite the `save` method.

AleCandido:
This should be merged/rebased on `main`.

src/qibocal/auto/operation.py
```diff
@@ -19,6 +19,8 @@
 """Name of the file where data acquired (arrays) by calibration are dumped."""
 JSONFILE = "conf.json"
 """Name of the file where data acquired (global configuration) by calibration are dumped."""
+RESULTFILE = "result.json"
```
AleCandido, line 22: Just for consistency:
```suggestion
RESULTSFILE = "results.json"
```

src/qibocal/auto/task.py
```diff
@@ -84,6 +84,10 @@ def update(self):
     def data(self):
         return self._data
 
+    @property
+    def results(self):
+        return self._results
```
AleCandido, line 89: I know in this way you (suggest more to) avoid overwriting from the outside, but sometimes as simpler approach is enough: you can just define `results` directly as an attribute, the alias is a bit redundant.

AleCandido: Of course same for `_data` above.

Especially because none of the two attributes (`_data` and `_results`) is declared in the `__init__`, at the moment.

src/qibocal/auto/operation.py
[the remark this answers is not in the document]
Jacfomg: For now I think this would only be a single routine. It depends on what you considerer big. The routine #383 would store around 2000 to 8000 floats, depending on the device, for each qubit to puts weights on the signal integration during the readout that need to be loaded into the device each time. I can try to overwrite the `save` method on that routine and see how it goes.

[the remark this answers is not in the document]
AleCandido: "big" definition depends on wrt to what, of course.

If we are talking about file size, I'd say below 100KB we don't care so much, and they are `1e5 / 8` floats.
But if it's about being useful as human-readable content, already few dozens are not small.

But are we calibrating 2000/8000 parameters?

src/qibocal/auto/task.py
andrea-pasquale: I know. For now both attributes are declared and defined in `task.run()`.

src/qibocal/auto/task.py
```diff
@@ -123,15 +115,15 @@ def run(
...
         else:
-            self._data: Data = operation.acquisition(parameters, platform=platform)
-        self._data.save(path)
+            self.data: Data = operation.acquisition(parameters, platform=platform)
+        self.data.save(path)
 
-        self._results: Results = operation.fit(self._data)
-        self._results.save(path)
-        return self._results
+        self.results: Results = operation.fit(self.data)
```
AleCandido, line 127: It is fine to have results on a completed task, and it makes sense, but then you should not set on the general one.

andrea-pasquale:
Thanks for the quick review @AleCandido.
I've implemented all the suggestions.
Some of the stuff that is implemented here is outdated wrt to #406. For the time being let's try to merge this and then we proceed with the review of #406.

src/qibocal/auto/task.py
```diff
@@ -119,12 +115,15 @@ def run(
...
         else:
-            self._data: Data = operation.acquisition(parameters, platform=platform)
-        self._data.save(path)
-        return operation.fit(self._data)
+            self.data: Data = operation.acquisition(parameters, platform=platform)
+        self.data.save(path)
+
+        self.results: Results = operation.fit(self.data)
+        self.results.save(path)
+        return self.results
```
AleCandido: Even the saving can be potentially done during `Completed` construction (using `__post_init__`), to remove from here all the references to a completed task: just return `data, results`, and we could even move the `.datapath()` method.

src/qibocal/auto/task.py
andrea-pasquale: Indeed. I've removed `results` as an attribute.
I could do the same for `data` but this change is already addressed in #406.

src/qibocal/auto/task.py
andrea-pasquale: Ok, I will give it a try.

AleCandido: Actually, at a second thought, this is a feasible option, but we should make this method a `Generator`, first yielding `data` and then `results`, not returning at the same time.

`data` should be saved before running the `fit` (in case of troubles, you never know...).

src/qibocal/auto/task.py
andrea-pasquale: Nevermind, I will implement https://github.com/qiboteam/qibocal/pull/403#discussion_r1242403320 and move also data outside of the class

AleCandido: You're right... I just forgot...
(and I actually found out I had a pending review there...)

src/qibocal/auto/task.py
andrea-pasquale: Exactly... for sure `data` needs to be saved before.
Isn't than weird to have a `run` method which is a generator?
I expect to execute `run` only once. Of course we can rename this method.

AleCandido: > I expect to execute `run` only once. Of course we can rename this method.

A generator is not only an iterable, but in general is an execution that can be interrupted.
So, we could split the routine in two steps, first acquisition and then fit, as it is done inside, and lift the burden of collecting the two to the caller. Otherwise, we could state here that the two things have to execute one after the other, for each `Task`, and use `next()` to execute the run the two steps.

I'm not saying one is better than the other, they are two ways to achieve the same result. The advantage of the generator is that it keeps an internal state among the steps (you don't have to return and pass again as a parameter), but this can also make it more complex to debug. Every use case has its own trade-offs.

AleCandido: > Ok, so you are suggesting something like this:

Almost

> At this point we could also define an intermediate object that collects also a task with data.

Indeed, but then I would not define a further class, nor a status. Just make them optional attributes, and now really interface with properties: we can hook the save in the setter.

src/qibocal/auto/operation.py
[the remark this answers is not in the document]
Jacfomg: Yup, we should be doing that. 

> Find the optimal demodulation weight function by measuring the average detector voltage as a function of time during the course of the readout pulse (https://www.nature.com/articles/s41586-019-1666-5)

At least for dummy sampling rate the are about 10KB, twice as much for Zurich and about 5 times for the RFSoC per qubit assuming a 2us readout pulse. They would not be human-readable in any case but they are not above what you mentioned.

Jacfomg (approved):

src/qibocal/bootstrap.py
```diff
@@ -35,13 +35,13 @@ def data_uncertainties(data, method=None, data_median=None, homogeneous=True):
                 if homogeneous
                 else [np.meadian(row) for row in data]
             )
-        precentile_interval = (
+        percentile_inteval = (
```
rodolfocarobene, line 38: Why this change?

rodolfocarobene (approved):
Thanks @andrea-pasquale for the PR, honestly everything looks good to me.

I was really looking for something to improve and noticed that there are some unused parameters in qibocal.auto.operation and that in qibocal.auto.operation, in qibocal.auto.task and in qibocal.auto.history there are several methods without docstring...

I am approving since these are actually things unrelated to this PR but if you are not planning to modify them soon, these should be easily fixable :-)

src/qibocal/bootstrap.py
andrea-pasquale: I think that this is a typo. I've never heard about "precentile_interval". But I could be wrong.

andrea-pasquale:
> Thanks @andrea-pasquale for the PR, honestly everything looks good to me.

Thanks for the review @rodolfocarobene.



> I was really looking for something to improve and noticed that there are some unused parameters in qibocal.auto.operation and that in qibocal.auto.operation, in qibocal.auto.task and in qibocal.auto.history there are several methods without docstring...

I can do it here, no problem.
Which parameters are you referring to in `qibocal.auto.operation`?

rodolfocarobene:
> I can do it here, no problem. Which parameters are you referring to in `qibocal.auto.operation`?

These! https://github.com/qiboteam/qibocal/actions/runs/5390135156/jobs/9785049841?pr=403#step:10:38

andrea-pasquale:
> > I can do it here, no problem. Which parameters are you referring to in `qibocal.auto.operation`?
> 
> These! https://github.com/qiboteam/qibocal/actions/runs/5390135156/jobs/9785049841?pr=403#step:10:38

I see. They are necessary otherwise https://github.com/qiboteam/qibocal/blob/save_fit/tests/test_auto.py will fail since
the functions do not respect the expected signature. I can disable pylint for these lines if you want.

rodolfocarobene:
> I can disable pylint for these lines if you want.

nono, fair enough. :-)

AleCandido:
One way or the other we will have to solve, it's not good to keep the warning.

But I also would like to avoid disabling Pylint. A better option would be to silence this warning in particular, because it can happen for all the interfaces classes. But I'd even like to keep them, if possible.

The most appealing solution, at the moment, it seems to be prefixing the variables with an underscore, but then we would be losing the kwargs compatibility.

At the moment, I'm not fully satisfied by any possible solution, but we can face the problem later on. Just leave the warning where it is.

PR #404, opened 2023-06-19
Fix declared dependencies

AleCandido:
A bunch of Qibocal dependencies have been left implicit.

To check dependencies, I used the following search:
```bash
rg import src/ |
   # ignore qibo and internal imports
   rg -v 'from qibo' | rg -v 'import qibo' | rg -v 'from \.' |
   # ignore known dependencies
   rg -v yaml | rg -v pandas | rg -v numpy | rg -v plotly | rg -v networkx | rg -v click | rg -v lmfit | rg -v jinja | rg -v pint | rg -v pydantic | rg -v scipy |
   # ignore a bunch of standard library modules
   rg -v pathlib | rg -v statistics | rg -v subprocess | rg -v getpass | rg -v datetime | rg -v base64 | rg -v uuid | rg -v urllib | rg -v socket | rg -v logging | rg -v shutil | rg -v typing | rg -v dataclasses | rg -v tempfile | rg -v importlib | rg -v functools | rg -v collections | rg -v copy | rg -v inspect | rg -v enum | rg -v argparse
```

The only remaining imports are `os`, `re`, and `abc` (which I manually checked, because the short name could have been ambiguous, ignoring more than what I wanted).

An even better check would have been to:
- delimit packages names with `\b` (to check for whole words, e.g. `\bscipy\b`)
- read the list of known dependencies automatically from `pyproject.toml` (but names might differ, since Python import packages are contained by PyPI distribution packages, and names should not coincide, like `pyyaml -> yaml`)
- parse the actual syntax tree, to obtain import statements (rather than grepping)

However, better one simple check then no checks.

Notice that I'm only checking `src/`, so not `tests/` (which further imports would belong to a different Poetry group), nor `serverscripts/` (which are not part of the package anyhow).

Checklist:
- [ ] Tests are passing.


pyproject.toml
```diff
@@ -18,32 +18,38 @@ classifiers = [
...
+pint = "^0.20.1"
 Pint-Pandas = "^0.3"
+click = "^8.1.3"
+jinja2 = "^3.1.2"
+plotly = "^5.15.0"
 dash = "^2.6.0"
 lmfit = "^1.0.3"
 skops = "^0.6.0"
 scikit-learn = "^1.2.1"
+setuptools = "^67.8.0"
```
AleCandido, line 36: Why do we depend on `setuptools`? @Edoardo-Pedicillo

Edoardo-Pedicillo: https://github.com/qiboteam/qibocal/pull/358#issuecomment-1568000778 as said in the comment, I am aware this is not a good solution, but I didn't find a better one.

AleCandido: Apart from being or not being a good solution, and why it is missing (i.e. Poetry removing it), I still don't know why (or where) we depend on `setuptools`.

Why did you have to add it? What was failing without it?

AleCandido: In the failing workflow there is no mention of `setuptools`:
https://github.com/qiboteam/qibocal/actions/runs/5114132038/jobs/9194100468

And my guess is that all the Pylint errors got fixed in the following commit by upgrading the Pylint version.
I believe you had a reason to also add `setuptools` dep, but I simply can not infer it from Git.

Edoardo-Pedicillo: > In the failing workflow there is no mention of setuptools:
https://github.com/qiboteam/qibocal/actions/runs/5114132038/jobs/9194100468

I tried to update only Pylint in the previous commit (https://github.com/qiboteam/qibocal/actions/runs/5111513096/jobs/9188551370#step:10:71), but I received the error `No module named 'pkg_resources'`, then I downgraded pylint in the commit related to the workflow you showed, but since I had not idea how to solve the new issues, I decided to upgrade pylint and to solve the `pkg_resources` issue adding manually `setuptools` in the toml.

AleCandido: We might get rid of this together with Dash (and it might happen before Dash people will solve the problem, judging from the speed they replied to @rodolfocarobene)

Edoardo-Pedicillo: Thank @AleCandido

Edoardo-Pedicillo (approved):

pyproject.toml
AleCandido: > Yes, I don't know if you noticed how many issues do they have open and how old are some of those...

Having many issues opened is not a problem per se', popular repos have more issues on average (the fraction of open issues is slightly more relevant).

The average age a bit more, because a healthy repo is either closing too old issues (that's the role of stale bot), or keep them alive with active discussion.

The biggest drawback is that is a library targeting a non-simple task, and there are only two people actively (q.b.) working on the project.
image

AleCandido: Not terrible activity, but not even wonderful. If this is not something we strictly need, I would rather look for a different solution (even because I already know how I would do it)

andrea-pasquale (approved):
Thanks @AleCandido.
LGTM.
Since I created conflicts by merging #372 I already fixed them here.

AleCandido:
> Since I created conflicts by merging #372 I already fixed them here.

Thank you very much @andrea-pasquale, this was a quick one after all :)

PR #406, opened 2023-06-19
Add operational modes

andrea-pasquale:
In this PR I propose to add different operational modes to split acquisition, fitting and report generation.
To achieve this I've added 3 new commands in qibocal:

- `qq acquire <runcard>` which performs the data acquisition (without generating the report)
- `qq fit <output_folder>` which performs the fitting/post-processing on a folder containing only the data acquisition
- `qq report <output_folder>` which generates the html report from an output folder

`qq` is now an empty command. To run autocalibration the default command will be `qq auto`.

In this PR I still need to fix tests and coverage.

Things to be fixed before merging:

- [x] restore possibility to call `qq fit` (currently it is not possible because the task.run uses `yield` and the first output will always be the `acquisition`)
- [x] fix RB to work with the new layout
- [x] fix tune landscape
- [x] restore iterative dump
- [x] fix loading and saving of single shot classification with classifiers
- [x] check that everything works with all protocols
- [x] implement recursive deserialization

Let me know what you think. 

Checklist:
- [x] Tests are passing.


pyproject.toml
```diff
@@ -91,6 +91,9 @@ build-backend = "poetry.core.masonry.api"
 [tool.poetry.scripts]
 qq = "qibocal:command"
 qq-upload = "qibocal:upload"
+qq-acquire = "qibocal:acquire"
+qq-report = "qibocal:report"
```
AleCandido, line 95: Do you have a reason for having multiple commands?

Otherwise, I'd suggest using a single one, and subcommands. It will look almost the same, but with a single entry point:
```sh
qq  # giving help with all subcommands
qq acquire  # instead of `qq-acquire`
```

This can be done (even nested) using [`@click.group`](https://click.palletsprojects.com/en/8.1.x/commands/), it makes it pretty simple (that's the reason why I started using `click` instead of `argparse`, with which it was possible, but a pain).
Required changes are minimal, since you should just replace:
```py
@click.command
def acquire(runcard, folder, force):
    ...
```
with:
```py
@qq.command()
def acquire(runcard, folder, force):
    ...
```
Btw, that's the reason why I called the module `_base.py`, instead of `cli.py`, since I split the subcommands over further files (and by now I'm even calling it just `base.py`, the underscore it's not worth it...).

andrea-pasquale: I didn't think about having subcommands. It sounds good to me.
`qq` by default should combine all the subcommands `acquire`, `fit` and `report`. I think that it should be possible.

AleCandido: Yes, it is.
https://click.palletsprojects.com/en/8.1.x/commands/#group-invocation-without-command

Personally, I would make it more explicit using something like `qq auto` (`auto` is just the first word it came to my mind), and keep the usual behavior of printing the help for `qq`.
But this is just a matter of taste, so you're free to decide on your own (in a sense, you're the Qibocal director :wink:)

andrea-pasquale: Perfect, thank you!

src/qibocal/auto/history.py
```diff
@@ -21,7 +22,12 @@ class Completed:
         be added
 
     """
+    data: Data
+    """Data acquired."""
     status: Status
+    """Routine status."""
+    results: Optional[Results] = None
```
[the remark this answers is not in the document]
andrea-pasquale: Actually here `Results` is optional only for a practical reason.
The report is generated using the `History` 
https://github.com/qiboteam/qibocal/blob/8823e74a323955186f94f80432e2906a2dfc7a6c/src/qibocal/cli/builders.py#L85
and with the new commands the user should also be able to generate the report without necessarily executing the fitting/post-processing.
Of course I could also drop this dependency on the `History` object and generate the report in another way.

src/qibocal/web/templates/template.html
```diff
@@ -160,7 +160,7 @@ <h1>{{ report.title }}</h1>
 
 
             {% set figures, fitting_report = report.single_qubit_plot(task_uid, qubit) %}
-            {% if "No fitting data" not in fitting_report %}
+            {% if fitting_report %}
```
AleCandido, line 163:
```suggestion
            {% if fitting_report is not None %}
```

src/qibocal/protocols/characterization/allxy/allxy.py
```diff
@@ -223,11 +223,11 @@ def _fit(_data: AllXYData) -> AllXYResults:
 
 
 # allXY
-def _plot(data: AllXYData, _fit: AllXYResults, qubit):
+def _plot(data: AllXYData, qubit, fit: AllXYResults = None):
     """Plotting function for allXY."""
 
     figures = []
-    fitting_report = "No fitting data"
+    fitting_report = ""
```
AleCandido, line 230:
```suggestion
    fitting_report = None
```

andrea-pasquale:
Since there are some conflicts and I'm modifying the base branch of this PR I would suggest you to wait to review this PR @stavros11 @AleCandido.
After we merge #403 I will fix conflicts and then you can review.

src/qibocal/protocols/characterization/allxy/allxy.py
```diff
@@ -223,11 +223,11 @@ def _fit(_data: AllXYData) -> AllXYResults:
 
 
 # allXY
-def _plot(data: AllXYData, _fit: AllXYResults, qubit):
+def _plot(data: AllXYData, qubit, fit: AllXYResults = None):
```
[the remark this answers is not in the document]
andrea-pasquale: This PR was tested only with `Qubit` but it should be `Union[Qubit, QubitPair]`

src/qibocal/utils.py
```diff
@@ -4,6 +4,20 @@
...
+        return int(key)
+    return key
+
+
+def my_eval(key):
+    if key.startswith("("):
+        test = key[key.find("(") + 1 : key.find(")")].replace(" ", "").split(",")
+        return tuple([cast_str_to_int(i) for i in test])
+
+    return cast_str_to_int(key)
```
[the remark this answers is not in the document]
andrea-pasquale: Yeah, I know that this is quite ugly.
It was what I was referring on telegram a few weeks ago.
Here the problem is just the fact I have a `tuple` key which is loaded as a `str`. This mess is just to the following:
`'0' -> 0`
or 
`'(0,1)' -> (0,1)`
Of course I could just use `eval`... but I know that it is not a good idea.

AleCandido:
@andrea-pasquale now that you merged #403, this should be merged/rebased on current `main`.

src/qibocal/cli/builders.py
```diff
@@ -17,33 +19,117 @@
...
-        runcard (path): path containing the runcard.
-        folder (path): path for the output folder.
-        force (bool): option to overwrite the output folder if it exists already.
-        update (bool): option to update platform after each routine.
-    """
+@dataclass
+class ExecutionMode(Enum):
+    acquire = auto()
+    fit = auto()
+    run = auto()
```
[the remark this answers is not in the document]
andrea-pasquale: I've renamed it to autocalibration (acquisition + fitting)

src/qibocal/web/templates/template.html
andrea-pasquale: It seems that the correct syntax for jinja2 is [`none`](https://stackoverflow.com/questions/19614027/jinja2-template-variable-if-none-object-set-a-default-value)

src/qibocal/auto/execute.py
```diff
@@ -121,23 +121,25 @@ def current(self):
...
         """
         self.head = self.graph.start
-
         while self.head is not None:
             task = self.current
-            task_execution = task.run(platform=self.platform, qubits=self.qubits)
             completed = Completed(task, Normal(), self.output)
-            completed.data = next(task_execution)
-            completed.results = next(task_execution)
+            task_execution = task.run(platform=self.platform, qubits=self.qubits)
```
AleCandido, line 135: Why do you create `completed` before running the task?

src/qibocal/auto/execute.py
```diff
@@ -121,23 +121,25 @@ def current(self):
...
+            if mode.name not in ["report"]:
```
AleCandido, line 136:
```suggestion
            if mode.name != "report":
```
Prefer the simplest operation.

src/qibocal/auto/execute.py
```diff
@@ -121,23 +121,25 @@ def current(self):
...
+                if mode.name in ["autocalibration", "fit"]:
+                    completed.results = next(task_execution)
             self.history.push(completed)
             self.head = self.next()
-            if self.platform is not None:
-                if self.update and task.update:
-                    self.platform.update(completed.results.update)
+            if mode.name in ["autocalibration", "fit"]:
+                if self.platform is not None:
+                    if self.update and task.update:
```
AleCandido, line 144: You can also use `and`:
```suggestion
            if mode.name in ["autocalibration", "fit"] and self.platform is not None and self.update and task.update:
```

AleCandido: If you want to group conditions, you can alias them with:
- the benefit of giving them a name
- the drawback of evaluating them (while `and` is short-circuit, like nested `if`s)

```suggestion
            update = self.update and task.update
            if mode.name in ["autocalibration", "fit"] and self.platform is not None and update:
```

src/qibocal/auto/history.py
```diff
@@ -34,16 +33,22 @@ class Completed:
...
-        os.makedirs(path)
+        if not os.path.isdir(path):
+            os.makedirs(path)
         return path
 
     @property
     def results(self):
         """Access task's results."""
+        if not (self.datapath / RESULTSFILE).is_file():
+            return self._results
```
AleCandido, line 48: I would return `self._results` every time that is not empty
```suggestion
        if self._results is not None:
            return self._results
```
No need of reloading.

If your goal is to also return `None` if there is no file, add it explicitly:

```suggestion
        if self._results is not None:
            return self._results
        if not (self.datapath / RESULTSFILE).is_file():
            return None
```

src/qibocal/auto/history.py
```diff
@@ -55,6 +60,11 @@ def results(self, results: Results):
     @property
     def data(self):
         """Access task's data."""
+        if not (self.datapath / DATAFILE).is_file():
+            return self._data
```
AleCandido, line 64: Same of above

src/qibocal/auto/history.py
```diff
@@ -79,6 +89,10 @@ class History(dict[tuple[Id, int], Completed]):
     def push(self, completed: Completed):
         """Adding completed task to history."""
         self[(completed.task.id, completed.task.iteration)] = completed
-        completed.task.iteration += 1
+
+        # FIXME: I'm not too sure why but the code doesn't work anymore
+        # with this line. We can postpone it when we will have the
+        # ExceptionalFlow.
+        # completed.task.iteration += 1
```
AleCandido, line 96: Do you have an example?
In principle, nothing should have changed in this respect.

src/qibocal/auto/operation.py
```diff
@@ -92,6 +94,20 @@ def to_json(self, path):
...
+    def load(cls, path):
+        if (path / JSONFILE).is_file():
+            params = json.loads((path / JSONFILE).read_text())
+            obj = cls(**params)
+        else:
+            obj = cls()
+        raw_data = np.load(path / DATAFILE)
+        for i in raw_data:
+            # FIXME: change eval asap
+            obj.data[my_eval(i)] = np.rec.array(raw_data[i])
```
AleCandido, line 107:
```suggestion
        for i, ar in raw_data.items():
            # FIXME: change eval asap
            obj.data[my_eval(i)] = np.rec.array(ar)
```

src/qibocal/auto/operation.py
```diff
@@ -92,6 +94,20 @@ def to_json(self, path):
...
+    @classmethod
+    def load(cls, path):
+        if (path / JSONFILE).is_file():
+            params = json.loads((path / JSONFILE).read_text())
+            obj = cls(**params)
+        else:
+            obj = cls()
+        raw_data = np.load(path / DATAFILE)
+        for i in raw_data:
+            # FIXME: change eval asap
```
AleCandido, line 106: Why do you need `eval` at all? Are they not just strings?

src/qibocal/auto/operation.py
```diff
@@ -133,6 +149,18 @@ def save(self, path):
...
         (path / RESULTSFILE).write_text(json.dumps(asdict(self), indent=4))
 
+    @classmethod
+    def load(cls, path):
+        params = json.loads((path / RESULTSFILE).read_text())
+
+        for key, elem in params.items():
+            if isinstance(elem, dict):
+                # FIXME: necessary since after loading QubitId is string and not int
+                # maybe convert all QubitIds into strings ?
```
AleCandido, line 159: Either that or convert all of them to `int`, such that you can just call `int()`, instead of `my_eval()`.

src/qibocal/utils.py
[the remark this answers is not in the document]
AleCandido: It is not a good idea in case of untrusted input. `eval` can execute arbitrary code, so you should always sanitize your input before passing to it, or just not do it.
This is true for websites, that receives input from external user. If we are planning to use Qibocal with data from someone else, it's also a good idea not using `eval` here.

However, you can deserialize with `pydantic`, since it's already a dependency. Just declare your format in a model.

src/qibocal/web/templates/template.html
```diff
@@ -160,7 +160,7 @@ <h1>{{ report.title }}</h1>
 
 
             {% set figures, fitting_report = report.single_qubit_plot(task_uid, qubit) %}
-            {% if "No fitting data" not in fitting_report %}
+            {% if fitting_report is not none %}
```
AleCandido, line 163:
```suggestion
            {% if fitting_report is not None %}
```

AleCandido (commented):
I still didn't review the routines, and I should take a second look at `builders.py`.

src/qibocal/web/templates/template.html
andrea-pasquale: With `None` it throws an error.
https://github.com/qiboteam/qibocal/pull/406#discussion_r1245340793

src/qibocal/auto/operation.py
andrea-pasquale: I believe that this is not an option since for some platforms we will need to use strings (qw25q).

src/qibocal/auto/execute.py
AleCandido: Ok, it doesn't change much. But they are swapped in `main`, I was wondering why they have been changed in here.

src/qibocal/auto/operation.py
andrea-pasquale: For some protocols they might be tuple: https://github.com/qiboteam/qibocal/blob/44fb65aeccd2a1254c5a7e1667d0ed85e8005d4c/src/qibocal/protocols/characterization/classification.py#L35

AleCandido: Ok, so they are essentially never strings.

However, to simplify the process, I'd suggest to stop using a `dict`, and instead just use a `list`.
To associate the values to some keys, we can just use another list, to be store in the JSON metadata (this can also be implemented by each routine, unless there is a consistent pattern to be factored out).

Consider that, if you have a list of keys `k` and a list of values `v`, to reconstruct the dictionary you only need to do `dict(zip(k, v))` (i.e. you can construct a dictionary from any iterable of pairs).

andrea-pasquale: > Ok, so they are essentially never strings.

Actually they can be strings. For example qm uses `A1, .., B1` as convection for naming qubits.

 

> However, to simplify the process, I'd suggest to stop using a `dict`, and instead just use a `list`.
> To associate the values to some keys, we can just use another list, to be store in the JSON metadata (this can also be implemented by each routine, unless there is a consistent pattern to be factored out).

I am not so sure about the list. I think that for readability the dict looks cleaner. Mainly because it some cases it will be a list of lists therefore it might be not trivial to specify the order. Another option will be to keep the dict in the frontend (the protocol) and when we store to file we can split the dict into keys and values.

AleCandido: There is no front-end, other than the runcard. The *user* of Qibolab writes a runcard and runs the CLI. People writing the routines are developers as well.

If we can't use `int`s, because sometimes we have to deal with string ids, then there is no point in discussing anything, we can't use a list (now, until `QubitId` will become a synonym of `int`).
We need a step in which the qubit can be called with the same name used in the lab, but this can be fully consumed by Qibolab. The moment we sort the qubits (that are always a finite amount) we can use integers, and contiguous integers `0, ..., n-1`. From there on we'll be able to start using integers, and avoid all the branching for conversions that is currently required (*if `int` do this, if `str` do that*).

src/qibocal/auto/operation.py
AleCandido: As written above: once sorted, the conversion is automatic. No need to propagate `str` wherever in the code, and add significantly more code to handle them.

src/qibocal/auto/history.py
AleCandido: If you have it, can you post directly the error for everyone?

At the moment, I don't have time to debug myself, but of course also the reproduction is useful.

andrea-pasquale:
After fighting with pydantic for a few days I manage to implement the `deserialization`. 
This should also all the problems related to string to tuple/int conversions and possible qubit ordering.
This is now the base class:
https://github.com/qiboteam/qibocal/blob/f4d02ebfe4ea0e11bb1e3226ac0baef1028dc8af/src/qibocal/auto/operation.py#L61-L124

and this is an example of an actual class

https://github.com/qiboteam/qibocal/blob/f4d02ebfe4ea0e11bb1e3226ac0baef1028dc8af/src/qibocal/protocols/characterization/classification.py#L28-L44

I had to introduce a dtype attribute because trying to perform an inheritance of a np.array with a custom dtype is not supported currently by pydantic.

I'm still using the following ugly functions to implement the deserialization in pydantic
https://github.com/qiboteam/qibocal/blob/f4d02ebfe4ea0e11bb1e3226ac0baef1028dc8af/src/qibocal/utils.py#L7-L17
I would be glad to remove them but I still need to figure out how.

What do you think @AleCandido?

andrea-pasquale:
@AleCandido whenever you have time let me know what you think about my attempt at deserializing with pydantic https://github.com/qiboteam/qibocal/pull/406#issuecomment-1620267832

src/qibocal/auto/operation.py
```diff
@@ -54,11 +58,26 @@ def load(cls, parameters):
...
+        smart_union = True
+
+    @validator("data", pre=True)
+    def validate_data(cls, data):
+        new_data = {}
+        for i, ar in data.items():
+            if "(" in i:
+                new_data[conversion(i)] = np.rec.array(ar)
+            else:
+                new_data[cast_to_int(i)] = np.rec.array(ar)
```
AleCandido, line 79: This should not be needed, because you should not deal with strings, but serialize as `int` or `list` directly.

src/qibocal/auto/operation.py
```diff
@@ -54,11 +58,26 @@ def load(cls, parameters):
...
+    class Config:
+        arbitrary_types_allowed = True
+        smart_union = True
+
+    @validator("data", pre=True)
+    def validate_data(cls, data):
+        new_data = {}
+        for i, ar in data.items():
+            if "(" in i:
+                new_data[conversion(i)] = np.rec.array(ar)
```
AleCandido, line 77: This one should be a `tuple()` call, because you got a list, but you want to use a tuple.

AleCandido (commented):
I see that most of the leftover troubles are coming from using the array names to serialize the keys.

My advice would be to use a hash or uid as keys for the array files, and instead save the mapping in the metadata, properly serialized.
If you wish, you could reconstruct the actual `data` mapping in the Pydantic equivalent of `__post_init__`

AleCandido:
> I had to introduce a dtype attribute because trying to perform an inheritance of a np.array with a custom dtype is not supported currently by pydantic.

These tools are sufficiently complex on their own, it is expected to find some limitations somewhere. If this is the price, it's quite cheap.

Regarding the other comment about the extra machinery, a solution should be proposed in the review.

src/qibocal/auto/operation.py
```diff
@@ -74,8 +93,9 @@ def __getitem__(self, qubit: Union[QubitId, tuple[QubitId, int]]):
     @property
     def global_params_dict(self):
```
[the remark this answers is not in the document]
andrea-pasquale: Fixed thanks.

src/qibocal/auto/operation.py
```diff
@@ -92,6 +112,17 @@ def to_json(self, path):
         if self.global_params_dict:
             (path / JSONFILE).write_text(json.dumps(self.global_params_dict, indent=4))
 
+    @classmethod
+    def load(cls, path):
+        data_dict = dict(np.load(path / DATAFILE))
```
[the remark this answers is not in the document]
andrea-pasquale: You are right, thank you.

src/qibocal/utils.py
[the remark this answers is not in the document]
andrea-pasquale: I'm resolving this since it is outdated.

andrea-pasquale:
@AleCandido I tried to implement what we discussed yesterday, this is just a first version.
The approach seems to work for all protocols except for:
- tune_landscape (I need to deserialize a `namedtuple`...)
- randomized_benchmarking 

Both should be visible in the tests failed.
Any idea on the `namedtuple`?
For the review I suggest to avoid looking into the protocols, just check the implementation at the high level.

src/qibocal/auto/history.py
```diff
@@ -32,13 +31,17 @@ class Completed:
     """Protocol data."""
     _results: Optional[Results] = None
     """Fitting output."""
+    data_time: Optional[float] = 0
+    """Protocol data."""
+    results_time: Optional[float] = 0
+    """Fitting output."""
```
AleCandido, line 37: If you declare them as `Optional`, then set the default to `None` (at least it's clear that is empty).

If you do not need to distinguish `0.` and `None`, and `0.` is already good enough, just avoid `Optional`.

src/qibocal/auto/execute.py
```diff
@@ -153,4 +150,4 @@ def run(self, mode):
                 and update
             ):
                 self.platform.update(completed.results.update)
-            yield acquisition_time, fit_time, task.id
+            yield completed.data_time, completed.results_time, task.id
```
AleCandido, line 153: In principle `task.id` should be enough to retrieve `completed` in the `.history`. So, you could avoid yielding back multiple values.

src/qibocal/auto/execute.py
```diff
@@ -132,18 +132,15 @@ def run(self, mode):
...
-                    platform=self.platform, qubits=self.qubits
-                )
-            if mode.name in ["autocalibration", "fit"]:
-                completed.results, fit_time = task.fit(completed.data)
+            task.run(
+                platform=self.platform,
+                qubits=self.qubits,
+                mode=mode,
+                completed=completed,
+            )
```
AleCandido, line 143: For better separation, since the initialization is pretty trivial at this point, you could avoid passing `completed` to `task.run()`, and make `task.run()` create it instead, and return once finished.

I.e. instead of
```py
completed = Completed(...)
task.run(..., completed)
```
just do
```py
completed = task.run(...)
```

The only element that is not available to `task.run()` is `self.output`, but:
- it seems to be the same for each task, and thus for each completed (can you confirm it?)
- it seems the only purpose of that field is to make `.datapath` and friends properties and full paths, but we can lift one of the conditions (i.e. either we pass the `self.output` as argument, or we just return a relative path to `self.output`, someone else's responsibility to make it absolute, accessing `self.output`)

src/qibocal/auto/task.py
```diff
@@ -136,6 +150,23 @@ def run(
             self.qubits = list(qubits)
         else:
             data, time = operation.acquisition(parameters, platform=platform)
-        yield data, time
-        results, time = operation.fit(data)
-        yield results, time
+        return data, time
+        # yield data, time
+        # results, time = operation.fit(data)
+        # yield results, time
```
AleCandido, line 156: As soon as you're done, drop the dead code (I'm not saying you should not commit it at all, it's fine to be flexible, but don't forget it).

src/qibocal/auto/execute.py
andrea-pasquale: When I was changing it I also felt that we could return `completed` when executing `task.run(...)`.


> * it seems to be the same for each task, and thus for each completed (can you confirm it?)

Indeed this is the case



> * it seems the only purpose of that field is to make `.datapath` and friends properties and full paths, but we can lift one of the conditions (i.e. either we pass the `self.output` as argument, or we just return a relative path to `self.output`, someone else's responsibility to make it absolute, accessing `self.output`)

I think that we can pass `self.output` as argument. I will give it a try now.

src/qibocal/cli/_base.py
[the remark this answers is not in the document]
andrea-pasquale: The current approach will all the builders was just a first attempt to make sure that everything worked as I expected. The function approach will of course be cleaner. I'm not sure whether we can do it for the `ReportBuilder`, mainly because it used to render the html https://github.com/qiboteam/qibocal/blob/e064c357c0af7a58fdf54001ddec45ef946da0e2/src/qibocal/web/report.py#L14

If I remove it I guess that we will have the pass all the variables/functions that are needed to correctly produce the html report. Let's see how it goes.

src/qibocal/auto/execute.py
AleCandido: > I think that we can pass `self.output` as argument. I will give it a try now.

I know we can, but one argument less is better than one argument more (it's not only about the number of arguments and characters, it's really the isolation: you should provide the minimal information possible to any function).

If `self.output` is going to be always the same, I would try to send it around as little as possible.

However, if you want to start like this do it, an argument to `.datapath` is still better than a further attribute in `Completed`, repeated over and over.

src/qibocal/auto/execute.py
```diff
@@ -134,12 +133,11 @@ def run(self, mode):
         while self.head is not None:
             task = self.current
             log.info(f"Running task {task.id}.")
-            completed = Completed(task, Normal(), self.output)
-            task.run(
+            completed = task.run(
                 platform=self.platform,
                 qubits=self.qubits,
+                folder=self.output,
```
AleCandido, line 139: This actually I wanted to avoid: in case you still want to set it in `completed` for a while (that eventually I'd like to avoid as well, but not necessarily now) just leave it `None` when you create the object, and override here with:
```py
completed.folder = self.output
```

src/qibocal/auto/execute.py
```diff
@@ -150,4 +148,4 @@ def run(self, mode):
                 and update
             ):
                 self.platform.update(completed.results.update)
-            yield completed.data_time, completed.results_time, task.id
+            yield task
```
AleCandido, line 151: Why not just the `id`? What else do you need from `task` outside?
(the idea is that usually you just want them to execute, and in case you need something more, you can always retrieve it from the `history`, given a suitable `id`)

src/qibocal/cli/_base.py
```diff
@@ -66,10 +74,40 @@ def auto(runcard, folder, force, update):
...
+
+    if platform is not None:
+        platform.connect()
+        platform.setup()
+        platform.start()
+
+    for task in executor.run(mode=ExecutionMode.autocalibration):
+        timing_task = {}
+        timing_task["acquisition"] = executor.history[task.uid].data_time
+        timing_task["fit"] = executor.history[task.uid].results_time
```
AleCandido, line 94: `executor.history[task.uid]` should be equal to `task`, in principle.

src/qibocal/auto/execute.py
andrea-pasquale: Yeah, I know. I read your comment. I was planning to change it afterwards.

src/qibocal/cli/_base.py
```diff
@@ -66,10 +74,40 @@ def auto(runcard, folder, force, update):
...
+        timing_task["tot"] = timing_task["acquisition"] + timing_task["fit"]
+        # TODO: Change this to task.uid which is JSON friendly
+        meta[task.id] = timing_task
```
AleCandido, line 97: All of this can be reconstructed from the `history`, instead of during the execution.

Just make a function that processes the `history` and extracts what you want :)

(in principle, you might have had to take the list of `taks.uid` to know in which order they have been executed, but since a few versions of Python is guaranteed that dictionaries are sorted by insertion, so even this information is available to `history` itself)

src/qibocal/auto/execute.py
andrea-pasquale: The problem is that for the history I am using `task.id` while for the timings I am using `task.uid`.
The difference is that task.uid is a tuple (task, iteration) while task.id is just the task.id. I cannot use task.uid for timings since I get an error when dumping to JSON since the key is a tuple. I added a `#TODO` in order to fix it.

src/qibocal/cli/_base.py
AleCandido: In case there will be nothing left in the loop, the iteration could just be performed by the `list` builtin:
```py 
executed = list(executor.run(mode=ExecutionMode.autocalibration))
```

src/qibocal/auto/execute.py
AleCandido: Why not returning `task.uid` then?

src/qibocal/auto/operation.py
```diff
@@ -113,12 +136,32 @@ def save(self, path):
 
     def to_npz(self, path):
         """Helper function to use np.savez while converting keys into strings."""
-        np.savez(path / DATAFILE, **{str(i): self.data[i] for i in self.data})
+        key = GenericKey()
+        np.savez(path / DATAFILE, **{key.dump(i): self.data[i] for i in self.data})
 
     def to_json(self, path):
```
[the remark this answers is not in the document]
andrea-pasquale: What do you mean exactly?
Given that data may contain some information which are common among all qubits/qubit pairs to avoid repeating the information in an array I thought that could be a good idea to separate the content of data in two file. One is the `npz` which contains long array of data while the common information are store in a json file.
https://github.com/qiboteam/qibocal/blob/b638d0420e3361962a7d4c46ae6143cc3d840005/src/qibocal/auto/operation.py#L32

src/qibocal/cli/_base.py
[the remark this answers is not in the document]
andrea-pasquale: I manage to implement everything here.
https://github.com/qiboteam/qibocal/blob/report_from_file/src/qibocal/cli/_base.py
For now I kept the `ReportBuilder` I think that removing it and passing lot's of argument is not necessary.
Perhaps there is some code repetition that could be avoided but overall it looks cleaner than before.

src/qibocal/auto/history.py
```diff
@@ -55,6 +60,14 @@ def results(self, results: Results):
     @property
     def data(self):
         """Access task's data."""
+        if (
+            not (self.datapath / DATAFILE).is_file()
+            and not (self.datapath / RB_DATAFILE).is_file()
```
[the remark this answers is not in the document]
andrea-pasquale: I was just playing around to be able to serialize and de-serialize also the RB.
If I don't manage to do it I will assign someone from the RB team.

src/qibocal/auto/execute.py
andrea-pasquale: Actually since I introduced a function for loading the timings into `meta` from `history` we don't need to return anything

src/qibocal/auto/operation.py
```diff
@@ -77,11 +84,26 @@ def load(cls, parameters):
         return cls(**parameters)
 
 
-class Data:
+class Data(BaseModel):
```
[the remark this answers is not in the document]
andrea-pasquale: In theory yes, I will give it a try to see if everything works. I'm expecting to change quite a lot of things since all the protocols were slightly modified to support pydantic :/

src/qibocal/auto/operation.py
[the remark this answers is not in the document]
AleCandido: > For example when I run a resonator spectroscopy I can do it at a specific amplitude which might be different to the one specified in the runcard. In this case I would like to save this value which is a single value for each qubit.

But these values are a function of the parameters and the platform (as everything should be, since a routine is getting nothing else as input).
The only thing that might change are the acquired data, because the system is a physical entity in the real world, and it is not perfectly reproducible.

So, instead of saving the computed value, expose the function to recompute it. Moreover, I don't know what is the current state of reproducibility for the platform, but if it is not, we should resume the diffs project.

src/qibocal/auto/execute.py
```diff
@@ -123,25 +123,34 @@ def current(self):
...
         """
         self.head = self.graph.start
-
         while self.head is not None:
+            acquisition_time = None
+            fit_time = None
             task = self.current
             log.info(f"Running task {task.id}.")
-            task_execution = task.run(platform=self.platform, qubits=self.qubits)
+            # task_execution = task.run(platform=self.platform, qubits=self.qubits)
```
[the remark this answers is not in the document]
AleCandido: Not sure we understood each other, but you did exactly what I attempted to describe. So thanks :)

src/qibocal/auto/history.py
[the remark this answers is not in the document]
AleCandido: Why the data are in a Pandas dataframe?

In any case, a dataframe is kind of an array, and it's easy to convert into it.
In particular, if you see here how it is constructed:
https://github.com/qiboteam/qibocal/blob/report_from_file/src/qibocal/protocols/characterization/randomized_benchmarking/standard_rb.py#L241
it already is made by a label (the associated depth) and an array (the samples).

As an easy solution, I would suggest turning this table of `depth | samples` into a map `${n}-${depth}: samples`, where `n` is an incremental number (in case there are equal depths).
In general, I'd avoid to store data in the array names, because it becomes too complicated quick enough. But as said in the other thread, there shouldn't be data to save in the acquisition phase, since complex computations should happen in the fit phase.

src/qibocal/cli/autocalibration.py
```diff
@@ -0,0 +1,73 @@
...
+    PLATFORM,
+    RUNCARD,
+    UPDATED_PLATFORM,
+    create_qubits_dict,
+    generate_meta,
+    generate_output_folder,
+)
+
+
+def run_autocalibration(card, folder, force, update):
```
AleCandido, line 23: Just a small note on names: try to be concise, if possible.

In this case, I would call it:
```suggestion
def autocalibrate(card, folder, force, update):
```

It contains the same level of information, and it doesn't sound bad, since it's still an action :)
(in general, identifiers are symbols, so we could even stretch a bit the English vocabulary and force to be verbs things that are not, analogously with names - for as long as it is clear enough)

src/qibocal/cli/fit.py
```diff
@@ -0,0 +1,42 @@
...
+def run_fit(folder, update):
+    """Post-processing analysis
+
+    Arguments:
+
+    - FOLDER: input folder.
+
+    """
+    # load path, meta, runcard and executor
+    path = pathlib.Path(folder)
```
AleCandido, line 24:
```suggestion
def run_fit(folder: Path, update):
    """Post-processing analysis

    Arguments:

    - FOLDER: input folder.

    """
```

Try to avoid converting as much as possible: if you need a `Path`, ask for it.

At some point, you'll hit the user. But if it's in Python, you should also ask for a `Path`. In this case, it is in the CLI, but `click` can directly provide you a `Path` (and check it exists or not, if you need it).

AleCandido: Also, dropping the conversion allows you to keep using `folder` as a name, even without shadowing it. And `folder` is more specific than `path` (slightly, but meaningfully).

src/qibocal/cli/_base.py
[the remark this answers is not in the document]
AleCandido: A better way would be to define a `dataclass` in `web.report`, declaring all the attributes you need, but with no methods, and populate that one in the `generate_report` function.

src/qibocal/cli/fit.py
andrea-pasquale: > In this case, it is in the CLI, but click can directly provide you a Path (and check it exists or not, if you need it).
Indeed this is correct. I've added the `[path_type` ](https://click.palletsprojects.com/en/8.1.x/parameters/).
Very useful!

Also from the discussion I forgot whether it is better to put `folder` or `path`. For now I converted everything to path.

src/qibocal/auto/serialize.py
```diff
@@ -0,0 +1,31 @@
...
+def serialize(raw: dict):
+    """JSON-friendly serialization for nested dict."""
+    for key, value in raw.items():
+        if isinstance(value, dict):
+            raw[key] = serialize(value)
+
+    return {
+        key if isinstance(key, str) else json.dumps(key): value
+        for key, value in raw.items()
+    }
```
AleCandido, line 22:
```suggestion
    return {
        key if isinstance(key, str) else json.dumps(key):
        value if not isinstance(value, dict) else serialize(value)
        for key, value in raw.items()
    }
```

In this way:
- you treat in the same way `key` and `value`
- you avoid the double iteration
- you avoid modifying the parameter
- you avoid copying the parameter

It seems a bit weird, but there are languages in which this is the only option (more or less for the reasons above).

src/qibocal/auto/serialize.py
```diff
@@ -0,0 +1,31 @@
+import json
+
+
+def deserialize(raw: dict):
+    """Deserialization of nested dict."""
+    for key, value in raw.items():
+        if isinstance(value, dict):
+            raw[key] = deserialize(value)
+
+    return {load(key): value for key, value in raw.items()}
```
AleCandido, line 10: Same as below: use the `if` expression, rather than modifying first the input object.

src/qibocal/auto/operation.py
```diff
@@ -189,25 +188,14 @@ def update(self) -> dict[str, ParameterValue]:
...
-        for result_name, result in asdict(self).items():
-            result_dict[result_name] = {}
-            if isinstance(result, dict):
-                for key, elem in result.items():
-                    result_dict[result_name][convert.dump(key)] = elem
-
-            else:
-                result_dict[result_name] = result
-        (path / RESULTSFILE).write_text(json.dumps(result_dict, indent=4))
+        (path / RESULTSFILE).write_text(json.dumps(serialize(asdict(self)), indent=4))
```
AleCandido, line 191:
```suggestion
        (path / RESULTSFILE).write_text(json.dumps(serialize(asdict(self))))
```
These files are not for manual inspection (mainly, at least). And it's always possible to easily reformat a JSON, even with standard CLI tools like `jq`.

Consider that each 4 spaces are 4 bytes (in UTF-8), so it would increase the size of data just for presentation purposes.
Let's reduce the required storage, and format on-demand.

src/qibocal/protocols/characterization/randomized_benchmarking/data.py
```diff
@@ -1,15 +1,16 @@
...
 
     def save(self, path):
-        super().to_csv(f"{path}/{self.__class__.__name__}.csv")
+        """Overwrite because qibocal action builder calls this function with a directory."""
+        super().to_json(f"{path}/{self.__class__.__name__}.json", default_handler=str)
 
     @classmethod
     def load(cls, path):
-        return read_csv(path)
+        return cls(pd.read_json(f"{path}/RBData.json"))
```
AleCandido, line 16: Thiis is inconsistent with the one above: once you're using `self.__class__.__name__` and once `RBData` explicitly.
Moreover, `__class__` and `__name__` are internal, and maybe not 100% reliable for this purpose.

Since they just have to be the same, the easiest solution is just to put a top-level constant:
```py
DATAFILE = "RBData"
```
(or `"RBData.json"`, if you prefer) and reuse the same in both.

src/qibocal/protocols/characterization/randomized_benchmarking/data.py
```diff
@@ -1,15 +1,16 @@
...
+class RBData(pd.DataFrame):
     """A pandas DataFrame child. The output of the acquisition function."""
 
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
 
     def save(self, path):
-        super().to_csv(f"{path}/{self.__class__.__name__}.csv")
+        """Overwrite because qibocal action builder calls this function with a directory."""
+        super().to_json(f"{path}/{self.__class__.__name__}.json", default_handler=str)
```
AleCandido, line 12: If `path` is a `pathlib.Path` variable, avoid composing it as a string. Use instead:
```py
path / "filename.json"
```

It's not incredibly relevant, but it's clean enough and allows you to run on Windows as well (someone else is doing the proper job for you, so why not).

src/qibocal/auto/serialize.py
```diff
@@ -0,0 +1,31 @@
+import json
+
+
+def deserialize(raw: dict):
+    """Deserialization of nested dict."""
+    return {
+        # TODO: don't apply load(key) if the key is an actual string
+        load(key): value if not isinstance(value, dict) else deserialize(value)
+        for key, value in raw.items()
+    }
```
andrea-pasquale, line 10: Thanks @AleCandido for all the feedback I really appreciate it.
I have one question concerning deserialization.
This approach generally works but it tends to fail when withing the nested dict there is a key which is an actual string. Is there a quick way in which I can distinguish an actual string from something like '0' or '(0,1)'?
I could try to put those cases inside the function `load` eventually..

(Btw, this is why tests are failing now. I've added also the report generation in tests to increase coverage and this issue came up)

AleCandido: Thanks for the tests, it's incredibly better to fail in the CI than failing in production.

About strings: in JSON strings should be enclosed by double quotes, so:
```py
assert json.loads('0') == 0
assert json.loads('"0"') == "0"
```
Everything that has double quotes in the serialization is a string, everything that does not have them should be something else.

To avoid exceptions, also serialize strings with `json.dumps()`
```py
assert json.dumps('0') == '"0"'
```

src/qibocal/protocols/characterization/randomized_benchmarking/data.py
```diff
@@ -9,8 +11,8 @@ def __init__(self, *args, **kwargs) -> None:
 
     def save(self, path):
         """Overwrite because qibocal action builder calls this function with a directory."""
-        super().to_json(f"{path}/{self.__class__.__name__}.json", default_handler=str)
+        super().to_json(path / DATAFILE, default_handler=str)
 
     @classmethod
     def load(cls, path):
-        return cls(pd.read_json(f"{path}/RBData.json"))
+        return cls(pd.read_json(f"{path}/{DATAFILE}"))
```
AleCandido: Same here concerning the path concatenation.

andrea-pasquale: Yep, I'm aware of it thanks. I just need to push the change.

src/qibocal/auto/history.py
```diff
@@ -1,70 +1,25 @@
...
 from .runcard import Id
-from .status import Status
-from .task import Task
+from .task import Completed
 
 
-@dataclass
-class Completed:
-    """A completed task."""
+def add_timings_to_meta(meta, history):
```
AleCandido, line 7:
```suggestion
def add_timings(meta, history):
```

Not my favorite name even like this, but I just wanted to remark conciseness once more :)

Being `meta` is chance, and it's already in the named of the variable both in the function and call site - the purpose of the function is just to add the timings to the parameter passed, whatever it is (provided it is a dictionary, of course).

AleCandido: Btw, I'd avoid getting meta as input: just create a new dictionary with the information.

If needed, you can merge them after the call.

There is a [dedicated operator](https://docs.python.org/3/library/stdtypes.html?highlight=dict#dict.values) to merge dictionaries in Python:
```py
merged = d1 | d2
```

Up to py3.8 the proper way would have been `{**d1, **d2}`

More info about merging dicts:
https://stackoverflow.com/a/26853961/8653979

src/qibocal/auto/history.py
```diff
@@ -1,70 +1,25 @@
...
-    task: Task
-    """A snapshot of the task when it was completed.
+        if "acquisition" not in meta[task_id] and completed.data_time > 0:
+            meta[task_id]["acquisition"] = completed.data_time
+        if "fit" not in meta[task_id] and completed.results_time > 0:
+            meta[task_id]["fit"] = completed.results_time
+        if "acquisition" in meta[task_id]:
+            meta[task_id]["tot"] = meta[task_id]["acquisition"]
+        if "fit" in meta[task_id]:
+            meta[task_id]["tot"] += meta[task_id]["fit"]
```
AleCandido, line 20: If you use a newly created dictionary, you do not even need all the checks, and just create a new `times` dictionary per task, to eventually append to the full dictionary to be returned.

src/qibocal/auto/history.py
```diff
@@ -79,6 +34,21 @@ class History(dict[tuple[Id, int], Completed]):
...
+    @property
+    def timing(self):
+        time = {}
+        for task_id, iteration in self:
+            completed = self[(task_id, iteration)]
+            time[task_id] = {}
+            time[task_id]["acquisition"] = completed.data_time
+            time[task_id]["fit"] = completed.results_time
+            time[task_id]["tot"] = completed.data_time + completed.results_time
+        return time
```
andrea-pasquale, line 54: To be removed

src/qibocal/cli/_base.py
```diff
@@ -45,28 +57,91 @@
...
+@click.argument(
+    "path", metavar="FOLDER", type=click.Path(exists=True, path_type=pathlib.Path)
+)
+def upload(path):
+    """Uploads output folder to server
+
+    Arguments:
+
+    - FOLDER: input folder.
+    """
```
AleCandido, line 144: Split even this function in a separate one, such that it could be called also from Python (not only from the CLI).

(I've never tried, and it might be that we don't need to split, in principle you could use `click` functions from Python as they are, but I would recommend keeping the Python API separate from the CLI "bindings" anyhow, for better separation)

src/qibocal/auto/serialize.py
andrea-pasquale: I manage to fix all tests, thank you!

src/qibocal/auto/history.py
andrea-pasquale: I see your point. There could be a few problems with just passing a dictionary.
Given that by default `data_time` and `results_time` are 0, I might overwrite a value in meta with 0 while it already had a value (for instance when running just fitting it will overwrite the acquisition time to 0). Which is the reason for each I implemented several checks below. 

A guess that I could have two separate functions. One for creating the timing dict and a second one for combining the current meta with the timing dict.

src/qibocal/auto/history.py
andrea-pasquale: I had a second look at this. 
As far as I understand the problem is here:
https://github.com/qiboteam/qibocal/blob/5fb37003d5f3af20d0c1f3f57160b55217eaa0d9/src/qibocal/auto/history.py#L37-L42
The error is generated by the fact that datapath depends directly on iteration which is increased when we push something to the history.
Given that this is related to exceptional flows I prefer to address this in a separate PR once we actually arrive at exceptional flow.

src/qibocal/auto/execute.py
andrea-pasquale: After giving it a try I disagree about passing `self.output` to `.datapath`. The reason for this is that currently `datapath` is a property, passing an argument means that everytime I need to access `data` or `results` I will need to pass a path...
I know that in this case we will avoid to set in `Completed`, however I don't see much benefits if then we need to pass it around all the time inside the class at the cost of converting several properties in methods.
Maybe I am missing something..

src/qibocal/auto/execute.py
andrea-pasquale: This makes sense instead

andrea-pasquale:
@AleCandido I believe that we should be more or less ready here.
I opened some issues for some open points, feel free to reply to the others whenever you have time.
I also plan to add documentation after #485 is merged to main .

AleCandido (approved):
If you want to merge, just do it. 
Everything else could be solved in separate issues.

However, I believe that the current priority of Qibocal is to reduce and simplify the infrastructure. We should try to what we are doing with the minimal (as reasonable amount of) code.

In particular, I would not care too much about what is inside `protocols/`, but attempt to polish everything else (eventually, I'd also like to review protocols, and improve the code in there, but in a sense we could start considering protocols as plugins, and actually expose a Plugin API for extending Qibocal).

src/qibocal/cli/autocalibration.py
```diff
@@ -0,0 +1,76 @@
...
+    PLATFORM,
+    RUNCARD,
+    UPDATED_PLATFORM,
+    create_qubits_dict,
+    generate_meta,
+    generate_output_folder,
+)
+
+
+def autocalibrate(card, folder, force, update):
```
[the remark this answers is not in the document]
andrea-pasquale: I know. Potentially the autocalibrate will become more complicated than a simple acquisition followed by a fitting. But for sure we could recover some parts from `acquire` and `fit`. Let's do it in a separate PR as you suggested.

src/qibocal/protocols/characterization/flux_dependence/utils.py
```diff
@@ -139,18 +141,34 @@ def flux_dependence_plot(data, fit, qubit):
...
-
             if fit.sweetspot[qubit] != 0:
-                fitting_report += f"{qubit} | Sweetspot: {fit.sweetspot[qubit]} V<br>"
+                fitting_report += (
+                    f"{qubit} | Sweetspot: {float(fit.sweetspot[qubit]):.3f} V<br>"
+                )
             else:
                 fitting_report += f"{qubit} | Sweetspot: Fitting not successful<br>"
 
+            params = {
```
AleCandido, line 151: Optimal top-level constant

Edoardo-Pedicillo (approved):
Thank @andrea-pasquale

PR #413, opened 2023-06-20
Remove statistics

### LINKED ISSUE #405: Replace `statistics` usage
The `statistics` standard library module is just a simple module in the standard library spirit of *batteries included*.
However, we are already working with arrays, and we depend on much better libraries, so we should use them, as spelled out in the stdlib itself:

> The module is not intended to be a competitor to third-party libraries such as [NumPy](https://numpy.org/), [SciPy](https://scipy.org/), or proprietary full-featured statistics packages aimed at professional statisticians such as Minitab, SAS and Matlab. It is aimed at the level of graphing and scientific calculators.
> https://docs.python.org/3/library/statistics.html

So, please, replace its only usage:

https://github.com/qiboteam/qibocal/blob/72435cbf202dcc4e79789f471314a8827e5c27c7/src/qibocal/protocols/characterization/utils.py#L190

with the equivalent SciPy function [`scipy.stats.mode`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.mode.html#scipy.stats.mode).

Edoardo-Pedicillo:
This PR closes #405 and tries to solve some issues related to the punch-out fitting function.
image

Checklist:
- [ ] Tests are passing.


andrea-pasquale (commented):
Thanks for fixing this issues @Edoardo-Pedicillo.
Since the fitting for the punchout attenuation should be more or less the same could you also refactor it in the same way?

src/qibocal/protocols/characterization/utils.py
```diff
@@ -171,3 +172,55 @@ def spectroscopy_plot(data, fit: Results, qubit):
...
+        middle_freq = (max_freq + min_freq) / 2
+
+        freq_hp = peak_freqs[peak_freqs < middle_freq]
+        freq_lp = peak_freqs[peak_freqs >= middle_freq]
+        freq_hp = mode(freq_hp)[0]
+        freq_lp = mode(freq_lp)[0]
+
+        ro_values = getattr(qubit_data, fit_type)[
+            np.argmax(qubit_data.msr[np.where(qubit_data.freq == freq_lp)[0]])
+        ]
```
andrea-pasquale, line 221: I think that this is not correct for the punchout over attenuation.
In that case we want the minimum value for the points at high attenuation (in this case high power).
Another possible solution could be to just invert the axis for the attenuation so that we get the same shape.

Edoardo-Pedicillo: > In that case we want the minimum value for the points at high attenuation (in this case high power).

I have implemented this one, but I cannot test it now.

DavidSarlle:
@Edoardo-Pedicillo thanks for implementing. I have tested everything with qblox. The fitting is correct for punchout attenuation and the code for punchout using amplitude also works.

So, please andrea, could we merge this PR as soon as possible?

PR #415, opened 2023-06-26
Branch for qibolab-paper hardware benchmarks

rodolfocarobene:
This PR should help coordinate the work for the hardware benchmarks for the qibolab paper.
At the moment I added some timing of:
* acquisition function
* fit function
* complete routine

I would also like to add a timing for the complete execution, but wasn't sure how to do it and where...
Also, for the moment I've just written loggers for timings, but we should probably dump things into files.

Moreover, I added a runcard with an hypothetical list of routines to execute. The last one "rotation circuit" is a routine that still does not exist, but we could easily write something integrated in qibocal (at elast in this branch)

### DIFF
--- a/runcards/benchmark.yml
+++ b/runcards/benchmark.yml
@@ -1,0 +1,118 @@
+platform: dummy
+
+qubits: [0]
+
+actions:
+
+  - id: resonator_spectroscopy high power
+    priority: 0
+    operation: resonator_spectroscopy
+    main: resonator punchout
+    parameters:
+      freq_width: 10_000_000
+      freq_step: 500_000
+      amplitude: 0.4
+      power_level: high
+      nshots: 1024
+      relaxation_time: 0
+
+  - id: resonator punchout
+    priority: 10
+    operation: resonator_punchout
+    main: resonator_spectroscopy low power
+    parameters:
+      freq_width: 5_000_000
+      freq_step: 1_000_000
+      min_amp_factor: 0.02
+      max_amp_factor: 2.4
+      step_amp_factor: 0.1
+      nshots: 1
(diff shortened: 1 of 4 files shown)


andrea-pasquale:
Thanks @rodolfocarobene for opening this PR.

> * I would also like to add a timing for the complete execution, but wasn't sure how to do it and where...

If by complete execution you mean the duration of the `qq` script this is already saved and shown in the html report.
See for example here:
https://github.com/qiboteam/qibocal/blob/6a4bf9e51a33e98c16157ec4ffc2af552fc4cb19/src/qibocal/cli/builders.py#L112-L116




> * Also, for the moment I've just written loggers for timings, but we should probably dump things into files.

With the new data structure it should be relatively easy to do it. Right now for each acquisition we save two files: a `npz` containing the raw data and a `json` containing general information. I would say that this could be included in the json file. 

After we merge  #403, which is just waiting for review, you have also the option to save the fitting results into a `json`. We can think about adding a field with the execution time. If you do this than in the fitting results you can also compute the total execution time as acquisition_time + fitting_time :)

Let me know if you need any help with these high level modifications from the qibocal side.

Jacfomg:
Should we also try to have some timing inside of the drivers to check the pulse translation or the creating of other things done there ?

rodolfocarobene:
> Should we also try to have some timing inside of the drivers to check the pulse translation or the creating of other things done there ? That way we can check where the time of the execution is being spent on and compare to how long that execution would take ideally.

It could be really useful if we are able to isolate the same operations on all the drivers... I'm not sure if it is possible.
In any case it would probably still be interesting for us

PR #421, opened 2023-07-01
Plot nicely when punchout fitting fails

Jacfomg:
Plot nicely when punchout fitting fails by removing the failed point from the plot at (0,0) 

Checklist:
- [ ] Tests are passing.


andrea-pasquale (commented):
Thanks for the proposal @Jacfomg.
I think that this PR might be not necessary after we merge #413, since it will remove the try except block for the punchout. Lets see which PR gets merged first.

PR #431, opened 2023-07-06
Fix double backend allocation

### LINKED ISSUE #428: Platform instantiated twice
Hi @andrea-pasquale ,
While debugging qblox driver I noticed the platform is being instantiated twice: 

the problem seems to be in auto/runcard.py lines 58 and 59
```python
    @cached_property
    def backend_obj(self) -> Backend:
        """Allocate backend."""
        set_backend(self.backend, self.platform)
        return construct_backend(self.backend, self.platform)
```
each of those lines result in the calling of the `create` method.
Below is a snapshot of the call stack for each of the calls:
image

andrea-pasquale:
Closes #428.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibocal/auto/runcard.py
+++ b/src/qibocal/auto/runcard.py
@@ -4,7 +4,7 @@ from typing import Any, NewType, Optional, Union
 
 from pydantic import Field
 from pydantic.dataclasses import dataclass
-from qibo.backends import Backend, construct_backend, set_backend
+from qibo.backends import Backend, GlobalBackend
 from qibolab.platform import Platform
 from qibolab.qubits import QubitId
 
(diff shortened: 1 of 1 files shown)


src/qibocal/auto/runcard.py
```diff
@@ -55,8 +55,8 @@ class Runcard:
     @cached_property
     def backend_obj(self) -> Backend:
         """Allocate backend."""
-        set_backend(self.backend, self.platform)
-        return construct_backend(self.backend, self.platform)
+        GlobalBackend.set_backend(self.backend, self.platform)
+        return GlobalBackend()
```
andrea-pasquale, line 59: After merging #432 here. Things might be confusing.
This is the fix for #428.

stavros11 (approved):
Looks good to me, thanks.

PR #432, opened 2023-07-06
Fix dispersive shift updated frequency

### LINKED ISSUE #430: Dispersive shift
The dispersive shift routine has something wrong.

The plot works, but the updated runcard has the wrong frequency:

image

runcard: 7371399127000000512

andrea-pasquale:
Closes #430.

I've also renamed `Frequency Shift` to `Chi` by dividing the shift by two (hopefully it is correct).

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibocal/protocols/characterization/dispersive_shift.py
+++ b/src/qibocal/protocols/characterization/dispersive_shift.py
@@ -327,7 +327,7 @@ def _plot(data: DispersiveShiftData, fit: DispersiveShiftResults, qubit):
         f"{qubit} | State one freq : {fit_data_1.frequency[qubit]*GHZ_TO_HZ:,.0f} Hz.<br>"
     )
     fitting_report = fitting_report + (
-        f"{qubit} | \u03A7 : {(fit_data_0.frequency[qubit]*GHZ_TO_HZ - fit_data_1.frequency[qubit]*GHZ_TO_HZ)/2:,.0f} Hz.<br>"
+        f"{qubit} | Chi : {(fit_data_0.frequency[qubit]*GHZ_TO_HZ - fit_data_1.frequency[qubit]*GHZ_TO_HZ)/2:,.0f} Hz.<br>"
     )
     fitting_report = fitting_report + (
         f"{qubit} | Best frequency : {fit.best_freq[qubit]*GHZ_TO_HZ:,.0f} Hz.<br>"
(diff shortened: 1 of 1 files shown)


rodolfocarobene:
Tests seems to be failing for unrelated reasons, but the routine now works (and improves the fidelity!)

rodolfocarobene (approved):

andrea-pasquale:
> Tests seems to be failing for unrelated reasons, but the routine now works (and improves the fidelity!)

I'll have a look at tests now, thanks for the feedback.

rodolfocarobene:
I approved, but obviously this should be merged after the tests pass.
Also, for a future change, I think would be nice to have in the characterization section "readout frequency" AND "resonator frequency"

andrea-pasquale:
> Also, for a future change, I think would be nice to have in the characterization section "readout frequency" AND "resonator frequency"

In theory we already have the information about the readout frequency in the readout pulse.
We should be more careful when updating the runcard. For this particular case I would say that the dispersive shift should only update the frequency of the readout pulse without touching the resonator frequency.
What do you think?

rodolfocarobene:
It makes sense, but then I would change the naming in the characterization section that is now called "readout frequency" to "resonator frequency".
Obviously doesn't change much, but now it's a bit misleading and we "loose" some information

andrea-pasquale:
> It makes sense, but then I would change the naming in the characterization section that is now called "readout frequency" to "resonator frequency". Obviously doesn't change much, but now it's a bit misleading and we "loose" some information

I agree. We already have `bare resonator frequency` even if we are not using it in the runcards I think.
At this point we could have `bare resonator frequency` and `dressed resonator frequency` in the `characterization` section and we keep the readout freq in the pulse.

Jacfomg:
> > Also, for a future change, I think would be nice to have in the characterization section "readout frequency" AND "resonator frequency"
> 
> In theory we already have the information about the readout frequency in the readout pulse. We should be more careful when updating the runcard. For this particular case I would say that the dispersive shift should only update the frequency of the readout pulse without touching the resonator frequency. What do you think?

I think we only care about the readout frequency, we may only store the resonator frequency if we were interested on doing some simulation, right ?

andrea-pasquale:
> I think we only care about the readout frequency, we may only store the resonator frequency if we were interested on doing some simulation, right ?

To run things yes we need only the readout frequency. I would say that there is some value in also storing the resonator frequency.

Edoardo-Pedicillo (approved):

Jacfomg (approved):

PR #433, opened 2023-07-06
Qibolab benchmarks

rodolfocarobene:
Replaces #415 for the benchmarks
I merged main and some fix for dispersive shift. Now the execution produces a file called "timings.json" with the important info.

### DIFF
--- a/runcards/benchmark.yml
+++ b/runcards/benchmark.yml
@@ -1,0 +1,118 @@
+platform: dummy
+
+qubits: [0]
+
+actions:
+
+  - id: resonator_spectroscopy high power
+    priority: 0
+    operation: resonator_spectroscopy
+    main: resonator punchout
+    parameters:
+      freq_width: 10_000_000
+      freq_step: 500_000
+      amplitude: 0.4
+      power_level: high
+      nshots: 1024
+      relaxation_time: 0
+
+  - id: resonator punchout
+    priority: 10
+    operation: resonator_punchout
+    main: resonator_spectroscopy low power
+    parameters:
+      freq_width: 5_000_000
+      freq_step: 1_000_000
+      min_amp_factor: 0.02
+      max_amp_factor: 2.4
+      step_amp_factor: 0.1
+      nshots: 1
+      relaxation_time: 50
+
+  - id: resonator spectroscopy low power
+    priority: 0
+    operation: resonator_spectroscopy
+    next: qubit spectroscopy
+    parameters:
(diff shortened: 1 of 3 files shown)


Jacfomg:
What do you think about those scans for the scaling ?

rodolfocarobene:
@stavros11 @Jacfomg I've pushed a runcard for the scalings in 1D. in every routine the points are `1,10,100,1000`. For the single  point sweeper sometimes there is the need of changing the fit functions so that they do not fail.

With a `timings.json` for every type of sweeper is very easy to achieve something like this:
image

If you can execute them it would be nice, if you have other proposals please feel free to propose them :-)

runcards/scaling/1d_sweepers.yml
```diff
@@ -0,0 +1,255 @@
...
+  #- id: rabi amplitude 1
+  #  priority: 00
+  #  operation: rabi_amplitude
+  #  main: rabi amplitude 10
+  #  parameters:
+  #    min_amp_factor: 0
+  #    max_amp_factor: 1.0000
+  #    step_amp_factor: 1
+  #    pulse_length: 40
+  #    relaxation_time: 5_000
```
stavros11, line 125: The relaxation times here are different than the other benchmark. Since we are not using a qubit it is fine to do even Rabi with very short relaxation time, just need to make sure we report it properly in the paper.

stavros11 (commented):
Thanks, I will try soon. The runcard is scaling/1d_sweepers.yml? There is another runcard for scaling in iqm that has different point values, we need to make sure everyone is using the same.

runcards/scaling/1d_sweepers.yml
rodolfocarobene: Yes that was my point, but in any case we can change it.

I used `scaling/1d_sweepers.yml` that I derived from the one proposed by Juan

andrea-pasquale:
What is the plan for this PR @Jacfomg @stavros11?
Keep in mind that now the timings are stored into the `meta.json` file. See https://github.com/qiboteam/qibocal/pull/442

Jacfomg:
> What is the plan for this PR @Jacfomg @stavros11? Keep in mind that now the timings are stored into the `meta.json` file. See #442

This also saved times inside of the drivers I think, but in the end we decided not to use them. So, I would say that this PR is only important now because it has the runcards we used for the timings. We should save them somewhere.

PR #441, opened 2023-07-12
Fix scale fits

rodolfocarobene:
t2 and spin echo still had the scale problem... not sure if it was already fixed elsewhere

Checklist:
- [ ] Tests are passing.


Jacfomg (approved):
Thanks!

PR #442, opened 2023-07-12
Adding logs

andrea-pasquale:
This PR adds some logs in order to show which protocol is being executed.
The timing of both the acquisition and the fitting are shown for each protocol.
Here is an example

image

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibocal/auto/operation.py
+++ b/src/qibocal/auto/operation.py
@@ -30,7 +30,6 @@ RESULTSFILE = "results.json"
 
 def show_logs(func):
     """Decorator to add logs."""
-
     @wraps(func)
     # necessary to maintain the function signature
     def wrapper(*args, **kwds):
--- a/src/qibocal/auto/operation.py
+++ b/src/qibocal/auto/operation.py
@@ -1,12 +1,16 @@
 import inspect
 import json
+import time
 from dataclasses import asdict, dataclass, fields
+from functools import wraps
 from typing import Callable, Generic, NewType, TypeVar, Union
 
 import numpy as np
 import numpy.typing as npt
 from qibolab.platform import Platform
 from qibolab.qubits import Qubit, QubitId
+
+from qibocal.config import log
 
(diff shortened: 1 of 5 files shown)


Jacfomg:
Usually the fit takes almost nothing, could we add a check to only log that time when is above a threshold (1s maybe) ?

rodolfocarobene:
I kinda liked the "timing.json" file that we are producing for the benchmarks... Now that the the process is nicely written do you think we could have it directly in qibocal or is it useless?

andrea-pasquale:
> Usually the fit takes almost nothing, could we add a check to only log that time when is above a threshold (1s maybe) ?

Yep, I can put a threshold. Maybe 1 second is too much (?)

andrea-pasquale:
> I kinda liked the "timing.json" file that we are producing for the benchmarks... Now that the the process is nicely written do you think we could have it directly in qibocal or is it useless?

I need to see if there is a way to retrieve the information from the decorator. I feel like we don't need another json.
Perhaps we could store it somewhere else.

src/qibocal/auto/operation.py
```diff
@@ -24,6 +28,25 @@
 """Name of the file where results are dumped."""
 
 
+def show_logs(func):
+    """Decorator to add logs."""
+
```
rodolfocarobene, line 33:
```suggestion
```

rodolfocarobene (approved):

src/qibocal/auto/operation.py
andrea-pasquale: Thanks for catching this.

andrea-pasquale:
Just to a give an update now the timings are stored in the `meta.json` file (formerly it was a `yaml` but since it should not be modified by the user I decided to converted it to json.)
I've also added a threshold as requested .
Thanks @rodolfocarobene for the review already.

Jacfomg (approved):
Great thanks!

PR #447, opened 2023-07-18
Classifiers

Edoardo-Pedicillo:
This PR ports the classifiers' module to the autocalibration framework.

Checklist:
- [x] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


rodolfocarobene:
Thanks @Edoardo-Pedicillo, the code looks good to me (and finally it's fast!), I have just a couple of general comments:
* we should add a detailed explanation of the routine in the documentation, with a list of the classifiers and a general explanation. For example, I did not really know which classifiers were available. Also, I have no idea what a ROC curve is or what AUC stands for... Maybe the names on the axis could help, but in any case I would also go for a documentation page
* Maybe the plotting of the ROC curves and of the benchmarks should be plotted only if more than one classifiers is used
* I got some warnings, I don't know if you can fix them / suppress them in some way. otherwise good anyway

```
The keyword argument 'parallel=True' was specified but no transformation for parallel execution was p
ossible.

To find out why, try turning on parallel diagnostics, see https://numba.readthedocs.io/en/stable/user
/parallel.html#diagnostics for help.

File "../qibocal/src/qibocal/protocols/characterization/utils.py", line 175:
@njit(["float64[:] (float64[:], float64[:])"], parallel=True, cache=True)
def cumulative(input_data, points):
^

  warnings.warn(errors.NumbaPerformanceWarning(msg,
```
```
/home/rodolfoc/venv/qibo/lib/python3.10/site-packages/skl2onnx/algebra/onnx_ops.py:159: UserWarning
: OpSchema.FormalParameter.typeStr is deprecated and will be removed in 1.16. Use OpSchema.FormalPa
rameter.type_str instead.
  tys = obj.typeStr or ''
/home/rodolfoc/venv/qibo/lib/python3.10/site-packages/skl2onnx/algebra/automation.py:154: UserWarni
ng: OpSchema.FormalParameter.isHomogeneous is deprecated and will be removed in 1.16. Use OpSchema.
FormalParameter.is_homogeneous instead.
  if getattr(obj, 'isHomogeneous', False):
/home/rodolfoc/venv/qibo/lib/python3.10/site-packages/jinja2/environment.py:485: UserWarning: OpSch
ema.FormalParameter.typeStr is deprecated and will be removed in 1.16. Use OpSchema.FormalParameter
.type_str instead.
  return getattr(obj, attribute)
```

rodolfocarobene (approved):
There are some conflicts, but in general looks good to me

src/qibocal/protocols/characterization/classification.py
```diff
@@ -142,208 +237,354 @@ def _acquisition(
...
+            )
+            title_text += (
+                f"q{qubit}/{model} | threshold: {qubit_model.threshold:.6f}<br>"
+            )
+            title_text += (
+                f"q{qubit}/{model} | fidelity: {fit.models[qubit][i].fidelity:.3f}<br>"
+            )
+            title_text += f"q{qubit}/{model} | assignment fidelity: {fit.models[qubit][i].assignment_fidelity:.3f}<br>"
+
+        fitting_report += title_text
```
andrea-pasquale, line 551: Here I think that the `q` before qubit can be removed.
I'm not a fan of using `qubit/qubit_fit`. Perhaps we can write somewhere (maybe above the table) the fit model.
Eventually you could have a richer table summarizing all the information shown in the plot below for different classifiers (adding a classifier column). I don't know if this is strictly necessary.

src/qibocal/protocols/characterization/classification.py
```diff
@@ -142,208 +237,354 @@ def _acquisition(
...
+                height=COLUMNWIDTH,
+                width=COLUMNWIDTH * 3,
+                title=dict(text="Benchmarks", font=dict(size=TITLE_SIZE)),
+            )
+
+        title_text = ""
+
+        if models_name[i] == "qubit_fit":
+            qubit_model = fit.models[qubit][i]
+            title_text += f"q{qubit}/{model} | average state 0: {np.round(qubit_model.iq_mean0,3)}<br>"
```
andrea-pasquale, line 538:
```suggestion
        title_text = "No fitting data"

        if models_name[i] == "qubit_fit":
            qubit_model = fit.models[qubit][i]
            title_text = f"q{qubit}/{model} | average state 0: {np.round(qubit_model.iq_mean0,3)}<br>"
```
In this way you can avoid to show the empty table for non qubit-fit classifiers.

src/qibocal/protocols/characterization/utils.py
```diff
@@ -256,3 +257,11 @@ def fit_punchout(data: Data, fit_type: str):
...
         ro_values[qubit] = ro_val
     return [low_freqs, high_freqs, ro_values]
+
+
+def get_color_state0(number):
+    return "rgb" + str(hls_to_rgb((-0.35 - number * 9 / 20) % 1, 0.6, 0.75))
+
+
+def get_color_state1(number):
+    return "rgb" + str(hls_to_rgb((-0.02 - number * 9 / 20) % 1, 0.6, 0.75))
```
andrea-pasquale, line 267: This is fine. However, keep in mind that sometimes it is difficult to distinguish the points from the prediction of the classifier. Perhaps you could lower the opacity on the classifier prediction.

tests/runcards/protocols.yml
```diff
@@ -222,6 +222,13 @@ actions:
     parameters:
       nshots: 10
 
+  - id: single shot classification
+    priority: 0
+    operation: single_shot_classification
+    parameters:
+      nshots: 10
+      savedir: "cls_results"
+      classifiers_list: ["qubit_fit","naive_bayes", "linear_svm"]
```
andrea-pasquale, line 231: Should't we test all classifiers here?

src/qibocal/fitting/classifier/run.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def __init__(self, mod, base_dir: pathlib.Path) -> None:
+        self.mod = mod
+        self.base_dir = base_dir
+        self.trainable_model = None
+
+    @property
+    def name(self):
+        r"""Model's name."""
+        return self.mod.__name__.split(".")[-1]
```
andrea-pasquale, line 65: Perhaps we could also add a `pretty_name` property to show something more nice in the plot.
For example instead of showing `linear_svm` we could show `Linear SVM`.

src/qibocal/fitting/classifier/run.py
```diff
@@ -0,0 +1,264 @@
...
+        r"""The saving path."""
+        return self.base_dir / self.name
+
+    @property
+    def hyperfile(self):
+        r"""The path where the hyperparameters are stored."""
+        return self.savedir / HYPERFILE
+
+    def dump(self):
+        return self.mod.dump  # (self.trainable_model, path)
```
andrea-pasquale, line 100: This comment can be removed, right?

src/qibocal/protocols/characterization/classification.py
```diff
@@ -142,208 +237,354 @@ def _acquisition(
...
-            np.concatenate((real_values_state1, real_values_state0))
-        )
-        real_values_combined.sort()
-
-        cum_distribution_state1 = cumulative(real_values_combined, real_values_state1)
-        cum_distribution_state0 = cumulative(real_values_combined, real_values_state0)
-        cum_distribution_diff = np.abs(
-            np.array(cum_distribution_state1) - np.array(cum_distribution_state0)
+        benchmark_table, y_test, x_test, models, names, hpars_list = run.train_qubit(
+            data, qubit
```
andrea-pasquale, line 263: Here we could add a log "Training quibt with classifier {classifier}", so that if the training is long the user can see it through the log.

src/qibocal/protocols/characterization/classification.py
```diff
@@ -142,208 +237,354 @@ def _acquisition(
...
-        assignment_fidelities,
+        fidelity,
+        assignment_fidelity,
+        data.savedir,
     )
 
 
 def _plot(
     data: SingleShotClassificationData, fit: SingleShotClassificationResults, qubit
 ):
```
andrea-pasquale, line 301: The plot looks good. I don't know if this is trivial to do, but I think that it could be nice to have an option to go back to the previous plot (just the scatter plot with classifier prediction). For example we could add a verbose option in the parameters.

andrea-pasquale (commented):
Thanks @Edoardo-Pedicillo!
Generally it looks good to me, I tested it with both RFSoC and qblox and it works.
A few minor comments.
Could you please also edit the example [runcard](https://github.com/qiboteam/qibocal/blob/classifiers/runcards/actions_qq.yml) by adding the new version of this protocol?

src/qibocal/protocols/characterization/classification.py
Edoardo-Pedicillo: There is already one here:
https://github.com/qiboteam/qibocal/blob/1b0ed76883da901117eab91d5764ec8f68d2a64f/src/qibocal/fitting/classifier/run.py#L234

tests/runcards/protocols.yml
Edoardo-Pedicillo: It is not strictly necessary, since the coverage increases and testing all the classifiers requires too much time.

src/qibocal/protocols/characterization/classification.py
Edoardo-Pedicillo: I have added the `show_contour` parameter.

src/qibocal/protocols/characterization/classification.py
Edoardo-Pedicillo: > I'm not a fan of using qubit/qubit_fit

I have removed it.

> Eventually you could have a richer table summarizing all the information shown in the plot below for different classifiers (adding a classifier column). I don't know if this is strictly necessary.

I decided to avoid it because it is possible to read them on the plot and also these data are useful only when compared with each other (in a plot).

src/qibocal/protocols/characterization/classification.py
```diff
@@ -142,208 +245,354 @@ def _acquisition(
...
-        np.min(state0_data.q),
-        np.min(state1_data.q),
+    grid = np.vstack([i_values.ravel(), q_values.ravel()]).T
+
+    fig = make_subplots(
+        rows=1,
+        cols=len(models_name),
+        horizontal_spacing=SPACING * 3 / len(models_name),
+        vertical_spacing=SPACING,
+        subplot_titles=(models_name),
```
andrea-pasquale, line 361: Also here it would be nice to use the pretty name.

andrea-pasquale (commented):
Thanks for all the updates @Edoardo-Pedicillo.

> I have added the `show_contour` parameter.

I think that maybe it would be better to show or not show directly the contour by toggling it on and off in the legend. By going to the previous I meant to show just the classification and the raw data without anything else (possibly by increasing the size of the plot).
The problem with having the show_contour parameter is that in order to show it or not show it you will need to rerun the experiment again.

I also have a small comment down below.

PR #450, opened 2023-07-26
error bars in Ramsey

Edoardo-Pedicillo:
As the title says.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `errors` ( qiboteam/qibolab#571)


src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -145,45 +177,29 @@ def _acquisition(
                     )
 
             # execute the pulse sequence
+            if params.nboot != 0:
+                averaging_mode = AveragingMode.SINGLESHOT
+            else:
+                averaging_mode = AveragingMode.CYCLIC
```
rodolfocarobene, line 183:
```suggestion
            averaging_mode = AveragingMode.SINGLESHOT if params.nboot != 0 else AveragingMode.CYCLIC
```

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -204,79 +220,92 @@ def _fit(data: RamseyData) -> RamseyResults:
...
+                x = waits
+                try:
+                    popt = fitting(x, y)
+                    delta_fitting = popt[2] / (2 * np.pi)
+                    delta_phys = data.detuning_sign * int(
+                        (delta_fitting - data.n_osc / data.t_max) * GHZ_TO_HZ
+                    )
+
+                    # FIXME: for qblox the correct formula is the following (there is a bug related to the phase)
+                    # corrected_qubit_frequency = int(qubit_freq + delta_phys)
```
rodolfocarobene, line 255: I think this is no longer the case

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -204,79 +220,92 @@ def _fit(data: RamseyData) -> RamseyResults:
...
+    t2_err = {}
+    delta_phys_av = {}
+    delta_phys_err = {}
 
     for qubit in qubits:
         qubit_data = data[qubit]
-        voltages = qubit_data.msr * V_TO_UV
-        times = qubit_data.wait
         qubit_freq = data.qubit_freqs[qubit]
+        msrs = np.array(qubit_data[["msr"]].tolist()) * V_TO_UV
```
rodolfocarobene, line 235: Do you really need to convert it in array and then to list?

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -204,79 +220,92 @@ def _fit(data: RamseyData) -> RamseyResults:
...
+            log.warning(f"ramsey_fit: the fitting was not succesful. {e}")
+            popts[qubit] = [0] * 5
+
+        if data.nboot == 0:
+            delta_fitting = popts[qubit][2] / (2 * np.pi)
             delta_phys = data.detuning_sign * int(
                 (delta_fitting - data.n_osc / data.t_max) * GHZ_TO_HZ
             )
             # FIXME: for qblox the correct formula is the following (there is a bug related to the phase)
             # corrected_qubit_frequency = int(qubit_freq + delta_phys)
```
rodolfocarobene, line 290: Also this could be removed probably

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -204,79 +220,92 @@ def _fit(data: RamseyData) -> RamseyResults:
...
-        fitted_parameters[qubit] = popt
-        corrected_qubit_frequencies[qubit] = corrected_qubit_frequency * HZ_TO_GHZ
-        t2s[qubit] = t2
-        freqs_detuning[qubit] = delta_phys
+            freq_av[qubit] = corrected_qubit_frequency
+            freq_err[qubit] = 0.0
+            t2_av[qubit] = t2
+            t2_err[qubit] = 0.0
+            delta_phys_av[qubit] = delta_phys
+            delta_phys_err[qubit] = 0.0
```
rodolfocarobene, line 299:
```suggestion
            freq_av[qubit] = corrected_qubit_frequency
            t2_av[qubit] = t2
            delta_phys_av[qubit] = delta_phys
            freq_err[qubit] = t2_err[qubit] = delta_phys_err[qubit] = 0.0
```

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -287,11 +316,25 @@ def _plot(data: RamseyData, fit: RamseyResults, qubit):
     fig = go.Figure()
     fitting_report = ""
 
-    qubit_data = data[qubit]
+    qubit_data = data.data[qubit]
+    waits = np.unique(data.data[qubit].wait)
+    msrs = np.array(qubit_data[["msr"]].tolist()).flatten() * V_TO_UV
```
rodolfocarobene, line 321: Here there is also the double conversion...

rodolfocarobene (commented):
Thanks a lot @Edoardo-Pedicillo, I'll test it as soon as possible. In the meantime, here are some minor comments

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -34,6 +35,8 @@ class RamseyParameters(Parameters):
     """Number of shots."""
     relaxation_time: Optional[int] = None
     """Relaxation time (ns)."""
+    nboot: int = 0
+    """Number of bootstrap samples"""
```
andrea-pasquale, line 39: Since we are planning to add errors to all protocols. We can consider to add this as a parameter above `Parameters` like we did for `update` and `qubits`. If this PR is just for the ramsey it is fine.

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -42,10 +45,16 @@ class RamseyResults(Results):
 
     frequency: dict[QubitId, float] = field(metadata=dict(update="drive_frequency"))
     """Drive frequency [GHz] for each qubit."""
+    error_frequency: dict[QubitId, float]
+    """Error of the drive frequency [GHz]."""
     t2: dict[QubitId, float]
     """T2 for each qubit [ns]."""
+    error_t2: dict[QubitId, float]
+    """Error T2 [ns]."""
```
andrea-pasquale, line 53: At this point we could also think about having a dict like `dict[QubitId, tuple[float]]`.
To avoid generating many attributes. This will then require to update also the `update` method in qibolab.

andrea-pasquale (commented):
Thanks @Edoardo-Pedicillo.
I like the fact that we were able to add errors without having to modify completely the protocol.
Feel free to propagate errors also in the other protocols.

Edoardo-Pedicillo:
image

Edoardo-Pedicillo:
@andrea-pasquale @AleCandido before propagating the error bars in the other routines, I have a question: the routine's parameter `nboot` is the number of bootstrap samples, it should be a parameter in each routine that uses this method. Should we implement it as a global variable like `qubits` that is passed separately in `acquisition` https://github.com/qiboteam/qibocal/blob/98745570e0333531e8cdd957c511f0c7e0f6ccee/src/qibocal/auto/task.py#L132-L134
or we should simply add it as an attribute of `Parameters` like `nshots`?
https://github.com/qiboteam/qibocal/blob/98745570e0333531e8cdd957c511f0c7e0f6ccee/src/qibocal/auto/operation.py#L60-L63
I prefer the second option because of code cleanliness and because I don't want to set a precedent, but in this way we have some problems with the inheritance of the attributes (that we could maybe solve by dropping Python 3.9), so we are forced to repeat the variable.
https://github.com/qiboteam/qibocal/blob/98745570e0333531e8cdd957c511f0c7e0f6ccee/src/qibocal/protocols/characterization/flipping.py#L27-L30
@andrea-pasquale @AleCandido

andrea-pasquale:
> @andrea-pasquale @AleCandido before propagating the error bars in the other routines, I have a question: the routine's parameter `nboot` is the number of bootstrap samples, it should be a parameter in each routine that uses this method. Should we implement it as a global variable like `qubits` ([#450 (comment)](https://github.com/qiboteam/qibocal/pull/450#discussion_r1285183491)) that is passed separately in `acquisition`
> 
> https://github.com/qiboteam/qibocal/blob/98745570e0333531e8cdd957c511f0c7e0f6ccee/src/qibocal/auto/task.py#L132-L134
> 
> 
> or we should simply add it as an attribute of `Parameters` like `nshots`?
> https://github.com/qiboteam/qibocal/blob/98745570e0333531e8cdd957c511f0c7e0f6ccee/src/qibocal/auto/operation.py#L60-L63
> 
> 
> I prefer the second option because of code cleanliness and because I don't want to set a precedent, but in this way we have some problems with the inheritance of the attributes (that we could maybe solve by dropping Python 3.9), so we are forced to repeat the variable.
> https://github.com/qiboteam/qibocal/blob/98745570e0333531e8cdd957c511f0c7e0f6ccee/src/qibocal/protocols/characterization/flipping.py#L27-L30

I know that this is tricky. I would prefer not to remove Python3.9.
Personally, given that the number of bootstrap is not a routine - specific parameter, I would prefer the first option.
Also because repeating the variable for all the protocols can become a nightmare.

Edoardo-Pedicillo:
> but are you using the information on the errors during the fit?

No, should I? I have avoided it because the fit with errors requires more time, and since we should repeat it hundreds of times, it is time-consuming.

I think this PR is ready for a review (we can change the error evaluation in another moment).

AleCandido:
> No, should I? I have avoided it because the fit with errors requires more time, and since we should repeat it hundreds of times, it is time-consuming.

Well, it's not mandatory, but it would make sense. Noisy points should drive less the parameters' values.
If then the points are all equivalently noisy, it will make little difference in practice. But a priori this is not something you know, unless you there is a specific reason.

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -120,18 +127,40 @@ def _acquisition(
...
         n_osc=params.n_osc,
         t_max=params.delay_between_pulses_end,
         detuning_sign=+1,
+        nboot=params.nboot,
         qubit_freqs=freqs,
     )
 
-    if params.n_osc != 0:
-        # sweep the parameter
+    if params.n_osc == 0 and params.nboot == 0:
```
andrea-pasquale, line 138: If `n_osc` is 0 we don't apply bootstrap, why?

src/qibocal/protocols/characterization/ramsey_sequences.py
```diff
@@ -68,21 +69,27 @@ def _acquisition(
                     * (params.n_osc)
                     / params.delay_between_pulses_end
                 )
-
+        if params.nboot != 0:
+            averaging_mode = AveragingMode.SINGLESHOT
+        else:
+            averaging_mode = AveragingMode.CYCLIC
```
andrea-pasquale, line 75: Again we can move it directly in `execute_pulse_sequence`

andrea-pasquale (commented):
Thanks @Edoardo-Pedicillo.
Generally it looks good to me.

andrea-pasquale:
We should try also to fix coverage

andrea-pasquale:
> @andrea-pasquale @AleCandido before propagating the error bars in the other routines, I have a question: the routine's parameter `nboot` is the number of bootstrap samples, it should be a parameter in each routine that uses this method. Should we implement it as a global variable like `qubits` ([#450 (comment)](https://github.com/qiboteam/qibocal/pull/450#discussion_r1285183491)) that is passed separately in `acquisition`
> 
> https://github.com/qiboteam/qibocal/blob/98745570e0333531e8cdd957c511f0c7e0f6ccee/src/qibocal/auto/task.py#L132-L134
> 
> 
> or we should simply add it as an attribute of `Parameters` like `nshots`?
> https://github.com/qiboteam/qibocal/blob/98745570e0333531e8cdd957c511f0c7e0f6ccee/src/qibocal/auto/operation.py#L60-L63
> 
> 
> I prefer the second option because of code cleanliness and because I don't want to set a precedent, but in this way we have some problems with the inheritance of the attributes (that we could maybe solve by dropping Python 3.9), so we are forced to repeat the variable.
> https://github.com/qiboteam/qibocal/blob/98745570e0333531e8cdd957c511f0c7e0f6ccee/src/qibocal/protocols/characterization/flipping.py#L27-L30

What was the final decision for this?

src/qibocal/protocols/characterization/ramsey.py
Edoardo-Pedicillo: No, if `n_osc` is 0 we can apply bootstrap, but if  `n_osc` and `nboot` are both zeros we can use the Sweepers.

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -204,80 +216,78 @@ def _fit(data: RamseyData) -> RamseyResults:
...
+                freq_measure[qubit] = (np.mean(new_freqs), np.std(new_freqs))
+                t2_measure[qubit] = (np.mean(t2s), np.std(t2s))
+                delta_phys_measure[qubit] = (
+                    np.mean(deltas_phys_list),
+                    np.std(deltas_phys_list),
+                )
+                popts[qubit] = np.mean(fit_out, axis=0).tolist()
+                y = np.array(y)
+                x = np.array(x)
+                chi2[qubit] = chi2_reduced(y, ramsey_fit(x, *popts[qubit]), error_bars)
```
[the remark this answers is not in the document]
Edoardo-Pedicillo: yes

tests/runcards/protocols.yml
```diff
@@ -218,6 +218,17 @@ actions:
       n_osc: 10
       nshots: 10
 
+  - id: ramsey detuned sequences
```
andrea-pasquale, line 221:
```suggestion
  - id: ramsey detuned sequences with bootstrap
```

tests/runcards/protocols.yml
```diff
@@ -238,7 +249,16 @@ actions:
       n_osc: 0
       nshots: 1024
 
-
+  - id: ramsey
```
andrea-pasquale, line 252:
```suggestion
  - id: ramsey with bootstrap
```

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -342,3 +391,48 @@ def _plot(data: RamseyData, fit: RamseyResults, qubit):
...
+            [1, 1, np.inf, np.pi, np.inf],
+        ),
+    )[0]
+    popt = [
+        (y_max - y_min) * popt[0] + y_min,
+        (y_max - y_min) * popt[1] * np.exp(x_min * popt[4] / (x_max - x_min)),
+        popt[2] / (x_max - x_min),
+        popt[3] - x_min * popt[2] / (x_max - x_min),
+        popt[4] / (x_max - x_min),
+    ]
```
andrea-pasquale, line 437: Here make sure to add also the latest improvements that got merged to main with #482
https://github.com/qiboteam/qibocal/blob/e7aed916b63365ff8e2b0de1dfa44b3f20ff9e80/src/qibocal/protocols/characterization/ramsey.py#L221-L268

andrea-pasquale (commented):
Thanks @Edoardo-Pedicillo.
LGTM, you can find some suggestions below than we can merge.

Edoardo-Pedicillo:
Locally the tests are passing, the only issue is in Github where they are stucked

AleCandido:
I'm not worried about MacOS, we know it is frequently taking long to acquire a runner.
But Ubuntu is taking more than 1h to run tests, do you have any idea about?

Btw, this branch is running with the released Qibolab, why than would we need a new version?

Edoardo-Pedicillo:
> I'm not worried about MacOS, we know it is frequently taking long to acquire a runner.
But Ubuntu is taking more than 1h to run tests, do you have any idea about?

No, I tried to run the test locally, and they took less than one minute. I suppose, it is a Github issue.

>Btw, this branch is running with the released Qibolab, why then would we need a new version?

The tests pass because they run the protocols in the `no-update` mode (maybe we have to change it), but in the `update` mode the routines will fail, if we do not release a new Qibolab version. https://github.com/qiboteam/qibocal/blob/616836035dd863d9277b1b55c6bb93a31590f836/tests/test_protocols.py#L28-L34

andrea-pasquale:
> But Ubuntu is taking more than 1h to run tests, do you have any idea about?

It happens also to me locally once or twice. I don't have a precise idea. Perhaps lowering the number of evaluations could help, but we should carefully test on hardware before doing so.

andrea-pasquale (approved):

Edoardo-Pedicillo:
Now `update` is `True` and the tests do not pass as expected.

AleCandido:
> Regarding this, I believe that we can set `update=True`, originally it was false since I was testing all protocols in a single execution and by keeping `update=True` there was a possibility that a wrong result in the fitting may raise some qibolab errors when executing the next routine.
> 
> To check that everything is working we can turn it True as soon as we make a new release of qibolab.

I would actually check both: `update=False` is a simpler test, because there is less coupling, but therefore is more informative if failing. `update=True` is a stronger test, and more informative if passing.

andrea-pasquale:
> I would actually check both: `update=False` is a simpler test, because there is less coupling, but therefore is more informative if failing. `update=True` is a stronger test, and more informative if passing.

It doesn't hurt so we should probably do it.

AleCandido:
In any case: the `update=False` is already working. ***If** we do not need* to fix `update=True` in order to use Qibocal, then decorate that test with `@pytest.mark.skip`, and open a dedicated issue.

src/qibocal/protocols/characterization/utils.py
```diff
@@ -259,6 +266,55 @@ def fit_punchout(data: Data, fit_type: str):
...
+        name (str): Variable's name.
+        value (float): Variable's value.
+        error (float): Error associated to the variable.
+        unit (str): Measurement unit. Default value `None`.
+        ndigits (int): Number of decimal digits to display when error is `None`
+            (i.e. it is not evaluated).
+    """
+    table = f"{qubit}| {name}: "
+    if value:
+        magnitude = floor(log10(abs(value)))  # number of non decimal digits in value
```
AleCandido, line 292: This is sufficiently complex to be worth a function on its own (you already have the docstring ;)

src/qibocal/protocols/characterization/utils.py
```diff
@@ -259,6 +266,55 @@ def fit_punchout(data: Data, fit_type: str):
...
+    else:
+        magnitude = 0
+    if error:
+        ndigits = max(significant_digit(error * 10 ** (-1 * magnitude)), 0)
```
AleCandido, line 296: Being this module in a broader scope than RB, invert the dependency: move the definition of `significant_digit` in here, and import this module on that side.

src/qibocal/protocols/characterization/utils.py
```diff
@@ -259,6 +266,55 @@ def fit_punchout(data: Data, fit_type: str):
...
+    Args:
+        qubit (QubitId): Qubit.
+        name (str): Variable's name.
+        value (float): Variable's value.
+        error (float): Error associated to the variable.
+        unit (str): Measurement unit. Default value `None`.
+        ndigits (int): Number of decimal digits to display when error is `None`
+            (i.e. it is not evaluated).
+    """
+    table = f"{qubit}| {name}: "
```
AleCandido, line 290: This variable is a bit misnamed: it should be called `row` rather than `table`.

src/qibocal/protocols/characterization/utils.py
```diff
@@ -259,6 +266,55 @@ def fit_punchout(data: Data, fit_type: str):
...
+        magnitude = floor(log10(abs(value)))  # number of non decimal digits in value
+    else:
+        magnitude = 0
+    if error:
+        ndigits = max(significant_digit(error * 10 ** (-1 * magnitude)), 0)
+        table += f"({round(value*10**(-1*magnitude), ndigits)} {chr(177)} {np.format_float_positional(round(error*10**(-1*magnitude), ndigits), trim = '-')})"
+    else:
+        table += f"{round(value*10**(-1* magnitude), ndigits)}"
+    if magnitude != 0:
+        table += f"* 10^{magnitude}"
```
AleCandido, line 301: Do you really need such a complex formatting function? Can't you just use something like `f"{value:.{ndigits}g} +- {error:.{ndigits}g}`?

In any case, any use of standard formatters to replace some digits manipulations would be welcome.

src/qibocal/protocols/characterization/utils.py
```diff
@@ -259,6 +266,55 @@ def fit_punchout(data: Data, fit_type: str):
...
+            (i.e. it is not evaluated).
+    """
+    table = f"{qubit}| {name}: "
+    if value:
+        magnitude = floor(log10(abs(value)))  # number of non decimal digits in value
+    else:
+        magnitude = 0
+    if error:
+        ndigits = max(significant_digit(error * 10 ** (-1 * magnitude)), 0)
+        table += f"({round(value*10**(-1*magnitude), ndigits)} {chr(177)} {np.format_float_positional(round(error*10**(-1*magnitude), ndigits), trim = '-')})"
```
AleCandido, line 297: Python supports Unicode (i.e. even in its source), if you want to use `±` just do:
```suggestion
        table += f"({round(value*10**(-1*magnitude), ndigits)} ± {np.format_float_positional(round(error*10**(-1*magnitude), ndigits), trim = '-')})"
```

src/qibocal/auto/operation.py
```diff
@@ -61,6 +61,8 @@ class Parameters:
     """Number of executions on hardware"""
     relaxation_time: float
     """Wait time for the qubit to decohere back to the `gnd` state"""
+    nboot: int
```
AleCandido, line 64: I still wonder whether bootstrap is so general, or if eventually we'll use it only for a selection of protocols.

src/qibocal/auto/operation.py
```diff
@@ -74,7 +76,10 @@ def load(cls, parameters):
             the linked outputs
 
         """
-        return cls(**parameters)
+        nboot = parameters.pop("nboot", 0)
+        par = cls(**parameters)
+        par.nboot = nboot
+        return par
```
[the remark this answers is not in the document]
Edoardo-Pedicillo: In theory yes, but this approach is not working in practice: although `RamseyParameters`  inherits from `Parameters`, the first is not able to set the value of `nboot` during the initialization, i.e., `RamseyParameters(...., nboot = 1000)` fails.

It seems this can be fixed in `Python` 3.10 with Dataclasses `kw_only` attribute.

src/qibocal/auto/operation.py
Edoardo-Pedicillo: Bootstrap should be used in all the routines that have as output some values with physical meaning, that now are the majority.

src/qibocal/auto/operation.py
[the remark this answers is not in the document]
AleCandido: Why do you need something to be keyword only? Could you make a simple as possible snippet failing? (i.e. with two classes, one inheriting from the other, and minimal number of parameters as possible)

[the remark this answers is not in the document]
Edoardo-Pedicillo:
```py

class Parent:
    b:float = 5
    
@dataclass
class Son(Parent):
    ...
        
foo = Son(b=10)

#Output: TypeError: __init__() got an unexpected keyword argument 'b'
```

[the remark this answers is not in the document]
Edoardo-Pedicillo: If Parent is a dataclass too it works, but it is impossible to add a non-default argument in `Son`. All this examples are run with python 3.9.16

```py

@dataclass
class Parent:
    b:float = 5
    
@dataclass
class Son(Parent):
    a: float
        
foo = Son(10, b = 10)
```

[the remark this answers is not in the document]
AleCandido: Ok, I see the problem now. But the bit I missed was that `Parameters` is not a dataclass, despite having class attributes. It should be a dataclass, otherwise class attributes have a different meaning (they would be the same among all instances, making them practically globals).

Concerning the default instead I know the problem, in this sense dataclasses are not incredibly nice with subclassing. The current solution is good enough.

src/qibocal/protocols/characterization/utils.py
Edoardo-Pedicillo: I tried different solutions, but the problem is the use of scientific notation by default. This is the result of your suggestion.
Screenshot from 2023-09-15 12-08-18
By the way, I am not proud of the implementation of this function. Maybe we should find better solutions; for example, we could use the `to_html` Pandas function to generate a table automatically.

AleCandido: It is true that like this is a bit of a mess, but I believe it would be sensible if you split the information in three columns: `Value`, `Error`, `Unit`

Edoardo-Pedicillo: The problem of your proposal is that for the time being the template of the table is fixed by this lines 
https://github.com/qiboteam/qibocal/blob/9fb02a27005db6242a136edadb189640ed36172a/src/qibocal/web/templates/template.html#L148-L175
We could add extra columns, and then update the tables for all the routines, at this point I suggest discussing new possible (and more flexible) solutions to generate the report tables in a dedicated issue and then apport the changes in a new PR.

Edoardo-Pedicillo:
@AleCandido This PR is ready to be merged. If you agree, could you approve it ?  :)

AleCandido (approved):
That's simple :)

PR #455, opened 2023-08-08
Fix dispersive shift (qubit as str)

rodolfocarobene:
Basically, the dispersive shift didn't work for qubitId that weren't number, this should solve it

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


andrea-pasquale (approved):
Thanks @rodolfocarobene.
Good catch. To avoid such errors we should perhaps add to `dummy` also string QubitIDs.

PR #462, opened 2023-08-15
Implement flux crosstalk matrix

stavros11:
I added the possibility to investigate crosstalk in the fluxlines in the existing `qubit_flux` routine, because it was needed to explain some of the mysteries in the calibration of qw25qB. It can be used by passing the new `flux_qubits` parameter, for example:
```yml
platform: qw25qB

qubits: ["B1", "B2"]

actions:
  
  - id: qubit flux crosstalk
    priority: 0
    # main: t1
    operation: qubit_flux
    parameters:
      freq_width: 20_000_000
      freq_step: 400_000
      bias_width: 0.8
      bias_step: 0.0075
      drive_amplitude: 0.01
      flux_qubits: ["B1", "B2", "B3", "B4", "B5"]
      nshots: 1000
      relaxation_time: 10000
```
will perform spectroscopy on qubits B1 and B2 (multiplexed) while sweeping the bias of B1, B2, ..., B5 one at a time. Note that this will result to five (in general `len(flux_qubits)`) executions, but I think the biases need to be swept one at a time in order to extract meaningful results.

If `flux_qubits` is not given the behavior of the routine will be the same as before. The only difference is the color of the fit (because I changed the order of the plots), but we could force the old color if we need. Also, the colorbar in the crosstalk case is a bit unreadable but I did not try to fix it yet.

The same routine can be used to create avoided crossings:

Screenshot from 2023-08-15 16-41-13

In this plot we are sweeping the bias of qubit B2 (high frequency) while performing spectroscopy and reading B1 (low frequency).

Checklist:
- [ ] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`


src/qibocal/protocols/characterization/flux_dependence/qubit_flux_dependence.py
```diff
@@ -143,40 +173,58 @@ def _acquisition(
...
+            Sweeper(
+                Parameter.bias,
+                delta_bias_range,
+                qubits=[platform.qubits[flux_qubit]],
+                type=SweeperType.OFFSET,
+            )
+            for flux_qubit in flux_qubits
+        ]
+        # create a DataUnits object to store the results,
+        # DataUnits stores by default MSR, phase, i, q
```
andrea-pasquale, line 202: Same here.

src/qibocal/protocols/characterization/flux_dependence/qubit_flux_dependence.py
```diff
@@ -143,40 +173,58 @@ def _acquisition(
...
+        bias_sweepers = [
+            Sweeper(
+                Parameter.bias,
+                delta_bias_range,
+                qubits=list(qubits.values()),
+                type=SweeperType.OFFSET,
+            )
+        ]
+        # create a DataUnits object to store the results,
+        # DataUnits stores by default MSR, phase, i, q
```
andrea-pasquale, line 187: I know that this is a remnant of the previous code, but we could probably delete these comments since they are outdated.

andrea-pasquale (commented):
Thanks @stavros11, generally it looks good to me.
You can find below a few comments.
Just a tiny remark. The colormaps tend to overlap when you use more than one flux qubit.
image
I think that it should be possible to separate the color maps.

stavros11:
Thanks for the review @andrea-pasquale. I fixed your comments and removed the phase from the plots in order to have the matrix structure, as discussed in a previous meetings. I also implemented the crosstalk matrix for the resonator spectroscopy routine since this was also a TODO. Note that I disabled the colorbar from the matrix plots because I was unable to properly align it in the general case (because the columns of the matrix depends on `len(flux_qubits)`).

Other than that, this should be ready. I am not sure why the tests/coverage are failing in the CI.

src/qibocal/protocols/characterization/flux_dependence/utils.py
```diff
@@ -172,6 +184,54 @@ def flux_dependence_plot(data, fit, qubit):
     return figures, fitting_report
 
 
+def flux_crosstalk_plot(data, fit, qubit):
+    figures = []
+    fitting_report = ""
```
andrea-pasquale, line 189:
```suggestion
    fitting_report = "No fitting data"
```
In this way the empty table will be removed.
I know that this is weird... in #406 you will simply be able to set it to None.

src/qibocal/protocols/characterization/flux_dependence/utils.py
```diff
@@ -172,6 +184,54 @@ def flux_dependence_plot(data, fit, qubit):
...
+                x=frequencies,
+                y=qubit_data.bias,
+                z=qubit_data.msr * V_TO_UV,
+            ),
+            row=1,
+            col=col + 1,
+        )
+
+        fig.update_xaxes(
+            title_text=f"{qubit}: Frequency (Hz)",
```
andrea-pasquale, line 217:
```suggestion
            title_text="Frequency (Hz)",
```
I think that the label is clear also without `qubit`

andrea-pasquale (approved):
Thanks @stavros11.
After relaunching the jobs the CI seems to work.
I have a few small suggestions.
I guess that the first step will be to perform also a post-processing in order to compute the cross-talk matrix. We can do it in a separate. I like already the possibility of visualizing the cross-talk in this way.

src/qibocal/protocols/characterization/flux_dependence/utils.py
```diff
@@ -172,6 +184,54 @@ def flux_dependence_plot(data, fit, qubit):
...
+            row=1,
+            col=col + 1,
+        )
+        fig.update_yaxes(title_text=f"{flux_qubit}: Bias (V)", row=1, col=col + 1)
```
andrea-pasquale, line 221:
```suggestion
        fig.update_yaxes(title_text=f"Qubit {flux_qubit} Bias (V)", row=1, col=col + 1)
```

src/qibocal/protocols/characterization/flux_dependence/utils.py
stavros11: Thanks, I didn't know that. Indeed, using `None` makes more sense.

stavros11:
> I guess that the first step will be to perform also a post-processing in order to compute the cross-talk matrix. We can do it in a separate.

Yes, it would be useful to add fitting to this routine. I didn't do it yet because it is not very trivial: if `{qubit} - {flux_qubit}` are not interacting then we should get wide parabolas (ideally straight lines if no crosstalk), which I guess can be used to calculate the crosstalk matrix. However, if `{qubit} - {flux_qubit}` interact, we may get avoided crossings which should be handled differently by the fitting.

andrea-pasquale:
> Yes, it would be useful to add fitting to this routine. I didn't do it yet because it is not very trivial: if `{qubit} - {flux_qubit}` are not interacting then we should get wide parabolas (ideally straight lines if no crosstalk), which I guess can be used to calculate the crosstalk matrix. However, if `{qubit} - {flux_qubit}` interact, we may get avoided crossings which should be handled differently by the fitting.

Right... and also the fitting for the standard protocol (one qubit - one flux protocol) is very unstable at the moment, we should also try to improve the fitting before trying to go for something like this.
Thanks for all the updates!

Edoardo-Pedicillo (approved):

PR #480, opened 2023-08-22
Runcards qibolab paper

Jacfomg:
I think those are the scans we did. 

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


runcards/scaling/1d_sweepers.yml
```diff
@@ -0,0 +1,292 @@
+platform: iqm5q
```
stavros11, line 1: Here maybe we should put
```suggestion
platform: dummy
```
to make it neutral, like the other file.

Also, I am not sure if we really need the folder as it only contains one file. Maybe we can just rename it `scaling_benchmarks.yml` or something like that and leave it under runcards.

stavros11 (commented):
Thanks for opening this. The scaling runcard is the same with the one I used, but `benchmarks.yml` is different. I believe in the end we used [benchmark_iqm5q_q0.yml](https://github.com/qiboteam/qibocal/pull/433/files#diff-f5b55b1130a20d311f2933901ccacce2b68103dcd30558677835be918553ba20) from #433, not [benchmark.yml](https://github.com/qiboteam/qibocal/pull/433/files#diff-c3e6f1df1daecd9fdbc965ba645a3669da52f82c7311408c6ab242a934ea7de4) that is copied here.

runcards/scaling/1d_sweepers.yml
Jacfomg: I left just in case we go for a 2d_sweepers thing in the future, but yeah right now it's a bit pointless.

stavros11: Sure, that is fine with me. Thanks for the updates.

stavros11 (approved):

runcards/scaling/1d_sweepers.yml
Jacfomg: I will add the dummy as well when I can push

andrea-pasquale (approved):
Thanks @Jacfomg @stavros11

PR #482, opened 2023-08-23
Fix problems with Rabi fitting

### LINKED ISSUE #475: Rabi Amplitude fitting function fails
Hi, 
the current implementation of the fitting function often fails:
image
I presume it is the result of not being able to generate several oscillations, but this often going to happen. As you can see in the plot, the pi pulse was calibrated for an amplitude of 0.5. With this, sweeping the amplitude from 0 to 1 only results in half an oscillation.
I suppose I could have calibrated the gate using a smaller amplitude, but that would have been at the expense of losing resolution.
Ideally we would like to have a fit that catered for both scenarios, not having to change the routing to ensure the fitting doesn't fail.

andrea-pasquale:
Closes #475.
After performing the FFT I added the [`find_peaks`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.find_peaks.html) primitive from `scipy`.
In the case where not oscillations are observed the initial guess for the oscillations is 0.5.

I tried it on hardware and it seems to work as expected.

image

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/characterization/rabi/amplitude.py
+++ b/src/qibocal/protocols/characterization/rabi/amplitude.py
@@ -166,7 +166,6 @@ def _fit(data: RabiAmplitudeData) -> RabiAmplitudeResults:
         index = local_maxima[0] if len(local_maxima) > 0 else None
         # 0.5 hardcoded guess for less than one oscillation
         f = x[index] / (x[1] - x[0]) if index is not None else 0.5
-        print(f)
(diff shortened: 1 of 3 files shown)


aorgazf (approved):
I have tested the fit and it works well now, thanks @andrea-pasquale

Edoardo-Pedicillo (approved):

PR #487, opened 2023-08-24
David/qubit flux track

DavidSarlle:
This PR implements the qubit spectroscopy flux tracking. 

For that, we first save the polinomial coefficients obtained during the fitting resonator spectroscopy flux  in the runcard.
Afterthat, the "track" option has been added to Qubit Spec Flux to enable the flux tracking. If it is enabled, we modifiy the RO frequency with the estimated value using the "resonator_polycoef_flux" parameters stored in the runcard, for each bias value, in order to track the qubit during the spectroscopy. 

Checklist:
- [ ] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `https://github.com/qiboteam/qibolab/pull/573/`

### DIFF
--- a/src/qibocal/protocols/characterization/flux_dependence/qubit_flux_dependence.py
+++ b/src/qibocal/protocols/characterization/flux_dependence/qubit_flux_dependence.py
@@ -40,6 +40,8 @@ class QubitFluxParameters(Parameters):
     """Relaxation time (ns)."""
     transition: Optional[str] = "01"
     """Flux spectroscopy transition type ("01" or "02"). Default value is 01"""
+    track: Optional[bool] = False
+    """Enable Qubit Flux spectroscopy trancking. tracking disabled by default"""
 
 
 @dataclass
@@ -133,6 +135,11 @@ def _acquisition(
     delta_frequency_range = np.arange(
         -params.freq_width // 2, params.freq_width // 2, params.freq_step
     )
+
+    delta_bias_range = np.arange(
+        -params.bias_width / 2, params.bias_width / 2, params.bias_step
+    )
+
     freq_sweeper = Sweeper(
         Parameter.frequency,
         delta_frequency_range,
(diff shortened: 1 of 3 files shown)


src/qibocal/protocols/characterization/flux_dependence/qubit_flux_dependence.py
```diff
@@ -133,51 +135,93 @@ def _acquisition(
...
+                )
+
+                # modify qubit resonator frequency
+                qubits[qubit].readout_frequency = freq_resonator
+
+                # modify qubit flux
+                qubits[qubit].flux = bias
+
+                # execute pulse sequence sweeping only qubit resonator
+                results = platform.sweep
```
DavidSarlle, line 202: @andrea-pasquale this oart of the code is not working. I do not know if is possible to execute 1d sweeper inside the bias loop. I am getting the error:

  File "/nfs/users/david.fuentes/qibocal/src/qibocal/protocols/characterization/flux_dependence/qubit_flux_dependence.py", line 215, in _acquisition
TypeError: 'method' object is not subscriptable

andrea-pasquale: I think that we need to move the parentheses on line 202.
The error seems that you are overwriting the sweep method.

DavidSarlle:
Seems that this way of setting the bias: 

qubits[qubit].flux = bias 

Is not working with qblox. The qubit flux track routine is ready and the routine has no errors, but the results are not ok. Executing the tracking mode, the bias is not changed in the platform and the same result is obtained for each value of the bias range. 

Pending to test with other drivers to see if it a driver problem or a platform problem.

andrea-pasquale:
> Seems that this way of setting the bias:
> 
> qubits[qubit].flux = bias
> 
> Is not working with qblox. The qubit flux track routine is ready and the routine has no errors, but the results are not ok. Executing the tracking mode, the bias is not changed in the platform and the same result is obtained for each value of the bias range.
> 
> Pending to test with other drivers to see if it is a driver problem or a platform problem.

@stavros11 do you know what could be the issue here?

DavidSarlle:
closed for https://github.com/qiboteam/qibocal/tree/david/qubit_tracking

PR #489, opened 2023-08-25
Fix classification after #447

andrea-pasquale:
This PR fixes the threshold value computed by `Qubit Fit`.
The problem was the one already detected in #305.
I've also decided to show the ROC curve only when more than one classifier is used.

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/characterization/classification.py
+++ b/src/qibocal/protocols/characterization/classification.py
@@ -351,7 +351,7 @@ def _plot(
     fig = make_subplots(
         rows=1,
         cols=len(models_name),
-        horizontal_spacing=SPACING / len(models_name) * 3,
+        horizontal_spacing=SPACING * 3/ len(models_name) * 3,
         vertical_spacing=SPACING,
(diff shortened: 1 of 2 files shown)


Edoardo-Pedicillo (approved):

src/qibocal/protocols/characterization/classification.py
```diff
@@ -351,17 +351,17 @@ def _plot(
     fig = make_subplots(
         rows=1,
         cols=len(models_name),
-        horizontal_spacing=SPACING * 3 / len(models_name),
+        horizontal_spacing=SPACING / len(models_name) * 3,
```
Edoardo-Pedicillo, line 354: With this, the benchmarks and the results plots are not aligned.
```suggestion
        horizontal_spacing=SPACING * 3/ len(models_name) * 3,
```

andrea-pasquale: Thank you

PR #503, opened 2023-08-29
Fix flipping fit

### LINKED ISSUE #502: Flipping fitting function fails
Hi!
The fitting function of the `flipping` routine does not seem to work.
image

Below is the action parameters used with `qw5q_gold_qblox`
```yml
  - id: flipping
    priority: 0 #70
    operation: flipping
    # next: t1
    parameters:
      nflips_max: 30
      nflips_step: 1
```

andrea-pasquale:
Closes #502.

image

I did not test if the update method works as expected, mainly because I was not able to find any reference to `Flipping`
in the literature.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/characterization/flipping.py
+++ b/src/qibocal/protocols/characterization/flipping.py
@@ -9,6 +9,7 @@ from qibolab.platform import Platform
 from qibolab.pulses import PulseSequence
 from qibolab.qubits import QubitId
 from scipy.optimize import curve_fit
+from scipy.signal import find_peaks
 
 from qibocal.auto.operation import Data, Parameters, Qubits, Results, Routine
 from qibocal.config import log
@@ -136,10 +137,9 @@ def flipping_fit(x, p0, p1, p2, p3):
     # Offset                        : p[1]
     # Period of oscillation         : p[2]
     # phase for the first point corresponding to pi/2 rotation   : p[3]
-    return np.sin(x * 2 * np.pi / p2 + p3) * p0 + p1
+    return np.sin(x * 2 * np.pi * p2 + p3) * p1 + p0
 
 
-# FIXME: not working
 def _fit(data: FlippingData) -> FlippingResults:
     r"""Post-processing function for Flipping.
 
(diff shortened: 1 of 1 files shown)


aorgazf:
Hi @andrea-pasquale,
Thanks for taking care of this issue.
I've just tested the fix and the fitting function now follows the data. However, the purpose of the fit is to be able to extract parameters and those are still not correct.

aorgazf:
The extracted parameters worked a while back, perhaps before the migration to the new architecture of qibocal.
I do not have any paper explaining the procedure, and I have not been able to find anything online. But the procedure is fairly simple, so it is not complicated to extract the correct amplitude from the fitting function parameters.
I suspect the problem may be in having switched from absolute values to a multiplication factor.

aorgazf (changes requested):
The amplitudes saved in the runcard after executing the function are not correct.

andrea-pasquale:
I went over the calculations and I've made some changes to the current fitting.
I tested it by detuning on purpose the amplitude and both the value and the sign of the correction seems to be correct.
Here is an example
image
@aorgazf let me know if it works also for you

src/qibocal/protocols/characterization/flipping.py
```diff
@@ -9,12 +9,15 @@
...
 from scipy.optimize import curve_fit
+from scipy.signal import find_peaks
 
 from qibocal.auto.operation import Data, Parameters, Qubits, Results, Routine
 from qibocal.config import log
 
 from .utils import V_TO_UV
 
+ZERO_PHASE_THRESHOLD = 1e-6
+
```
Edoardo-Pedicillo, line 20: Could you explain the meaning of this constant in a comment ?

Edoardo-Pedicillo (changes requested):

andrea-pasquale:
Thanks @Edoardo-Pedicillo. It seems to be working as expected.

Edoardo-Pedicillo (approved):

Edoardo-Pedicillo:
@aorgazf this PR is blocking the new `Qibocal` release, please have a look as soon as possible. Thanks !

scarrazza:
Looks good, thanks.

PR #504, opened 2023-08-29
fix dumping in classifers

Edoardo-Pedicillo:
This PR is related to #406.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


andrea-pasquale (changes requested):
Thanks @Edoardo-Pedicillo.
If I run `qq report` or `qq auto` I get the following error:
```sh
UnboundLocalError: local variable 'fig_benchmarks' referenced before assignment
```
Could you please have a look?

src/qibocal/cli/builders.py
```diff
@@ -240,7 +240,7 @@ def _run(self, mode: ExecutionMode):
             timing_task["tot"] = timing_task["acquisition"] + result_time
             self.meta[task_id] = timing_task
             # TODO: reactivate iterative dump
-            # self.dump_report()
+            self.dump_report()
```
andrea-pasquale, line 243: I remember why I commented this out.
It fails when you run more than one action.
Until I solve it in #406 it is better to keep it commented.
```suggestion
            # self.dump_report()
```

andrea-pasquale (approved):
Thanks @Edoardo-Pedicillo, it works as expected now.

PR #517, opened 2023-09-11
Minor change in rabi duration

stavros11:
Changes the initial pulse duration to be the `pulse_duration_start` given in the runcard, because initializing with a hardcoded value of 4ns (<16ns) causes the QM driver to use baking mode for sweeping duration which is slower and I would prefer to avoid when not needed. It is more convenient to control the initial pulse from the action runcard.

I think this will not cause any issues with other drivers, since `pulse_duration_start` is really the shortest duration to be played in this routine.

Checklist:
- [ ] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`


Edoardo-Pedicillo (approved):

andrea-pasquale (approved):
Thanks @stavros11.
Having an hardcoded value here also didn't make too much sense.

andrea-pasquale:
Coverage is decreasing due to a problem with Ramsey that we need to solve... we can merge this PR.

PR #520, opened 2023-09-14
Add fit to Chevron plot

andrea-pasquale:
This PR implements a post-processing operation for the chevron protocol in order to find the best point for the CZ.

The post-processing works as follows:

1. Perform FFT for each amplitude value
2. Find amplitude at which the freq has the minimum value
3. Compute duration by fitting cosinusoidal function at the point of minimum freq

This is the result on hardware
image

I still need to implement the update method, after that it should be ready to merge.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


igres26 (approved):

igres26:
That is a good choice for fitting. More usage might expose where it fails, but for that we might want it merged.

src/qibocal/protocols/characterization/two_qubit_interaction/cz_virtualz.py
```diff
@@ -249,7 +249,7 @@ def _fit(
     for pair in pairs:
         pair_data = data[pair]
         qubits = next(iter(pair_data))[:2]
```
AleCandido, line 251: You might not need this variable any longer.

src/qibocal/protocols/characterization/two_qubit_interaction/cz_virtualz.py
```diff
@@ -280,7 +280,7 @@ def _fit(
                 fitted_parameters[target_q, control_q, "X"][2]
                 - fitted_parameters[target_q, control_q, "I"][2]
             )
-            virtual_phase[qubits][target_q] = fitted_parameters[
+            virtual_phase[pair][target_q] = -fitted_parameters[
```
AleCandido, line 283: Why did you change sign here?

src/qibocal/protocols/characterization/two_qubit_interaction/utils.py
```diff
@@ -1,8 +1,61 @@
+from statistics import median_high
+
+import numpy as np
 from qibolab.qubits import Qubit, QubitId
+from scipy.signal import find_peaks
+
+RANDOM_HIGH_VALUE = 1e6
+"""High value to avoid None when computing FFT."""
```
AleCandido, line 8: I like the variable name and description, definitely explanatory xD

(it's funny, but I truly like it, at least we know where it is coming from)

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
```diff
@@ -182,6 +190,49 @@ def _aquisition(
     return data
 
 
+def _fit(data: ChevronData) -> ChevronResults:
+    durations = {}
+    amplitudes = {}
+    pairs = data.pairs
+    for pair in pairs:
+        pair_data = data[pair]
+        qubits = next(iter(pair_data))[:2]
```
AleCandido, line 199: You may not need even this occurrence.

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
```diff
@@ -182,6 +190,49 @@ def _aquisition(
...
+            )
+
+            # guess amplitude computing FFT
+            amplitude, index, delta = fit_flux_amplitude(
+                msr_matrix, np.unique(amps), np.unique(times)
+            )
+
+            # estimate duration by rabi curve at amplitude previously estimated
+            def cos(x, omega, phase, amplitude, offset):
+                return amplitude * np.cos(x * omega + phase) + offset
```
AleCandido, line 218: This function is fully stand-alone, it doesn't capture anything from the outer scope. You could even define it top-level and briefly document.

src/qibocal/protocols/characterization/two_qubit_interaction/utils.py
```diff
@@ -1,8 +1,61 @@
...
+
+        samplingFrequency = 1 / time_step
+        values = np.arange(int(len(y) / 2))
+        timePeriod = len(y) / samplingFrequency
+        frequencies = values / timePeriod
+        f = frequencies[index] if index is not None else RANDOM_HIGH_VALUE
+        fs.append(2 * np.pi * f)
+
+    low_freq_interval = np.where(fs == np.min(fs))
+    amplitude = median_high(amps[low_freq_interval])
```
AleCandido, line 58: Is there a reason to use `median_high` instead of just the median?

src/qibocal/protocols/characterization/two_qubit_interaction/utils.py
```diff
@@ -1,8 +1,61 @@
...
     """Order a pair of qubits by drive frequency."""
     if qubits[pair[0]].drive_frequency > qubits[pair[1]].drive_frequency:
         return pair[1], pair[0]
     return pair[0], pair[1]
+
+
+def fit_flux_amplitude(matrix, amps, times):
+    """Estimate amplitude of Chevron plot by computing
+    FFT for each amplitude and chooses the amplitude
+    which lowers the frequency.
```
AleCandido, line 21: I know it's annoying, but the algorithm is definitely not trivial: it would be nice to have a complete description in English (rather than Python) of the whole algorithm.

Maybe we could discuss it before going through the process. Or you can use me as a guinea pig to test the description.

src/qibocal/protocols/characterization/two_qubit_interaction/utils.py
```diff
@@ -1,8 +1,61 @@
...
+        # Select the frequency with the highest peak
+        index = (
+            int(local_maxima[0][np.argmax(peak_heights)] + 1)
+            if len(local_maxima[0]) > 0
+            else None
+        )
+
+        samplingFrequency = 1 / time_step
+        values = np.arange(int(len(y) / 2))
+        timePeriod = len(y) / samplingFrequency
```
AleCandido, line 52: @stavros11 the JS disease is affecting more and more people!

(just fun, but use the Python naming convention, i.e. snake case for variables `samplingFrequency` -> `sampling_frequency`)

AleCandido (commented):
Of course, the critical part is the fit (as for the name of the PR), and we might need to iterate a bit.

All the other modifications are fine :)

src/qibocal/protocols/characterization/two_qubit_interaction/utils.py
andrea-pasquale: Yes, the reason is that I want an amplitude which is between the one that I have swept.
If I take the standard median and I sweep a even number of amplitudes I don't get a value in the sweep but the mean of two values. See for example [here](https://numpy.org/doc/stable/reference/generated/numpy.median.html).

The reason for this is that after I found a good guess for this amplitude value I perform a simple cos fit around the row corresponding to this amplitude to find the optimal duration of the CZ gate. Of course I already have this information from the FFT in theory but in order to be more accurate I prefer to do a one dimensional fit.

I could have chose also `median_low` it should not change that much. I didn't have a chance to test on hardware which of the two is better. I'm guessing there might not be a unique answer.

AleCandido: I see your point.

However, after that you still need to search for the index, and here you are also adopting the strategy to take the first one in case there are multiple ones (which could be very relevant for the result in the case of non-monotonic functions).

I believe you could achieve something similar, and in simple cases the exact same, just by taking the `np.median`, and then look for the first argument which is crossing that value (i.e. a sign change in `ar - np.median(ar)`, e.g. using https://stackoverflow.com/a/4843931/8653979).

The problem is that for monotonic functions they should be all equivalent, but if they are not monotonic the `median_high` might be actually rather unstable, because it sorts the values in a non-local way.
E.g. think about a bump function, for which you expect the median to be crossed twice. Whether `median_high` is located on one side or the other of the bump might heavily depend on the points you sample for the evaluation.

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
```diff
@@ -182,6 +190,49 @@ def _aquisition(
...
+def _fit(data: ChevronData) -> ChevronResults:
+    durations = {}
+    amplitudes = {}
+    pairs = data.pairs
+    for pair in pairs:
+        pair_data = data[pair]
+        qubits = next(iter(pair_data))[:2]
+        pair_amplitude = []
+        pair_duration = []
+        for target, control, setup in data[pair]:
```
stavros11, line 202: These may be naive questions because I have not followed the previous development of this routine, but out of curiosity: What are we looping over here? Isn't the `target` and `control` fixed for a specific `pair`? Also what is `setup`?

stavros11 (approved):
Thanks for implementing this. As a test I tried it on the chevron we got with QM on qw5q_gold last February (one of the first in TII) and it gave: amplitude=0.998, duration=28.7 which is acceptable. Here is the final curve fit in the amplitude slice (done quickly in a notebook):

curve_fit

We should address @AleCandido's comments related to code, such as the redundant `qubit` variables, before merging, but I agree with @igres26 that in order to evaluate how good the fitting is we need to try it in practice. So I would not block this PR for this, particularly because now we don't have any QPUs to check. We can always improve the fitting later, as we did with other routines.

src/qibocal/protocols/characterization/two_qubit_interaction/utils.py
stavros11: > @stavros11 the JS disease is affecting more and more people!

I think @andrea-pasquale copied this function from somewhere. Otherwise, I will stop going to the office. :mask:

andrea-pasquale: > I think @andrea-pasquale copied this function from somewhere. Otherwise, I will stop going to the office. 😷

Indeed this is the case. You got me. I was having a really hard time trying to figure out the frequency from the FFT so I decided to copy-and-paste this from somewhere...
I will adopt the our usual variable convention

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
andrea-pasquale: Here `setup` is for sure wrong. It makes more sense in the `tune_landscape` where setup could be either "I" or "X" depending on what you do on the control qubit.
In this case the data structure for the chevron is something like this:

https://github.com/qiboteam/qibocal/blob/a9539262fec3718bbfbd434c7aed363f79aae78f/src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py#L57-L69

Basically the idea would be to store the information such that I save in the array the information in this way: `(qubit_pair, qubit_measured)`. Another possible solution would be to modify the dtype and have something like this:

```py
ChevronType = np.dtype(
    [("amp", np.float64), ("length", np.float64), ("prob_high", np.float64), ("prob_low", np.float64)]
)
```
I don't remember why I chose the first option, actually it might be better to do something like this instead. Let me know if you agree @stavros11.

src/qibocal/protocols/characterization/two_qubit_interaction/cz_virtualz.py
andrea-pasquale: Here the correct value should be with the minus sign. For some reason I was adding this minus sign only in the plot function when the value is shown in the table. Now that I'm performing also the update I need to have this sign already correct in the `Results` object.

https://github.com/qiboteam/qibocal/blob/a9539262fec3718bbfbd434c7aed363f79aae78f/src/qibocal/protocols/characterization/two_qubit_interaction/cz_virtualz.py#L361-L367

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
```diff
@@ -176,47 +173,41 @@ def _aquisition(
...
-            times = data[pair][target, control, setup].length
-            msr_matrix = (
-                data[pair][target, control, setup]
-                .prob.reshape(len(np.unique(times)), len(np.unique(amps)))
-                .T
-            )
+
+        for qubit in pair:
+            amps = data[pair].amp
+            times = data[pair].length
```
AleCandido, line 201: If these only depend on `pair` they should be outside the `for` loop

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
```diff
@@ -235,32 +226,47 @@ def cos(x, omega, phase, amplitude, offset):
 
 def _plot(data: ChevronData, fit: ChevronResults, qubit):
     """Plot the experiment result for a single pair."""
-    colouraxis = ["coloraxis", "coloraxis2"]
+    colouraxis = ["coloraxis2", "coloraxis1"]
```
AleCandido, line 229: Great top level constant

src/qibocal/protocols/characterization/two_qubit_interaction/utils.py
andrea-pasquale: Personally, given that we cannot test on hardware I think it is better if we explore this solution in the future. Let's see how stable is the current fit. I'll open an issue.

src/qibocal/protocols/characterization/two_qubit_interaction/utils.py
andrea-pasquale: Feel free to have a look now.

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
andrea-pasquale: I implemented the second approach. Now it is way more clean and easier to understand.

andrea-pasquale:
@AleCandido I addressed all your comments but the improvements on the median for which I opened #535. 
If everything else looks good for you I will merge soon given that I already have 2 approvs

src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py
stavros11: Thanks for the update. I looked in more detail and now I understand what was happening. I agree that the current implementation is easier to understand.

src/qibocal/protocols/characterization/two_qubit_interaction/utils.py
AleCandido: Thanks, the moment the issue is available, feel free to close the conversation.

andrea-pasquale: #535

AleCandido (approved):
Ok, since we postponed the `fit` improvements, the rest is pretty much innocuous, and the implementation is fine.

Let's merge :)

andrea-pasquale:
> Ok, since we postponed the `fit` improvements, the rest is pretty much innocuous, and the implementation is fine.
> 
> Let's merge :)

I saw that you approved the comment above so I merged...
Thanks for the feedback.

Issue #498, opened 2023-08-29
Towards autocalibration


Edoardo-Pedicillo:
This issue summarizes all the next steps required to reach a minimal autocalibration prototype.
In brackets, I tried to estimate their completion times.
Please take this evaluation as purely indicative to design a timeline. 
I remember that #336 and @aorgazf 's issues report
other useful features that can improve the quality of the autocalibration project.

## Missing Routines

- [ ]  `Chevron` fitting (1 week). #490
- [ ]  Parameters update in `Tune Landscape` (1 week). #491
- [ ]  Amplitude Readout Optimization (2 weeks). #492
- [ ]  TWPA characterization routine (2 weeks). #478
- [ ]  Avoided Crossing (2 weeks). #493
- [ ]  SNZ (2 weeks). #494
- [ ]  Leakage Optimization (2 weeks). #495
- [ ]  Concluding Cryoscope routine (2 weeks). #460
- [ ] Error bars propagation (2 weeks). #450
- [ ] CHSH (1 week). #452
- [ ] Readout Mitigation Matrix (1 week). #452
- [ ] Flux Crosstalk matrix (1 week). #462

## Missing Cyclic Graph Features

- [ ] Single Node Optimizer (2 weeks). #496
- [ ] Exceptional Flow (1 month). #497

## General Maintenance and Improvements

- [ ] Operational modes (1 week). #406
- [ ] Documentation (2 weeks). #485

andrea-pasquale:
Closing this since at the moment the DAG is not supported anymore.

PR #521, opened 2023-09-14
Ro amplitude

Edoardo-Pedicillo:
Amplitude Readout Optimization #498 
Screenshot from 2023-09-14 12-52-36

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


src/qibocal/protocols/characterization/readout_optimization/resonator_amplitude.py
```diff
@@ -0,0 +1,211 @@
...
+    def unique_amplitudes(self, qubit: QubitId) -> np.ndarray:
+        return np.unique(self.data[qubit]["amp"])
+
+
+@dataclass
+class ResonatorAmplitudeResults(Results):
+    """Result class for `resonator_amplitude` protocol."""
+
+    lowest_errors: dict[QubitId, list]
+    best_amp: dict[QubitId, list]
```
andrea-pasquale, line 78: Missing docstrings

src/qibocal/protocols/characterization/readout_optimization/resonator_amplitude.py
```diff
@@ -0,0 +1,211 @@
...
+
+def _acquisition(
+    params: ResonatorAmplitudeParameters,
+    platform: Platform,
+    qubits: Qubits,
+) -> ResonatorAmplitudeData:
+    r"""
+    Data acquisition for resoantor amplitude optmization.
+    This protocol perform a classification protocol for amplitude optimization
+    with step amplitude_step.
```
andrea-pasquale, line 89: I will explain a little bit more what we are doing here since we are not doing only an acquisition

src/qibocal/protocols/characterization/readout_optimization/resonator_amplitude.py
```diff
@@ -0,0 +1,211 @@
...
+        qubits (dict): dict of target :class:`Qubit` objects to be characterized
+
+    Returns:
+        data (:class:`ResonatorAmplitudeData`)
+    """
+
+    data = ResonatorAmplitudeData()
+    for qubit in qubits:
+        error = 1
+        old_amp = platform.qubits[qubit].native_gates.MZ.amplitude
```
andrea-pasquale, line 103: Given that all the routine is looping over this for loop it might be worth to have a single qubit_acquisition_function. We could discuss it later today.

andrea-pasquale: This is True assuming that we want to get rid of multiplex. I still don't know what is the best solution.

src/qibocal/protocols/characterization/readout_optimization/resonator_amplitude.py
```diff
@@ -0,0 +1,211 @@
...
+            data.append_data(
+                qubit=qubit,
+                amp=new_amp,
+                state=states,
+                i=i_values,
+                q=q_values,
+                errors=error,
+            )
+            platform.qubits[qubit].native_gates.MZ.amplitude = old_amp
+            new_amp += params.amplitude_step
```
andrea-pasquale, line 155: Here it is nice to have this sort of iterative fitting. For the time being we can keep it. For the future it would be nice to have a way to move how the acquisition and the fitting interact with each other coded outside the acquisition or the fitting function.

andrea-pasquale (commented):
Thanks @Edoardo-Pedicillo, generally it looks good to me.
You can find some comments, mainly ideas.
I still need to test it on hardware

src/qibocal/protocols/characterization/readout_optimization/resonator_amplitude.py
```diff
@@ -0,0 +1,211 @@
...
+    return data
+
+
+def _fit(data: ResonatorAmplitudeData) -> ResonatorAmplitudeResults:
+    qubits = data.qubits
+    best_amps = {}
+    lowest_err = {}
+    for qubit in qubits:
+        data_qubit = data[qubit]
+        index_best_err = np.argmin(data_qubit["errors"])
```
Jacfomg, line 165: I suggest not to select the best amplitude based on the smaller error alone. You could use this to check if there are points with a bit more error but smaller ro amplitude and they should be more stable for readout.

Edoardo-Pedicillo: I understand your point, but I think this is a little bit difficult to quantify. I tried to solve this issue by adding an amplitude stop value. In this way, if you need a low amplitude level, you can set up a small interval to find the best value. Let me know if you have better ideas.

andrea-pasquale: I think for now this is fine.

andrea-pasquale:
@Edoardo-Pedicillo make sure to implement also the update properly.
See #526 on how to do it (it should quite trivial to do).

src/qibocal/protocols/characterization/readout_optimization/resonator_amplitude.py
Jacfomg: Checking all the other points and if they are 0.01 (or something) bigger but lower amplitude to switch to those instead of the global minimum. But, yeah let's keep like this for now.

andrea-pasquale: Actually there is already a threshold:
https://github.com/qiboteam/qibocal/blob/1cbf6fefa24264316aa3aa6ecbbc289514d990bb/src/qibocal/protocols/characterization/readout_optimization/resonator_amplitude.py#L105
Or are you suggesting something different?

Jacfomg: My bad, I probably didn't see it when I reviewed

Jacfomg (approved):

src/qibocal/protocols/characterization/readout_optimization/resonator_amplitude.py
```diff
@@ -0,0 +1,219 @@
...
+        ar["q"] = q
+        ar["state"] = state
+        ar["errors"] = errors
+        if qubit in self.data.keys():
+            self.data[qubit] = np.append(self.data[qubit], np.rec.array(ar))
+        else:
+            self.data[qubit] = np.rec.array(ar)
+
+    def unique_amplitudes(self, qubit: QubitId) -> np.ndarray:
+        return np.unique(self.data[qubit]["amp"])
```
andrea-pasquale, line 71: If this is not used I would removed it. I think this is why coverage is not passing.

andrea-pasquale (approved):

PR #526, opened 2023-09-20
Include update methods in all protocols

andrea-pasquale:
This PR aims at removing the [`platform.update`](https://github.com/qiboteam/qibolab/blob/8a4c670653058999ff7c6fb43935a0f9c93bdf7f/src/qibolab/platform.py#L85)method in qibolab by delegating to qibocal the update of the platform.
This is possible since qibocal has access to all the primitives in qibolab to peform this operation.

@AleCandido @stavros11 I think that this is more or less what we discussed yesterday.
I'm encapsulating each parameter update in a function in [update.py](https://github.com/qiboteam/qibocal/blob/refactor_update/src/qibocal/update.py) for two reasons:

1. some routines update the same parameters
2. the update sometimes is not trivial see for example the update of the readout frequency
https://github.com/qiboteam/qibocal/blob/36eeb5dff39142ccdff57e4333cf5fb1d36b68f3/src/qibocal/update.py#L8-L16
4. it is easier to write tests for each update_* function as I am doing[ here](https://github.com/qiboteam/qibocal/blob/refactor_update/tests/test_update.py)
 
For now I implemented the new proposal only for resonator spectroscopy.
If you agree with it I will propagate the change for all protocols in qibocal.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


src/qibocal/update.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def update_readout_frequency(results: dict[QubitId, float], platform: Platform):
+    """Update readout frequency value in platform for each qubit in results."""
+    for qubit, freq in results.items():
+        mz = platform.qubits[qubit].native_gates.MZ
+        freq_hz = int(freq * GHZ_TO_HZ)
+        mz.frequency = freq_hz
+        if mz.if_frequency is not None:
+            mz.if_frequency = freq_hz - platform.get_lo_readout_frequency(qubit)
```
stavros11, line 15: I would prefer
```suggestion
            mz.if_frequency = freq_hz - qubit.readout.lo_frequency
```
because the `platform.get_*`/`platform.set_*` will be deprecated at some point.

stavros11 (commented):
Thanks for implementing this @andrea-pasquale. It looks good to me, so I would say to go ahead with the rest of methods (after @AleCandido's review). 

The only thing that seems a bit annoying is that the for loop over qubits is repeated in all update methods. One way to solve this would be to lift it to `Task.update_platform` but I am not sure if this will create problems in the longer term, maybe for two-qubit gate routines, etc. where we may need full flexibility when defining `_update`.

src/qibocal/update.py
```diff
@@ -0,0 +1,28 @@
...
+from qibolab.qubits import QubitId
+
+from qibocal.protocols.characterization.utils import GHZ_TO_HZ
+
+
+def update_readout_frequency(results: dict[QubitId, float], platform: Platform):
+    """Update readout frequency value in platform for each qubit in results."""
+    for qubit, freq in results.items():
+        mz = platform.qubits[qubit].native_gates.MZ
+        freq_hz = int(freq * GHZ_TO_HZ)
```
AleCandido, line 12: I wonder if it would be simpler to standardize the units even before writing the results. I.e. in `_fit` rather than in `_update`, at least the platform and the history would contain the same information.

The only drawback is that we would not preserve the original information. But since this is already a post-processing of the acquired data, it should not be truly relevant.

AleCandido:
I agree with @stavros11, this is mostly fine, in that the overall structure is perfect (at least I'd not have any better proposal right now).
We might slightly improve the helper update functions (maybe), but I believe they should be in Qibocal. So, we'll always be in time to improve them.

Please, keep going. And thanks for this :)

src/qibocal/update.py
andrea-pasquale: Yes, I was planning to remove it. Thanks for catching this.

src/qibocal/update.py
andrea-pasquale: > I wonder if it would be simpler to standardize the units even before writing the results. I.e. in `_fit` rather than in `_update`, at least the platform and the history would contain the same information.

It is doable, however as a first step I will prefer to just propagate the changes for each protocol. We could standardize the units later on.

src/qibocal/auto/task.py
```diff
@@ -94,6 +94,9 @@ def update(self):
         """Local update parameter."""
         return self.action.update
 
+    def update_platform(self, results: Results, platform: Platform):
+        self.operation.update(results=results, platform=platform)
```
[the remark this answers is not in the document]
andrea-pasquale: I think that the problem with this is that `self.update` by itself is not enough to determine whether we need to perform the update or not.
https://github.com/qiboteam/qibocal/blob/56607c71d4e69ab2dccee50371b254aa18f7df18/src/qibocal/auto/execute.py#L144-L150
Here in the executor `self.update` is the global update flag and also the mode cannot be inferred from the task since it is a property of the `Executor`.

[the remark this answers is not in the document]
AleCandido: Here `self.update` is the `task.update` of the executor (because `self` is a `Task`).

The idea is to move it locally, such that it will be checked inside the `task.update_platform` call.

src/qibocal/update.py
AleCandido: Then move into a separate issue :)

andrea-pasquale:
In principle this is done, I think.
I will see if I can find a way to repeat those for loops every time.
If I don't manage to do it, I will open an issue.

src/qibocal/update.py
```diff
@@ -0,0 +1,175 @@
...
+        platform.qubits[qubit].t2_spin_echo = int(t2_spin_echo)
+
+
+def drag_pulse_beta(results: dict[QubitId, float], platform: Platform):
+    """Update beta parameter e value in platform for each qubit in results."""
+    for qubit, beta in results.items():
+        shape = platform.qubits[qubit].native_gates.RX.shape
+        rel_sigma = re.findall(r"[\d]+[.\d]+|[\d]*[.][\d]+|[\d]+", shape)[0]
+        drag_pulse = pulses.Drag(rel_sigma=rel_sigma, beta=beta)
+        platform.qubits[qubit].native_gates.RX.shape = repr(drag_pulse)
```
andrea-pasquale, line 167: @stavros11 I had to these weird shenanigans since apparently `native_gates.RX.shape` will give me the serial of the pulse shape and not the pulse shape object by itself.
Is there any reason to return the str and not the object itself?
I'm guessing that this is mostly related to the serialization of the native pulse

AleCandido (approved):
Thanks @andrea-pasquale for this. To me, it's perfect (and it has been incredibly fast).

The problem of the loops (repeated operations for qubits and pairs) in a sense is a long-standing one, because it is related to the multi-qubit calibration, for which we never had a very clear plan.
The message is: if there is not already one, open a dedicated issue, and we will solve anyhow in a dedicated PR.

The only tiny remark is that you're mixing this relevant and general update, with something specific to the Chevron, that maybe would have been worth to examine in a separate PR as well.

andrea-pasquale:
> The problem of the loops (repeated operations for qubits and pairs) in a sense is a long-standing one, because it is related to the multi-qubit calibration, for which we never had a very clear plan.
> The message is: if there is not already one, open a dedicated issue, and we will solve anyhow in a dedicated PR.

Actually I might be close to find a solution for this. 

> The only tiny remark is that you're mixing this relevant and general update with something specific to the Chevron, that maybe would have been worth to examine in a separate PR as well.

Yeah, I know. I just did it since https://github.com/qiboteam/qibocal/pull/520 is ready to merge I just need a second approval and I already fixed a few things there. I didn't want to open a second PR just for the chevron.

Finally, this PR does not tackle the update of the twpa parameters since the interface is currently being developed in https://github.com/qiboteam/qibolab/pull/585

andrea-pasquale:
> Actually I might be close to find a solution for this.

For loop implemented :)

AleCandido:
> For loop implemented :)

Saw it, that's definitely a possible solution.

I wonder if it's worth to have the whole routines working just for individual qubit/pair.
I believe the case of the `fit` will be similar to the `update`, but maybe it is convenient to make the acquisition in parallel (or maybe not, in which case we could really write the routines per qubit/pair, and keep the multi-thing at the level of the executor - in a separate PR, of course).

AleCandido (approved):

AleCandido:
@andrea-pasquale since the Chevron might require slightly more time, would you consider removing it from here, such that we can immediately merge this in `main`?
If you wish, operationally:
- make another branch `update_chevron` that duplicates this
- edit this to remove the Chevron related part
- merge this into `main`
- merge `update_chevron` into `chevron_fit`
- merge `main` into `update_chevron` (in principle, this step should do nothing)

Or is there any other limitation for which we need the Chevron to merge the update?

src/qibocal/update.py
```diff
@@ -0,0 +1,151 @@
...
+
+def t2_spin_echo(t2_spin_echo: float, platform: Platform, qubit: QubitId):
+    """Update mean excited state value in platform for specific qubit."""
+    platform.qubits[qubit].t2_spin_echo = int(t2_spin_echo)
+
+
+def drag_pulse_beta(beta: float, platform: Platform, qubit: QubitId):
+    """Update beta parameter e value in platform for specific qubit."""
+    shape = platform.qubits[qubit].native_gates.RX.shape
+    rel_sigma = re.findall(r"[\d]+[.\d]+|[\d]*[.][\d]+|[\d]+", shape)[0]
```
stavros11, line 143: I have not tested but this can probably be simplified using
```suggestion
    pulse = platform.qubits[qubit].native_gates.RX.pulse(start=0)
    rel_sigma = pulse.shape.rel_sigma
```
This will basically delegate the shenanigans to [qibolab's pulse constructor](https://github.com/qiboteam/qibolab/blob/352a4691a503b394cd12f57e455e6cddaed7aa10/src/qibolab/pulses.py#L951), but better to have them in a single place instead of two.

stavros11 (approved):
> Or is there any other limitation for which we need the Chevron to merge the update?

I think they are independent, so it should be possible to merge this before the Chevron fit. However, given that the Chevron fit is an enhancement and does not break any existing functionality, I am not sure if it is worth going through the trouble reversing the order now. The fit most likely won't be perfect but if it is on main it will be easier to test and improve.

The current PR also looks good to be merged.

> I wonder if it's worth to have the whole routines working just for individual qubit/pair. I believe the case of the `fit` will be similar to the `update`, but maybe it is convenient to make the acquisition in parallel (or maybe not, in which case we could really write the routines per qubit/pair, and keep the multi-thing at the level of the executor - in a separate PR, of course).

For `fit` it may be possible, however for the acquisition it may be complex. With the current qibolab, if you want to take advantage of multiplex, you need to create a `PulseSequence` that targets multiple qubits and send it in a single `platform.execute_pulse_sequence` call. Lifting that to the executor will reduce our flexibility when writing routines, for example what if you want to execute two different sequences. There may be a better design that allows this, that I don't see.

src/qibocal/update.py
andrea-pasquale: Thanks @stavros11, I'll give it a try.

andrea-pasquale:
> > Or is there any other limitation for which we need the Chevron to merge the update?
> 
> I think they are independent, so it should be possible to merge this before the Chevron fit. However, given that the Chevron fit is an enhancement and does not break any existing functionality, I am not sure if it is worth going through the trouble reversing the order now. The fit most likely won't be perfect but if it is on main it will be easier to test and improve.

In theory they are independent, I decided to make this PR point to `chevron_fit` only because I could already implement the update for the chevron given that I the fitting is working in that branch. 

> > I wonder if it's worth to have the whole routines working just for individual qubit/pair. I believe the case of the `fit` will be similar to the `update`, but maybe it is convenient to make the acquisition in parallel (or maybe not, in which case we could really write the routines per qubit/pair, and keep the multi-thing at the level of the executor - in a separate PR, of course).
> 
> For `fit` it may be possible, however for the acquisition it may be complex. With the current qibolab, if you want to take advantage of multiplex, you need to create a `PulseSequence` that targets multiple qubits and send it in a single `platform.execute_pulse_sequence` call. Lifting that to the executor will reduce our flexibility when writing routines, for example what if you want to execute two different sequences, on multiple qubits each. There may be a better design that allows this though.

I agree with pretty much everything that @stavros11 said.
The main problem here is multiplexing. A possible way to solve it would be to have another function before `acquisition` something like `setup` which should prepare the pulse sequence so that in the `acquisition` we merge and execute multiple pulse sequences...
I would say that this is out of scope of this PR, but we should definitely think about it.

AleCandido:
> I would say that this is out of scope of this PR, but we should definitely think about it.

I agree it is out of scope for the PR, but it is not even urgent right now in general. We can rediscuss when we'll have lifted the main bottlenecks for autocalibration (and this is not one of them).

AleCandido:
I'd like this to be merged asap, in order to:
- remove `update` from Qibolab
- avoid merging another `update` extension in https://github.com/qiboteam/qibolab/pull/581
- avoid breaking it right after merging

So, if you want to merge #520 before, that's even fine. But if possible, let's do it soon.

Otherwise, the proposal above is because I do not feel much pressure about merging #520, but I'd like to merge this soon.
If you wish, I could follow the procedure above myself, and you can provide feedback.

andrea-pasquale:
> I'd like this to be merged asap, in order to:
> 
> * remove `update` from Qibolab
> * avoid merging another `update` extension in [David/qubit tracking qibolab#581](https://github.com/qiboteam/qibolab/pull/581)
> * avoid breaking it right after merging
> 
> So, if you want to merge #520 before, that's even fine. But if possible, let's do it soon.
> 
> Otherwise, the proposal above is because I do not feel much pressure about merging #520, but I'd like to merge this soon. If you wish, I could follow the procedure above myself, and you can provide feedback.

I'm about to push a few fixes here but in principle both this PR and #520 are ready to merge. I prefer to merge #520 first I don't think that it is necessary to merge this one and afterwards #520.

andrea-pasquale:
As soon as CI passes I will merge also this one.

AleCandido:
> As soon as CI passes I will merge also this one.

Thank you very much :D

PR #532, opened 2023-09-22
ramsey error with probabilities

Edoardo-Pedicillo:
image

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


AleCandido:
Thanks a lot @Edoardo-Pedicillo, the result is wonderful!

I will properly review, but if you can fix the tests, I believe we can merge this very soon.

Just a tiny improvement: the variance of the chi2 distribution also depends only on the number of degrees of freedom, in particular it is $2n$ https://en.wikipedia.org/wiki/Chi-squared_distribution.
Thus, the variance of the reduced chi2 is $2/n$, and so its standard deviation is $\sqrt{2/n}$. If you could compute this number and put aside the chi2 you computed in the table it would be nice as a comparison.

In this case, e.g., the standard deviation should be $\sim 0.22$, or something like. So the chi2 might be a bit large, but really just a bit (it is within $2\sigma$), and I'd consider it perfect for our purposes :)

AleCandido:
You got it!

If you're ready now, I will review asap. Just ask for it :)

Edoardo-Pedicillo:
image

@aorgazf I attempted to implement your suggestions from the Friday meeting. There are three possible ways to "disable" the error bars in the report page:
- have two traces, one with error and the second one without them,
- add a flag in the routine parameters to disable the error,
- use a lighter color for the error bars to increase the readability,

Personally, I don't like the second one because the errors will be evaluated by default (I would like to avoid too many branches). The last one is not a real solution, but it is the one I prefer.
If you have other suggestions, we can discuss them in this PR before propagating the changes in all the other routines.

@AleCandido @andrea-pasquale  Because of the warm-up, I can test only the fit and plot methods (thanks to @andrea-pasquale ), we can start the review now and then we will merge after testing on chips.

andrea-pasquale:
Thanks @Edoardo-Pedicillo. I will have a look at the PR.

> * have two traces, one with errors and the second one without them,
> * add a flag in the routine parameters to disable the errors,
> * use a lighter color for the error bars to increase the readability,

At this point a 4th option could be to use [errors bands](https://plotly.com/python/continuous-error-bars/).

In this way we could keep them while maintaining readability.

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -156,17 +150,30 @@ def _acquisition(
...
         )
-        for qubit in qubits:
+        for i, qubit in enumerate(qubits):
             result = results[ro_pulses[qubit].serial]
+            probs = []
+            errors = []
+            for i, wait in enumerate(waits):
+                prob, error = eval_prob(
+                    result.voltage_i[i], result.voltage_q[i], platform, qubit
+                )
```
andrea-pasquale, line 164: I believe that if you put `AcquisitionType.DISCRIMINATION` you can already get probabilities since the return type should be [`SampleResults`](https://github.com/qiboteam/qibolab/blob/18cf3221cf3800ab7c36626d1daa13dcd1dee324/src/qibolab/result.py#L104). In this way you can avoid to code `eval_prob`.
If for a particolar driver this approach does not work we should fix it in qibolab.

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -156,17 +150,30 @@ def _acquisition(
...
             sweeper,
         )
-        for qubit in qubits:
+        for i, qubit in enumerate(qubits):
             result = results[ro_pulses[qubit].serial]
+            probs = []
+            errors = []
+            for i, wait in enumerate(waits):
+                prob, error = eval_prob(
+                    result.voltage_i[i], result.voltage_q[i], platform, qubit
```
andrea-pasquale, line 163: Be careful here since you are shadowing the variable `i` if it is not needed you should just iterate over `qubits` and not `enumerate(qubits)`

andrea-pasquale (commented):
Thanks @Edoardo-Pedicillo.
I did a quick review. You can find some comments belows.
When I tried to test it on multiple qubits it seems that the error bars appear only on certain qubits. Probably there is a bug somewhere.

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -70,24 +69,20 @@ class RamseyData(Data):
...
-        ar["msr"] = np.array([msr])
-        ar["phase"] = np.array([phase])
-        if qubit in self.data:
-            self.data[qubit] = np.rec.array(np.concatenate((self.data[qubit], ar)))
-        else:
-            self.data[qubit] = np.rec.array(ar)
+        ar = np.empty(prob.shape, dtype=RamseyType)
+        ar["wait"] = np.array([wait])
+        ar["prob"] = np.array([prob])
+        ar["errors"] = np.array([errors])
```
AleCandido, line 83: Why do you need the extra dimension? (i.e. why not `np.array([errors])`)

Most likely it is just me being slow, but it's easier to ask :)

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -185,21 +192,24 @@ def _acquisition(
...
-                data.register_qubit(
-                    qubit,
-                    wait=wait,
-                    msr=result.magnitude,
-                    phase=result.phase,
+                prob, error = eval_prob(
+                    result.voltage_i, result.voltage_q, platform, qubit
                 )
+                errors[i].append(error)
+                probs[i].append(prob)
```
AleCandido, line 204:
```py
            for qubit, error_, prob_ in zip(qubits, errors, probs):
                data.register_qubit(
                prob, error = eval_prob(
                    qubit,
                    result.voltage_i, result.voltage_q, platform, qubit
                    wait=wait,
                    msr=result.magnitude,
                    phase=result.phase,
                )
                )
                error_.append(error)
                prob_.append(prob)
```

If you can, avoid indexing in iteration, and prefer to iterate simultaneously.

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -185,21 +192,24 @@ def _acquisition(
...
+
+        for i, qubit in enumerate(qubits):
+            data.register_qubit(
+                qubit,
+                wait=np.array(waits),
+                prob=np.array(probs[i]),
+                errors=np.array(errors[i]),
+            )
```
AleCandido, line 212: Same of the above, just use `zip()`.

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -230,71 +240,34 @@ def _fit(data: RamseyData) -> RamseyResults:
...
-            t2_measure[qubit] = (t2, None)
-            delta_phys_measure[qubit] = (delta_phys, None)
-            popts[qubit] = popt
-
+        probs = np.concatenate(qubit_data[["prob"]].tolist())
+        try:
+            popt, perr = fitting(waits, probs, qubit_data.errors)
+        except:
+            popt = [0, 0, 0, 0, 1]
+            perr = [1] * 5
```
AleCandido, line 248: These values should be top-level constants and documented.

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -307,24 +280,19 @@ def _plot(data: RamseyData, qubit, fit: RamseyResults = None):
...
-    msrs = qubit_data[["msr"]].tolist()
-    if data.nboot != 0:
-        msrs = np.reshape(msrs, (len(waits), -1)) * V_TO_UV
-        nsamples = msrs.shape[1]
-        error_bars = np.std(msrs, axis=1).flatten() / np.sqrt(nsamples)
-        msrs = np.mean(msrs, axis=1).flatten()
-    else:
-        msrs = np.reshape(msrs, (len(waits))) * V_TO_UV
-        error_bars = np.zeros(len(msrs))
+    probs = qubit_data[["prob"]].tolist()
```
AleCandido, line 283: You don't need to make it a list
```suggestion
    probs = qubit_data[["prob"]]
```

You're processing with NumPy even after.

src/qibocal/protocols/characterization/ramsey_sequences.py
```diff
@@ -69,28 +70,31 @@ def _acquisition(
...
                 nshots=params.nshots,
                 relaxation_time=params.relaxation_time,
                 acquisition_type=AcquisitionType.INTEGRATION,
-                averaging_mode=averaging_mode,
+                averaging_mode=AveragingMode.SINGLESHOT,
             ),
         )
-        for qubit in qubits:
+
+        for i, qubit in enumerate(qubits):
```
AleCandido, line 84: `zip()`

src/qibocal/protocols/characterization/ramsey_sequences.py
```diff
@@ -69,28 +70,31 @@ def _acquisition(
...
-                qubit,
-                wait=wait,
-                msr=np.array([result.magnitude]),
-                phase=np.array([result.phase]),
-            )
+            prob, error = eval_prob(result.voltage_i, result.voltage_q, platform, qubit)
+            errors[i].append(error)
+            probs[i].append(prob)
+
+    for i, qubit in enumerate(qubits):
```
AleCandido, line 90: `zip()`

src/qibocal/protocols/characterization/utils.py
```diff
@@ -315,10 +317,18 @@ def chi2_reduced(
     errors: npt.NDArray,
     dof: float = None,
 ):
+    if np.count_nonzero(errors) == 0:
+        return 100
```
AleCandido, line 321: Please, document this function, and including this value.

src/qibocal/protocols/characterization/utils.py
```diff
@@ -315,10 +317,18 @@ def chi2_reduced(
...
+
     if dof is None:
         dof = len(observed) - 1
 
-    return np.sum(np.square((observed - estimated) / errors)) / dof
+    chi2 = np.sum(np.square((observed - estimated) / errors)) / dof
+
+    if chi2 == np.inf:
+        return 100_000
```
AleCandido, line 329: And even this number, please :)

src/qibocal/protocols/characterization/ramsey_sequences.py
```diff
@@ -69,28 +70,31 @@ def _acquisition(
...
+        data.register_qubit(
+            qubit,
+            wait=np.array(waits),
+            prob=np.array(probs[i]),
+            errors=np.array(errors[i]),
+        )
```
AleCandido, line 96: This seems very similar (maybe identical, didn't check) to what is done in `ramsey.py`.

Could you factor out in a dedicated function?

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -70,24 +69,20 @@ class RamseyData(Data):
...
-        if qubit in self.data:
-            self.data[qubit] = np.rec.array(np.concatenate((self.data[qubit], ar)))
-        else:
-            self.data[qubit] = np.rec.array(ar)
+        ar = np.empty(prob.shape, dtype=RamseyType)
+        ar["wait"] = np.array([wait])
+        ar["prob"] = np.array([prob])
+        ar["errors"] = np.array([errors])
+
+        self.data[qubit] = np.rec.array(ar)
```
AleCandido, line 85: This pattern is now present in many places in the code (due to the usage of rec arrays).

It could be worth to abstract it in a small common functions.
(such that if we then find a more optimal/convenient way of doing it, we'll change in a single place, as usual)

src/qibocal/protocols/characterization/ramsey.py
Edoardo-Pedicillo: Without a `tolist()`,  `probs` will be a recarray and the plot is not rendered as expected.

src/qibocal/protocols/characterization/ramsey_sequences.py
Edoardo-Pedicillo: It is the same, I would like to avoid a new function initialization, because this routine should be removed as soon as all the driver can support sweepers and run directly `ramsey.py`

src/qibocal/protocols/characterization/ramsey.py
Edoardo-Pedicillo: No sorry, it was my mistake, it is not necessary.

src/qibocal/protocols/characterization/ramsey.py
AleCandido: Ok, this is because you're slicing with a list
```suggestion
    probs = qubit_data["prob"]
```

AleCandido: Moreover, avoiding the list you probably can skip also the `np.reshape` right after (but about this I'm not sure, it would only if the current shape were `(len(waits), 1)` and you're just reshaping to get rid of the extra dimension).

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -14,10 +14,14 @@
 
 from qibocal import update
 from qibocal.auto.operation import Data, Parameters, Qubits, Results, Routine
-from qibocal.bootstrap import bootstrap
-from qibocal.config import log
 
-from .utils import GHZ_TO_HZ, V_TO_UV, chi2_reduced, fill_table
+from .utils import GHZ_TO_HZ, chi2_reduced, fill_table
+
+POPT_EXCEPTON = [0, 0, 0, 0, 0]
```
andrea-pasquale, line 20:
```suggestion
POPT_EXCEPTION = [0, 0, 0, 0, 0]
```

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -230,71 +232,34 @@ def _fit(data: RamseyData) -> RamseyResults:
...
-            freq_measure[qubit] = (corrected_qubit_frequency, None)
-            t2_measure[qubit] = (t2, None)
-            delta_phys_measure[qubit] = (delta_phys, None)
-            popts[qubit] = popt
-
+        probs = np.concatenate(qubit_data[["prob"]].tolist())
+        try:
+            popt, perr = fitting(waits, probs, qubit_data.errors)
+        except:
+            popt = POPT_EXCEPTON
```
andrea-pasquale, line 239:
```suggestion
            popt = POPT_EXCEPTION
```

src/qibocal/protocols/characterization/ramsey.py
Edoardo-Pedicillo: We could add `register_qubit` in the `Data` method, maybe something like the following function should be general enough.

``` python   
def register_qubit(self, qubit, dtype, *args):
        """Store output for single qubit."""
        # to be able to handle the non-sweeper case
        ar = np.empty(np.shape(args[0]), dtype=dtype)
        for i, name in enumerate(dtype.names):
            ar[name] = args[i]
        if qubit in self.data:
            self.data[qubit] = np.rec.array(np.concatenate((self.data[qubit], ar)))
        else:
            self.data[qubit] = np.rec.array(ar)
```

Eventually, we could also have a `dtype` attribute in `Data`.

andrea-pasquale: I like this a lot. If it works for the majority of the protocols we could give it a try.

Edoardo-Pedicillo:
> At this point a 4th option could be to use [errors bands](https://plotly.com/python/continuous-error-bars/).
In this way we could keep them while maintaining readability.

Ok, `error bands` seem the best option. I have used the color in the example because it fits well, feel free to suggest others.

andrea-pasquale:
@Edoardo-Pedicillo did you check also the small errors for the frequencies? Probably there is some factor 10^9 which is wrong.

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -155,15 +156,19 @@ def _acquisition(
...
         for qubit in qubits:
-            result = results[ro_pulses[qubit].serial]
+            probs = results[qubit].probability()
+            errors = [np.sqrt(prob * (1 - prob) / params.nshots) for prob in probs]
             data.register_qubit(
-                qubit, wait=waits, msr=result.magnitude, phase=result.phase
+                qubit,
+                wait=np.array(waits),
+                prob=np.array(probs),
+                errors=np.array(errors),
```
andrea-pasquale, line 171:
```suggestion
                wait=waits,
                prob=probs,
                errors=errors,
```
It seems to work also without an explicit cast to `np.array`.

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -70,20 +74,18 @@ class RamseyData(Data):
...
     """Qubit freqs for each qubit."""
     data: dict[QubitId, npt.NDArray] = field(default_factory=dict)
     """Raw data acquired."""
 
-    def register_qubit(self, qubit, wait, msr, phase):
+    def register_qubit(self, qubit, wait, prob, errors):
         """Store output for single qubit."""
         # to be able to handle the non-sweeper case
-        ar = np.empty(msr.shape, dtype=RamseyType)
+        ar = np.empty(np.shape(prob), dtype=RamseyType)
```
andrea-pasquale, line 85:
```suggestion
        shape = (1,) if np.isscalar(prob) else prob.shape
        ar = np.empty(shape, dtype=RamseyType)
```
If you do this than you can remove also the cast to `np.array` in the acquisition (see comments below).

src/qibocal/protocols/characterization/ramsey.py
andrea-pasquale, line 171:
```suggestion
                wait=waits,
                prob=probs,
                errors=errors,
```
If you implement my correction in `register_qubit` it will work.

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -184,23 +189,20 @@ def _acquisition(
...
+                prob = results[qubit].probability()
+                error = np.sqrt(prob * (1 - prob) / params.nshots)
                 data.register_qubit(
                     qubit,
-                    wait=wait,
-                    msr=result.magnitude,
-                    phase=result.phase,
+                    wait=np.array([wait]),
+                    prob=np.array([prob]),
+                    errors=np.array([error]),
```
andrea-pasquale, line 204:
```suggestion
                    wait=wait,
                    prob=prob,
                    errors=error,
```
Same as above

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -230,122 +232,78 @@ def _fit(data: RamseyData) -> RamseyResults:
...
-            name="Voltage",
-            showlegend=True,
-            legendgroup="Voltage",
-        )
+            go.Scatter(
+                x=np.concatenate((waits, waits[::-1])),
+                y=np.concatenate((probs + error_bars, (probs - error_bars)[::-1])),
+                fill="toself",
+                fillcolor="rgba(0,100,80,0.2)",
+                line=dict(color="rgba(255,255,255,0)"),
```
andrea-pasquale, line 294: This should be top level constants, right?
Or eventually you can move them to a utils file if we are planning to use the same for error bands of all protocols

src/qibocal/protocols/characterization/ramsey_sequences.py
```diff
@@ -69,27 +68,25 @@ def _acquisition(
...
+            prob = results[qubit].probability()
+            error = np.sqrt(prob * (1 - prob) / params.nshots)
             data.register_qubit(
                 qubit,
-                wait=wait,
-                msr=np.array([result.magnitude]),
-                phase=np.array([result.phase]),
+                wait=np.array([wait]),
+                prob=np.array([prob]),
+                errors=np.array([error]),
```
andrea-pasquale, line 89:
```suggestion
                wait=wait,
                prob=prob,
                errors=error,
```

src/qibocal/protocols/characterization/utils.py
```diff
@@ -269,6 +271,8 @@ def fit_punchout(data: Data, fit_type: str):
 
 def eval_magnitude(value):
     """number of non decimal digits in `value`"""
+    if value == 0:
+        return 0
```
andrea-pasquale, line 275: Here it would be better to consider also the case where this could be `np.infinity`.
I was having some issues with CI about this.
See for example here https://github.com/qiboteam/qibocal/actions/runs/6348675855/job/17246250863

tests/test_utils.py
```diff
@@ -18,3 +22,7 @@ def test_fill_table():
     output_no_error = f"{qubit}| {name}: 1.0* 10^1<br>"
     assert fill_table(qubit, name, value, error, unit) == output
     assert fill_table(qubit, name, value, None) == output_no_error
+
+
+def test_eval_magnitude():
+    assert 0 == eval_magnitude(0)
```
andrea-pasquale, line 28: I would also add a test with a generic value instead of testing only with 0

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -155,15 +156,19 @@ def _acquisition(
...
-                averaging_mode=AveragingMode.CYCLIC,
+                acquisition_type=AcquisitionType.DISCRIMINATION,
+                averaging_mode=AveragingMode.SINGLESHOT,
             ),
             sweeper,
         )
         for qubit in qubits:
-            result = results[ro_pulses[qubit].serial]
+            probs = results[qubit].probability()
+            errors = [np.sqrt(prob * (1 - prob) / params.nshots) for prob in probs]
```
andrea-pasquale, line 166: This errror is derived from the binomial distribution, right?
I would add at least a comment to document it.

andrea-pasquale (commented):
Thanks @Edoardo-Pedicillo for all the updates.
I only need to check the error propagation than I will approve

src/qibocal/protocols/characterization/utils.py
AleCandido: [`np.isfinite`](https://numpy.org/doc/stable/reference/generated/numpy.isfinite.html#numpy-isfinite)

src/qibocal/protocols/characterization/ramsey.py
Edoardo-Pedicillo: I have opened #541

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -230,122 +236,78 @@ def _fit(data: RamseyData) -> RamseyResults:
...
-            t2_measure[qubit] = (t2, None)
-            delta_phys_measure[qubit] = (delta_phys, None)
-            popts[qubit] = popt
-
+        probs = qubit_data["prob"]
+        try:
+            popt, perr = fitting(waits, probs, qubit_data.errors)
+        except:
+            popt = POPT_EXCEPTION
+            perr = PERR_EXCEPTION
```
AleCandido, line 244: In principle, if the fit actually fails, it would be meaningful to "return" an error.

We do not want necessarily to raise an exception (though it could be an option), but we want to give back the control to the executor, that should gracefully stop the execution.

However, this will become easier with the exceptional flow (not because the exc-flow is required, but because we have to design with @andrea-pasquale a strategy for handling the outcome anyhow).

Edoardo-Pedicillo: If you want, I can remove the try-except lines, at the beginning they were used to generate the report in any case, now we can do it with `qq report`.

andrea-pasquale: I suggest to postpone this until we discuss about the exceptional flow.
It is not worth to do something temporary just for a single routine.

src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -439,12 +402,23 @@ def fitting(x: list, y: list) -> list:
...
+        delta_y * popt[1] * np.exp(x_min * popt[4] / delta_x),
+        popt[2] / delta_x,
+        popt[3] - x_min * popt[2] / delta_x,
+        popt[4] / delta_x,
+    ]
+    perr = np.sqrt(np.diag(perr))
+    perr = [
+        delta_y * perr[0],
+        delta_y
+        * np.exp(-x_min * popt[4] / delta_x)
```
andrea-pasquale, line 418:
```suggestion
        * np.exp(x_min * popt[4] / delta_x)
```
If I did the error propagation correctly this might be a + sign

andrea-pasquale (approved):
Thanks @Edoardo-Pedicillo, LGTM.
Just a tiny error on the error propagation (if I did not do a mistake myself).

AleCandido (approved):

Issue #533, opened 2023-09-23
Allow routines to access the folder where data will be saved


aorgazf:
Hi,
While developing a routine to perform the T1_T2_vs_Temperature experiment, would have liked to be able to save data to the same folder where the report is saved. The routine performs measurements for long periods and extracts quite a lot of data. To minimise the risk of loosing data, the routine saves the data corresponding to each measurement (t1, ramsey_t2, qubit spectroscopy) after they are taken.
The problem is that the folder where the report is computed is currently calculated only after the routine is finished, so there is no way to access it. If possible, it would be nice to have it calculated beforehand and have it accessible to routines.

Thanks.

PR #534, opened 2023-09-25
T1 and T2 vs Temperature Experiment

aorgazf:
A new routine to carry out an experiment to measure T1 and T2 as a function of QPU Temperature.
It controls BF temperature and therefore it requires a driver for it in qibolab.
https://arxiv.org/abs/1105.4652
https://arxiv.org/abs/1105.4652
Checklist:
- [ ] Tests are passing.
    - [x] Qibo: `master`
    - [ ] Qibolab: `main`


src/qibocal/protocols/characterization/t1_t2_vs_temperature.py
```diff
@@ -0,0 +1,571 @@
...
+        results = platform.execute_pulse_sequence(
+            sequence,
+            ExecutionParameters(
+                nshots=nshots,
+                relaxation_time=relaxation_time,
+                acquisition_type=AcquisitionType.DISCRIMINATION,
+            ),
+        )
+        h_bar = 6.62607015e-34
+        Kb = 1.380649e-23
```
andrea-pasquale, line 225: This should be top level constants

src/qibocal/protocols/characterization/t1_t2_vs_temperature.py
```diff
@@ -0,0 +1,571 @@
...
+    # effective_temperature: dict[QubitId, list[float]]
+    # """Effective Temperature"""
+    # frequency: dict[QubitId, list[float]]
+    # """Effective Temperature"""
+    # T1: dict[QubitId, list[int]]
+    # """Effective Temperature"""
+    # T2: dict[QubitId, list[int]]
+    # """Effective Temperature"""
+    # timestamp: list[float]
+    # """Timestamp"""
```
andrea-pasquale, line 158: To be completed I guess?

src/qibocal/protocols/characterization/t1_t2_vs_temperature.py
```diff
@@ -0,0 +1,571 @@
...
+                max_heater_power: 3000    # uW
+                min_heater_power: 0       # uW
+                stabilization_time: 3600  # s
+                num_datapoints: 24
+                nshots: 50_000
+                relaxation_time: 50_000   # ns
+
+    """
+
+    def _run_effective_temperature(
```
andrea-pasquale, line 202: I think that the _run_* functions could also be defined outside this scope.
They should also be documented.

src/qibocal/protocols/characterization/t1_t2_vs_temperature.py
```diff
@@ -0,0 +1,571 @@
...
+        )
+
+        operation: Routine = qubit_spectroscopy
+        qubit_spectroscopy_params: QubitSpectroscopyParameters = (
+            QubitSpectroscopyParameters(
+                freq_width=10_000_000,
+                freq_step=100_000,
+                drive_duration=5_000,
+                drive_amplitude=0.05,
+                nshots=2000,  # params.nshots,
```
andrea-pasquale, line 256: These parameters are hardcoded. The same goes for all other `_run_*` functions.
I think that they should either defined as top level constants or they should be listed among the parameters of this protocol.
Moreover, you are changing the number of shots compared to the shots that the user is passing as argument.

src/qibocal/protocols/characterization/t1_t2_vs_temperature.py
```diff
@@ -0,0 +1,571 @@
...
+            autoshift=True,
+            side="right",
+        ),
+    )
+    figures.append(fig)
+
+    fitting_report = ""
+    # fitting_report += (
+    # f"{qubit} | Effective Temperature: {float(fit.T_eff[qubit])*1000:.3f} mK<br>"
+    # )
```
andrea-pasquale, line 565: Remove commented lines

src/qibocal/protocols/characterization/t1_t2_vs_temperature.py
```diff
@@ -0,0 +1,571 @@
...
+            showlegend=True,
+            legendgroup="Effective Temperature",
+            yaxis="y3",
+        ),
+        row=1,
+        col=2,
+        secondary_y=False,
+    )
+
+    # fig.update_yaxes(title_text="Coherance (ns)", row=1, col=1)
```
andrea-pasquale, line 532: Remove commented line

src/qibocal/protocols/characterization/t1_t2_vs_temperature.py
```diff
@@ -0,0 +1,571 @@
...
+folder = Path("t1_t2_vs_temperature_data") / Path("data")
+folder.mkdir(exist_ok=True)
+
+
+@dataclass
+class T1T2vsTemperatureParameters(Parameters):
+    """T1T2vsTemperature runcard inputs."""
+
+    temperature_controller_ip: Optional[str] = "192.168.0.197"
+    """IP address of BF Temperature Controller."""
```
andrea-pasquale, line 58: Shoudn't the IP be on the platform runcard or inside a create function?

src/qibocal/protocols/characterization/t1_t2_vs_temperature.py
```diff
@@ -0,0 +1,571 @@
...
+    params: T1T2vsTemperatureParameters,
+    platform: Platform,
+    qubits: Qubits,
+) -> T1T2vsTemperatureData:
+    """
+    This method acquires the data necessary to determine how the qubits T1 and T2 vary with Temperature.
+    The routine controls the temperature of the fridge and triggers the execution of several other routines
+    once the temperature of the fridge is stable.
+
+    TODO
```
andrea-pasquale, line 171: For the correction syntax to render properly todo see for example here https://github.com/qiboteam/qibocal/blob/46346a104c07f5ff74fa4076e239a9dcc77983e7/src/qibocal/auto/task.py#L179

src/qibocal/protocols/characterization/t1_t2_vs_temperature.py
```diff
@@ -0,0 +1,571 @@
...
+        h_bar = 6.62607015e-34
+        Kb = 1.380649e-23
+        probability = {}
+        effective_temperature = {}
+        for qubit in qubits:
+            result: SampleResults = results[ro_pulses[qubit].serial]
+            P1 = probability[qubit] = float(result.probability(1))
+            P0 = 1 - P1
+            w01 = frequencies[qubit]
+            T = effective_temperature[qubit] = -h_bar * w01 / Kb / np.log(P1 / P0)
```
andrea-pasquale, line 233: I guess that this part could be seen as a fitting (?)

andrea-pasquale (commented):
Thanks @aorgazf for implementing this.

A few points that we should discuss.

1. Is it possible to test this protocol using a dummy platform? I don't think so given that we are using the driver for the fridge but it would be very nice to find a way to test this.
2. Given that this protocol is calling several other protocols I wonder whether it is possible to have a runcard with qubit spectroscopy, T1, and T2 and then create a script that changes the temperature of the frdige... 
3. About #533 I will have a look. I guess that the problem with this routine is the fact that it is acting as a monitoring tool (which we already discussed in the past)...

1.

PR #536, opened 2023-09-25
Fix update not working using `qq fit`

andrea-pasquale:
While performing some tests I noticed that when using `qq fit` the update is not performed correctly.
This PR fixes this issue and it also simplifies the structure of `Task`.

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`


stavros11 (approved):
Thank you for the fix. I am not very familiar with the `auto` part of the code, so it would be great if someone more familiar with it also reviews, but I do not see anything annoying in terms of code.

I also tested the update with one and two routines in the runcard and it worked as expected for the dummy platform. On main it does not work in either case. It may be useful to add a test that checks whether update works.

Edoardo-Pedicillo (approved):
Thank @andrea-pasquale

Issue #522, opened 2023-09-15
Report table refactoring


Edoardo-Pedicillo:
Implementing new and more flexible solutions to generate the report tables.
We could use the `to_html` Pandas function to generate a table automatically.

https://github.com/qiboteam/qibocal/pull/450#discussion_r1326993451

PR #546, opened 2023-10-05
html table generator

Edoardo-Pedicillo:
This PR solves #522, for the time being it is used only in time of flight routine.
Please, @andrea-pasquale @AleCandido have a look before propagating it.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


src/qibocal/protocols/characterization/utils.py
```diff
@@ -364,3 +366,19 @@ def significant_digit(number: float):
         position = max(position, np.ceil(-np.log10(abs(np.imag(number)))))
 
     return int(position)
+
+
+def table_html(data: dict):
```
AleCandido, line 371: Add a docstring :)

src/qibocal/protocols/characterization/utils.py
```diff
@@ -364,3 +366,19 @@ def significant_digit(number: float):
...
+    ncols = len(fitting_report.columns)
+    css = pd.DataFrame(
+        [["td_styles"] * ncols],
+        index=fitting_report.index,
+        columns=fitting_report.columns,
+    )
+    style = Styler(fitting_report, uuid_len=0, cell_ids=False)
+    style.set_td_classes(css)
+    style.set_table_attributes('class="fitting-table"')
+    fitting_report = style.hide(axis="index").to_html()
```
AleCandido, line 382: Do you really need to style it?

src/qibocal/web/templates/template.html
```diff
@@ -144,34 +144,8 @@ <h1>{{ report.title }}</h1>
...
-                                                </tr>
-                                              {% endif %}
-                                            {% endfor %}
-                                          </tbody>
-                                        </table>
-                                      </div>
-                                  {% endif %}
-                                  {{ figures }}
+            {{ fitting_report }}
+            {{ figures }}
```
AleCandido, line 148: Much nicer, thanks :D

src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
```diff
@@ -154,9 +154,13 @@ def _plot(data: TimeOfFlightReadoutData, qubit, fit: TimeOfFlightReadoutResults)
             line_dash="dash",
             line_color="grey",
         )
-
-        fitting_report = f"{qubit} | Time of flight(ns) : {fit.fitted_parameters[qubit] * S_TO_NS}<br>"
-
+        fitting_report = table_html(
+            {
+                "qubit": qubit,
+                "Parameter": "Time of flights(ns)",
```
andrea-pasquale, line 160:
```suggestion
                "Parameter": "Time of flights [ns]",
```

src/qibocal/protocols/characterization/utils.py
andrea-pasquale: I think that this is needed if we want to maintain the same style. Eventually we can move this directly in `template.html` (?)

andrea-pasquale (approved):
Thanks @Edoardo-Pedicillo.
LGTM, I have just a few comments.

I think that having a generic html as `fitting_report` is a good idea so that people can choose whatever they want. 
Before merging it, make sure to update also the documentation in particular the example on how to add a new protocol.

src/qibocal/protocols/characterization/utils.py
AleCandido: Indeed, I would not style the tables one by one. Just add the class and use a global CSS.

Edoardo-Pedicillo: > Indeed, I would not style the tables one by one. Just add the class and use a global CSS.

The styler is used only to add the classes according to [src/qibocal/web/static/styles.css](https://github.com/qiboteam/qibocal/blob/3bd6404e6c9ab35e725d219a6630bb6bd8a9794a/src/qibocal/web/static/styles.css)

AleCandido: You can style a child of the class you're putting in the HTML.

Just use a different selector:
```css
.div-fitting table {
  ...
}
```

src/qibocal/protocols/characterization/allxy/allxy.py
```diff
@@ -223,7 +223,7 @@ def _plot(data: AllXYData, qubit, fit: AllXYResults = None):
     """Plotting function for allXY."""
 
     figures = []
-    fitting_report = None
+    fitting_report = ""
```
AleCandido, line 226: I saw you now modified this already in plenty of places.

However, even when refactoring, try to avoid the option that requires touching all the routines.
Before they were all `None`, and this could have been fine as well: if you really need an empty string, just write
```py
report = fitting_report if fitting_report is not None else ""
```
in the single (or few) consumer(s) of the returned values.

src/qibocal/web/static/styles.css
```diff
@@ -149,15 +149,15 @@
   border-radius: 15px;
 }
 
-.th_styles
+.div-fitting th
```
AleCandido, line 152:
```suggestion
.fitting-table th
```

Use the closer parent, stay more local. No need to go beyond.

src/qibocal/web/static/styles.css
```diff
@@ -149,15 +149,15 @@
...
 {
   background-color: #f0e6ff;
   padding: 8px;
   text-align: center;
   font-family: system-ui,-apple-system,"Segoe UI",Roboto,"Helvetica Neue","Noto Sans","Liberation Sans",Arial,sans-serif,"Apple Color Emoji","Segoe UI Emoji","Segoe UI Symbol","Noto Color Emoji";
 }
 
-.td_styles
+.div-fitting td
```
AleCandido, line 160:
```suggestion
.fitting-table td
```

src/qibocal/protocols/characterization/utils.py
```diff
@@ -373,3 +376,61 @@ def significant_digit(number: float):
...
+        data (dict): the keys will be converted into table entries and the
+        values will be the columns of the table.
+
+    Return:
+        str
+    """
+    fitting_report = pd.DataFrame(data)
+    fitting_report = fitting_report.to_html(
+        classes="fitting-table", index=False, border=0
+    )
```
AleCandido, line 434:
```suggestion
    fitting_report = pd.DataFrame(data).to_html(
        classes="fitting-table", index=False, border=0
    )
```
Not much to save characters, but to avoid shadowing.

src/qibocal/protocols/characterization/utils.py
```diff
@@ -373,3 +376,61 @@ def significant_digit(number: float):
...
+    fitting_report = '<div class="div-fitting">' + fitting_report + "</div>"
+    return fitting_report
```
AleCandido, line 436:
```suggestion
    return '<div class="div-fitting">' + fitting_report + "</div>"
```
Same of above.

src/qibocal/protocols/characterization/utils.py
```diff
@@ -373,3 +376,61 @@ def significant_digit(number: float):
...
+        values will be the columns of the table.
+
+    Return:
+        str
+    """
+    fitting_report = pd.DataFrame(data)
+    fitting_report = fitting_report.to_html(
+        classes="fitting-table", index=False, border=0
+    )
+    fitting_report = '<div class="div-fitting">' + fitting_report + "</div>"
```
AleCandido, line 435: This is mostly used for centering the table:
https://github.com/qiboteam/qibocal/blob/19e029070c93458c2719974e7dcaf8e5966a1fd5/src/qibocal/web/static/styles.css#L109-L118

Font and margins could be moved to the table itself, i.e. `fitting-table` (btw, `margin-top` could be included in `margins`, just writing `margin: 40px 0 0 -100px;` not sure why you spell out both).

The leftover is only `display: flex` and vertical and horizontal centering.
However, the table is already an item of a larger container, and it could be vertically and horizontally centered as such.

In case this is done, you would not need to touch Pandas output at all, and get rid of some more HTML inside the code.

AleCandido (commented):
HTML comments.

AleCandido:
Is the report still working as expected with the new CSS?
(with CSS you often have to iterate to get a nice result)

Edoardo-Pedicillo:
> Is the report still working as expected with the new CSS?

Yes, thank you

src/qibocal/protocols/characterization/allxy/allxy.py
Edoardo-Pedicillo: I will keep it in mind for the next refactoring :)

andrea-pasquale: I was also going to comment on this...
I guess that by using `""` you have been able to remove this `if` statement in the html template.
Personally I prefer `None` but an empty string works as well.

andrea-pasquale (approved):
Thanks @Edoardo-Pedicillo for this.
It looks good to me, you can find below a few small comments.

scarrazza:
Could we do something to fix coverage before merging?

AleCandido:
> Could we do something to fix coverage before merging?

Improving coverage is always a good thing.

But in this case the percentage is dropping because some lines that were tested have been removed, but there are no new lines that are not tested.
So, for cases like this one, I believe we should add also here a threshold. This is definitely a false positive.

src/qibocal/protocols/characterization/utils.py
```diff
@@ -373,3 +376,56 @@ def significant_digit(number: float):
...
+    """This function converts a dictionary into an HTML table.
+
+    Args:
+        data (dict): the keys will be converted into table entries and the
+        values will be the columns of the table.
+
+    Return:
+        str
+    """
+    return pd.DataFrame(data).to_html(classes="fitting-table", index=False, border=0)
```
AleCandido, line 431: Since this function is just one line (which is not a bad thing) and it is always used together with the previous one, maybe it is worth to merge the two of them.

AleCandido (approved):
A few comments, but it is generally fine :)

src/qibocal/protocols/characterization/utils.py
Edoardo-Pedicillo: For the time being I agree with you, but I prefer to keep them separated to have more flexibility, if we want different table entries we should repeat this line each time.

PR #549, opened 2023-10-05
Adding "coherence"  protocols with probabilities and error bands. 

andrea-pasquale:
In this PR I'm planning to add new protocols where the y axes correspond to probabilities and not longer MSR.
In this way we can use again the error given by the binomial distribution.
Protocols to modify:

- [x] T1
- [x] T2
- [x] SpinEcho
- [x] Zeno
- [x] Restore Ramsey with MSR

For the time being I'm keeping the ones with MSR. In the future we can think about removing them and just compute T1, T2, .. after single shot classification.

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/characterization/coherence/spin_echo.py
+++ b/src/qibocal/protocols/characterization/coherence/spin_echo.py
@@ -12,7 +12,7 @@ from qibocal import update
 from qibocal.auto.operation import Parameters, Qubits, Results, Routine
 
 from ..utils import V_TO_UV
-from .t1 import T1Data
+from .t1_msr import T1MSRData
 from .utils import exp_decay, exponential_fit
 
 
@@ -42,7 +42,7 @@ class SpinEchoResults(Results):
     """Raw fitting output."""
 
 
-class SpinEchoData(T1Data):
+class SpinEchoData(T1MSRData):
     """SpinEcho acquisition outputs."""
 
 
--- a/src/qibocal/protocols/characterization/coherence/t1.py
+++ b/src/qibocal/protocols/characterization/coherence/t1.py
@@ -134,7 +134,7 @@ def _acquisition(params: T1Parameters, platform: Platform, qubits: Qubits) -> T1
 
     for qubit in qubits:
         probs = results[ro_pulses[qubit].serial].probability(state=1)
-        errors = [np.sqrt(prob * (1 - prob) / params.nshots) for prob in probs]
+        errors = np.sqrt(probs * (1 - probs) / params.nshots)
         data.register_qubit(qubit, wait=ro_wait_range, prob=probs, error=errors)
 
     return data
--- a/src/qibocal/protocols/characterization/coherence/t1.py
+++ b/src/qibocal/protocols/characterization/coherence/t1.py
@@ -67,10 +67,7 @@ class T1Data(Data):
         ar["wait"] = wait
         ar["prob"] = prob
         ar["error"] = error
-        if qubit in self.data:
-            self.data[qubit] = np.rec.array(np.concatenate((self.data[qubit], ar)))
-        else:
-            self.data[qubit] = np.rec.array(ar)
+        self.data[qubit] = np.rec.array(ar)
 
 
 def _acquisition(params: T1Parameters, platform: Platform, qubits: Qubits) -> T1Data:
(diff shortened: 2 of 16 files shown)


src/qibocal/protocols/characterization/coherence/spin_echo.py
```diff
@@ -100,22 +100,25 @@ def _acquisition(
...
             ),
         )
 
         for qubit in qubits:
-            result = results[ro_pulses[qubit].serial]
-            data.register_qubit(
-                qubit, wait=wait, msr=result.magnitude, phase=result.phase
-            )
+            prob = results[ro_pulses[qubit].serial].probability(state=0)
+            probs[qubit].append(prob)
```
Edoardo-Pedicillo, line 110: Should we use the excited state probabilities?

src/qibocal/protocols/characterization/coherence/t1.py
```diff
@@ -127,17 +126,16 @@ def _acquisition(params: T1Parameters, platform: Platform, qubits: Qubits) -> T1
...
         sweeper,
     )
 
     for qubit in qubits:
-        result = results[ro_pulses[qubit].serial]
-        data.register_qubit(
-            qubit, wait=ro_wait_range, msr=result.magnitude, phase=result.phase
-        )
+        probs = results[ro_pulses[qubit].serial].probability(state=1)
+        errors = [np.sqrt(prob * (1 - prob) / params.nshots) for prob in probs]
```
Edoardo-Pedicillo, line 137: if `prob` is a np array, the for loop is unnecessary.

src/qibocal/protocols/characterization/coherence/t2.py
```diff
@@ -97,15 +94,16 @@ def _acquisition(
...
+            averaging_mode=AveragingMode.SINGLESHOT,
         ),
         sweeper,
     )
 
     for qubit in qubits:
-        result = results[ro_pulses[qubit].serial]
-        data.register_qubit(qubit, wait=waits, msr=result.magnitude, phase=result.phase)
+        probs = results[ro_pulses[qubit].serial].probability(state=1)
+        errors = [np.sqrt(prob * (1 - prob) / params.nshots) for prob in probs]
```
Edoardo-Pedicillo, line 105: Also here you can use np arrays

src/qibocal/protocols/characterization/coherence/utils.py
```diff
@@ -63,3 +63,47 @@ def exponential_fit(data, zeno=None):
...
+                maxfev=2000000,
+                bounds=(
+                    [-2, -2, 0],
+                    [2, 2, np.inf],
+                ),
+                sigma=data[qubit].error,
+            )
+            popt = popt.tolist()
+            perr = np.sqrt(np.diag(perr))
+            perr[2] = 0.1
```
Edoardo-Pedicillo, line 98:
```suggestion
```
I suppose this should be deleted.

src/qibocal/protocols/characterization/coherence/utils.py
```diff
@@ -63,3 +63,47 @@ def exponential_fit(data, zeno=None):
...
+        except Exception as e:
+            log.warning(f"Exp decay fitting was not succesful. {e}")
+            popt = [0] * 3
+            dec = 5
+            perr = [1] * 3
+
+        fitted_parameters[qubit] = popt
+        dec = popt[2]
+        decay[qubit] = (dec, perr[2])
```
Edoardo-Pedicillo, line 107: Since this formula is used to fit https://github.com/qiboteam/qibocal/blob/e48ec1cf66f18eac13de3f26b09c19213fbe94a2/src/qibocal/protocols/characterization/coherence/utils.py#L9C1-L10C1
the decay error should be perr[2]/popt[2]^2. I suggest to change directly `exp_decay` in the following way
```python
def exp_decay(x, *p):
    return p[0] - p[1] * np.exp(-1 * x/p[2])
```
in this way you get directly the decay error without error propagation.

src/qibocal/protocols/characterization/coherence/spin_echo.py
andrea-pasquale: Let's wait until we deploy on hardware to check what is the correct one.

src/qibocal/protocols/characterization/coherence/utils.py
andrea-pasquale: Indeed you are right, thanks!

src/qibocal/protocols/characterization/coherence/utils.py
andrea-pasquale: Yup

GabrielePalazzo (approved):
Nothing to comment from my side. Tests are passing and the code seems correct.

Edoardo-Pedicillo (approved):

PR #554, opened 2023-10-11
abstract register qubit

### LINKED ISSUE #541: Abstract register_qubit
https://github.com/qiboteam/qibocal/pull/532#discussion_r1336983033

Edoardo-Pedicillo:
This PR closes #541 

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/auto/operation.py
+++ b/src/qibocal/auto/operation.py
@@ -143,6 +143,19 @@ class Data:
 
         return obj
 
+    def register_qubit(self, dtype, *args, **kwargs):
+        """Store output for single qubit."""
+        # to be able to handle the non-sweeper case
+        ar = np.empty(np.shape(kwargs[list(kwargs.keys())[0]]), dtype=dtype)
+        for key, value in kwargs.items():
+            ar[key] = value
+        if len(args) == 1:
+            args = args[0]
+        if args in self.data:
+            self.data[args] = np.rec.array(np.concatenate((self.data[args], ar)))
+        else:
+            self.data[args] = np.rec.array(ar)
+
 
 @dataclass
 class Results:
--- a/src/qibocal/fitting/classifier/data.py
+++ b/src/qibocal/fitting/classifier/data.py
@@ -17,10 +17,11 @@ def generate_models(data, qubit, test_size=0.25):
         - x_test: Test inputs.
         - y_test: Test outputs.
     """
-    qubit_data = data.data[qubit]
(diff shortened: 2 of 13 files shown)


Edoardo-Pedicillo:
@AleCandido @andrea-pasquale Could you please take a quick look ? Sorry for the commented code, I will remove it as soon as possible

andrea-pasquale:
> @AleCandido @andrea-pasquale Could you please take a quick look at the new `register_qubit` function? Sorry for the commented code, I will remove it as soon as possible

I think that it looks good.
A possible enhancement could be to have a clearer distinction between `keys` and `values` in `data`.
Currently `keys` are treated as `args` while `values` are treated as `kwargs`. 

https://github.com/qiboteam/qibocal/blob/6d0ac49e8d6c2097ff67c36c5b29141abfc69842/src/qibocal/auto/operation.py#L146-L157

We could introduce an argument called `key` or `index` to pass the key and for the rest we can keep `kwargs`.
I think that this could be relevant given that for some protocols the key could be also a tuple (especially for two qubit gates protocols).
Let me know what you think.

AleCandido:
Unfortunately, I will be unable to have look until late next week, sorry

andrea-pasquale:
> Unfortunately, I will be unable to have look until late next week, sorry

Don't worry @AleCandido. Thank you for letting us know. I will put @GabrielePalazzo as a reviewer.

src/qibocal/auto/operation.py
```diff
@@ -143,6 +143,19 @@ def load(cls, path):
 
         return obj
 
+    def register_qubit(self, dtype, *args, **kwargs):
```
GabrielePalazzo, line 146: Discriminating arguments by how they are passed to the function looks too ambiguous to me.
I would change the function arguments to something like this:
```
def register_qubit(self, dtype, keys_of_selfdata: list, keys_of_ar: dict):
```

GabrielePalazzo (changes requested):
I agree with @andrea-pasquale.

Apart from the comment below, everything else seems good to me.

GabrielePalazzo (approved):

andrea-pasquale (approved):
Thanks @Edoardo-Pedicillo, I tested the code when I was calibrating qw5q_gold and it looks good to me. As soon as you fixed conflicts we should be ready to merge.

PR #543, opened 2023-10-03
Qutrit classifier

Edoardo-Pedicillo:
As the title says

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.
- [ ] Compatibility with Qibo modules (Please edit this section if the current pull request is not compatible with the following branches).
    - [ ] Qibo: `master`
    - [ ] Qibolab: qiboteam/qibolab#615
    - [ ] Qibolab_platforms_qrc: `main`

PR #558, opened 2023-10-12
Adding "mean excited state population / effective qubit temperature" protocol

rodolfocarobene:
Hi!

Since I saw that you implemented the RX12, I tried to code a routine to compute the average excited state population and the effective qubit temperature. These are interesting parameter for characterization, although they are not useful for calibration and do not lead to update any parameter.
The experiment is taken from page 46-47 of https://knowledge.uchicago.edu/record/2942/files/Dixit_uchicago_0330D_15670.pdf.

If you prefer not to merge it, since you have other things to take care of, I can understand. Eventually I will fork qibocal for the purpose of adding new experiments without increasing your workload.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


rodolfocarobene:
I thought this was ready, but I just noticed that I have to use probabilities (P(e)) and not magnitude... 
So I suppose that I have to use a three state classifier that, if I remember correctly, you are already developing.
I'll wait till then

andrea-pasquale:
Thank you very much @rodolfocarobene!
Indeed there is a prototype for the qutrit classifier in #543.
Currently I don't think that we want to plug the qutrit classifier directly in qibolab, in general if a routine needs the qutrit classification the idea will be that you run first the qutrit classification and then you run after the protocol that needs qutrit classification. 

I will have a look probably next week but the code already looks good from my side.

In theory we can have a first estimate for a temperature like parameter by looking at MSR until the classifier is ready.

Edoardo-Pedicillo:
@rodolfocarobene, the #543 has been merged, feel free to change the status of this PR and assign the reviewers.

rodolfocarobene:
> @rodolfocarobene, the #543 has been merged, feel free to change the status of this PR and assign the reviewers.

Oh nice, thanks for notifying me! I will fix conflicts and try to complete this asaic

andrea-pasquale:
> Oh nice, thanks for notifying me! I will fix conflicts and try to complete this asaic

Thanks! Btw, we have also added a RabiEF experiment (rabi between $\ket{1}$ and $\ket{2}$). If I understood correctly for your protocol we can use the same acquisition protocol with the only difference that you don't prepare $\ket{1}$ before scanning the amplitude.

rodolfocarobene:
Hey, I've looked into the RabiEF acquisition but for the moment I would maintain mine. Because RabiEF works with MSR and also because it works with amplitude; it should still work fine with amplitude, but I would prefer to use the same parameters of the thesis from which I took the experiment.

However, the routine is still a draft... mainly because I don't know how to do the classification:
https://github.com/qiboteam/qibocal/blob/9fcde2863a82147e46b86921315b00619e4e81e9/src/qibocal/protocols/characterization/excited_state_population.py#L179-L191

I tried looking into the qutrit and into qibocal.fitting.classifiers but without much gain. 
Could you give me some help when you can? Here on GitHub or, eventually, we could have a quick chat.
Obviously when/if you can.  Thaanks :-)

rodolfocarobene:
Closing it since it's useless to have a PR stalled forever, I will maybe retry in the future :-)

PR #561, opened 2023-10-16
Now excluding differences lower than -1*st.dev.

GabrielePalazzo:
There was a typo in how the lorentzian offset guess was computed.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


Edoardo-Pedicillo (commented):
Since the formula of the offset is the same for the two branches, you could evaluate it outside the if-else.

GabrielePalazzo:
> Since the formula of the offset is the same for the two branches, you could evaluate it outside the if-else.

Indeed.

andrea-pasquale (approved):

PR #576, opened 2023-10-24
2q clifford filtered rb

Jacfomg:
Draft missing:

- Single qubit gates
- Filter function
- A way to reduce code from single qubit clifford filtered RB

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


andrea-pasquale:
Are we planning to start from here for the 2q RB? If not I suggest to close it.

Jacfomg:
We need some stuff from here and I assume we should go for 1q filtered 1st too. And even before that we need to save the circuits or indexes properly on std RB.

PR #578, opened 2023-10-24
Refactor qubit flux and crosstalk routines

Edoardo-Pedicillo:
As the title says
Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


andrea-pasquale:
image
Just a two quick comments since you are working on this routine:

- we should change the units of measure from Hz to GHz
- the range on the xaxis should be different for each qubit while they appear to be the same

Edoardo-Pedicillo:
>  - we should change the units of measure from Hz to GHz

Done

andrea-pasquale (commented):
I just realized that I forgot to push some comments.
They should be easy to fix.

andrea-pasquale (approved):
Thanks @Edoardo-Pedicillo, LGTM.
If it is easy to fix the coverage could you give it a try?
Otherwise we can just merge.

Edoardo-Pedicillo:
The lost of coverage is due to the anharmonicity that is 0 in the dummy runcard.
https://github.com/qiboteam/qibocal/blob/5011129569ca9233a2b34753eaff35d47a8fb925/src/qibocal/protocols/characterization/flux_dependence/qubit_crosstalk.py#L85-L88
If we change it, we will lose the coverage of the first branch.

andrea-pasquale:
> The lost of coverage is due to the anharmonicity that is 0 in the dummy runcard.
> 
> https://github.com/qiboteam/qibocal/blob/5011129569ca9233a2b34753eaff35d47a8fb925/src/qibocal/protocols/characterization/flux_dependence/qubit_crosstalk.py#L85-L88
> 
> 
> If we change it, we will lose the coverage of the first branch.

I see. Nevermind than. Feel free to merge as soon as you have 2 approvs

GabrielePalazzo (approved):

PR #579, opened 2023-10-24
Reducing import overhead

andrea-pasquale:
After talking with @stavros11 we realized that now importing qibocal takes longer than usual.
After profiling the import we discover that importing `skops` introduces an overhead of around 2 seconds.
Here is the profiling with skops:
image
This PR reduces the import time by moving `skops` as a relative import.
Here is the updated profiling:
image

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/fitting/classifier/qubit_fit.py
+++ b/src/qibocal/fitting/classifier/qubit_fit.py
@@ -4,7 +4,6 @@ from pathlib import Path
 
 import numpy as np
 import numpy.typing as npt
-import skops.io as sio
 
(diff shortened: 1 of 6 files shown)


Edoardo-Pedicillo (approved):
Thank @andrea-pasquale

stavros11 (approved):
Thanks @andrea-pasquale. I see a decrease of 1+ sec on total execution time when I run with dummy on my machine, maybe more in the cluster.

PR #580, opened 2023-10-24
David/twpa opt

DavidSarlle:
This Pr implements the TWPA characterization obtaining the best fidelity while sweeping frequency and power.

image

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/characterization/readout_optimization/twpa_calibration/frequency_power.py
+++ b/src/qibocal/protocols/characterization/readout_optimization/twpa_calibration/frequency_power.py
@@ -1,5 +1,4 @@
 from dataclasses import dataclass, field
-from typing import Optional
 
 import numpy as np
 import numpy.typing as npt
@@ -117,7 +116,7 @@ def _acquisition(
                 )
 
                 classification_result = classification._fit(classification_data)
-    
+
                 data.register_qubit(
                     TwpaFrequencyPowerType,
                     (qubit),
--- a/src/qibocal/protocols/characterization/readout_optimization/twpa_calibration/frequency_power.py
+++ b/src/qibocal/protocols/characterization/readout_optimization/twpa_calibration/frequency_power.py
@@ -6,12 +6,9 @@ import numpy.typing as npt
 import plotly.graph_objects as go
 from qibolab.platform import Platform
 from qibolab.qubits import QubitId
-from sklearn.model_selection import train_test_split
 
 from qibocal import update
 from qibocal.auto.operation import Data, Parameters, Qubits, Results, Routine
-from qibocal.fitting.classifier.qubit_fit import QubitFit
-from qibocal.fitting.classifier.run import benchmarking
 from qibocal.protocols.characterization import classification
 from qibocal.protocols.characterization.utils import HZ_TO_GHZ, table_dict, table_html
 
@@ -33,6 +30,7 @@ class TwpaFrequencyPowerParameters(Parameters):
     relaxation_time: Optional[int] = None
     """Relaxation time (ns)."""
 
+
 TwpaFrequencyPowerType = np.dtype(
     [
         ("freq", np.float64),
@@ -40,6 +38,7 @@ TwpaFrequencyPowerType = np.dtype(
         ("assignment_fidelity", np.float64),
     ]
 )
+
 
 @dataclass
 class TwpaFrequencyPowerData(Data):
@@ -53,6 +52,7 @@ class TwpaFrequencyPowerData(Data):
     """Frequencies for each qubit."""
     powers: dict[QubitId, float] = field(default_factory=dict)
     """Frequencies for each qubit."""
+
 
 @dataclass
 class TwpaFrequencyPowerResults(Results):
@@ -146,12 +146,12 @@ def _acquisition(
 def _fit(data: TwpaFrequencyPowerData) -> TwpaFrequencyPowerResults:
     """Extract fidelity for each configuration qubit / param.
     Where param can be either frequency or power."""
-    
+
     best_freq = {}
     best_power = {}
     best_fidelity = {}
     qubits = data.qubits
-    
+
     for qubit in qubits:
         data_qubit = data[qubit]
         index_best_err = np.argmax(data_qubit["assignment_fidelity"])
@@ -160,6 +160,7 @@ def _fit(data: TwpaFrequencyPowerData) -> TwpaFrequencyPowerResults:
         best_power[qubit] = data_qubit["power"][index_best_err]
 
     return TwpaFrequencyPowerResults(best_freq, best_power, best_fidelity)
+
 
 def _plot(data: TwpaFrequencyPowerData, fit: TwpaFrequencyPowerResults, qubit):
     """Plotting function that shows the assignment fidelity
@@ -184,8 +185,14 @@ def _plot(data: TwpaFrequencyPowerData, fit: TwpaFrequencyPowerResults, qubit):
             )
         )
 
-        fig = go.Figure([go.Heatmap(x=frequencies * HZ_TO_GHZ, y=powers, z=fidelities, name="Fidelity")])
-        
+        fig = go.Figure(
+            [
+                go.Heatmap(
+                    x=frequencies * HZ_TO_GHZ, y=powers, z=fidelities, name="Fidelity"
+                )
+            ]
+        )
+
         fig.update_layout(
             showlegend=True,
             xaxis_title="TWPA Frequency [GHz]",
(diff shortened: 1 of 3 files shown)


Jacfomg:
It would be nice to scan the SNR instead of best fidelity, just as we do with the VNA. To avoid this unstable points ypu think were causing iqm to get those results, right ?

DavidSarlle:
@Jacfomg thanks for the comments. 

Finally I have detected that the cause of the decresement in the qubit fidelities is not caused by the TWPA. It is caused by the cross talk between q2 and the rest of the qubits in the iqm chip. When any single qubit gate is made over q0, q1, q3 or q4 the fidelities are good consistently with the current characterization, but when we do sigle qubit gates at the same time, including q2, all the fidelities decrease 20-30%. That results are shown consistenly in all conditions (other chips working, different fridge temp, different hours...) So, I am pretty sure that the problem is not the TWPA. 

Even with that, we have coded this function with the assg. fidelitie as the target to find the best power/freq parameters instead of the SNR. I agree that it could be better, but following the 2 routines that we alredy have in qibocal main, I coded this one. It is good enoygh for the moment and the routine helps alot in the TWPA characterization. So I want to merge it into main asap.

src/qibocal/protocols/characterization/readout_optimization/twpa_calibration/frequency_power.py
```diff
@@ -0,0 +1,213 @@
...
+    frequency_step: float
+    """Frequency step to be probed."""
+    power_width: float
+    """Power total width."""
+    power_step: float
+    """Power step to be probed."""
+    nshots: Optional[int] = None
+    """Number of shots."""
+    relaxation_time: Optional[int] = None
+    """Relaxation time (ns)."""
```
andrea-pasquale, line 31: You can remove those two attributes. They should be imported automatically from `Parameters`

src/qibocal/protocols/characterization/readout_optimization/twpa_calibration/frequency_power.py
```diff
@@ -0,0 +1,213 @@
...
+    with step frequency_step.
+
+    Args:
+        params (:class:`TwpaFrequencyParameters`): input parameters
+        platform (:class:`Platform`): Qibolab's platform
+        qubits (dict): dict of target :class:`Qubit` objects to be characterized
+
+    Returns:
+        data (:class:`TwpaFrequencyData`)
+    """
```
andrea-pasquale, line 84: Outdated docstring

src/qibocal/protocols/characterization/readout_optimization/twpa_calibration/frequency_power.py
```diff
@@ -0,0 +1,213 @@
...
+        for qubit in qubits:
+            platform.qubits[qubit].twpa.local_oscillator.frequency = (
+                initial_twpa_freq[qubit] + freq
+            )
+
+        for power in power_range:
+            for qubit in qubits:
+                platform.qubits[qubit].twpa.local_oscillator.power = (
+                    initial_twpa_power[qubit] + power
+                )
```
andrea-pasquale, line 114: Here I believe you can only 1 for loop around qubits:
```py
for freq in freqs:
    for power in powers:
        set_freq(...)
        set_power(...)
        


```

andrea-pasquale (commented):
Thanks @DavidSarlle, I need to test on hardware but the code looks generally good to me.
To avoid decreasing coverage could you please add this routine also in this [runcard](https://github.com/qiboteam/qibocal/blob/main/tests/runcards/protocols.yml)? In this way it will automatically tested using the dummy platform in GitHub.

andrea-pasquale (commented):
Thanks @DavidSarlle, I need to test on hardware but the code looks generally good to me.
To avoid decreasing coverage could you please add this routine also in this [runcard](https://github.com/qiboteam/qibocal/blob/main/tests/runcards/protocols.yml)? In this way it will automatically tested using the dummy platform in GitHub.

Jacfomg:
This may be a routine used only to finetune and not very reliable and taking long time for what I realized because it depends on the calibrate qubit states, meaning a higher number of shots than usual to avoid getting lucky on some points let's say. Also, you have no guarantees that this good points will not be just distorting the clouds leaving an unusable qubit. I will strongly avoid touching the TWPA power outside the routine where we get the SNR on the resonator and just finetune the frequency using the 1D routine on most cases. What do you think @andrea-pasquale  and @Edoardo-Pedicillo ?

andrea-pasquale:
> This may be a routine used only to finetune and not very reliable and taking long time for what I realized because it depends on the calibrate qubit states, meaning a higher number of shots than usual to avoid getting lucky on some points let's say. Also, you have no guarantees that this good points will not be just distorting the clouds leaving an unusable qubit. I will strongly avoid touching the TWPA power outside the routine where we get the SNR on the resonator and just finetune the frequency using the 1D routine on most cases. What do you think @andrea-pasquale and @Edoardo-Pedicillo ?

I agree with the fact that it should be used only for finetuning, and the issues that @Jacfomg presented are true.
I saw several plots with 99% fidelity which were mainly the results of a very weird/lucky calibration of the twpa.
Having said this, the major issue I believe is to have protocols to calibrate twpas from scratch using SNR.

We can merge this protocol as it is but we should avoid to run long scans with is, since as we saw it can be very unstable.

DavidSarlle:
> Thanks @DavidSarlle, I need to test on hardware but the code looks generally good to me. To avoid decreasing coverage could you please add this routine also in this [runcard](https://github.com/qiboteam/qibocal/blob/main/tests/runcards/protocols.yml)? In this way it will automatically tested using the dummy platform in GitHub.

Sure. Done. thanks

src/qibocal/protocols/characterization/readout_optimization/twpa_calibration/frequency_power.py
DavidSarlle: REmeoved

src/qibocal/protocols/characterization/readout_optimization/twpa_calibration/frequency_power.py
DavidSarlle: Updated docstrings

src/qibocal/protocols/characterization/readout_optimization/twpa_calibration/frequency_power.py
DavidSarlle: Done

DavidSarlle:
> > This may be a routine used only to finetune and not very reliable and taking long time for what I realized because it depends on the calibrate qubit states, meaning a higher number of shots than usual to avoid getting lucky on some points let's say. Also, you have no guarantees that this good points will not be just distorting the clouds leaving an unusable qubit. I will strongly avoid touching the TWPA power outside the routine where we get the SNR on the resonator and just finetune the frequency using the 1D routine on most cases. What do you think @andrea-pasquale and @Edoardo-Pedicillo ?
> 
> I agree with the fact that it should be used only for finetuning, and the issues that @Jacfomg presented are true. I saw several plots with 99% fidelity which were mainly the results of a very weird/lucky calibration of the twpa. Having said this, the major issue I believe is to have protocols to calibrate twpas from scratch using SNR.
> 
> We can merge this protocol as it is but we should avoid to run long scans with is, since as we saw it can be very unstable.

Obviously should be use for fine retunning, because it depends on the initial calibration from the lab using the VNA and also depends on the pi pulse calibration. But even with that, for recalibrating all these routines are useful and provide a method to re callibrate without using the VNA and be present in the lab. 

Regarding the duration, as all the routines, depends on the range that you scan. As I said, as you should have an initial callibration from the lab and the VNA, it should not be a large scan

Jacfomg:
> Regardiong the inestability points, I have also seen this points using the twpa freq or power. But with the 2d scan is much easier to identify this regions and be sure that you are in the correct area for a given power and freq.

The main problem is that we would be trying to measure these unstable points relying on the results of another unstable routine, the calibrate qubits states, so this be way trickier. I would say that the first step for the finetune would be the resonator SNR with power and frequency on the TWPA.

DavidSarlle:
@andrea-pasquale can we merge this PR. Until we do not implement other methods like the resonator spec vs TWPA freq/power and fit the best SNR point, this routine is still useful for the people that is re-calibrating as we can see what are the unestable points and we can refine the TWPA once the pi pulses are defined.

Edoardo-Pedicillo (approved):
Thank @DavidSarlle LGTM

Jacfomg (approved):
I hope approving this routine will mean that we won't get the others, but still, you are calibrating using an unstable figure of merit, so if you are going to use try to have a big number of shots above 10_000 to 50_000 at least to get reproducible results and not just some lucky blobs that may misslead the calibration.

andrea-pasquale (approved):

PR #582, opened 2023-10-25
kernels

Jacfomg:
Prototype routine to test the kernel computation for integration using zhinst functions.
Also, results.npz and result.json will be generated, which I didn't do correctly and all routines crash.

Tests will fail until we merge https://github.com/qiboteam/qibolab/pull/752 to get kernels in the dummy.
 
Checklist:
- [ ] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`


src/qibocal/auto/operation.py
andrea-pasquale, line 1: At this point if the implementation of `Results` is mirroring `Data` we could define an abstraction layer instead of duplicating code.

src/qibocal/protocols/characterization/signal_experiments/calibrate_state_discrimination.py
```diff
@@ -0,0 +1,226 @@
...
+    ] = field(default_factory=dict)
+
+    def register_qubit(self, qubit, state, msr, phase, i, q):
+        """Store output for single qubit."""
+        ar = np.empty(i.shape, dtype=CalibrateStateDiscriminationType)
+        ar["msr"] = msr
+        ar["phase"] = phase
+        ar["i"] = i
+        ar["q"] = q
+        self.data[qubit, state] = np.rec.array(ar)
```
andrea-pasquale, line 75: I think that we can remove this and the protocol should still work

src/qibocal/protocols/characterization/signal_experiments/calibrate_state_discrimination.py
```diff
@@ -0,0 +1,226 @@
...
+    for qubit in qubits:
+        qd_pulses[qubit] = platform.create_RX_pulse(qubit, start=0)
+        ro_pulses[qubit] = platform.create_qubit_readout_pulse(
+            qubit, start=qd_pulses[qubit].duration
+        )
+        sequence_0.add(ro_pulses[qubit])
+        sequence_1.add(qd_pulses[qubit])
+        sequence_1.add(ro_pulses[qubit])
+
+    # create a DataUnits objects to store the results
```
andrea-pasquale, line 111: Remove outdated comment

src/qibocal/protocols/characterization/signal_experiments/calibrate_state_discrimination.py
```diff
@@ -0,0 +1,226 @@
...
+            result = results[ro_pulses[qubit].serial]
+            # store the results
+            data.register_qubit(
+                qubit=qubit,
+                state=i,
+                msr=result.magnitude,
+                phase=result.phase,
+                i=result.voltage_i,
+                q=result.voltage_q,
+            )
```
andrea-pasquale, line 147: This should be changed if you drop your implementation of `register_qubit`.
You can have a look at the other routines to see how it should work.

Edoardo-Pedicillo:
Is this PR in draft or under review?

Jacfomg:
Well, it was draft because the results were not the expected, I'm waiting for comments from Zurich

src/qibocal/auto/operation.py
```diff
@@ -200,15 +202,53 @@ class Cmd1Res(Results):
 
     """
 
+    def __post_init__(self):
+        if "data" not in self.__dict__:
+            self.data: Optional[
+                dict[Union[tuple[QubitId, int], QubitId], npt.NDArray]
+            ] = None
```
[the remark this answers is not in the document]
Jacfomg: No, because then it was complaning that I was giving after other non default arguments

src/qibocal/protocols/characterization/signal_experiments/calibrate_state_discrimination.py
```diff
@@ -0,0 +1,219 @@
...
+from qibolab import AcquisitionType, AveragingMode, ExecutionParameters
+from qibolab.platform import Platform
+from qibolab.pulses import PulseSequence
+from qibolab.qubits import QubitId
+
+from qibocal.auto.operation import Data, Parameters, Qubits, Results, Routine
+
+
+@dataclass
+class CalibrateStateDiscriminationParameters(Parameters):
```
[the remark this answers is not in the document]
Jacfomg: ikr, but run out of ideas, do you have any ?

Jacfomg:
It needs laboneq 2.20.0 version that will get merged soon. And testing on hardware.

src/qibocal/protocols/characterization/signal_experiments/calibrate_state_discrimination.py
```diff
@@ -0,0 +1,213 @@
+from dataclasses import dataclass, field
+from typing import Optional
+
+import numpy as np
+import numpy.typing as npt
+import plotly.graph_objects as go
+from laboneq.analysis import calculate_integration_kernels
```
AleCandido, line 7: This is of course specific for zhinst. Keep at least a TODO, since it will have to be wrapped in Qibolab.

Jacfomg: It just some math on an output array that should be compatible with every device. But, I agree that once it works we should try to reproduce without using their function to avoid using their stuff, right ?

Jacfomg: I removed the laboneq function so this can be reviewed.

src/qibocal/protocols/characterization/signal_experiments/calibrate_state_discrimination.py
andrea-pasquale, line 1: Here you need to rename `msr` -> `signal`.

src/qibocal/protocols/characterization/signal_experiments/calibrate_state_discrimination.py
```diff
@@ -0,0 +1,220 @@
...
+
+        figures.append(fig)
+
+    return figures, fitting_report
+
+
+def _update(
+    results: CalibrateStateDiscriminationResults, platform: Platform, qubit: QubitId
+):
+    pass
```
andrea-pasquale, line 216: After we release `qibolab` you can define a proper update, right?

src/qibocal/protocols/characterization/signal_experiments/calibrate_state_discrimination.py
```diff
@@ -0,0 +1,220 @@
...
+    kernel_state_zero = {}
+    for qubit in qubits:
+        traces = []
+        for i in range(2):
+            # This is due to a device limitation on the number of samples
+            trace = (
+                data[qubit, i]["i"][: (len(data[qubit, i]["i"]) // 16) * 16]
+                + 1j * data[qubit, i]["q"][: (len(data[qubit, i]["i"]) // 16) * 16]
+            )
+            traces.append(trace)
```
andrea-pasquale, line 152: Uhm, can this functionality be integrated in qibolab?

andrea-pasquale: Or at least define `16` as a global variable

src/qibocal/protocols/characterization/signal_experiments/calibrate_state_discrimination.py
```diff
@@ -0,0 +1,220 @@
...
+        for our use case where we only want to discirminate between state 0 and 1
+        """
+        # Calculate the optimal kernel
+        kernel = np.conj(traces[0] - traces[1])
+
+        # Normalize the kernel
+        max_abs_weight = max(np.abs(kernel))
+        kernel *= 1 / max_abs_weight
+
+        kernel_state_zero[qubit] = kernel
```
andrea-pasquale, line 164: Here I would like to know more about this. Do you have a reference?

src/qibocal/protocols/characterization/signal_experiments/calibrate_state_discrimination.py
```diff
@@ -0,0 +1,220 @@
...
+                showlegend=True,
+                legendgroup="kernel state 0",
+            ),
+            row=1,
+            col=1,
+        )
+
+        fig.update_layout(
+            showlegend=True,
+            uirevision="0",  # ``uirevision`` allows zooming while live plotting
```
andrea-pasquale, line 203:
```suggestion
```

andrea-pasquale (commented):
Thanks @Jacfomg.
I have a few suggestions, do you have any reports that you can share about this protocol?

src/qibocal/protocols/characterization/signal_experiments/calibrate_state_discrimination.py
Jacfomg: No, I checked laboneq functions computing kernels and the procedure is similar, they also take into account up to 4 traces for multistate discrimination, whereas this works for just for qubits.

src/qibocal/protocols/characterization/signal_experiments/calibrate_state_discrimination.py
Jacfomg: I think it should, but I don't know if it's just a Zurich thing or all the devices behave in the same way.

src/qibocal/auto/operation.py
```diff
@@ -180,38 +202,36 @@ def register_qubit(self, dtype, data_keys, data_dict):
...
+    """Generic runcard update."""
 
-    def save(self, path):
-        """Store results to json."""
-        (path / RESULTSFILE).write_text(json.dumps(serialize(asdict(self))))
+    def __post_init__(self):
+        if "data" not in self.__dict__:
+            self.data: Optional[
+                dict[Union[tuple[QubitId, int], QubitId], npt.NDArray]
+            ] = {}
```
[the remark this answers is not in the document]
Jacfomg: It was related to this https://github.com/qiboteam/qibocal/pull/582#discussion_r1379628350 .

andrea-pasquale (approved):

src/qibocal/protocols/characterization/signal_experiments/calibrate_state_discrimination.py
```diff
@@ -0,0 +1,228 @@
...
+    # sequence_1: RX - MZ
+
+    data = CalibrateStateDiscriminationData(resonator_type=platform.resonator_type)
+
+    # FIXME: Multiplex Raw acquisition may not work properly ???
+    for qubit in qubits:
+        sequence_0 = PulseSequence()
+        sequence_1 = PulseSequence()
+        ro_pulses = {}
+        qd_pulses = {}
```
GabrielePalazzo, line 96: Why do you need to use dictionaries here?

src/qibocal/auto/operation.py
```diff
@@ -98,68 +97,96 @@ def load(cls, input_parameters):
...
+        self._to_json(path, filename)
+        self._to_npz(path, filename)
 
-    def _to_npz(self, path):
+    def _to_npz(self, path: Path, filename: str):
         """Helper function to use np.savez while converting keys into strings."""
-        np.savez(path / DATAFILE, **{json.dumps(i): self.data[i] for i in self.data})
+        if hasattr(self, "data"):
+            np.savez(
+                path / filename,
```
GabrielePalazzo, line 129:
```suggestion
                path / f"{filename}.npz",
```
I think it is better to explicitly sate the extension, even if numpy adds it automatically.

Edoardo-Pedicillo (approved):

src/qibocal/protocols/characterization/signal_experiments/calibrate_state_discrimination.py
andrea-pasquale: `ro_pulses` is usually used to retrieve results from pulse sequence execution. 
`qd_pulses` we can probably drop.

andrea-pasquale: Also they should be declared outside the for loop for sure

Jacfomg: Yes, only qd_pulses would be needed. They could be outside but we don't care about they once we added the data with `.register_qubit` so I think it would be the same

andrea-pasquale:
Thanks @GabrielePalazzo for the review. Your comments have been addressed.

GabrielePalazzo (approved):

PR #595, opened 2023-10-30
Add platform and backend entries to qq commands

### LINKED ISSUE #593: QPU performance bot
We need a performance measurement tool which will extract parameters after execution on QPU for a given platform runcard. This program should store at least T1, T2 and fidelities.

andrea-pasquale:
Closes #593.
This PR adds the possibility to specify the platform name and the backend directly from command line:

```sh
qq auto <runcard> --platform dummy --backend qibolab
```
Currently if those values are not specified from CLI they will be read from the runcard if they are present.
If they are not specified also in the runcard the default values wil be "qibolab" for backend and "dummy" for platform.

I still need to update documentation but in theory this should be ready in terms of implementation.

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`


andrea-pasquale:
I uploaded a first runcard for monitoring without RB.
In order to run it should be sufficient to do:
```sh
qq auto runcards/monitor.yml --platform <name_of_the_platform>
```
It will be executed on qubits available for the platform simultaneously.

If we don't care about the `html` report we could also run something like this
```sh
qq acquire runcards/monitor.yml --platform <name_of_the_platform> -o <output_folder>
qq fit <output_folder> --update
```
which should the same but it will not include the html report.

src/qibocal/auto/runcard.py
```diff
@@ -47,21 +44,9 @@ class Runcard:
     """Structure of an execution runcard."""
 
     actions: list[Action]
-    qubits: Union[list[QubitId], list[tuple[QubitId, QubitId]]]
+    qubits: Union[list[QubitId], list[tuple[QubitId, QubitId]]] = None
```
stavros11, line 47: Strictly speaking here we need
```suggestion
    qubits: Optional[Union[list[QubitId], list[tuple[QubitId, QubitId]]]] = None
```
since it can be `None`.

src/qibocal/cli/_base.py
```diff
@@ -43,15 +44,27 @@ def command():
     help="Use --no-update option to avoid updating iteratively the platform."
     "With this option the new runcard will not be produced.",
 )
-def auto(runcard, folder, force, update):
+@click.option(
+    "--platform",
+    default=None,  # FIXME: choose default value
```
stavros11, line 49: Do we still need this FIXME? I think the `None` case is properly handled in the function.

src/qibocal/cli/_base.py
```diff
@@ -43,15 +44,27 @@ def command():
...
 
     Arguments:
 
      - RUNCARD: runcard with declarative inputs.
     """
-    card = yaml.safe_load(runcard.read_text(encoding="utf-8"))
-    autocalibrate(card, folder, force, update)
+    runcard = Runcard.load(yaml.safe_load(runcard.read_text(encoding="utf-8")))
+    plat = platform if platform is not None else runcard.platform
+    back = backend if backend is not None else runcard.backend
```
stavros11, line 66: I think another way to implement this PR would be
```py
if platform is not None:
    runcard.platform = platform
if backend is not None:
    runcard.backend = backend
```
One advantage of that is that you can leave the `GlobalBackend.set_backend()` manipulations in runcard.py instead of doing them in both acquisition and autocalibration, so the total change of this PR would be less.

Other than that, I do not have strong arguments that this is better than the current implementation, it may be not.

src/qibocal/auto/runcard.py
```diff
@@ -47,21 +44,9 @@ class Runcard:
...
     backend: str = "qibolab"
     platform: str = "dummy"
```
stavros11, line 49: Regarding the default (used when the platform is not specified, either through cli or the runcard), we could also consider using the `"QIBO_PLATFORM"` environment variable, particularly if @scarrazza is planning to use this for slurm.

One way would be to set
```suggestion
    platform: Optional[str] = None
```
and let qibo find the platform from the environment variable (this may need some fix from the qibo side).

Another way would be
```suggestion
    platform: str = os.environ("QIBO_PLATFORM", "dummy")
```
which will use the variable if exists, otherwise falls back to dummy. This is closer to the current behavior.

If we do this we will also be able to submit the same job/runcard on different queues and qibocal will know the proper platform based on the slurm (assuming it is properly set and working).

stavros11 (approved):
Looks good to me, thanks. Some suggestions in the comments, particularly the one about `"QIBO_PLATFORM"` may be useful.

src/qibocal/cli/_base.py
andrea-pasquale: Indeed, I like this approach better compared to mine.

src/qibocal/cli/_base.py
andrea-pasquale: I don't need them anymore I think

src/qibocal/auto/runcard.py
andrea-pasquale: It makes more sense to have `QIBO_PLATFORM` instead of `dummy`.
Is the prologue now working as it should @scarrazza?
If so I believe that I can commit @stavros11 's suggestion.

scarrazza: Yes, expect for queues based on the same virtual node.

scarrazza:
@andrea-pasquale could you please update the lock with the latest qibo release?

andrea-pasquale:
> @andrea-pasquale could you please update the lock with the latest qibo release?

Both qibo and qibolab, right?

scarrazza:
Sure, then we merge and release.

andrea-pasquale:
Lock updated @scarrazza

scarrazza:
Thanks.

scarrazza (approved):

PR #596, opened 2023-10-30
Platform default parameters read from qibolab platforms

### LINKED ISSUE #594: Nshots defaults to None
I thing this may be related to routines like rabi amplitude, which uses nshots to compute stuff, crashing. This should default to the nshots parameters on the platform.

```py

DEFAULT_PARENT_PARAMETERS = {
    "nshots": None,
    "relaxation_time": None,
}
"""Default values of the parameters of `Parameters`"""


```

GabrielePalazzo:
Closes #594 

Parameters not explicitly defined in the qibocal runcard are read from qibolab runcards.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/auto/task.py
+++ b/src/qibocal/auto/task.py
@@ -135,9 +135,9 @@ class Task:
         task_qubits = self._allocate_local_qubits(qubits, platform)
 
         if self.parameters.nshots is None:
-            self.parameters.nshots = platform.settings.nshots
+            self.action.parameters["nshots"] = platform.settings.nshots
         if self.parameters.relaxation_time is None:
-            self.parameters.relaxation_time = platform.settings.relaxation_time
+            self.action.parameters["relaxation_time"] = platform.settings.relaxation_time
 
         try:
             operation: Routine = self.operation
--- a/src/qibocal/auto/task.py
+++ b/src/qibocal/auto/task.py
@@ -19,9 +19,9 @@ from .operation import (
     RESULTSFILE,
     Data,
     DummyPars,
+    Parameters,
     Qubits,
     QubitsPairs,
-    Parameters,
     Results,
     Routine,
     dummy_operation,
@@ -134,12 +134,14 @@ class Task:
     ):
         completed = Completed(self, Normal(), folder)
         task_qubits = self._allocate_local_qubits(qubits, platform)
-        
+
         # get parent and child class data_members
-        parameters_keys = set(self.operation.parameters_type.__dict__["__annotations__"].keys()) | set(Parameters.__dict__["__annotations__"].keys())
+        parameters_keys = set(
+            self.operation.parameters_type.__dict__["__annotations__"].keys()
+        ) | set(Parameters.__dict__["__annotations__"].keys())
         # read all settings defined in platform-specific qibolab runcard
         for setting, value in platform.settings.__dict__.items():
-            # save values non defined in qibocal runcard, 
+            # save values non defined in qibocal runcard,
             # but that are used by the class
             if setting not in self.action.parameters:
                 if setting in parameters_keys:
--- a/src/qibocal/auto/task.py
+++ b/src/qibocal/auto/task.py
@@ -137,7 +137,9 @@ class Task:
         if self.parameters.nshots is None:
             self.action.parameters["nshots"] = platform.settings.nshots
         if self.parameters.relaxation_time is None:
-            self.action.parameters["relaxation_time"] = platform.settings.relaxation_time
+            self.action.parameters[
+                "relaxation_time"
+            ] = platform.settings.relaxation_time
 
         try:
             operation: Routine = self.operation
--- a/src/qibocal/auto/task.py
+++ b/src/qibocal/auto/task.py
@@ -21,6 +21,7 @@ from .operation import (
     DummyPars,
     Qubits,
     QubitsPairs,
+    Parameters,
     Results,
     Routine,
     dummy_operation,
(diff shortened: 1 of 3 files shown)


src/qibocal/auto/task.py
```diff
@@ -134,6 +135,18 @@ def run(
...
+        parameters_keys = set(
+            self.operation.parameters_type.__dict__["__annotations__"].keys()
+        ) | set(Parameters.__dict__["__annotations__"].keys())
+        # read all settings defined in platform-specific qibolab runcard
+        for setting, value in platform.settings.__dict__.items():
+            # save values non defined in qibocal runcard,
+            # but that are used by the class
+            if setting not in self.action.parameters:
+                if setting in parameters_keys:
+                    self.action.parameters[setting] = value
```
andrea-pasquale, line 148: I think that this can be done more easily...
What about this?
```suggestion
        if self.parameters.nshots is None:
            self.parameters.nshots = platform.settings.nshots

        if self.parameters.relaxation_time is None:
            self.parameters.relaxation_time = platform.settings.relaxation_time
```
I tested and it seems to work as expected.
You can also remove the deepcopy that you did before I believe.

andrea-pasquale (commented):
Thanks @GabrielePalazzo for implementing this.
Your solution works but it is a bit convoluted.
I put in a comment another possible solution which seems easier to implement.
Let me know if it works also for you.

src/qibocal/auto/task.py
GabrielePalazzo: My only concern with this solution is that if in the future we add more parameters to `qibolab` runcards, it is going to pass under the radar, since this behaviour is not well tested. 
By the way, I’m not satisfied with the tests I drafted yesterday, so I need to think some way to implement some testing.

andrea-pasquale: Currently I believe that the only common parameters between all the routines are `nshots` and `relaxation_time`.

Jacfomg: For testing, we just need to add some routines in the test runcard settings those fields to none, `rabi_amplitude` would be one that could work as it uses `params.nshots` for the fit.

Edoardo-Pedicillo:
```suggestion
for setting in parameters_keys:
            if setting not in self.action.parameters:
                if hasattr(platform.settings, setting):
                    self.action.parameters[setting]=getattr(platform.settings, setting)

```
I like your solution. I understand that for a few parameters, maybe the @andrea-pasquale approach is better, but this is more general. I suggest not using the dictionaries for better readability.

Edoardo-Pedicillo: I know that maybe it is too late but if @andrea-pasquale agrees I prefer to restore your solution, in this way it is easier to add default parameters. With my suggested changes (they should work) the only convoluted part is the `parameters_keys` definition, I didn't find a better solution for now.

GabrielePalazzo: I agree. I'll commit your suggestion as soon as he replies.

Jacfomg: I would go with @andrea-pasquale solution as it is the one I already implemented in qibolab and I suggested the nshots and relaxation_time as parameters for all routines and I could thing of any other that needs to be shared.

andrea-pasquale: To be more specific this approach will only work with attributes in `platform.settings` which are only 3 at the moment:
https://github.com/qiboteam/qibolab/blob/53737e12aa7400b22e100eabf538fab9a77667d6/src/qibolab/platform.py#L23-L32
And the sampling rate should not be there I believe.

Jacfomg: Also, maybe before discussing which one to use we should get one that passes tests, right ?

GabrielePalazzo: Ok, @andrea-pasquale's solution doesn't work. @Edoardo-Pedicillo's and mine work.
The problem is that `self.parameters()` calls `<SomeChildParameterClass>.load()` which pops elements from its input argument.

I think we should definitely avoid changing the value of function arguments, e.g.:
```
    @classmethod
    def load(cls, input_parameters):
        """Load parameters from runcard.

        Possibly looking into previous steps outputs.
        Parameters defined in Parameters class are removed from `parameters`
        before `cls` is created.
        Then `nshots` and `relaxation_time` are assigned to cls.

        .. todo::

            move the implementation to History, since it is required to resolve
            the linked outputs

        """
        default_parent_parameters = deepcopy(DEFAULT_PARENT_PARAMETERS)
        parameters = deepcopy(input_parameters)
        for parameter, value in default_parent_parameters.items():
            default_parent_parameters[parameter] = parameters.pop(parameter, value)
        instantiated_class = cls(**parameters)
        for parameter, value in default_parent_parameters.items():
            setattr(instantiated_class, parameter, value)
        return instantiated_class
```

GabrielePalazzo: With the change in `Parameters.load()` all solutions pass test locally on dummy, so it's just a matter of how the code is self-explainable and robust to future changes. I'll leave it up to you.

andrea-pasquale:
Thanks for all the updates @GabrielePalazzo.
As soon as you fixed tests we should be ready to merge I think.

GabrielePalazzo:
> Thanks for all the updates @GabrielePalazzo. As soon as you fixed tests we should be ready to merge I think.

Apparently, `test_auto.py` fails because some runcards (e.g.: `vertical.yaml`) do not have `operation` (required by `self.parameters`). Moving the new lines to the `try`/`except` block makes tests pass, but I don't know if you like this solution.

andrea-pasquale:
> Apparently, `test_auto.py` fails because some runcards (e.g.: `vertical.yaml`) do not have `operation` (required by `self.parameters`). Moving the new lines to the `try`/`except` block makes tests pass, but I don't know if you like this solution.
 We are already doing something like this here 
https://github.com/qiboteam/qibocal/blob/578f07b1c20288564fe91f533afe68225de9a499/src/qibocal/auto/task.py#L137-L142C26
So I guess that it is fine.

src/qibocal/auto/task.py
```diff
@@ -135,6 +135,13 @@ def run(
         task_qubits = self._allocate_local_qubits(qubits, platform)
 
         try:
+            if self.parameters.nshots is None:
+                print(self.action)
```
Edoardo-Pedicillo, line 139:
```suggestion
```

Edoardo-Pedicillo (approved):
LGTM

Jacfomg (approved):

andrea-pasquale (approved):

PR #607, opened 2023-11-07
Adding effective qubit temperature to some protocols

### LINKED ISSUE #606: Add effective qubit temperature to `readout_characterization` protocol
As title says.
I believe that this should be useful especially for #534 (eventually we could also add it to the single_shot_classification)

andrea-pasquale:
Closes #606.

@stavros11 perhaps we could consider adding a qubit effective temperature attribute to `Qubit` in qibolab?

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/characterization/utils.py
+++ b/src/qibocal/protocols/characterization/utils.py
@@ -46,7 +46,7 @@ def effective_qubit_temperature(
 
     """
     try:
-        temp = -HBAR * qubit_frequency / np.log(prob_1 / prob_0) / KB
+        temp = -HBAR * qubit_frequency / (np.log(prob_1 / prob_0) * KB)
     except ZeroDivisionError:
         # TODO: choose appropriate value to return
         temp = -1
--- a/src/qibocal/protocols/characterization/utils.py
+++ b/src/qibocal/protocols/characterization/utils.py
@@ -58,12 +58,12 @@ def effective_qubit_temperature(
     error_prob_1 = np.sqrt(prob_1 * (1 - prob_1) / nshots)
     try:
         temp = -HBAR * qubit_frequency / (np.log(prob_1 / prob_0) * KB)
+        dT_dp0 = temp / prob_0 / np.log(prob_1 / prob_0)
+        dT_dp1 = temp / prob_1 / np.log(prob_1 / prob_0)
+        error = np.sqrt((dT_dp0 * error_prob_0) ** 2 + (dT_dp1 * error_prob_1) ** 2)
     except ZeroDivisionError:
-        # TODO: choose appropriate value to return
-        temp = -1
-    dT_dp0 = temp / prob_0 / np.log(prob_1 / prob_0)
-    dT_dp1 = temp / prob_1 / np.log(prob_1 / prob_0)
-    error = np.sqrt((dT_dp0 * error_prob_0) ** 2 + (dT_dp1 * error_prob_1) ** 2)
+        temp = np.nan
+        error = np.nan
     return temp, error
 
 
@@ -425,7 +425,12 @@ def significant_digit(number: float):
             is ``>= 1``, ``= 0`` or ``inf``.
     """
 
-    if np.isinf(np.real(number)) or np.real(number) >= 1 or number == 0:
+    if (
+        np.isinf(np.real(number))
+        or np.real(number) >= 1
+        or number == 0
+        or np.isnan(number)
+    ):
         return -1
 
     position = max(np.ceil(-np.log10(abs(np.real(number)))), -1)
--- a/src/qibocal/protocols/characterization/utils.py
+++ b/src/qibocal/protocols/characterization/utils.py
@@ -36,21 +36,35 @@ COLORBAND_LINE = "rgba(255,255,255,0)"
 
 
 def effective_qubit_temperature(
-    prob_0: np.array, prob_1: np.array, qubit_frequency: float
+    prob_0: np.array, prob_1: np.array, qubit_frequency: float, nshots: int
 ):
     """Calculates the qubit effective temperature.
 
-    The formula used in this one:
+    The formula used is the following one:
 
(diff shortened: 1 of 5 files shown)


andrea-pasquale:
I would like to test on hardware to see if the results make sense but in theory this should be ready for a first review.
I am not sure about my implementation in the single shot classification, could you please have a look whenever you have time @Edoardo-Pedicillo?

src/qibocal/protocols/characterization/utils.py
```diff
@@ -27,11 +28,31 @@
...
+):
+    """Calculates the qubit effective temperature.
+
+    The formula used in this one:
+
+    kB Teff = - hbar qubit_freq / ln(prob_1/prob_0)
+
+    """
+    try:
+        temp = -HBAR * qubit_frequency / np.log(prob_1 / prob_0) / KB
```
GabrielePalazzo, line 49:
```suggestion
        temp = -HBAR * qubit_frequency / (np.log(prob_1 / prob_0) * KB)
```
Could you group all divisions?

andrea-pasquale: Sure

src/qibocal/protocols/characterization/utils.py
```diff
@@ -27,11 +28,31 @@
...
+    The formula used in this one:
+
+    kB Teff = - hbar qubit_freq / ln(prob_1/prob_0)
+
+    """
+    try:
+        temp = -HBAR * qubit_frequency / (np.log(prob_1 / prob_0) * KB)
+    except ZeroDivisionError:
+        # TODO: choose appropriate value to return
+        temp = -1
```
Edoardo-Pedicillo, line 52: Maybe it is better to use np.inf or a really high value if the first option breaks the code. I am not sure but `temp = -1` is a possible value (not physically), if we consider the statistical fluctuations and a chip not calibrated.

Edoardo-Pedicillo (commented):
LGTM, only one small comment. Let's wait the test on hardware to merge.

src/qibocal/protocols/characterization/utils.py
andrea-pasquale: Yeah I also was not sure about what to put here.
I will put a random high value, I don't really trust `np.inf` for these applications.

stavros11:
> @stavros11 perhaps we could consider adding a qubit effective temperature attribute to `Qubit` in qibolab?

Fine with me as it could also be interesting to report in the monitoring tool. I am just not sure how stable this is with time, but in the end it only depends on the state populations and qubit frequency, so it is probably as stable as the other things we are reporting. Feel free to open a PR in qibolab to add tihs.

andrea-pasquale:
I manage to test on hardware: http://login.qrccluster.com:9000/KFl3IYM3Rm-TJpc8ySmvMw==/
After fixing the constants it seems to work.
There is only a small issue: with an uncalibrated qubit the population of 1 might exceed the population of 0s resulting in a negative temperature...
How do we want to handle this case @Edoardo-Pedicillo @GabrielePalazzo?
We could keep it to show that the chip is uncalibrated. Or we should find a way to handle this... maybe with #617

Edoardo-Pedicillo:
Since it happens when the chip is not calibrated, it is not a big problem. I suggest evaluating the uncertainties of the probabilities and propagate the error to the temperature.

src/qibocal/protocols/characterization/utils.py
andrea-pasquale: I decided to put `np.nan`

src/qibocal/protocols/characterization/utils.py
```diff
@@ -27,11 +28,45 @@
...
+        temp (float): effective temperature
+        error (float): error on effective temperature
+
+    """
+    error_prob_0 = np.sqrt(prob_0 * (1 - prob_0) / nshots)
+    error_prob_1 = np.sqrt(prob_1 * (1 - prob_1) / nshots)
+    try:
+        temp = -HBAR * qubit_frequency / (np.log(prob_1 / prob_0) * KB)
+        dT_dp0 = temp / prob_0 / np.log(prob_1 / prob_0)
+        dT_dp1 = temp / prob_1 / np.log(prob_1 / prob_0)
```
GabrielePalazzo, line 62: Shouldn't there be a minus sign in `dT_dp1`? It doesn't affect the error, since it is squared.
Could you also reduce the number of division operations?

src/qibocal/protocols/characterization/utils.py
```diff
@@ -327,16 +362,33 @@ def round_report(
...
 
 
+def format_error_single_cell(measure: tuple):
+    """Helper function to print mean value and error in one line."""
+    # extract mean value and error
+    mean = measure[0][0]
+    error = measure[1][0]
+    if all("e" in number for number in measure[0] + measure[1]):
+        magn = mean.split("e")[1]
+        return f"({mean.split('e')[0]} ± {error.split('e')[0]}) e{magn}"
```
GabrielePalazzo, line 388: If this is meant to be used only in html code, you can also use this.
```suggestion
        return f"({mean.split('e')[0]} ± {error.split('e')[0]})*10<sup>{magn}</sup>"
```

src/qibocal/protocols/characterization/utils.py
andrea-pasquale: In theory yes, but as you said it is squared. For completeness I will add the minus sign.
About divisions... personally I prefer this approach compared to doing something like ` a / (b*c*d)` purely based on the fact that the number of char are less.

src/qibocal/protocols/characterization/utils.py
andrea-pasquale: I think this line is not rendered in HTML but I am not sure. @Edoardo-Pedicillo?

GabrielePalazzo (approved):

Edoardo-Pedicillo (approved):

PR #610, opened 2023-11-07
Sequence unrolling in standard RB

stavros11:
Requires qiboteam/qibolab#618. I temporarily modified the pyproject.toml to include this dependency.

I added the unrolling as an option as it may be useful to benchmark with and without.
I only tested locally with the dummy. I am planning to test and also benchmark with instruments when they are available. So far this is expected to work well only with qblox.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `unrolling`


andrea-pasquale:
@stavros11 do you think that we can have this for the 0.0.7? Ideally I believe that release should happen at the beginning of january

stavros11:
I fixed the conflicts, this should be ready.

andrea-pasquale (approved):
Thanks @stavros11, LGTM.
I will try it on hardware as soon as we have a calibrated chip.
You kept the `deploy on hardware` label because you still need to run on it on hardware or because you have not been able to run in on a calibrated chip?

stavros11:
> You kept the `deploy on hardware` label because you still need to run on it on hardware or because you have not been able to run in on a calibrated chip?

I tried it with qw5q_gold and it works but the results look a bit different with and without unrolling:
* unrolling true: http://login.qrccluster.com:9000/0slZJRS1STu4uspfS4_ZjA==
* unrolling false: http://login.qrccluster.com:9000/jAv6Sh-9REygVy17AjtdyQ==

so I would prefer to also try with a calibrated qubit which has a reasonable (not just noise) RB using the main branch, to make sure it is consistent here.

Edoardo-Pedicillo (approved):

Issue #508, opened 2023-09-02
Plugin API


AleCandido:
_Originally posted by @AleCandido in https://github.com/qiboteam/qibocal/pull/406#pullrequestreview-1608320306_

`protocols/` are a fundamental part of Qibocal, but at the same time they are just separately developed components, that should adhere to a given specification.

For this reason, Qibocal could be simply extended even with protocols written outside of it and registered at runtime via a given API.

To treat them on the same ground, we could even distribute protocols together with Qibocal, but in a separate package (and automatically registering them, e.g. telling the CLI to import that package by default, unless an explicit option is given, or something like that).

andrea-pasquale:
We should have a similar mechanism for validators.
_Originally posted by @AleCandido in https://github.com/qiboteam/qibocal/pull/617#discussion_r1391564165_

PR #617, opened 2023-11-09
Implementing validation using Chi squared

andrea-pasquale:
First attempt at introducing a validation in Qibocal through chi2.
Here is a possible runcard:

```yaml
platform: dummy

qubits: [0,1]


actions:

  - id: ramsey
    priority: 0
    operation: ramsey
    validator:
      scheme: chi2
      parameters:
        chi2_max_value: 0.1
    parameters:
      delay_between_pulses_start: 16 # must be a multiple of 4 incl 0
      delay_between_pulses_end: 1000
      delay_between_pulses_step: 4 # must be a multiple of 4
      n_osc: 10
```
For now I am just planning to add the `chi2` attribute in all protocols that implements errors.
Currently I believe that this works for `ramsey`, `rabi amplitude` and `rabi length`.
I tried to make it already quite validation scheme agnostic in case we want to add other validation schemes.
I am still not sure about how to handle the `Broken` status case. For now the easiest possibility will be just to fail
(which is not implemented in the code yet). We could add a handler which should specify how to act in the `Broken` solution. 
Also everything now should be intended only for a single qubit.
The PR is still missing tests and documentation which I will add once we agree on what to do.
Suggestions are welcomed :)

TODOs:

- [x] add chi2 protocols with errors
- [x] define exit condition with `Broken()` status

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`

### DIFF
--- a/doc/source/getting-started/validation.rst
+++ b/doc/source/getting-started/validation.rst
@@ -1,0 +1,52 @@
+How to add validation to your protocol?
+=======================================
+
+In Qibocal there is the possibility to add a validation step
+during the execution of your protocols.
+Here is an example of a runcard which validates the results through
+:math:`\chi^2`.
+
+.. code-block:: yaml
+
+   actions:
+    - id: t1
+      priority: 0
+      operation: t1
+      validator:
+        scheme: chi2
+        parameters:
+            chi2_max_value: 1
+      parameters:
+        ...
+
+The execution will be interrupted in this case if the :math:`\chi^2` exceed
+`chi_max_value`.
+
+Adding a custom validator at runtime
+------------------------------------
+
+.. TO COMPLETE
+The user is also able to define a custom validator through the python API.
+For example let's suppose that we want to write a validator that implements
+the following condition: "stop if T1 is less than 10us".
+
+.. code-block:: python
+
+   from qibocal.protocols.characterization import Operation
+   from qibolab.qubits import QubitId
+   from qibocal.auto.status import Normal, Failure
+   from qibocal.auto.validators import VALIDATORS
+
+   # retrieve t1 object
+   t1 = Operation["t1"].value
+
+   def t1_threshold(results: t1.results_type, qubit: QubitId, t1_min: float = 10e4):
+        # store T1 value for qubit
+        t1 = results.t1[qubit][0]
+
+        if t1 > t1_min:
+            return Normal()
+        else:
+            return Failure()
+
+    VALIDATORS["t1_threshold"] = t1_threshold
--- a/doc/source/getting-started/validation.rst
+++ b/doc/source/getting-started/validation.rst
@@ -21,32 +21,3 @@ Here is an example of a runcard which validates the results through
 
 The execution will be interrupted in this case if the :math:`\chi^2` exceeds
 `chi_max_value`.
-
-Adding a custom validator at runtime
-------------------------------------
-
-.. TO COMPLETE
-The user is also able to define a custom validator through the python API.
-For example let's suppose that we want to write a validator that implements
-the following condition: "stop if T1 is less than 10us".
-
-.. code-block:: python
-
-   from qibocal.protocols.characterization import Operation
-   from qibolab.qubits import QubitId
-   from qibocal.auto.status import Normal, Failure
-   from qibocal.auto.validators import VALIDATORS
-
-   # retrieve t1 object
-   t1 = Operation["t1"].value
-
-   def t1_threshold(results: t1.results_type, qubit: QubitId, t1_min: float = 10e4):
-        # store T1 value for qubit
-        t1 = results.t1[qubit][0]
-
-        if t1 > t1_min:
-            return Normal()
-        else:
-            return Failure()
-
-    VALIDATORS["t1_threshold"] = t1_threshold
--- a/doc/source/getting-started/validation.rst
+++ b/doc/source/getting-started/validation.rst
@@ -19,7 +19,7 @@ Here is an example of a runcard which validates the results through
       parameters:
         ...
 
-The execution will be interrupted in this case if the :math:`\chi^2` exceed
+The execution will be interrupted in this case if the :math:`\chi^2` exceeds
 `chi_max_value`.
 
 Adding a custom validator at runtime
--- a/src/qibocal/auto/execute.py
+++ b/src/qibocal/auto/execute.py
@@ -10,6 +10,7 @@ from qibocal.config import log
 from .graph import Graph
 from .history import History
 from .runcard import Id, Runcard
+from .status import Broken
 from .task import Qubits, Task
 
 
@@ -140,6 +141,10 @@ class Executor:
                 mode=mode,
             )
             self.history.push(completed)
+            if isinstance(completed.status, Broken) and mode.name == "autocalibration":
+                log.warning("Stopping execution due to error in validation.")
+                yield task.uid
+                break
             self.head = self.next()
             update = self.update and task.update
             if (
--- a/src/qibocal/auto/execute.py
+++ b/src/qibocal/auto/execute.py
@@ -10,7 +10,7 @@ from qibocal.config import log
 from .graph import Graph
 from .history import History
 from .runcard import Id, Runcard
-from .status import Broken
+from .status import Failure
 from .task import Qubits, Task
 
 
@@ -141,7 +141,7 @@ class Executor:
                 mode=mode,
             )
             self.history.push(completed)
-            if isinstance(completed.status, Broken) and mode.name == "autocalibration":
+            if isinstance(completed.status, Failure) and mode.name == "autocalibration":
                 log.warning("Stopping execution due to error in validation.")
                 yield task.uid
                 break
--- a/src/qibocal/auto/task.py
+++ b/src/qibocal/auto/task.py
@@ -27,7 +27,7 @@ from .operation import (
 )
 from .runcard import Action, Id
 from .status import Normal, Status
-from .validation import ValidationSchemes
+from .validation import ValidationSchemes, Validator
 
 MAX_PRIORITY = int(1e9)
 """A number bigger than whatever will be manually typed. But not so insanely big not to fit in a native integer."""
@@ -88,9 +88,9 @@ class Task:
         return Operation[self.action.operation].value
 
     @property
-    def validator(self):
+    def validator(self) -> Validator:
         """Validator object from ValidationSchemes Enum"""
-        if self.action.update is None:
+        if self.action.validator is None:
             return None
         return ValidationSchemes[self.action.validator["scheme"]].value.load(
             self.action.validator
@@ -100,7 +100,7 @@ class Task:
         if self.validator is not None:
             status = {}
             for qubit in self.qubits:
-                status[qubit] = self.validator(results, qubit)
+                status[qubit] = self.validator.__call__(results, qubit)
 
     @property
     def main(self):
@@ -175,7 +175,8 @@ class Task:
 
         if mode.name in ["autocalibration", "fit"]:
             completed.results, completed.results_time = operation.fit(completed.data)
-            self.validate(completed.results)
+            if self.validator is not None:
+                self.validate(completed.results)
 
         return completed
 
--- a/src/qibocal/auto/task.py
+++ b/src/qibocal/auto/task.py
@@ -86,16 +86,19 @@ class Task:
 
         return Operation[self.action.operation].value
 
-    def validate(self, results: Results):
-        if self.action.validator is not None:
-            status = {}
-            for qubit in self.qubits:
-                status[qubit] = self.action.validator.__call__(results, qubit)
-            # exit if any of the qubit state is Failure
-            if any(isinstance(stat, Failure) for stat in status.values()):
-                return Failure()
-            else:
-                return Normal()
+    def validate(self, results: Results) -> Optional[Status]:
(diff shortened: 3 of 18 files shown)


src/qibocal/auto/task.py
```diff
@@ -86,6 +87,21 @@ def operation(self):
...
         return Operation[self.action.operation].value
 
+    @property
+    def validator(self) -> Validator:
+        """Validator object from ValidationSchemes Enum"""
+        if self.action.validator is None:
+            return None
+        return ValidationSchemes[self.action.validator["scheme"]].value.load(
+            self.action.validator
+        )
```
AleCandido, line 97: If the `scheme` attribute is required, make `validator` already a `dataclass`, and deserialize `scheme` in it.

Moreover, this procedure could also be offloaded to the object in which `self.action.validator` is deserialized into, and leave here
```suggestion
        return self.action.validator.load()
```

AleCandido: Notice that you already have such an object, is exactly the `Validator` class - I'm just suggesting to make it a `pydantic.dataclasses.dataclass` and deserialize in it the possible parameters as well.

Actually, you would not even need to `load` in that case, and you would not even need the `property` at all, just return the `Validator` instance created during the runcard deserialization.

AleCandido (commented):
Essentially, I'm proposing to move part of the code you added to the task to the deserialization phase.

src/qibocal/auto/task.py
andrea-pasquale: Good point. In fact I called it `validator` before and `pydantic` complained saying that they have already a validator...

andrea-pasquale: One quick question. I tried to implement the deserialization in `Runcard` and it almost works....
The only problem is that I would like to deserialize depending on the `scheme`. For example in this case if I pass `scheme: chi2` I would like to deserialize using `chi2` in `ValidationSchemes` which will be mapped to `Chi2Validator`. Right now I am only able to deserialize to the abstract class `Validator`, which I believe it is what I should put in `Runcard`.

https://github.com/qiboteam/qibocal/blob/dcc5abdc5108cad4ecf5df01f481bfb0e11df484/src/qibocal/auto/validation.py#L61-L64

I was looking for a way to deserialize like this on poetry but I could not find it.
Have you already encountered this problem @AleCandido?

AleCandido: I guess with "Poetry" you meant "Pydantic". I believe there are two ways of doing it:

1. attempt to deserialize a `Union`
2. implement custom deserialization in `Validator`, e.g. using a [pre/post hook](https://docs.pydantic.dev/latest/concepts/dataclasses/#initialization-hooks) (or a combination of the two)

I see problems in both, in the sense that no one is a perfect solution out of the box. 
The `Union` would work perfectly if you do not allow for arbitrary parameters, and distinguish the various validators based on their parameters (rather than based on a tag). But this is not always sensible, since you might want to have different validators logic with the same parameters.
The custom deserialization is slightly annoying, because you have to write more custom logic, and it seems there are no hooks in which you get the parameters and output the object (at least for dataclasses), so you should first let Pydantic construct the `Validator` object (base class), and then create a new object of the given type in the post hook.

The problem is that, in this case, your type is intrinsically a `Union` (not an `Enum`, that is a single type, but different values, well [supported](https://docs.pydantic.dev/latest/api/standard_library_types/#enum) by Pydantic), but since you're discriminating them on a value (not the key, but the value in the dictionary) this should be done *at runtime* (so it can't be inferred from the type).

AleCandido: Actually, an alternative would be to give up on the different subclasses, and have a single class `Validator`, while the logic is delegated to functions (so you would only pick the function dynamically, and Pydantic can safely construct the `Validator` object, that will later look up for the function).

Same with classes, of course, if you give up on subclassing, and use composition instead (so you will later load in the validator the validation logic as an attribute).
But if functions are already working, they are just simpler.

andrea-pasquale: > I guess with "Poetry" you meant "Pydantic".

Yes, indeed I meant "Pydantic"...

Thanks for detailed summary.

> Actually, an alternative would be to give up on the different subclasses, and have a single class `Validator`, while the logic is delegated to functions (so you would only pick the function dynamically, and Pydantic can safely construct the `Validator` object, that will later look up for the function).

I guess this will be the easiest option. I don't really feel like spending too much time trying to implement 1. or 2.
Eventually if we see that deserializing with Pydantic is too complicated we can always fall back to the present solution already implemented in this PR.

andrea-pasquale: I manage to implement the approach you suggested. Here is the updated runcard:

```yaml
platform: dummy

qubits: [0,1]


actions:

  - id: ramsey
    priority: 0
    operation: ramsey
    validator:
      scheme: chi2
      parameters:
        chi2_max_value: 0.1
    parameters:
      delay_between_pulses_start: 16 # must be a multiple of 4 incl 0
      delay_between_pulses_end: 1000
      delay_between_pulses_step: 4 # must be a multiple of 4
      n_osc: 10
```

AleCandido: It looks great :D

andrea-pasquale: Also this validator now only works for results. With this we should be able to also have validators for data directly. For example we could test whether we are seeing only noise (for example using some SNR).
The missing step would be to define some sort of handler, but we can keep it for another PR.

AleCandido: > With this approach we should be able to also have validators for data directly.

In principle yes, but this could be done even during the `fit` call, if relevant for the routine, so (for the time being) I would avoid it.
It is true that we could make the single routines more composable, creating some building blocks that could be added (as the SNR). But keep it on the side, and let's try to converge on the exceptional flow first.

src/qibocal/auto/validation.py
```diff
@@ -0,0 +1,26 @@
...
+from typing import NewType, Optional, Union
+
+from qibolab.qubits import QubitId, QubitPairId
+
+from . import validators
+from .operation import Results
+from .status import Status
+
+ValidatorId = NewType("ValidatorId", str)
+"""Identifier for validator object."""
```
AleCandido, line 12: For the time being, I'd say that is fine like this.

However, just be aware of the limitations: here we rely on all the validators to be in the same module.

We could instead create an `Enum` (or even just a `list`) as we are doing for the routines.
(in particular, for the plugin system #508, it would be interesting to give the user the ability to extend the set of validators, that is even more interesting than implementing new routines, that I expect to occur seldom - so either we can extend the `Enum` dynamically, or we should use a `list`)

AleCandido: Just checked, it's not possible with standard library, only with third-party `Enum` classes.

I'm inclined to move towards a list/dictionary...
https://stackoverflow.com/a/33680929/8653979

src/qibocal/auto/execute.py
```diff
@@ -10,6 +10,7 @@
 from .graph import Graph
 from .history import History
 from .runcard import Id, Runcard
+from .status import Broken
```
AleCandido, line 13: My fault (most likely) but maybe `Failure` (or something like) is more appropriate than `Broken`...

src/qibocal/auto/validation.py
andrea-pasquale: I moved everything to a dictionary. I can also implement something a bit more fancy to add new validators, but I guess that what we have now should be enough to be able to add validators.

src/qibocal/auto/execute.py
andrea-pasquale: Also `Broken` could be acceptable but `Failure` is better for sure.

andrea-pasquale:
This should be ready to review.
I'm just planning to improve a bit the documentation.

doc/source/getting-started/validation.rst
```diff
@@ -0,0 +1,52 @@
...
+   def t1_threshold(results: t1.results_type, qubit: QubitId, t1_min: float = 10e4):
+        # store T1 value for qubit
+        t1 = results.t1[qubit][0]
+
+        if t1 > t1_min:
+            return Normal()
+        else:
+            return Failure()
+
+    VALIDATORS["t1_threshold"] = t1_threshold
```
AleCandido, line 52: Unfortunately, this would not work for the CLI, since the Python file would never be executed.
That's why we need a mechanism to *register* "plugins" (e.g. even just a list of paths of scripts to be executed before running).

AleCandido: It should not be hard to implement a minimal (and temporary) version.

Just define a list top-level, and then you can just dynamically import the scripts in there:
```py
import importlib.util
import sys

for n, plugin in enumerate(plugins):
	spec = importlib.util.spec_from_file_location(f"qibolab_plugin_{n}", plugin)
	module = importlib.util.module_from_spec(spec)
	sys.modules[module_name] = module
	spec.loader.exec_module(module)
```
and that's it.

https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly

AleCandido: If you want to avoid the arbitrary module name you could ask the user for a dictionary (but I would mangle, i.e. prefix, the same with `qibolab_plugin_` in order to avoid collisions)

doc/source/getting-started/validation.rst
```diff
@@ -0,0 +1,52 @@
...
+      priority: 0
+      operation: t1
+      validator:
+        scheme: chi2
+        parameters:
+            chi2_max_value: 1
+      parameters:
+        ...
+
+The execution will be interrupted in this case if the :math:`\chi^2` exceed
```
GabrielePalazzo, line 22:
```suggestion
The execution will be interrupted in this case if the :math:`\chi^2` exceeds
```

src/qibocal/auto/task.py
```diff
@@ -86,6 +86,17 @@ def operation(self):
 
         return Operation[self.action.operation].value
 
+    def validate(self, results: Results):
+        if self.action.validator is not None:
```
GabrielePalazzo, line 90: What do you return if `validator` is None?

src/qibocal/auto/validators/chi2.py
```diff
@@ -0,0 +1,37 @@
...
+from qibocal.config import raise_error
+
+from ..operation import Results
+from ..status import Failure, Normal
+
+CHI2_MAX = 0.05
+"""Max value for accepting fit result."""
+
+
+def chi2(
```
GabrielePalazzo, line 15: Shall we rename it with a more specific name? I mean, from a chi2 function I would expect to get the chi2, i.e.: a `float` and not a `Status`.
Anyway, this is probably not a priority, since in all other modules we are going to call `VALIDATORS` and this function name is not going to exit this folder.

src/qibocal/protocols/characterization/coherence/t2.py
```diff
@@ -168,7 +184,15 @@ def _plot(data: T2Data, qubit, fit: T2Results = None):
...
-            table_dict(qubit, ["T2 [ns]"], [fit.t2[qubit]], display_error=True)
+            table_dict(
+                qubit,
+                [
+                    "T2 [ns]",
+                    "chi2 reduced",
+                ],
+                [fit.t2[qubit], fit.chi2[qubit]],
+                display_error=True,
+            )
```
GabrielePalazzo, line 195: I spotted a `"Time (ns)"` a few lines below, could you change it to `"Time [ns]"`?

src/qibocal/auto/validators/chi2.py
andrea-pasquale: > Anyway, this is probably not a priority, since in all other modules we are going to call `VALIDATORS` and this function name is not going to exit this folder.

Yes, eventually we could rename this function `check_chi2` and then write the dict as `{"chi2" : check_chi2}`

src/qibocal/auto/task.py
AleCandido: In practice `None`, because it's the default output of any Python function when it does not return.

A better layout might be:
```suggestion
        if self.action.validator is None:
            return None
```
i.e. an early return. And deindent the following.

If you'd annotate the return value as `Option[...]` it would even be beneficial.

GabrielePalazzo: I agree.

andrea-pasquale: Fixed

doc/source/getting-started/validation.rst
AleCandido: Code-wise, it is more or less the exact same.

But it is more reproducible within the config file, since it is easier to trace all the elements if they are in a single place (otherwise you'll find yourself asking someone "but do you have that variable set?" while it's easier if they should just post a single file)

andrea-pasquale: At this point I prefer to move this directly #508, I will add a comment about validators.

AleCandido (approved):

GabrielePalazzo (approved):

tests/test_validators.py
```diff
@@ -0,0 +1,64 @@
...
+        {
+            "id": "single shot",
+            "priority": 0,
+            "operation": "single_shot_classification",
+            "parameters": {
+                "nshots": 10,
+            },
+        },
+    ],
+}
```
Edoardo-Pedicillo, line 42: Can `DUMMY_CARD` be a runcard in tests/platform ?

PR #629, opened 2023-11-14
Sequence unrolling in allXY

stavros11:
Similar to #610 but for allXY. Also requires https://github.com/qiboteam/qibolab/pull/618 but I did not change the pyproject since we will wait for the qibolab release anyway.

@GabrielePalazzo @Jacfomg if you have calibrated/used iqm5q recently could you please check if this works for you? In principle you don't need calibrated qubits, we just need to confirm that the behavior is the same with `unrolling: True` and `unrolling: False`. On qw5q_gold I get the same behavior and execution time drops from 65 to 11 seconds.

Note that here I changed `np.array([z_proj])` back to `np.array(z_proj)` so that it works for the dummy and Zurich that return the probability as array. For qblox you need to change back to `np.array([z_proj])`. This is regardless of whether you use unrolling or no (see qiboteam/qibolab#647).

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `unrolling`


Jacfomg (approved):
We can't test the same behavior with zurich until we fix the sequences on zurich. But, since the unrolling is now in qibolab we can merge, right ?

tests/runcards/protocols.yml
```diff
@@ -432,6 +432,13 @@ actions:
       beta_param: null
       nshots: 10
 
+  - id: allXY
```
andrea-pasquale, line 435:
```suggestion
  - id: allXY unrolling
```
I would suggest this change just to have a better understanding when running tests

andrea-pasquale (approved):
Thanks @stavros11.
I confirm that it works as expected both on qblox and zh.

As a small note in would be nice to propagate this option also for the `allxy_drag_pulse_tuning` protocol. We can also do it in a separate PR and not here if you prefer not do it now.

stavros11:
Thank you for the reviews.

> As a small note it would be nice to propagate this option also for the `allxy_drag_pulse_tuning` protocol. We can also do it in a separate PR and not here if you prefer not do it now.

Unless this routine is used a lot and we would really benefit by doing unrolling (which I don't think so), I would prefer to postpone because after some discussions with @AleCandido we may change how the results are returned by `platform.execute_pulse_sequences` and this may break this implementation. This probably won't happen before qibolab 0.1.6 (or later), but still better having to update one place than many.

PR #630, opened 2023-11-15
Changed the lorentzian fit library to curve_fit

GabrielePalazzo:
Parameters obtained with scipy's `curve_fit` should be very similar to the ones from `lmfit`.
`curve_fit` seems slightly faster (as expected).

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/characterization/utils.py
+++ b/src/qibocal/protocols/characterization/utils.py
@@ -2,7 +2,6 @@ from colorsys import hls_to_rgb
 from enum import Enum
 from typing import Union
 
-import lmfit
 import numpy as np
 import numpy.typing as npt
 import pandas as pd
@@ -10,6 +9,7 @@ import plotly.graph_objects as go
 from numba import njit
 from plotly.subplots import make_subplots
 from qibolab.qubits import QubitId
+from scipy.optimize import curve_fit
 from scipy.stats import mode
 
 from qibocal.auto.operation import Data, Results
@@ -64,7 +64,6 @@ def lorentzian(frequency, amplitude, center, sigma, offset):
 def lorentzian_fit(data, resonator_type=None, fit=None):
     frequencies = data.freq * HZ_TO_GHZ
     voltages = data.signal
-    model_Q = lmfit.Model(lorentzian)
 
     # Guess parameters for Lorentzian max or min
     # TODO: probably this is not working on HW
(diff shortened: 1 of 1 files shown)


src/qibocal/protocols/characterization/utils.py
```diff
@@ -87,25 +86,28 @@ def lorentzian_fit(data, resonator_type=None, fit=None):
...
+    }
     # fit the model with the data and guessed parameters
     try:
-        fit_res = model_Q.fit(
-            data=voltages, frequency=frequencies, params=guess_parameters
+        fit_parameters, _ = curve_fit(
+            lorentzian,
+            frequencies,
+            voltages,
+            p0=list(model_parameters.values()),
```
GabrielePalazzo, line 101: Here I'm converting the `dict` to a `list` and a few lines below I basically re-convert it back into a `dict`, in order to keep compatibility with the previous implementation.
The fastest way probably is to use just lists, but I'm not sure if it impacts code readability somehow.

Edoardo-Pedicillo: > The fastest way probably is to use just lists, but I'm not sure if it impacts code readability somehow.

Use only lists

andrea-pasquale: @GabrielePalazzo as you fix this comment and conflicts we should be able to merge this.

src/qibocal/protocols/characterization/utils.py
```diff
@@ -87,25 +86,25 @@ def lorentzian_fit(data, resonator_type=None, fit=None):
...
+        model_parameters, _ = curve_fit(
+            lorentzian,
+            frequencies,
+            voltages,
+            p0=model_parameters,
         )
-        # get the values for postprocessing and for legend.
-        return fit_res.best_values["center"], fit_res.best_values
 
     except:
```
GabrielePalazzo, line 104: Last comment. I'm not in favor of catching all exceptions. Should we catch more specific exceptions (e.g. `RuntimeError`)?

andrea-pasquale: Sure, we can give it a try with the current CI.

AleCandido:
Even if it were not faster at all (but not much worse) this would have been an improvement anyhow, since it reduces the number of deps (and SciPy is definitely more reliable).

andrea-pasquale (approved):
Thanks @GabrielePalazzo for implementing this.
LGTM.

src/qibocal/protocols/characterization/utils.py
GabrielePalazzo: I ran some tests and I can only get a `RuntimeError`. Let me know if it also works for you.

Edoardo-Pedicillo:
@GabrielePalazzo when the conflicts are fixed, we are ready for a review.

Edoardo-Pedicillo (approved):

Issue #631, opened 2023-11-15
Calibration issues


Edoardo-Pedicillo:
This issue collects some bugs encountered while calibrating:

~- [ ] Rabi fit unstable: the curve starts to increase slowly and the fit is not able to handle this.~

![image](https://github.com/qiboteam/qibocal/assets/62623482/99c3cfd6-0534-4262-9ea3-61a5a3176860)

 - [x] qq fit not working:
```
qq[Qibo 0.2.2|ERROR|2023-11-15 16:27:41]: Platform IQM5q does not exist.
Traceback (most recent call last):
  File "/nfs/users/edoardo.pedicillo/routine/bin/qq", line 8, in <module>
    sys.exit(command())
  File "/nfs/users/edoardo.pedicillo/routine/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/nfs/users/edoardo.pedicillo/routine/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/nfs/users/edoardo.pedicillo/routine/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/nfs/users/edoardo.pedicillo/routine/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/nfs/users/edoardo.pedicillo/routine/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/nfs/users/edoardo.pedicillo/qibocal/src/qibocal/cli/_base.py", line 150, in fit
    fitting(folder, update)
  File "/nfs/users/edoardo.pedicillo/qibocal/src/qibocal/cli/fit.py", line 29, in fit
    GlobalBackend.set_backend(backend=meta["backend"], platform=meta["platform"])
  File "/nfs/users/edoardo.pedicillo/routine/lib/python3.10/site-packages/qibo/backends/__init__.py", line 84, in set_backend
    cls._instance = construct_backend(backend, platform, runcard)
  File "/nfs/users/edoardo.pedicillo/routine/lib/python3.10/site-packages/qibo/backends/__init__.py", line 35, in construct_backend
    return QibolabBackend(platform, runcard)
[...]

PR #632, opened 2023-11-16
Fix operational modes

Edoardo-Pedicillo:
This PR fixes the operational modes issues in #631.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


src/qibocal/protocols/characterization/flux_dependence/avoided_crossing.py
```diff
@@ -177,106 +179,109 @@ def _plot(data: AvoidedCrossingData, fit: AvoidedCrossingResults, qubit):
             row=1,
             col=i + 1,
         )
+        if fit:
```
AleCandido, line 182: Maybe
```suggestion
        if fit is not None:
```
?

(and in case, the parameter's type should be updated with `Optional[...]`)

src/qibocal/cli/report.py
```diff
@@ -40,8 +40,7 @@ def report(path):
     qubits = create_qubits_dict(qubits=runcard.qubits, platform=platform)
 
     # load executor
-    executor = Executor.load(runcard, path, qubits=qubits)
-
+    executor = Executor.load(runcard, path, qubits=qubits, platform=platform)
```
AleCandido, line 43: I'm not sure why `platform` should be required for the report.

I guess it's not working otherwise, but maybe we could fix it in another way. Or do we really need some information from the platform?

src/qibocal/protocols/characterization/flux_dependence/avoided_crossing.py
```diff
@@ -117,22 +118,23 @@ def _fit(data: AvoidedCrossingData) -> AvoidedCrossingResults:
...
         qubit_pair = tuple(qubit_pair)
+        fits[qubit_pair] = {}
         low = qubit_pair[0]
         high = qubit_pair[1]
         # Fit the 02*2 curve
         curve_02 = np.array(curves[high, "02"]) * 2
         x_02 = np.unique(qubit_data[high, "02"]["bias"])
         fit_02 = np.polyfit(x_02, curve_02, 2)
-        fits[qubit_pair, "02"] = fit_02.tolist()
+        fits[qubit_pair]["02"] = fit_02.tolist()
```
AleCandido, line 131: We might want to use a better data structure than a string to index this information.

Maybe there is no need of making something fully general (using tuples and integers), since the options will always be these.
But exactly because we have a limited amount of options, a simple enum could do the job.

src/qibocal/protocols/characterization/flux_dependence/avoided_crossing.py
```diff
@@ -177,106 +180,109 @@ def _plot(data: AvoidedCrossingData, fit: AvoidedCrossingResults, qubit):
...
+                    name=f"Parabola {transition}",
+                ),
+                row=1,
+                col=i + 1,
+            )
+    figures.append(heatmaps)
 
-        # the fit of the parabola in 02 transition was done doubling the frequencies
-        heatmaps.add_trace(
+    if fit:
```
AleCandido, line 215:
```suggestion
    if fit is not None:
```

src/qibocal/protocols/characterization/flux_dependence/avoided_crossing.py
```diff
@@ -177,106 +180,109 @@ def _plot(data: AvoidedCrossingData, fit: AvoidedCrossingResults, qubit):
...
-            qubit,
-            ["CZ bias", "iSwap bias"],
-            [np.round(cz[:, 0], 3), np.round(iswap[:, 0], 3)],
+        figures.append(parabolas)
+        fitting_report = table_html(
+            table_dict(
+                qubit,
+                ["CZ bias", "iSwap bias"],
+                [np.round(cz[:, 0], 3), np.round(iswap[:, 0], 3)],
+            )
```
AleCandido, line 284: Please, try to split each plot to its own function. At least, it will be easier to track which information is consumed in each one (and easier to read in general)

src/qibocal/cli/report.py
andrea-pasquale: Good point. We should look into that. I think we should be able to remove it.

src/qibocal/protocols/characterization/flux_dependence/avoided_crossing.py
Edoardo-Pedicillo: I tried to use the enum, I am not sure that I did what you expected.

andrea-pasquale (approved):

AleCandido (approved):

PR #642, opened 2023-11-24
Fixing punchout plots

### LINKED ISSUE #641: Punchout plot does not appear
I think something in the punchout plot was broken in one of the latest merges (possibly #627, but I am not sure at all). When I run with `qq auto` the plot does not appear in the report, even though no error raised. When I run with `qq acquire` + `qq report` the plot appears normally. So maybe the fitting breaks something?

andrea-pasquale:
Closes #641.
Thanks for spotting this @stavros11. There was a factor 1e9 wrong for the fitted point.
Now I've fixed it and I've also added the possibility to toggle the fitted point.
Let me know if this works for you.

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/characterization/resonator_punchout.py
+++ b/src/qibocal/protocols/characterization/resonator_punchout.py
@@ -225,6 +225,8 @@ def _plot(data: ResonatorPunchoutData, qubit, fit: ResonatorPunchoutResults = No
                     color="gray",
                     symbol="circle",
                 ),
+                name="Estimated readout point",
+                showlegend=True,
             )
         )
         fitting_report = table_html(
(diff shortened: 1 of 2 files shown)


GabrielePalazzo (approved):
I just tested it and the units seem correct. Thank you!

stavros11 (approved):
Thanks @andrea-pasquale, indeed this fixes the issue. I also think the toggle is useful. 

The funny thing is that actually the fit doesn't work for any of the plots I got (it is giving almost 0 amplitude and in some cases completly wrong frequency).

andrea-pasquale:
> The funny thing is that actually the fit doesn't work for any of the plots I got (it is giving almost 0 amplitude and in some cases completly wrong frequency).

Yeah, I know that the fit for the punchout works very poorly...
We should definitely try to improve it. I'll open an issue for this (if there isn't one already).

PR #643, opened 2023-11-24
Remove numba from cumulative

### LINKED ISSUE #637: Drop Numba (if possible)
Numba is a mandatory dependency in Qibocal:
https://github.com/qiboteam/qibocal/blob/8decaec48af4dc166bf69c6ebd1d20a7950d86e9/pyproject.toml#L52-L55
(btw, please lift mandatory ones above the optional ones)

However, it is used only in a single place:
https://github.com/qiboteam/qibocal/blob/8decaec48af4dc166bf69c6ebd1d20a7950d86e9/src/qibocal/protocols/characterization/utils.py#L248-L264

We should try to rewrite `cumulative` in pure NumPy (avoiding the `for` loop, the only non-NumPy operation), and drop the whole dep on Numba.

Consider that Numba is coming with the whole `llvmlite`, that is a full-blown compiler on its own...

stavros11:
Fixes #637. I don't fully understand why, but this simplification gives me exactly the same results with our current function in all examples I tried (including many tests with random numbers). If you can confirm, we can remove numba from the dependencies.

Btw, if there's something wrong and we end up keeping numba, you may have noticed that sometimes this gives an annoying warning
```
The keyword argument 'parallel=True' was specified but no transformation for parallel execution was possible.
```
If I understand this correctly, the parallelization is not properly implemented. One potential solution is to use [`prange`](https://numba.pydata.org/numba-doc/0.11/prange.html) in the `cumulative` loop:
```py
for i in prange(len(input_data)):
```
At least that's what we are using in qibojit to parallelize.

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`


src/qibocal/protocols/characterization/utils.py
```diff
@@ -244,23 +243,12 @@ def norm(x_mags):
...
-    # data and points sorted
-    prob = []
-    app = 0
-
-    for val in input_data:
-        app += np.maximum(np.searchsorted(points[app::], val), 0)
-        prob.append(float(app))
-
-    return np.array(prob)
+    return np.searchsorted(np.sort(points), np.sort(input_data))
```
AleCandido, line 251: Thanks for this, I imagined something like, but I thought that the `+= np.maximum(..., 0)` would have required more workarounds...

We could brainstorm soon to check if we could prove this to be the exact same.

AleCandido:
Actually, I thought that `parallel` would have affected compilation, but I was recalling wrong.

So, I believe that the operation, as it was written, was not parallelizable, because you were using `app` on the RHS, and modifying `app` on the LHS. So it is intrinsically sequential.

However, writing it in terms of pure vectorial operations should be almost equivalent to make it parallel (if it's operating on arrays, it is parallel - but if it is parallel, it should not be that hard to recast in terms of arrays).

src/qibocal/protocols/characterization/utils.py
stavros11: > Thanks for this, I imagined something like, but I thought that the `+= np.maximum(..., 0)` would have required more workarounds...

I think the `np.maximum(..., 0)` is not needed because `np.searchsorted` returns indices, which are probably always positive. What I don't understand is why the `[app::]` indexing seems to have no effect.

AleCandido: Yeah, sorry.

I also tried to trick it, but it is always returning an index. Possibly a wrong one, if assumptions do not hold (naively, I expected that it was returning `-1` if it was not found, but that's not what is happening -> it always returns a valid index).

And yes, I mentioned it below, but I forgot it here: `[app::]` is the one making it sequential and weird.
But thinking about it, maybe it is even redundant, since the array should be sorted. Even though it seems like it is cutting more than I would have expected...

AleCandido (approved):
Ok, regarding the `app` saga:

- the first `app` is ineffective, since `0`
- then `app` is incremented with the first index found for `val` (that is a valid index, since the array is sorted and the assumptions hold)
- this is cut from the array, when searching for the next `val`
- but the next `val` is strictly greater, since `input_data` is sorted
- and since also `points` is sorted, then it is guaranteed that the index for the second `val` would be greater-equal than the index for the first one
- ... (third, fourth, ...)

So, the `app` business is a performance improvement, since it allows you to search in a shorter array.
In the typical case, it is cutting a factor of 2, because you explore the triangle instead of the square.

But exploring the triangle requires you to go sequentially, because the exact shape of the triangle depends on the actual arrangement of numbers. The square instead is parallelizable.

Once `app` is stripped, and `np.maximum` is not needed, @stavros11 solution should be the correct one (the `.append()` and array conversion is automatically performed by NumPy).

AleCandido:
Actually, @stavros11, could you even drop the Numba dependency before merging?

stavros11:
Thanks for the explanation @AleCandido, that makes sense. I removed numba from the dependencies, but let's wait for another confirmation that this does not break anything before we merge.

AleCandido:
Perfect, I agree.

As soon as @andrea-pasquale or @Edoardo-Pedicillo will confirm, we can merge.

src/qibocal/protocols/characterization/utils.py
```diff
@@ -244,23 +243,12 @@ def norm(x_mags):
     return (x_mags - np.min(x_mags)) / (np.max(x_mags) - np.min(x_mags))
 
 
-@njit(["float64[:] (float64[:], float64[:])"], parallel=True, cache=True)
 def cumulative(input_data, points):
     r"""Evaluates in data the cumulative distribution
     function of `points`.
     WARNING: `input_data` and `points` should be sorted data.
```
Edoardo-Pedicillo, line 249:
```suggestion
```

Edoardo-Pedicillo (approved):
@stavros11 thank you for this PR ! The performances of the two implementations are similar. Just one small comment.

andrea-pasquale (approved):
Thanks @AleCandido and @stavros11 for looking into this.
Let's merge.

PR #659, opened 2023-12-07
Improve rabi period estimate

### LINKED ISSUE #656: Inaccurate fitted amplitude in rabi_amplitude_msr
Hi, 
I've noticed that the fit of the amplitude of the pi pulse in `rabi_amplitude_msr` returns a value that is close, but incorrect.
image
As you may see in the picture, the returned value is 0.436, but the argmax of the fitting curve is 0.466.
I've tried it on several qubits and it always gives a slightly smaller value.

[20231205_spinq10q_.zip](https://github.com/qiboteam/qibocal/files/13577196/20231205_spinq10q_.zip)

andrea-pasquale:
First attempt to close #656.
This PR improves the guess on the period by considering also the phase.

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`


src/qibocal/protocols/characterization/rabi/utils.py
```diff
@@ -190,3 +190,17 @@ def extract_rabi(data):
...
+
+    https://github.com/qiboteam/qibocal/issues/656
+    """
+
+    naive_half_period = period / 2
+    # solution of cos (2 pi x / period + phase) = +/- 1 for k in [-2,2]
+    solutions = [naive_half_period * (k - phase / np.pi) for k in [-2, -1, 0, 1, 2]]
+    # chose solution closest to naive_half period
+
+    return solutions[np.argmin(np.abs(solutions - naive_half_period))]
```
AleCandido, line 206: I don't see how this depends on `naive_half_period` at all (and thus period), apart for the normalization.

```py
solutions - naive_half_period
```
is even wrong, because `solutions` is not an array, but even the value of the `k`th element would be:
```py
naive_half_period * (k - phase / np.pi - 1)
```
and the minimum of the absolute value will be the minimum of the absolute value of `(k - phase / np.pi - 1)`.

So, instead of looking for the minimum of something, you could choose the closest integer to `phase / np.pi - 1`, just by rounding.

andrea-pasquale: It took me a while to understand it but now I should have done what you suggested :)

andrea-pasquale:
Tested on hardware and it seems to work as expected.

GabrielePalazzo (approved):

src/qibocal/protocols/characterization/rabi/utils.py
```diff
@@ -190,3 +190,13 @@ def extract_rabi(data):
...
         return "length", "Time [ns]", rabi_length_function
     raise RuntimeError("Data has to be a data structure of the Rabi routines.")
+
+
+def period_correction_factor(phase: float):
+    """Correct period by taking phase into account.
+
+    https://github.com/qiboteam/qibocal/issues/656
+    """
+    # solution of cos (2 pi x / period + phase) = +/- 1 for k in [-2,-1,0,1,2]
```
AleCandido, line 200: There is no `k` in your current formula :)

src/qibocal/protocols/characterization/rabi/utils.py
```diff
@@ -190,3 +190,13 @@ def extract_rabi(data):
...
+    # with k that gets correction closest to period
+    return np.round(phase / np.pi + 1) - phase / np.pi
```
AleCandido, line 202:
```py
np.round(phase / np.pi + 1) == np.round(phase / np.pi) + 1
np.round(x) - x == - np.modf(x)[0]
```
I was sure there was a function for the fractional part in NumPy, but unfortunately it's a bit less explicit than that (since it's returning both the fractional and the integral part - most likely because it also computes the second to get the first, internally).

```suggestion
    return 1 - np.modf(phase / np.pi)[0]
```

src/qibocal/protocols/characterization/rabi/utils.py
AleCandido: Seriously, the main problem is that I don't see clearly how this equation translates in the expression returned.

Maybe spelling out the steps, even in the docstring, would be beneficial.

andrea-pasquale: Yes, you are right. I will add the comment.
The translation should be the following:

$$ \cos(2 \pi x / T + \phi ) = \pm 1 \longrightarrow 2 \pi x / T + \phi = k \pi$$

AleCandido: This is the part I imagined, I was missing what is the *minimal correction* (minimal in which sense, how do you compute the correction to $T$)

src/qibocal/protocols/characterization/rabi/utils.py
```diff
@@ -193,10 +193,24 @@ def extract_rabi(data):
...
+    :math:`\cos(2 \pi x / T + \phi) = \pm 1` which will give us
+    the following solution
+
+    .. math::
+
+        x = ( k - \phi / \pi) T / 2
+
+
+    for integer :math:`k`, which is chosen such that we get the smallest
+    correction compared to :math:`T/2`.
```
AleCandido, line 212:
```suggestion
    multiplicative correction compared to :math:`T/2`.
```

src/qibocal/protocols/characterization/rabi/utils.py
AleCandido: Now it's perfect, thank you :)

AleCandido (approved):

PR #665, opened 2023-12-11
Add singleshot unrolling

Jacfomg:
I needed something to test the unrolling https://github.com/qiboteam/qibolab/pull/685 for Zurich with only uncalibrated qubits.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/characterization/classification.py
+++ b/src/qibocal/protocols/characterization/classification.py
@@ -185,11 +185,9 @@ def _acquisition(
             ro_pulses[qubit] = platform.create_qubit_readout_pulse(
                 qubit, start=RX_pulses[qubit].finish
             )
-            if state == 0:
-                sequence.add(ro_pulses[qubit])
-            elif state == 1:
+            if state == 1:
                 sequence.add(RX_pulses[qubit])
-                sequence.add(ro_pulses[qubit])
+            sequence.add(ro_pulses[qubit])
 
         sequences.append(sequence)
         all_ro_pulses.append(ro_pulses)
--- a/src/qibocal/protocols/characterization/classification.py
+++ b/src/qibocal/protocols/characterization/classification.py
@@ -177,8 +177,7 @@ def _acquisition(
     # taking advantage of multiplexing, apply the same set of gates to all qubits in parallel
     sequences, all_ro_pulses = [], []
     for state in [0, 1]:
-        sequences.append(PulseSequence())
-        all_ro_pulses.append({})
+        sequence = PulseSequence()
         RX_pulses = {}
         ro_pulses = {}
         for qubit in qubits:
@@ -187,12 +186,13 @@ def _acquisition(
                 qubit, start=RX_pulses[qubit].finish
             )
             if state == 0:
-                sequences[-1].add(ro_pulses[qubit])
-                all_ro_pulses[-1][qubit] = ro_pulses[qubit]
+                sequence.add(ro_pulses[qubit])
             elif state == 1:
-                sequences[-1].add(RX_pulses[qubit])
-                sequences[-1].add(ro_pulses[qubit])
-                all_ro_pulses[-1][qubit] = ro_pulses[qubit]
+                sequence.add(RX_pulses[qubit])
+                sequence.add(ro_pulses[qubit])
+
+            sequences.append(sequence)
+            all_ro_pulses.append(ro_pulses)
 
     data = SingleShotClassificationData(
         nshots=params.nshots,
--- a/src/qibocal/protocols/characterization/classification.py
+++ b/src/qibocal/protocols/characterization/classification.py
@@ -191,8 +191,8 @@ def _acquisition(
                 sequence.add(RX_pulses[qubit])
                 sequence.add(ro_pulses[qubit])
 
-            sequences.append(sequence)
-            all_ro_pulses.append(ro_pulses)
+        sequences.append(sequence)
+        all_ro_pulses.append(ro_pulses)
 
     data = SingleShotClassificationData(
         nshots=params.nshots,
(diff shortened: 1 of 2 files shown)


src/qibocal/protocols/characterization/classification.py
```diff
@@ -172,20 +175,24 @@ def _acquisition(
...
+            ro_pulses[qubit] = platform.create_qubit_readout_pulse(
+                qubit, start=RX_pulses[qubit].finish
+            )
+            if state == 0:
+                sequences[-1].add(ro_pulses[qubit])
+                all_ro_pulses[-1][qubit] = ro_pulses[qubit]
+            elif state == 1:
+                sequences[-1].add(RX_pulses[qubit])
+                sequences[-1].add(ro_pulses[qubit])
+                all_ro_pulses[-1][qubit] = ro_pulses[qubit]
```
AleCandido, line 195: To make it slightly simpler, instead of appending an empty container, and then addressing it as the last element, just make a variable for it, and append at the end.

I.e. instead of:
```py
for ...:
    xs.append({})
    xs[-1]["a"] = "b"
```
do
```py
for ...:
    x = {}
    x["a"] = "b"
    xs.append(x)
```

src/qibocal/protocols/characterization/classification.py
```diff
@@ -187,12 +186,13 @@ def _acquisition(
...
+                sequence.add(ro_pulses[qubit])
             elif state == 1:
-                sequences[-1].add(RX_pulses[qubit])
-                sequences[-1].add(ro_pulses[qubit])
-                all_ro_pulses[-1][qubit] = ro_pulses[qubit]
+                sequence.add(RX_pulses[qubit])
+                sequence.add(ro_pulses[qubit])
+
+            sequences.append(sequence)
+            all_ro_pulses.append(ro_pulses)
```
AleCandido, line 195: Pay attention: `ro_pulses` is defined outside the loop (so you're appending the same object over and over)

src/qibocal/protocols/characterization/classification.py
```diff
@@ -196,41 +203,35 @@ def _acquisition(
...
+            platform.execute_pulse_sequence(sequence, options) for sequence in sequences
+        ]
+
+    for ig, (state, ro_pulses) in enumerate(zip([0, 1], all_ro_pulses)):
+        for qubit in qubits:
+            serial = ro_pulses[qubit].serial
+            if params.unrolling:
+                result = results[serial][ig]
+            else:
+                result = results[ig][serial]
```
stavros11, line 225: Not related to this PR, but I noticed from the allXY that this is a bit weird (and annoying). @Jacfomg do you think it makes sense to change the results of the unrolling to follow the non-unrolling structure: list of dicts, instead of dict of lists.

Jacfomg: I dont have any strong opinions right now on either list of dicts or the other but it is indeed annoying so I would try to have both return the same.

src/qibocal/protocols/characterization/classification.py
Jacfomg: Yeah, it was working because they are always the same and I didn't notice

src/qibocal/protocols/characterization/classification.py
AleCandido: I also agree with @Jacfomg (and I guess it was already the original motivation of @stavros11): it does not matter which, but they should be the same.

(actually, it might matter, but at least less - so you could fix it both ways, and, if we will have a good enough reason to pick the other one later on, we will consider changing it)

src/qibocal/protocols/characterization/classification.py
```diff
@@ -172,20 +175,24 @@ def _acquisition(
...
+        for qubit in qubits:
+            RX_pulses[qubit] = platform.create_RX_pulse(qubit, start=0)
+            ro_pulses[qubit] = platform.create_qubit_readout_pulse(
+                qubit, start=RX_pulses[qubit].finish
+            )
+            if state == 0:
+                sequence.add(ro_pulses[qubit])
+            elif state == 1:
+                sequence.add(RX_pulses[qubit])
+                sequence.add(ro_pulses[qubit])
```
hay-k, line 192:
```suggestion
            if state == 1:
                sequence.add(RX_pulses[qubit])
            sequence.add(ro_pulses[qubit])
```

src/qibocal/protocols/characterization/classification.py
```diff
@@ -196,41 +203,35 @@ def _acquisition(
...
-            ClassificationType,
-            (qubit),
-            dict(i=result.voltage_i, q=result.voltage_q, state=[1] * params.nshots),
-        )
+    if params.unrolling:
+        results = platform.execute_pulse_sequences(sequences, options)
+    else:
+        results = [
+            platform.execute_pulse_sequence(sequence, options) for sequence in sequences
+        ]
```
hay-k, line 217: Is there actually any benefit of un-unenrolled execution of singleshot? Why not do unrolled execution always?

andrea-pasquale: I think that as long as some of the drivers do not fully support the unrolling is better to keep them both. Otherwise you won't be able to run some protocols with those drivers. Right now only qblox and zurich support the unrolling correct, @stavros11?

src/qibocal/protocols/characterization/classification.py
andrea-pasquale: +1

andrea-pasquale (commented):
Thanks @Jacfomg.
Find below some comments.

src/qibocal/protocols/characterization/classification.py
andrea-pasquale: As soon as we fix this point we should be ready to merge. Let's try to return the same in both protocols.

andrea-pasquale (approved):
Thanks @Jacfomg, LGTM.
As soon as we resolve the last conversation we should be ready to merge.

Edoardo-Pedicillo (approved):

PR #666, opened 2023-12-11
Ramsey unrolling

Jacfomg:
It adds unrolling to the detuned ramsey

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


src/qibocal/protocols/characterization/ramsey.py
```diff
@@ -161,31 +166,67 @@ def _acquisition(
                     errors=errors,
                 ),
             )
-    else:
+    if params.n_osc != 0:
```
hay-k, line 169: There is a lot of repeated code under the cases `n_osc == 0` and `n_osc != 0`. I would like to suggest unifying them, or moving the repeated functionality into a function. It would be even nicer if the `ramsey_signal` also benefits from such unification. Unfortunately I am not able to understand what is the meaning of `n_osc` from its documentation, so cannot suggest a more concrete refactoring. Let's discuss.

hay-k:
I don't see any unit tests. (in https://github.com/qiboteam/qibocal/pull/665 also)

tests/runcards/protocols.yml
```diff
@@ -380,19 +380,20 @@ actions:
...
       delay_between_pulses_step: 1
       nshots: 10
 
-  - id: ramsey detuned sequences
+  - id: ramsey_signal_detuned
     priority: 0
-    operation: ramsey_sequences
+    operation: ramsey_signal
     parameters:
+      unrolling: True
```
andrea-pasquale, line 393: @hay-k this is the uni test. In this `yaml` we put all the protocols that are later tested in [tests/test_protocols](https://github.com/qiboteam/qibocal/blob/ramsye_unrolling/tests/test_protocols.py).

src/qibocal/protocols/characterization/ramsey.py
andrea-pasquale: Thanks @hay-k. We should definitely find a way to improve it. Just for reference `n_osc` stands for `number of oscillations`. In the ramsey protocol by putting a non zero integer value we include a detuning through the relative phase that will generate `n_osc` number of oscillations in the plot.  https://github.com/qiboteam/qibocal/blob/528c3b91cae0f2734da39b77431c70323d66d525/src/qibocal/protocols/characterization/ramsey.py#L190-L195
Having said this, I believe that in `ramsey` we differentiate between the `n_osc==0` and the `n_osc!=0` case because in the first case the protocol can be implemented through a simple sweeper on the start of the second RX90 pulse while in the second case, if I remember correctly, there are some limitations in our drivers that prevents us from sweeping start and relative phase at the same time. @stavros11 feel free to correct me if I said something wrong.

Of course the best solution for qibocal should be to keep the code as clean as possible and let the drivers handle these specific cases.

andrea-pasquale (commented):
I have not tested the code yet. 
It seems ok. @Jacfomg you removed the `ramsey_sequences` because we are assuming that all the drivers will be compatible either with sweepers or with sequence unrolling, correct?

src/qibocal/protocols/characterization/ramsey.py
hay-k: I see, thanks for the explanation! I think, in general, the drivers (and their developers, of course) will benefit a lot from a more unified and consistent implementation. Anyways, that's a story for another day. I believe we should still simplify the code in this particular instance, even without improving anything else anywhere else. That should be doable, and I am here to assist. @Jacfomg probably best if we discuss AFK.

Jacfomg:
> I have not tested the code yet. It seems ok. @Jacfomg you removed the `ramsey_sequences` because we are assuming that all the drivers will be compatible either with sweepers or with sequence unrolling, correct?

Yeah, I would say it is a fair assumption. It has been there for far too long and I'm not aware of it being used.

tests/runcards/protocols.yml
hay-k: hmm, let's discuss this, because if I understand those tests correctly, they are more like integration or end-to-end tests, rather than unit tests. But in any way this seems like a discussion that should be kept away from this PR, not to hinder its progress.

src/qibocal/protocols/characterization/ramsey.py
Jacfomg: Sure, it's sounds like a good plan.

stavros11 (approved):

tests/runcards/protocols.yml
```diff
@@ -380,19 +380,30 @@ actions:
...
       delay_between_pulses_end: 50
       delay_between_pulses_step: 1
+      n_osc: 2
       nshots: 10
 
-  - id: ramsey detuned sequences
+  - id: ramsey_detuned_sequences
     priority: 0
-    operation: ramsey_sequences
+    operation: ramsey_seqeunces
```
andrea-pasquale, line 392:
```suggestion
    operation: ramsey_sequences
```

src/qibocal/protocols/characterization/__init__.py
```diff
@@ -36,7 +36,6 @@
 from .rabi.length_sequences import rabi_length_sequences
 from .rabi.length_signal import rabi_length_signal
 from .ramsey import ramsey
-from .ramsey_sequences import ramsey_sequences
 from .ramsey_signal import ramsey_signal
 from .randomized_benchmarking.standard_rb import standard_rb
 from .readout_characterization import readout_characterization
```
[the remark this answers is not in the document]
Jacfomg: Thanks for fixing

andrea-pasquale (approved):

Issue #589, opened 2023-10-26
Simultaneous routine execution


Jacfomg:
@andrea-pasquale, @Edoardo-Pedicillo  and @ingoroth 
Run different routines on different qubits at the same time

PR #668, opened 2023-12-12
Exceptional flow: handlers prototype

andrea-pasquale:
This PR implements `handlers` required by the exceptional flow. 

Here is an example of a possible runcard:

```yaml
max_iterations: 10

qubits: [0]

actions:
  - id: t1
    priority: 0
    operation: t1
    main: rabi
    parameters:
      delay_before_readout_start: 16
      delay_before_readout_end: 2000
      delay_before_readout_step: 20

  - id: rabi
    priority: 0
    operation: rabi_amplitude
    main: single shot
    parameters:
      min_amp_factor: 0.0
      max_amp_factor: 2.0
      step_amp_factor: 0.02
      pulse_length: 40
      relaxation_time: 100_000
      nshots: 1024
    validator:
      scheme: chi2
      parameters:
        thresholds: [0.1, 10]
      outcomes: [["t1", {"nshots" : 100}]]
    update: false

  - id: single shot
    priority: 0
    operation: single_shot_classification
    parameters:
      nshots: 10000
```
In this particular case when the validation on the `rabi` protocol is not satisfied the head of the graph goes to the `t1` node. Given that this could cause an infinite loop I've also added a `MAX_ITERATIONS` constant (which could eventually be the same node if necessary).

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`


src/qibocal/auto/execute.py
```diff
@@ -7,12 +7,16 @@
...
 
+from ..config import raise_error
 from .graph import Graph
 from .history import History
 from .runcard import Id, Runcard
 from .status import Failure
 from .task import Qubits, Task
 
+MAX_ITERATIONS = 5
+"""Maximum number of iteration for single task."""
```
AleCandido, line 18: This will have to be exposed to the runcard (not necessarily in this PR)

src/qibocal/auto/history.py
```diff
@@ -33,9 +33,15 @@ class History(dict[tuple[Id, int], Completed]):
...
     def push(self, completed: Completed):
         """Adding completed task to history."""
-        self[(completed.task.id, completed.task.iteration)] = completed
-
-        # TODO: re-implement this
-        # completed.task.iteration += 1
+        iteration = self.iteration_counter(completed.task.id)
+        self[(completed.task.id, iteration)] = completed
+
+    def iteration_counter(self, task_id: Id):
```
AleCandido, line 39:
```suggestion
    def iterations(self, task_id: Id):
```

AleCandido: However, why do you need this function?

The `Task` object already has an `iteration` attribute
https://github.com/qiboteam/qibocal/blob/2e85e0a4f496892212c616e78534c378de918e97/src/qibocal/auto/task.py#L42-L43
So, if you have a `Task` at hand, you already know the number of iterations.

I.e.
```py
self.history.iteration_counter(task.id) == task.iteration
```
(potentially with a +1 on one of the two sides, because of 0-based indices)

If this is false, then the attribute is invalid.

andrea-pasquale: In theory yes, but this is the only solution that I was able to find.
If you see the previous implementation when we pushed to the history we were implementing the +1 on `task.iteration`. This approach is not valid anymore for two reasons:

1.  In this while loop: https://github.com/qiboteam/qibocal/blob/330aac1b8b7945687c4d834e47ff73f8ba7fd35c/src/qibocal/auto/execute.py#L137-L151 the `task.run` method generates a new instance of `Completed`
2. when such instance is created a `deepcopy` of the `task` is performed
 https://github.com/qiboteam/qibocal/blob/330aac1b8b7945687c4d834e47ff73f8ba7fd35c/src/qibocal/auto/task.py#L148

Therefore, if you consider two iterations of the while loop where we are supposed to iterate over a protocol twice I think that currently there is no way to let one node aware of the other unless we check the history.

To be consistent when I push to the history I should also update `task.iteration` with `self.history.iteration_counter(task.id).

Hopefully the explanation was not too confusing :)

src/qibocal/auto/execute.py
andrea-pasquale: Since it will not require too much work I can do it here.

src/qibocal/auto/history.py
andrea-pasquale: In https://github.com/qiboteam/qibocal/pull/668/commits/d3458cf6458890e5cfb263fe0043a7f505a18b27 I tried to simplify a bit the code.

src/qibocal/auto/execute.py
```diff
@@ -133,24 +136,45 @@ def run(self, mode):
...
+                    log.warning("Stopping execution due to error in validation.")
+                    yield completed.task.uid
+                    break
+                else:
+                    log.info(
+                        f"Handling validation error by executing node {task.action.handler.id}"
+                    )
+
+                    self.head = task.action.handler.id
+                    yield completed.task.uid
```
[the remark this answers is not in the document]
AleCandido: However, I missed the `break`, and I missed that before was always present, as it should (the regular `yield` is happening below).

I would actually do even more at this point: I'd rather reconcile these `yield`s with the one at the bottom.
And provide a mechanism to `break`, if needed (e.g. `try... except` with a custom error).

src/qibocal/auto/execute.py
```diff
@@ -133,24 +136,45 @@ def run(self, mode):
...
             if (
                 mode.name in ["autocalibration", "fit"]
                 and self.platform is not None
                 and update
             ):
-                task.update_platform(results=completed.results, platform=self.platform)
-            yield task.uid
+                completed.task.update_platform(
+                    results=completed.results, platform=self.platform
+                )
```
AleCandido, line 178: Should we lift even the `update_platform`? All in all, it makes sense that a platform is updated by a completed task.
```suggestion
                completed.update_platform(self.platform)
```

src/qibocal/auto/execute.py
```diff
@@ -133,24 +136,45 @@ def run(self, mode):
         self.head = self.graph.start
         while self.head is not None:
             task = self.current
-            log.info(f"Executing mode {mode.name} on {task.id}.")
+            task.iteration = self.history.iterations(task.id)
+            if self.history.iterations(task.id) > MAX_ITERATIONS:
+                raise_error(
+                    ValueError,
+                    f"Maximum number of iterations {MAX_ITERATIONS} reached!",
+                )
```
AleCandido, line 144: This you could push inside `task.run`

src/qibocal/auto/execute.py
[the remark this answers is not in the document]
andrea-pasquale: Your solution is also missing the repositioning of the `head` of the graph.
Maybe the easiest solution would be to let validate return directly the position in graph (in the non-exceptional flow we can simply return the current task id).
```suggestion
            if mode.name == "autocalibration":
                self.head = self.completed.validate()
```
Where validate could be something like this:

```py
def validate(self) -> TaskId:
        """Check status of completed and handle Failure using handler."""
        if isinstance(self.status, Failure):
            try:
                return self.task.action.handler.id
            except AttributeError:
                log.error("Stopping execution because of error in validation without handler")
                return None
        return self.task.id
```
By putting the head to `None` we also have an easy condition to exit the while loop without using `break`.

src/qibocal/auto/history.py
AleCandido: Yes, now I see. I expected it to be more predictable, but we don't have (and don't need) to have global task objects to keep this information (since it's already contained in the `.history`).

In principle, we could have had a `defaultdict` with the counter (`id: count`), but it's nothing more than a cache for what is done in the `history`.

Let's keep it as it is. Sorry for the noise.

src/qibocal/auto/execute.py
andrea-pasquale: Do we prefer a global parameter or something local attached to a specific routine?

AleCandido: For the time being, just a global parameter. If we'll ever need something more fine-grained, we will introduce it later (with no other rationale of keeping it simple).

andrea-pasquale: Implemented.

src/qibocal/auto/task.py
```diff
@@ -243,5 +247,20 @@ def data(self, data: Data):
...
 
-    def __post_init__(self):
-        self.task = copy.deepcopy(self.task)
+    def update_platform(self, platform: Platform):
+        """Perform update on platform' parameters by looping over qubits or pairs."""
+        if self.task.update:
+            for qubit in self.task.qubits:
+                self.task.operation.update(self.results, platform, qubit)
+
+    def validate(self) -> TaskId:
```
AleCandido, line 256:
```suggestion
    def validate(self) -> Optional[TaskId]:
```

src/qibocal/auto/task.py
```diff
@@ -243,5 +247,20 @@ def data(self, data: Data):
...
+        """Check status of completed and handle Failure using handler."""
+        if isinstance(self.status, Failure):
+            try:
+                return self.task.action.handler.id
+            except AttributeError:
+                log.error(
+                    "Stopping execution because of error in validation without handler"
+                )
+                return None
```
AleCandido, line 265: It is pretty elegant to catch errors in Python, rather than checking, but in this case I would say it's not practical.

Indeed, whatever missing attribute would be raised, even inside the `.handler` it would generate an `AttributeError` indistinguishable from the one you are expecting.

Since in this case it costs very little, I would rather check:
```suggestion
            handler = self.task.action
            if handler is not None:
                return handler.id
            log.error(
                "Stopping execution because of error in validation without handler"
            )
            return None
```

andrea-pasquale:
I've added a test to check that the `handler` works as expected.
I think that it would be useful to add a tutorial about validators and handlers.
I don't know if I want to do it here or in a separate PR. Apart from that I believe that this PR is now ready to review.

src/qibocal/auto/task.py
```diff
@@ -163,8 +165,13 @@ def run(
...
             if operation.platform_dependent and operation.qubits_dependent:
                 completed.data, completed.data_time = operation.acquisition(
-                    parameters, platform=platform, qubits=task_qubits
+                    parameters,
+                    platform=platform,
+                    qubits=self._allocate_local_qubits(qubits, platform),
                 )
+                # need to reassign qubit since when task
+                # is passed it is deepcopied
+                completed.task.qubits = self.qubits
```
AleCandido, line 174: I'm not sure I understood this point

src/qibocal/auto/execute.py
```diff
@@ -133,24 +134,31 @@ def run(self, mode):
...
+            )
             completed = task.run(
+                max_iterations=self.max_iterations,
                 platform=self.platform,
                 qubits=self.qubits,
                 folder=self.output,
                 mode=mode,
             )
+            print("TASK QUBITS", task.qubits)
+            print("COMPLETED TASK QUBITS", completed.task.qubits)
```
AleCandido, line 151: Please, remove debug prints

tests/runcards/handlers.yml
```diff
@@ -0,0 +1,38 @@
...
+      step_amp_factor: 0.02
+      pulse_length: 40
+      relaxation_time: 100_000
+      nshots: 1024
+    validator:
+      scheme: chi2
+      parameters:
+        chi2_max_value: 0.01
+    handler:
+      id: t1
```
AleCandido, line 31: Only now I took the time to take a proper look at the general picture.

I believe it is not extremely useful to have this kind of "handlers". Essentially:
- you're defining an object that is only telling you where to jump, with a single node (it is a prototype, it's fine, but there is no room for the case in which there are multiple possible outcomes)
- the object should be directly related to the validator, but in this layout it is completely independent

AleCandido: All in all, I'd rather refactor including the information in the `handler` into the `validator` itself.

You could have a general key, in which you set some nodes to jump into, and then encode the information about which to choose in the order, in agreement with the specific key
```yaml
validator:
  scheme: ...
  outcomes: [id1, id2, id1, id3, ...]
```
such that your validator will return the index to select.

You could decide to keep it even more free form, and leave the whole key structure up to the specific validator, or some intermediate solution (e.g. there will be an `outcomes` key, but whose type depends on the scheme).
But the more structure you could give, the more you can use (the trade-off is between complexity and flexibility, and what is what depends on the specific case).

In any case, the main message is that:
1. there is some information that should be specified in the *program* (the runcard), and these are the specific `id`s, that are specific to the program
2. and some that belongs to the building blocks (the validators), that is how to make a choice

Since 2. is potentially repeated, we can reuse validators.
And, definitely, we can not reuse `id`s. But their meaning is bound to the validator.

AleCandido (changes requested):
Sorry, I'm asking you to review the *handlers* mechanism. Maybe I'm not fully understanding your goal, possibly because it's a too minimal prototype.

However, many of the things done here are actually useful, and not strictly related to the *handlers*, so I'd try to merge this PR, even in the case in which there will be no handler at all.

src/qibocal/auto/execute.py
andrea-pasquale: My bad. I forgot to save this file before committing.

tests/runcards/handlers.yml
andrea-pasquale: About unifying handler and validator I agree. 
I am not sure about the multiple outcomes... if we follows Sergi's structure I think that if there is an issue with one node you should more or less jump into the node directly above in the tree and if something doesn't work again you can keep going up in the tree. I assume that there is already a graph specified in the runcard and the handler is just trying to go back and forward in the graph (maybe my interpretation is wrong).
Anyway, maybe it is better to discuss offline and define a common strategy :)

src/qibocal/auto/task.py
AleCandido: Unfortunately, I just realized that `Task._allocate_local_qubits()`, and the functions called inside, have no docstrings.
They are not considered part of the public API, but I would try to always have minimal docstrings, just to have hints to understand the workflow of the functions, without having to jump over the whole code base for doing it.

However, having jumped around...

It seems to me that the *allocated qubits* are just a function of the `platform`, and the `qubits` parameters. And nothing is modifying them inside the `.run()` method, prior to that call.
If this is the case, then you could just trigger the update of `self.qubits` even before instantiating `Completed`.

But since in general the return value is going to be a dictionary, and only the keys are retained in `self.qubits`, I'd like to understand more about how this information is flowing. But maybe we can discuss in private.

tests/runcards/handlers.yml
AleCandido: It's not wrong, but I believe that diagram is just one possible *program* for the *autocalibrator*.

In general, you may be in the situation in which you can discriminate multiple different final states, and react in different ways accordingly.

andrea-pasquale:
@AleCandido @Edoardo-Pedicillo @GabrielePalazzo this should be ready to review.
@AleCandido I've implemented your suggestions after our meeting. 
Currently since there is only the $\chi^2$ validator available the workflow should be something like this:
```py
THRESHOLDS = [t_1,t_2,t_3,...,t_N]
OUTCOMES=[o_1,...,o_{N-1}]
```
Assuming that the thresholds are ordered if `t < t_1` the executor will follow the normal flow returning as status `Normal()`.
If `t_i < t < t_{i+1}` the executor will perform an exceptional flow by running node `o_i` with its own parameters if provided. Otherwise the output should be `Failure()`.
I still need to update the documentation by writing an example on how to use it and maybe add more robust tests.
Before doing so, I just wanted to check with you if something else should be modified. 
Concerning #678 I've already opened #682 which is pointing to this branch (let's focus on this first).

AleCandido:
> Assuming that the thresholds are ordered

What happens if they are not? (I've not yet read the code, so I truly don't know - I'm about to do it)

I'm not a huge fan of excessive checks, but when it comes to UI, I'd try to keep

src/qibocal/auto/task.py
```diff
@@ -243,5 +239,29 @@ def data(self, data: Data):
...
+            for qubit in self.task.qubits:
+                self.task.operation.update(self.results, platform, qubit)
+
+    def validate(self) -> tuple[Optional[TaskId], Optional[dict]]:
+        """Check status of completed and handle Failure using handler."""
+        if self.task.action.validator is not None:
+            status = []
+            for target in self.task.qubits:
+                # TODO: how to handle multiple targets?
+                # dummy solution for now: take the mode.
```
AleCandido, line 254: Having in mind the idea of the stack of tasks to be executed, if we have different outcomes per qubit (i.e. different task ids, not only different parameters), we should split them in groups by task id, and just put each group on the stack as a separate task.

The step after this one would be to be able to merge tasks in the stack, if eventually they rejoin for some reason.

This would be more friendly even in the case of parallel executors (though I wonder if that is even reasonable, I would expect the acquisition to take most of the time, and the chip is only one), but I would avoid thinking about that now.

src/qibocal/auto/runcard.py
```diff
@@ -51,9 +54,15 @@ class Runcard:
...
     actions: list[Action]
+    """List of action to be executed."""
     qubits: Optional[Union[list[QubitId], list[tuple[QubitId, QubitId]]]] = None
+    """Qubits to be calibrated."""
     backend: str = "qibolab"
+    """Qibo backend."""
     platform: str = os.environ.get("QIBO_PLATFORM", "dummy")
+    """Qibolab platform."""
+    max_iterations: int = MAX_ITERATIONS
+    """Maximum number of iterations."""
```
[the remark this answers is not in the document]
andrea-pasquale: At this point maybe we could already plug this functionality inside the code.

src/qibocal/auto/execute.py
```diff
@@ -28,6 +27,8 @@ class Executor:
     """Qubits to be calibrated."""
     platform: Platform
     """Qubits' platform."""
+    max_iterations: int
+    """Maximum number of iterations."""
```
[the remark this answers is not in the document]
andrea-pasquale: For sure it needs to be documented properly.

src/qibocal/auto/task.py
```diff
@@ -201,6 +194,9 @@ class Completed:
     results_time: float = 0
     """Fitting output."""
 
+    def __post_init__(self):
+        self.task = copy.deepcopy(self.task)
```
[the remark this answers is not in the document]
andrea-pasquale: Yes, I remember playing around with this `deepcopy` to check if I can remove it but I failed :(

src/qibocal/auto/task.py
andrea-pasquale: There is already regarding parallel executions https://github.com/qiboteam/qibocal/issues/589 so this is something that we might look at. For this PR specifically I might be happy with just a "proper" implementation for a single qubit.

src/qibocal/auto/validation.py
```diff
@@ -1,30 +1,64 @@
...
+            return Failure(), None
+        elif index == 0:
+            # for chi2 (to be generalized for other validators):
+            # if chi2 is less than first threshold the status is normal
+            return Normal(), None
+
+        # else we return outcomes [index-1] since outcomes outcomes[i] is
+        # the output of thresholds[index+1], given that for the first threshold
+        # the status is Normal.
+        return self.outcomes[index - 1]
```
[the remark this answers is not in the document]
andrea-pasquale: -1 sounds good to me. So then if we kill normal shall we keep only  `Failure` as a status?
When I was introducing these changes I also thought about removing `Status` directly but I don't know if this is the right time to do it.

src/qibocal/auto/validators/chi2.py
```diff
@@ -3,35 +3,31 @@
...
 CHI2_MAX = 0.05
 """Max value for accepting fit result."""
 
 
 def check_chi2(
     results: Results,
-    qubit: Union[QubitId, QubitPairId, list[QubitId]],
-    chi2_max_value=None,
+    target: Union[QubitId, QubitPairId, list[QubitId]],
+    thresholds: list = [CHI2_MAX],
```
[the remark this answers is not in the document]
andrea-pasquale: It looked a bit fishy when I typed it... 
What should the proper solution in this case? How can I initialize a function with a default list?

tests/test_task_options.py
```diff
@@ -46,22 +44,22 @@ def modify_card(card, qubits=None, update=None):
...
-    _, _ = task.operation.acquisition(task.parameters, platform, global_qubits)
-    if local_qubits:
-        assert task.qubits == local_qubits
-    else:
-        assert task.qubits == QUBITS
+#     _, _ = task.operation.acquisition(task.parameters, platform, global_qubits)
+#     if local_qubits:
+#         assert task.qubits == local_qubits
+#     else:
+#         assert task.qubits == QUBITS
```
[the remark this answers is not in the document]
andrea-pasquale: I forgot about this test. I will fix it.

src/qibocal/auto/validators/chi2.py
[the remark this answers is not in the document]
andrea-pasquale: https://web.archive.org/web/20200221224620id_/http://effbot.org/zone/default-values.htm
I will put `None` and add a if statement. It seems to be the best solution.

andrea-pasquale:
> I'm not a huge fan of excessive checks, but when it comes to UI, I'd try to keep the behavior explicit. Thus, I see two options:
> 
>     * _just **fail** if they are unsorted_, that has the benefit of warning the user about unexpected runcards (possibly errors)
> 
>     * _silently **sort** them_, that will make less annoying adding threshold, but gives up on error detection
> 
> 
> Whatever behavior, it has to be declared in the docs of the chi2 validator.

At this point I prefer to fail if they are unsorted. Given that each threshold or threshold interval will be associated with an outcomes it is better if you are consistent. If you allow to have unsorted thresholds that we will also need to sort the outcomes. Let's stick with sorted thresholds.

src/qibocal/auto/runcard.py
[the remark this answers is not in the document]
AleCandido: What do you mean exactly with *inside the code*?

src/qibocal/auto/validators/chi2.py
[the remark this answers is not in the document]
AleCandido: Indeed, sorry for the delay, but that would have been my proposal for a Pythonic solution.

src/qibocal/auto/validation.py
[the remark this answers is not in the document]
AleCandido: In this PR, I'd do neither of them, since killing `Normal` would mean to kill the directed graph, including the basic sequential execution.

But I'm leaning more and more towards it.

Moreover, if you kill `Normal` you wouldn't even need the `-1`: all the tasks in chi2 will be equal, not only the intermediate one. But the largest segment, which will correspond to a failure.

Concerning `Failure` itself, I don't know. We could use `None` as special value, that is fine, but I'd keep the concept: there should be a mechanism through which a validator can interrupt the whole execution, just because it ran out of possible scenarios and solutions.

src/qibocal/auto/task.py
AleCandido: The second part definitely (last paragraph), but the first one no.

Or not in a specific sense: it's not related to running multiple executors in parallel, but it is related to run some tasks on multiple qubits.
I.e. if we want to handle multiple outcomes, and we have routines acting on multiple qubits, we should account also for diverging ones.

Of course, if we are not even using the multiple qubits (and I know we mostly are not doing it) that is definitely beyond.

AleCandido:
> If you allow to have unsorted thresholds that we will also need to sort the outcomes.

Yes, but that's not too difficult to implement: if you have two lists to sort together, you just `zip` the lists (essentially transpose) and then you could sort the pairs (2-tuples).

Sorting of tuples happen lexicographically, so they will be sorted according to the first components (saving ties, but if they tie we don't care - and it won't happen with thresholds).

Then you could `zip` again (transpose once more) and you get your sorted lists. With consistent sorting.

With NumPy you could use `np.argsort` for that (first get the indices, then sort both).

src/qibocal/auto/runcard.py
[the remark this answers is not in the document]
andrea-pasquale: By inside the code I just mean to implement it. In this case to implement a local `max_iterations` parameter for each action.

andrea-pasquale:
> Yes, but that's not too difficult to implement: if you have two lists to sort together, you just `zip` the lists (essentially transpose) and then you could sort the pairs (2-tuples).
> 
> Sorting of tuples happen lexicographically, so they will be sorted according to the first components (saving ties, but if they tie we don't care - and it won't happen with thresholds).
> 
> Then you could `zip` again (transpose once more) and you get your sorted lists. With consistent sorting.
> 
> With NumPy you could use `np.argsort` for that (first get the indices, then sort both).

I know that it is doable, I just think that it will be very hard for a user to write an unsorted list with the corresponding unsorted outcomes.
Comparing a "sorted" runcard
```yaml
thresholds: [0, 1, 2, 3]
outcomes: [a, b, c]
```
with the "unsorted"

```yaml
thresholds: [3, 1, 2, 0]
outcomes: [c, a, b]
```
I find it very difficult that a human will produce the second runcard having in mind the first. Maybe it is just my way of thinking, but I suggest to keep using sorted list and raise an error if they are not.

src/qibocal/auto/validation.py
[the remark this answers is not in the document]
andrea-pasquale: > In this PR, I'd do neither of them, since killing `Normal` would mean to kill the directed graph, including the basic sequential execution.

I agree. In https://github.com/qiboteam/qibocal/pull/668/commits/fa3de8e8cef73300243f4b6e6928b41255bf4a1a I introduced the -1 for `Normal`.

Let me open an issue to remember to remove `Normal` and `Failure` in the future.

src/qibocal/auto/validation.py
```diff
@@ -1,30 +1,62 @@
...
+        Possible Returns are:
+            - (Failure, None) which stops the execution.
+            - (Normal, None) which corresponds to the normal flow
+            - (task, dict) which moves the head to task using parameters in dict.
+        """
+        index = self.method(results=results, target=target, **self.parameters)
+
+        if index == -1:
+            # -1 denotes Normal()
+            return Normal(), None
```
AleCandido, line 54:
```suggestion
        if index == -1:
            return Normal(), None
```
I'd say that the message is clear enough even without the comment :)

src/qibocal/auto/validation.py
```diff
@@ -1,30 +1,62 @@
...
+        else:
+            try:
+                return self.outcomes[index]
+            except (TypeError, IndexError):
+                # TypeError to handle the case where index is None
+                # IndexError to handle the case where index not in outcomes
+                log.error("Stopping execution due to error in validation.")
+                return Failure(), None
```
AleCandido, line 62:
```suggestion
        try:
            return self.outcomes[index]
        except (TypeError, IndexError):
            # TypeError to handle the case where index is None
            # IndexError to handle the case where index not in outcomes
            log.error("Stopping execution due to error in validation.")
            return Failure(), None
```
If you return in the `if`, you don't need the `else` (Pylint would flag it)

AleCandido (commented):
Small aesthetic changes

src/qibocal/auto/validators/chi2.py
```diff
@@ -3,35 +3,31 @@
...
-        if chi2 < chi2_max_value:
-            return Normal()
-        else:
-            return Failure()
+        chi2 = getattr(results, "chi2")[target][0]
+        for threshold in sorted(thresholds):
+            if chi2 < threshold:
+                return thresholds.index(threshold)
+
+        return None
```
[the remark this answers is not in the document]
andrea-pasquale: I force the thresholds to be sorted in https://github.com/qiboteam/qibocal/pull/668/commits/e7b73cd120456469976175ed542e1438670c2b83. There might be a simpler way of doing it. I used pydantic since I wanted to make this check as soon as the runcard is loaded and not after the acquisition or fit. Let me know if you agree.

src/qibocal/auto/runcard.py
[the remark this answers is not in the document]
andrea-pasquale: Actually I'm having second thoughts about this,  perhaps it is enough to keep it global. I would like to see what happens after we deploy this PR on hardware to see if you need this flexibility at the level of `Action`.

andrea-pasquale:
> > Yes, but that's not too difficult to implement: if you have two lists to sort together, you just `zip` the lists (essentially transpose) and then you could sort the pairs (2-tuples).
> > Sorting of tuples happen lexicographically, so they will be sorted according to the first components (saving ties, but if they tie we don't care - and it won't happen with thresholds).
> > Then you could `zip` again (transpose once more) and you get your sorted lists. With consistent sorting.
> > With NumPy you could use `np.argsort` for that (first get the indices, then sort both).
> 
> I know that it is doable, I just think that it will be very hard for a user to write an unsorted list with the corresponding unsorted outcomes. Comparing a "sorted" runcard
> 
> ```yaml
> thresholds: [0, 1, 2, 3]
> outcomes: [a, b, c]
> ```
> 
> with the "unsorted"
> 
> ```yaml
> thresholds: [3, 1, 2, 0]
> outcomes: [c, a, b]
> ```
> 
> I find it very difficult that a human will produce the second runcard having in mind the first. Maybe it is just my way of thinking, but I suggest to keep using sorted list and raise an error if they are not.

Going back to the possibility to check if the list is already ordered.

AleCandido:
> I find it very difficult that a human will produce the second runcard having in mind the first. Maybe it is just my way of thinking, but I suggest to keep using sorted list and raise an error if they are not.

Yes, on that I agree with you. The use case is whether you want to add something in the middle of a long list, and you don't want to count:
```yaml
thresholds: [0, 100, 200, 300, 400, 500, 600]
outcomes: [a, b, c, d, e, f]
```
and then:
```yaml
thresholds: [0, 100, 200, 300, 400, 500, 600, 250]
outcomes: [a, b, c, d, e, f, h]
```

This is just the practical case that comes to my mind. But it's really a matter of UI, and I don't care that much. Even requesting sorted it's not bad.

src/qibocal/auto/task.py
andrea-pasquale: Resolving this as implemented in #682

src/qibocal/auto/execute.py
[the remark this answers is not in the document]
andrea-pasquale: Documentation added

src/qibocal/auto/validators/chi2.py
```diff
@@ -1,37 +1,70 @@
...
+#         thresholds = [CHI2_MAX]
+
+#     try:
+#         chi2 = results.chi2[target][0]
+#         idx = np.searchsorted(thresholds, chi2)
+#         return idx - 1
+
+#     except AttributeError:
+#         log.error(f"Chi2 validation not available for {type(results)}")
+#         return None
```
AleCandido, line 70: If unused, drop this.

AleCandido (approved):
It looks alright :)

Thanks for the PR!

src/qibocal/auto/runcard.py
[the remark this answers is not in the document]
AleCandido: Good approach: let's limit the complexity to what we really need.

src/qibocal/auto/validators/chi2.py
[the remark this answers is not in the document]
AleCandido: Agreed

doc/source/getting-started/validation.rst
```diff
@@ -1,10 +1,14 @@
...
+==============================
 
-In Qibocal there is the possibility to add a validation step
-during the execution of your protocols.
-Here is an example of a runcard which validates the results through
-:math:`\chi^2`.
+Validation of a single node
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Validation can be added to a node in the runcard by adding an entry ``validator``.
```
Edoardo-Pedicillo, line 7:
```suggestion
The validation step can be added to a node in the runcard by adding an entry ``validator``.
```

src/qibocal/auto/validation.py
```diff
@@ -1,30 +1,62 @@
...
+    parameters: Optional[dict] = field(default_factory=dict)
+    """"Validator parameters."""
+    outcomes: Optional[list[Outcome]] = field(default_factory=list)
+    """Depending on the validation we jump into one of the possible nodes."""
+    method: Callable[[Results, Target], Union[Status, str]] = field(init=False)
+    """Validation method corresponding to the scheme attribute."""
+
+    def __post_init__(self):
+        # initialize validator method (it will also check parameters with pydantic)
+        self.method = VALIDATORS[self.scheme](**self.parameters)
```
Edoardo-Pedicillo, line 35: For more modularity, `Validator` could be an abstract class and `VALIDATORS` be an enum class collecting all the `Validator`(s) provided by Qibocal (like `Routines`). 

I understand that now `VALIDATORS` is a dictionary for mutability, but I think we don't need it, since the executor should play with a `Validator` and it doesn't care if it is provided by Qibocal or by the user.

andrea-pasquale: Initially `VALIDATORS` was an `Enum` we decided to go for a `dict` just to have the possibility to modify it at `runtime` through a plugin mechanism #508. I know that we still need a proof-of-concept for the plugin mechanism but until we do we can keep the `dict` approach @alecandido

alecandido: Unfortunately, it's not possible to extend `Enum`s...

https://docs.python.org/3/howto/enum.html#restricted-enum-subclassing

Edoardo-Pedicillo: > I understand that now `VALIDATORS` is a dictionary for mutability, but I think we don't need it, since the executor should play with a `Validator` and it doesn't care if it is provided by Qibocal or by the user.

Could you elaborate more? I don't see the problem with my approach, `VALIDATORS` could just contain the Validators provided by Qibocal and `Validator` is an abstract class. Then, in some way the Executor should be able to allocate the user's Validator and use it. My idea is more oriented to an API approach, so I am not sure it fits well with Qibocal.

Edoardo-Pedicillo (approved):

PR #672, opened 2023-12-13
Return url when running `qq upload`

### LINKED ISSUE #671: Option to keep job alive if a routine fails
While working on https://github.com/qiboteam/qibolab_platforms_qrc/pull/73 I noticed the following things that may be useful in other applications:
1. We could have an option in `qq auto` (and maybe `qq acquire`), similar to `--no-update`, that keeps the execution alive and goes to the next routine if one of the routines fails. I think now the whole job will stop if a routine fails. I am not sure if there is a clean way to implement this other than the ugly try-(global) except.
2. Completely unrelated to 1: the [function that uploads reports](https://github.com/qiboteam/qibocal/blob/5f101333aeef1d589f5f7c3875a20477978d37b5/src/qibocal/cli/upload.py#L26) currently uses logging to print the url and returns nothing. It could be useful to return at least the url as a string, for applications that use this as part of a Python script, not via the `qq upload` cli, and would like to pass the url somewhere else (such as the bot post on GitHub).

Not a priority, I am not sure if there are other applications, but at least 2 is very simple to fix (just need a `return url`).

andrea-pasquale:
Closes #671 (point 2).

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/cli/upload.py
+++ b/src/qibocal/cli/upload.py
@@ -91,3 +91,4 @@ def upload_report(path: pathlib.Path, tag: str, author: str):
 
     url = urljoin(ROOT_URL, randname)
     log.info(f"Upload completed. The result is available at:\n{url}")
+    return url


stavros11 (approved):
Thanks @andrea-pasquale. Following our discussion you can close the issue with this PR since there is no strong motivation to do point 1.

GabrielePalazzo (approved):

PR #674, opened 2023-12-18
Adding ramsey sequences signal protocol

### LINKED ISSUE #673: Restore ramsey_sequence with signal
After discussing with @Boulos-Alfakes, we think that it could be useful to add a `ramsey_sequences_signal` protocol to be able to run large scans without having memory issues.

andrea-pasquale:
Closes #673.
This protocol could be useful to run long scans without using sweepers. It might be replaced by the unrolled version #666 soon. 

@GabrielePalazzo @PiergiorgioButtarini whenever you have time could test it on qblox to see if the output is the same?
Here is an example of a runcard:

```yaml
actions:
    - id: ramsey
      priority: 0
      operation: ramsey_sequences_signal #ramsey_signal
      # main: single shot classification
      parameters:
        delay_between_pulses_start: 16 # must be a multiple of 4 incl 0
        delay_between_pulses_end: 200
        delay_between_pulses_step: 25 # must be a multiple of 4
        n_osc: 5

```

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


GabrielePalazzo (approved):
Everything seems ok to me (tested on qblox).

stavros11:
Out of curiosity, is this only needed for doing non-detuned Ramsey with sequences without sweepers? Because last week Boulos also contacted me about a similar issue and I remember that using the existing Ramsey would fall back to sequences if I set `n_osc > 0`.

andrea-pasquale:
> Out of curiosity, is this only needed for doing non-detuned Ramsey with sequences without sweepers? Because last week Boulos also contacted me about a similar issue and I remember that using the existing Ramsey would fall back to sequences if I set `n_osc > 0`.

When he spoke to me I think that the major issue was the fact that we wanted to run long scans, therefore he needed to use the `*_sequences` protocol, showing on the y axis the voltages. Before this PR we were supporting only one ramsey protocol with sequences that was using probabilities and not signal. I don't know if he needs to run a detuned one or not.

stavros11:
Yes, I think the existing ramsey_signal uses sweeper when `n_osc == 0` and sequences when `n_osc > 0`: https://github.com/qiboteam/qibocal/blob/0e617fb4dfe531141046c5a2429da03a17cb38bf/src/qibocal/protocols/characterization/ramsey_signal.py#L144

So in the end it should be equivalent to the new routine here for `n_osc > 0`.

andrea-pasquale:
Oh yes, you are right. I will double check with @Boulos-Alfakes and I can close the PR.

Edoardo-Pedicillo (approved):
Thank @andrea-pasquale, LGTM

PiergiorgioButtarini (approved):

andrea-pasquale:
I think that probably we will not need this PR. Let me close it for now and if it will be necessary I will open the PR again.

PR #677, opened 2023-12-20
Hotfix for rabi period estimate

### LINKED ISSUE #656: Inaccurate fitted amplitude in rabi_amplitude_msr

andrea-pasquale:
Closes #656.
I think there was a problem with some simplification in the correction formula for the rabi period estimate.

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/characterization/rabi/utils.py
+++ b/src/qibocal/protocols/characterization/rabi/utils.py
@@ -213,4 +213,5 @@ def period_correction_factor(phase: float):
 
     """
 
-    return np.round(1 + phase / np.pi) - phase / np.pi
+    x = phase / np.pi
+    return np.round(1 + x) - x
--- a/src/qibocal/protocols/characterization/rabi/utils.py
+++ b/src/qibocal/protocols/characterization/rabi/utils.py
@@ -213,4 +213,4 @@ def period_correction_factor(phase: float):
 
     """
 
-    return 1 - np.modf(phase / np.pi)[0]
+    return np.round(1 + phase / np.pi) - phase / np.pi


src/qibocal/protocols/characterization/rabi/utils.py
```diff
@@ -213,4 +213,4 @@ def period_correction_factor(phase: float):
 
     """
 
-    return 1 - np.modf(phase / np.pi)[0]
+    return np.round(1 + phase / np.pi) - phase / np.pi
```
AleCandido, line 216:
```suggestion
    x = phase / np.pi
    return np.round(1 + x) - x
```
Any better name than `x` is a further improvement.

AleCandido (approved):

AleCandido:
My bad, the two functions are actually equal only in a certain range, and I see why you want the current one
round-vs-modf

I should have tested better...

Another option, that is exactly what you're doing, would be:
```py
1.5 - ((x + 0.5) % 1)
```
round-vs-modf-vs-mod

But it is really a matter of taste...

andrea-pasquale:
Thanks for checking @AleCandido.
Indeed I saw the same with behavior with `np.modf`. For the time being let's keep `np.round`.

AleCandido:
Btw, is there why you fixed the codomain between -0.5 and 1.5?

andrea-pasquale:
> Btw, is there a reason why you fixed the codomain between 0.5 and 1.5?

Which codomain?

AleCandido:
> Which codomain?

The y-axis in the plot (the codomain of the function you're modifying in this PR)

andrea-pasquale:
> > Which codomain?
> 
> The y-axis in the plot (the codomain of the function you're modifying in this PR).
> 
> Essentially, the amplitude is fixed to be 1 by `np.round()` (and all its friends), and you're deciding to center the range at 0, and to make `f(0) = 1`.

I see. Maybe it is better to go a step behind. The (half-)period corrected taking into account the phase should be the following:

$$ \frac{T^*}{2} = \underbracket{\Big(  k - \frac{\phi}{\pi}\Big) }_{\xi} \frac{T}{2}  $$

and since I now that this correction will be small I want to find $k \in \mathbb{Z}$ which minimizes this correction. Which means that I need to find $k$ s.t. $k - \frac{\phi}{\pi} \approx 1$. Therefore $k = \text{int}( 1 + \phi / \pi)$ and the total correction will be  $\xi = \text{int}( 1 + \phi / \pi) - \phi/\pi$. The problem as far as I understood is how to treat this conversion to `int`.

Given that we want $k$ to be as close as possible to the float this is why I think that it is better to use `np.round` instead of `np.modf`. Here are more plots.
image
image

AleCandido:
Yes, I keep forgetting it is a multiplicative correction, that's why you want the value to be around 1.

Edoardo-Pedicillo (approved):

PR #679, opened 2023-12-25
Fix TOF routine with new sampling rate approach

PiergiorgioButtarini:
This PR try to put a patch on issue https://github.com/qiboteam/qibocal/issues/680.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


src/qibocal/protocols/characterization/signal_experiments/time_of_flight_readout.py
```diff
@@ -21,6 +21,8 @@ class TimeOfFlightReadoutParameters(Parameters):
     """Amplitude of the readout pulse."""
     window_size: Optional[int] = 10
     """Window size for the moving average."""
+    sampling_rate: Optional[int] = 1e9
```
stavros11, line 24:
```suggestion
    sampling_rate: int = 1e9
```

stavros11 (commented):
Thanks for spotting this @PiergiorgioButtarini.
Another approach is to expose the sampling rate from qibolab via the platform and do `platform.sampling_rate` here. This will require a PR in qibolab (I can do this), but it may be slightly more user friendly as the user will not need to pass the correct sampling rate via the runcard (which is in some sense is also duplicated, as the instruments already have their sampling rate).
That being said, I am not sure why this routine needs the sampling rate. From a quick look at this PR it seems to me that it is only being used to normalize the units of time, but we could also use arbitrary units and let the user interpret them according to the used instruments.

PiergiorgioButtarini:
Thanks @stavros11, yes it's only for normalization purpose but I think it is more user friendly to do it rather than making the user interpret the time in arbitrary units.

andrea-pasquale (changes requested):
Thanks @PiergiorgioButtarini for this fix.
Similarly to @stavros11, I also cannot understand why the sampling rate should be a parameter.  Merely because you can pass any parameter and it will not be set in the instruments, right? If the sampling rate is needed just for normalization but it cannot be modified at runtime we should just find a way to expose it from the platform, for example as @stavros11 was suggesting here.

> Another approach is to expose the sampling rate from qibolab via the platform and do `platform.sampling_rate` here. This will require a PR in qibolab (I can do this), but it may be slightly more user friendly as the user will not need to pass the correct sampling rate via the runcard (which is in some sense is also duplicated, as the instruments already have their sampling rate).

The other solution is to completely remove it and show on the $x$ axis arbitrary units.

I slightly prefer to first solution but I am fine with both.

PiergiorgioButtarini:
The PR is now taking into account the features introduced in https://github.com/qiboteam/qibolab/pull/732.

andrea-pasquale:
> The PR is now taking into account the features introduced in [qiboteam/qibolab#732](https://github.com/qiboteam/qibolab/pull/732).

Thanks @PiergiorgioButtarini for the updates. Tests will probably fail now since in the workflows qibocal is using the stable release of qibolab (this is why I added the `require qibolab release` label). As soon we make a new release of qibolab we can merge this PR.

andrea-pasquale (approved):

Edoardo-Pedicillo (approved):

PR #682, opened 2023-12-27
Use only `QubitId` or `QubitPairId` in protocols

andrea-pasquale:
This PR simplifies the layout by considering only lists of `QubitId` or of `QubitPairId` in qibocal.
Before the code was often jumping from `list[QubitId` to `dict[QubitId, Qubit]` see https://github.com/qiboteam/qibocal/pull/668#discussion_r1434327128.
Moreover, I decided to rename the argument qubit(s) into target(s) since it was a bit weird for two qubit gates protocol to call each single element qubit instead of qubit pair. I thought that `target` should be generic enough to accomodate the meaning of `qubit` or `qubit_pair`. In each function the type is correctly specified by the type hint.

Given that I modified a lot of files I want to test this PR on deeply on hardware before merging.

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`


andrea-pasquale:
The PR is quite big given that I had to change something in all protocols.
For a short summary I check the first comment on the PR. 
Maybe renaming `qubits` -> `targets` it could be misleading but I think that it should be an improvement with respect to qubits.

AleCandido (approved):
Every time that there is a change affecting all the routines, I have a gut reaction towards rejection.

But in this case is mostly an extensive renaming. It's a huge diff, but if I'm not wrong (that is easier, because of the size) it should be pretty much straightforward.

The base idea is simple and sensible.

Edoardo-Pedicillo (approved):
Thank @andrea-pasquale

PR #689, opened 2024-01-12
Refactor drag pulse calibration

### LINKED ISSUE #645: Refactor drag pulse calibration
See pag. 109-110 in https://qudev.phys.ethz.ch/static/content/science/Documents/phd/PhD_Thesis_Baur_Matthias.pdf

andrea-pasquale:
Closes #645 

- [x] unrolling 
- [x] errors

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`


andrea-pasquale:
Some results on hardware:
http://login.qrccluster.com:9000/H7hD8OBlSp2O-Uzve-1kvg==/

The fit still needs to be improved while the assignment fidelity tends to lower a bit.

andrea-pasquale:
Currently the fit is performed using a cosine curve, however it seems that adding also an exponential decay seems to improve the quality of the fit:
image

Thoughts @igres26?

andrea-pasquale:
@stavros11 @hay-k indeed using the relative [phase](http://login.qrccluster.com:9000/4UC5gDs7RIagLv2ExeKIEQ==) or the [negative amplitude](http://login.qrccluster.com:9000/R5gHZs8sTyORNNSgqM-xgg==) gives the same results. For now I will keep using the relative phase.

andrea-pasquale:
I think the PR should be ready to review.
There is a problem with unrolling, we can drop it if we don't manage to solve it.

src/qibocal/protocols/characterization/drag.py
```diff
@@ -0,0 +1,283 @@
...
+    elif not params.unrolling:
+        results = [
+            platform.execute_pulse_sequence(sequence, options) for sequence in sequences
+        ]
+
+    for ig, (beta, ro_pulses) in enumerate(zip(beta_param_range, all_ro_pulses)):
+        for qubit in targets:
+            serial = ro_pulses[qubit].serial
+            if params.unrolling:
+                result = results[serial][0]
```
Jacfomg, line 121:
```suggestion
                result = results[serial][ig]
```

src/qibocal/protocols/characterization/drag.py
```diff
@@ -0,0 +1,283 @@
...
+from scipy.optimize import curve_fit
+from scipy.signal import find_peaks
+
+from qibocal import update
+from qibocal.auto.operation import Data, Parameters, Results, Routine
+from qibocal.config import log
+
+from .utils import COLORBAND, COLORBAND_LINE, chi2_reduced, table_dict, table_html
+
+# TODO: implement unrolling
```
Jacfomg, line 20:
```suggestion
```

src/qibocal/protocols/characterization/drag.py
```diff
@@ -0,0 +1,283 @@
...
+
+    data: dict[QubitId, npt.NDArray[DragPulseTuningType]] = field(default_factory=dict)
+    """Raw data acquired."""
+
+
+def _acquisition(
+    params: DragPulseTuningParameters,
+    platform: Platform,
+    targets: list[QubitId],
+) -> DragPulseTuningData:
```
Jacfomg, line 68: I would like something shorter like `DragTuning` or `DragPulse` but fell free to ignore.

Jacfomg (approved):
>[!NOTE]
> It looks good to me.

igres26 (commented):
I have tried to see if it improved the allxy data. 

Before: http://login.qrccluster.com:9000/AOngNCPwRRmskMRfQEua4Q==/
Drag: http://login.qrccluster.com:9000/0V3aJpG1TEyS-nt4rctP9Q==/
After: http://login.qrccluster.com:9000/14aE99oKR5Cn3ueGOPb46g==/

It seems to improve qubit 0 and not really the other ones. We can looks a bit into this in more detail before merging?

The routine works well, but we should make sure it works well in conjunction with other results.

andrea-pasquale:
> I have tried to see if it improved the allxy data.
> 
> Before: http://login.qrccluster.com:9000/AOngNCPwRRmskMRfQEua4Q==/ Drag: http://login.qrccluster.com:9000/0V3aJpG1TEyS-nt4rctP9Q==/ After: http://login.qrccluster.com:9000/14aE99oKR5Cn3ueGOPb46g==/
> 
> It seems to improve qubit 0 and not really the other ones. We can looks a bit into this in more detail before merging?
> 
> The routine works well, but we should make sure it works well in conjunction with other results.

Yes, you are right. For qubit 2 and 3 the results are unclear. The protocol is doing the same for all qubits. The only noticable difference is that qubit 2 and 3 have a negative parameter, maybe this is causing the issue. Any ideas @stavros11?

If we can confirm that the issue is only in the driver I would proceed to merge otherwise let's wait.

Edoardo-Pedicillo (approved):
LGTM

stavros11:
> Yes, you are right. For qubit 2 and 3 the results are unclear. The protocol is doing the same for all qubits. The only noticable difference is that qubit 2 and 3 have a negative parameter, maybe this is causing the issue. Any ideas @stavros11?

Thanks for checking @igres26 @andrea-pasquale. Looking at the platform parameters used in the "after drag" report provided above, you are indeed using negative values in the DRAG definition for qubits 2 and 3. I have not confirmed, but most likely this triggers qiboteam/qibolab#826 and it is ignoring the minus sign during deserialization, which will eventually lead to the wrong shape.

Note that this is a `PulseShape` issue, not a driver issue, so it will most likely appear on all instruments. We originally planned this for 0.2, and it is already fixed there, but since it is becoming a problem I will try to have a look if we can easily fix it for 0.1.

andrea-pasquale:
> Thanks for checking @igres26 @andrea-pasquale. Looking at the platform parameters used in the "after drag" report provided above, you are indeed using negative values in the DRAG definition for qubits 2 and 3. I have not confirmed, but most likely this triggers [qiboteam/qibolab#826](https://github.com/qiboteam/qibolab/issues/826) and it is ignoring the minus sign during deserialization, which will eventually lead to the wrong shape.

Thanks @stavros11, since the fit looked good for all qubits I thought it could have been a qibolab issue.



> Note that this is a `PulseShape` issue, not a driver issue, so it will most likely appear on all instruments. We originally planned this for 0.2, and it is already fixed there, but since it is becoming a problem I will try to have a look if we can easily fix it for 0.1.

Depending on when qibolab 0.2 will happen it would be nice to have the fix in qibolab if people are planning to use DRAG. Even if you don't make a new 0.1.* release for qibolab we could keep at least qibocal main in sync with qibolab main.

PR #698, opened 2024-01-26
Updating mechanism to dump platforms

andrea-pasquale:
This updates the mechanism to dump platforms modified by https://github.com/qiboteam/qibolab/pull/758

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/cli/acquisition.py
+++ b/src/qibocal/cli/acquisition.py
@@ -3,7 +3,7 @@ import json
 from dataclasses import asdict
 
 import yaml
-from qibolab.serialize import dump_runcard
+from qibolab.serialize import dump_platform
 
 from ..auto.execute import Executor
 from ..auto.history import add_timings_to_meta
@@ -39,7 +39,8 @@ def acquire(runcard, folder, force):
     meta = generate_meta(backend, platform, path)
     # dump platform
     if backend.name == "qibolab":
-        dump_runcard(platform, path / PLATFORM)
+        (path / PLATFORM).mkdir(parents=True, exist_ok=True)
+        dump_platform(platform, path / PLATFORM)
 
     # dump action runcard
     (path / RUNCARD).write_text(yaml.safe_dump(asdict(runcard)))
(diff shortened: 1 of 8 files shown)


andrea-pasquale:
Given that tests are passing only here after updating the `poetry.lock` (because of https://github.com/qiboteam/qibolab/pull/758). I suggest to merge in the following order:

#698 -> #679 -> #688

In this way we should be able to merge to main without breaking CI.

alecandido:
> Given that tests are passing only here after updating the `poetry.lock` (because of [qiboteam/qibolab#758](https://github.com/qiboteam/qibolab/pull/758)). I suggest to merge in the following order:
> 
> #698 -> #679 -> #688
> 
> In this way we should be able to merge to main without breaking CI.

As usual, I'd rather avoid merging PRs in other PRs, because otherwise you will have to review the code twice (so the original changes will be masked by those already approved).

Can't we just merge this in `main`?

andrea-pasquale:
> As usual, I'd rather avoid merging PRs in other PRs, because otherwise you will have to review the code twice (so the original changes will be masked by those already approved).

I was suggesting just because each PRs contains very small changes.


> Can't we just merge this in `main`? I mean, after rebasing on main just the three commits of this PR.

This is also an option. You mean to close the other two PRs and merge this one directly in main?

alecandido:
> This is also an option. You mean to close the other two PRs and merge this one directly in main?

I didn't know they were small. If they are, we could even do that.

No, my idea was to use `git rebase -i main`, and only retain the last 3 commits, that are those only present in this PR.
I.e. instead of merging the full stack in `main` in a single shot (reviewing altogether or step by step), we could revert the order of the commits.

However, if they are few changes, just merge all of them in a single PR, i.e. merge this one and #679 in #688, and then merge #688, or equivalently close the other two and point this to `main`.

andrea-pasquale:
> However, if they are few changes, just merge all of them in a single PR, i.e. merge this one and #679 in #688, and then merge #688, or equivalently close the other two and point this to `main`.

I see, let me merge everything into #688 and then merge that to main. Thanks!

PR #705, opened 2024-02-06
Adding `try-except` block in flux fits

andrea-pasquale:
To avoid errors which stop the execution when running flux dependence experiments.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/characterization/flux_dependence/qubit_flux_dependence.py
+++ b/src/qibocal/protocols/characterization/flux_dependence/qubit_flux_dependence.py
@@ -226,7 +226,6 @@ def _fit(data: QubitFluxData) -> Optional[QubitFluxResults]:
                 "Lowering the value of `threshold` in `extract_*_feature`"
                 "should fix the problem."
             )
-            return None
 
     return QubitFluxResults(
         frequency=frequency,
--- a/src/qibocal/protocols/characterization/flux_dependence/qubit_flux_dependence.py
+++ b/src/qibocal/protocols/characterization/flux_dependence/qubit_flux_dependence.py
@@ -50,8 +50,8 @@ class QubitFluxResults(Results):
     """Sweetspot for each qubit."""
     frequency: dict[QubitId, float]
     """Drive frequency for each qubit."""
-    d: dict[QubitId, float]
-    """Asymmetry."""
+    asymmetry: dict[QubitId, float]
+    """Asymmetry between junctions."""
     fitted_parameters: dict[QubitId, dict[str, float]]
     """Raw fitting output."""
     matrix_element: dict[QubitId, float]
@@ -180,7 +180,7 @@ def _fit(data: QubitFluxData) -> Optional[QubitFluxResults]:
     qubits = data.qubits
     frequency = {}
     sweetspot = {}
-    d = {}
+    asymmetry = {}
     matrix_element = {}
     fitted_parameters = {}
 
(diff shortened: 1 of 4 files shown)


src/qibocal/protocols/characterization/flux_dependence/qubit_flux_dependence.py
```diff
@@ -203,20 +203,27 @@ def _fit(data: QubitFluxData) -> QubitFluxResults:
...
+                    data.qubit_frequency[qubit], qubit_data.bias
+                ),
+                maxfev=100000,
+            )[0]
+            fitted_parameters[qubit] = popt.tolist()
+            frequency[qubit] = popt[0] * GHZ_TO_HZ
+            d[qubit] = popt[1]
+            sweetspot[qubit] = popt[3]
+            matrix_element[qubit] = popt[2]
+        except:
```
hay-k, line 221: Can you add the specific type of the exception that is expected here? 

Also, perhaps it is a good idea to log the error message?

src/qibocal/protocols/characterization/flux_dependence/qubit_flux_dependence.py
```diff
@@ -203,20 +203,27 @@ def _fit(data: QubitFluxData) -> QubitFluxResults:
...
+                biases,
+                frequencies * HZ_TO_GHZ,
+                bounds=utils.qubit_flux_dependence_fit_bounds(
+                    data.qubit_frequency[qubit], qubit_data.bias
+                ),
+                maxfev=100000,
+            )[0]
+            fitted_parameters[qubit] = popt.tolist()
+            frequency[qubit] = popt[0] * GHZ_TO_HZ
+            d[qubit] = popt[1]
```
hay-k, line 218: what is `d`? :D Why does it not have a more descriptive name?

andrea-pasquale: A bit of physicist biased notation for sure.
It should be the asymmetry between the two junctions which is denoted in the literature with $d$.
Let me change it to `asymmetry`.

src/qibocal/protocols/characterization/flux_dependence/qubit_flux_dependence.py
```diff
@@ -203,25 +204,34 @@ def _fit(data: QubitFluxData) -> QubitFluxResults:
...
+            sweetspot[qubit] = popt[3]
+            matrix_element[qubit] = popt[2]
+        except ValueError as e:
+            log.error(
+                f"Error in qubit_flux protocol fit: {e} "
+                "The threshold for the SNR mask is probably too high. "
+                "Lowering the value of `threshold` in `extract_*_feature`"
+                "should fix the problem."
+            )
+            return None
```
andrea-pasquale, line 229: I'm having second thoughts about this `return None`.
For sure like this it will not work because I'm also looping over `qubits`, which I didn't consider before...
So here it will return `None` and remove all data of any previous fit successful for other qubits.
A proposal could be to just append the key is the fitting fail and check in the plotting functions (which tackle only one qubit at a time) if that particular `qubit` is in `Results`. 
I will also try to fix the platform update.
Thoughts @hay-k?

Jacfomg (approved):
Thanks @andrea-pasquale

andrea-pasquale:
@hay-k when you have time could you have a look at this PR? I would like to have this merged in the next release.

src/qibocal/protocols/characterization/flux_dependence/qubit_flux_dependence.py
hay-k: sorry, didn't reply to your question earlier. I believe your latest changes address this, and we discussed it AFK and. Let me know if any concern still remains.

hay-k (approved):

PR #597, opened 2023-11-01
Changing RB data

Jacfomg:
Changing RB data to stop using Pandas and align with the rest of qibocal. Still work in progress, but it runs.

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.
- [ ] Compatibility with Qibo modules (Please edit this section if the current pull request is not compatible with the following branches).
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`
    - [ ] Qibolab_platforms_qrc: `main`

PR #707, opened 2024-02-08
Partial refactor in `fit` functions

andrea-pasquale:
In the current version of the code the fitting functions populate a specific instance of `Results` depending on the protocol executed. `Results` is intended as class where attributes are of the form `dict[Union[QubitId, QubitPairId]` and each attribute usually corresponds to a parameter that will be written in the platform. Whenever the fitting failed for a particular qubit in some instances we were constructing some sort of default parameters to pass in `Results` for some protocols.
Given that those parameters were totally nonphysical I decided to append in `Results` only parameters from successful fits. Everything else has been updated to support this fix.
With this approach if the fitting fails for a specific qubit it will not be shown in the corresponding HTML reports or stored.

I would like to merge after this #597 to avoid some weird shenanigans that I am doing here.

 
Checklist:
- [x] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


src/qibocal/protocols/characterization/randomized_benchmarking/standard_rb.py
```diff
@@ -75,17 +75,21 @@ def __post_init__(self):
...
-    fit_uncertainties: tuple[float, float, float]
+    fit_uncertainties: dict[QubitId, tuple[float, float, float]]
     """Fitting parameters uncertainties."""
-    error_bars: Optional[Union[float, list[float]]] = None
+    error_bars: dict[QubitId, Optional[Union[float, list[float]]]] = None
     """Error bars for y."""
 
+    # FIXME: fix this after https://github.com/qiboteam/qibocal/pull/597
+    def __contains__(self, qubit: QubitId):
+        return True
```
andrea-pasquale, line 91: This is the shenanigan that I want to remove before merging...
For this I need #597.

src/qibocal/auto/operation.py
```diff
@@ -222,6 +222,12 @@ def load(cls, path: Path):
 class Results(AbstractData):
     """Generic runcard update."""
 
+    def __contains__(self, qubit: Union[QubitId, QubitPairId]):
+        """Checking if qubit is in Results."""
+        if isinstance(qubit, list):
```
hay-k, line 227: why would it be a list?

src/qibocal/cli/report.py
```diff
@@ -77,8 +77,12 @@ def routine_qubits(self, task_id: TaskId):
     def single_qubit_plot(self, task_id: TaskId, qubit: QubitId):
         """Generate single qubit plot."""
         node = self.history[task_id]
+        if node.results is not None:
+            fit = node.results if qubit in node.results else None
+        else:
+            fit = None
```
hay-k, line 83:
```suggestion
        fit = node.results if node.results and qubit in node.results else None
```

src/qibocal/protocols/characterization/qubit_spectroscopy.py
```diff
@@ -121,11 +122,14 @@ def _fit(data: QubitSpectroscopyData) -> QubitSpectroscopyResults:
...
-        )
-        frequency[qubit] = freq
-        fitted_parameters[qubit] = fitted_params
+        try:
+            freq, fitted_params = lorentzian_fit(
+                data[qubit], resonator_type=data.resonator_type, fit="qubit"
+            )
+            frequency[qubit] = freq
+            fitted_parameters[qubit] = fitted_params
+        except RuntimeError:
```
hay-k, line 131: What is the exact error message you expect to get here, or in the other places where lorentzian fit is done? 

I was going to suggest to move the try-catch logic inside the `lorentzian_fit` function to avoid code repetition, but I realized that there is already a try-catch in it that catches RutimeError. So I am wondering what is left for the try-catch here.

src/qibocal/auto/operation.py
andrea-pasquale: Initially I added this line because tests were failing for `readout_mitigation_matrix`.
After digging through the code I understood that there was an error in the type hint of `Action` that was causing tests to fail. 
For two qubit/ multi-qubit protocols it could be a list, the choice `qubit` is a bit unfortunate... I will change to something else. Any suggestions?

src/qibocal/protocols/characterization/qubit_spectroscopy.py
andrea-pasquale: You are right...
In https://github.com/qiboteam/qibocal/pull/707/commits/cab4435779da4ccd54aef834934973d1bcd7e097 I tried to remove some code duplication by moving the `try-except` inside `lorentzian_fit`. Let me know if you like it.

src/qibocal/auto/operation.py
```diff
@@ -222,6 +222,10 @@ def load(cls, path: Path):
 class Results(AbstractData):
     """Generic runcard update."""
 
+    def __contains__(self, key: Union[QubitId, QubitPairId, tuple[QubitId, ...]]):
+        """Checking if qubit is in Results."""
+        return all(key in getattr(self, field.name) for field in fields(self))
+
```
Edoardo-Pedicillo, line 228: I would suggest to explain better the purposes of this function (taking in consideration non expert users). For more generality, you could check if an attribute is a dictionary and apply the qubit check only there.

andrea-pasquale: Done here https://github.com/qiboteam/qibocal/pull/707/commits/97d4227a15733eff572f3f29eb10ae3c690f02c9

src/qibocal/auto/operation.py
```diff
@@ -222,6 +222,23 @@ def load(cls, path: Path):
 class Results(AbstractData):
     """Generic runcard update."""
 
+    def __contains__(self, key: Union[QubitId, QubitPairId, tuple[QubitId, ...]]):
+        """Checking if qubit is in Results.
+
+        Used by the ReportBuilder to check whether an instance of Results
```
hay-k, line 228: I would avoid putting in the docstring of a function where and why it is used. This type of documentation is bound to get outdated without being noticed for a long time :D You can put this somewhere in `ReportBuilder` itself.

src/qibocal/auto/operation.py
```diff
@@ -276,7 +293,7 @@ def data_type(self):
 
     @property
     def results_type(self):
-        """ "Results object type return by data acquisition."""
+        """Results object type return by data acquisition."""
```
hay-k, line 296:
```suggestion
        """Results object type returned by data acquisition."""
```

src/qibocal/auto/operation.py
andrea-pasquale: https://github.com/qiboteam/qibocal/pull/707/commits/69a8b4381282e9dbee71b6d01577ad2980c86b6e

hay-k (approved):

Edoardo-Pedicillo (approved):

PR #710, opened 2024-02-12
Dump platform iteratively

### LINKED ISSUE #709: Dump platform after each node
In order to stop and start calibration from a specific node.
Useful feature to split the calibration process.

andrea-pasquale:
Closes #709.
A `platform` folder will now appear under each `protocol` folder in the report.
I've also made tests a bit more robust by checking if certain files/folders are present.

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`

### DIFF
--- a/src/qibocal/auto/task.py
+++ b/src/qibocal/auto/task.py
@@ -240,8 +240,8 @@ class Completed:
         if self.task.update and update:
             for qubit in self.task.qubits:
                 self.task.operation.update(self.results, platform, qubit)
-            (self.datapath / "platform").mkdir(parents=True, exist_ok=True)
-            dump_platform(platform, self.datapath / "platform")
+            (self.datapath / PLATFORM_FOLDER).mkdir(parents=True, exist_ok=True)
+            dump_platform(platform, self.datapath / PLATFORM_FOLDER)
 
     def validate(self) -> tuple[Optional[TaskId], Optional[dict]]:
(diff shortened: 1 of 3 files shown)


src/qibocal/auto/task.py
```diff
@@ -33,6 +34,8 @@
 """A number bigger than whatever will be manually typed. But not so insanely big not to fit in a native integer."""
 TaskId = tuple[Id, int]
 """Unique identifier for executed tasks."""
+PLATFORM_FOLDER = "platform"
```
hay-k, line 37: minor: I would prefer to call it `PLATFORM_DIR` or `PLATFORM_DIRECTORY`

hay-k (approved):

alecandido (approved):

src/qibocal/auto/task.py
alecandido: @hay-k is there a reason for that? I always used the two interchangeably...

(i.e. for me it's also fine your proposal, no problem with that)

hay-k: @alecandido yeah, I myself use them interchangeably as well, especially when speaking. In code like this, I prefer to use dir/directory because it is the strict file system terminology. Folder seems to be more like a concept popularized by Microsoft Windows, is not necessarily a file system directory, and has some graphical/UI related connotations attached to it.

PR #714, opened 2024-02-15
qq compare command to compare two qibocal reports

GabrielePalazzo:
Example of a combined report containing a resonator spectroscopy and a resonator flux dependence:
http://login.qrccluster.com:9000/T5rbRzsmTRmKLtOSaa03DQ==

Another example:
http://login.qrccluster.com:9000/o73U5W5dTuKIjl_8wCkTzg==/

Only experiments with the same id are compared. All other experiments are discarded.
Tables contain results from both experiments side-by-side.
Experiments with scatterplots (e.g.: resonator spectroscopy) are shown in the same figure (with overlapping points and lines).
Plots of experiments with heatmaps (e.g.: resonator punchout) are vertically stacked.

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`


andrea-pasquale:
Missing `lxml` dependency.

Edoardo-Pedicillo:
Hi @GabrielePalazzo, what is the status of this PR ?

GabrielePalazzo:
> Hi @GabrielePalazzo, what is the status of this PR ?

Comparison of single sweeper experiments (like resonator spectroscopy) works.
Flux experiments or readout characterization are not working (we should probably not compare heatmap plots). If I have some spare time later this week I’ll try to make it ready for review.

Edoardo-Pedicillo:
> Flux experiments or readout characterization are not working (we should probably not compare heatmap plots)

I agree. If you want, you can just put the two plots near each other.

andrea-pasquale:
> > Flux experiments or readout characterization are not working (we should probably not compare heatmap plots)
> 
> I agree. If you want, you can just put the plots of the two reports near each other.

I was going to suggest the same. Another possible idea would be to come up with some sort of ratio plot to see the differences between the two outcomes. I don't know how feasible it is.

GabrielePalazzo:
The new code is not covered by tests. I'll try to write some.

Edoardo-Pedicillo (approved):

andrea-pasquale:
> @GabrielePalazzo, could you resolve the conflicts? Thanks

For conflicts let me know if you need help. I recently changed some stuff related to the report generation.

src/qibocal/cli/compare.py
```diff
@@ -0,0 +1,184 @@
...
+            for plot in plots[1:]:
+                merged_plot_data = merged_plot_data + plot[0].data
+            fig = go.Figure(data=merged_plot_data, layout=plots[0][0].layout).to_html()
+        elif plots[0][0].data[0].type == "heatmap":
+            html_block = lxml.html.fromstring(plots[0][0].to_html())
+            html_block[1].remove(html_block[1][0])
+            for i, plot in enumerate(plots):
+                single_report_html_block = lxml.html.fromstring(plot[0].to_html())
+                html_block[1].insert(i, single_report_html_block[1][0])
+            fig = lxml.etree.tostring(html_block).decode("ascii")
```
alecandido, line 141: In addition to my previous comment.

The explicit HTML manipulation seems to be localized here.
Isn't it possible to avoid the HTML manipulation at all?

Moreover, it seems that the function is:
- extracting the original HTML from the first plot, then into an object
- removing the plot itself (I guess)
- reinserting all the plots in the HTML as consecutive nodes
- converting back to an HTML string

Can't you just manipulate the Plotly objects, and stack them vertically?

pyproject.toml
```diff
@@ -36,6 +36,7 @@ seaborn = { version = "^0.12.2", optional = true }
 pydot = { version = "^1.4.2", optional = true }
 skl2onnx = { version = "^1.14.0", optional = true }
 pyyaml = "^6.0"
+lxml = "^5.1.0"
```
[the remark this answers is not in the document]
GabrielePalazzo: Yeah, for some reason I found it easier to use `lxml` for manipulating the html code, but I can try to switch to a standard package.

src/qibocal/cli/compare.py
GabrielePalazzo: From what I remember, vertically stacking two plots from HTML reports was giving weird layouts (e.g.: plots overlapping one on top of the other).

alecandido: I mean, doing something like this
https://plotly.com/python/subplots/#stacked-subplots

Since, initially, you should have the Plotly objects (those on which you're calling the `.to_html()` method)

GabrielePalazzo: I think I previuosly had issues with subplots size.
I tried again, using the default plotly settings as much as possible and it seems to be working just fine.

http://login.qrccluster.com:9000/o73U5W5dTuKIjl_8wCkTzg==/

tests/test_cli.py
```diff
@@ -1,11 +1,16 @@
+import json
 import pathlib
+import re
 
 import pytest
 from click.testing import CliRunner
+from lxml.html import parse
```
alecandido, line 7: Not needed after ea60341

pyproject.toml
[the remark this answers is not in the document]
alecandido: Ok, now it is not needed any longer, after ea60341

src/qibocal/cli/compare.py
alecandido: Great, thanks :D

pyproject.toml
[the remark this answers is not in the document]
GabrielePalazzo: It is still required by pandas for `read_html` when I merge the tables from the two reports (here: https://github.com/qiboteam/qibocal/blob/258f415869208d325eba92d623d7c3e268ae2630/src/qibocal/cli/compare.py#L168).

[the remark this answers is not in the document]
alecandido: If Pandas requires it, it will be its own dependency :)

(thus transitive for Qibocal)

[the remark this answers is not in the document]
alecandido: I believe you should add the `html` extra:
https://pandas.pydata.org/docs/getting_started/install.html#html
I will do it, and you can check if you agree.

In general, it is also interesting to check the following:
https://pandas.pydata.org/docs/user_guide/io.html#io-html-gotchas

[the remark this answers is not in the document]
GabrielePalazzo: It is an explicit optional dependency starting from pandas v2.0.0. For versions >= 1.4 and < 2.0 I don't see it as a dependency, even though in pandas' documentation it is clearly stated that it's required: https://pandas.pydata.org/pandas-docs/version/1.4/getting_started/install.html#html.

pyproject.toml
```diff
@@ -17,11 +17,11 @@ classifiers = [
...
 [tool.poetry.dependencies]
 python = ">=3.9,<3.12"
-qibolab = {git = "https://github.com/qiboteam/qibolab.git", branch = "main" }
-qibo ="^0.2.6"
+qibolab = { git = "https://github.com/qiboteam/qibolab.git", branch = "main" }
+qibo = "^0.2.6"
 numpy = "^1.26.4"
 scipy = "^1.10.1"
-pandas = "^1.4.3"
+pandas = { version = "^1.4.3", extras = ["html"] }
```
GabrielePalazzo, line 24: @alecandido Are you sure this works?. I just tried to create a virtual environment from scratch, and `poetry install` does not install `lxml`.
If I change pandas version to `^2.0.0` then poetry does install lxml.

@andrea-pasquale is there a reason not to update pandas?

alecandido: Sorry, you are right.

As I said in the meeting, I expected Poetry to complain about the missing extra, but it didn't happen.
It didn't complain, and the extra was not there (other than the docs, I also checked the wheel metadata).

alecandido: Now it should be fixed by ade846f

tests/test_cli.py
alecandido: Here `parse` is still being used explicitly for testing.

We could maybe test in different ways (e.g. testing the `Report` class, instead of the HTML content), just leaving the HTML report generation to the template, and test in the same way of the usual reports (or not test them at all, according to what is done).

However, since the dependency is still required here, I left it in the `test` group, for the time being.

GabrielePalazzo: Indeed. I'll try to change the tests.

pyproject.toml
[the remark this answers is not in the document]
alecandido: Thanks @GabrielePalazzo for the input!

src/qibocal/cli/compare.py
```diff
@@ -0,0 +1,194 @@
...
+        """Merge plots from different reports.
+
+        Scatter plots are plotted in the same figure.
+        Heatmaps are vertically stacked.
+        """
+        if plots[0][0].data[0].type == "scatter":
+            merged_plot_data = plots[0][0].data
+            for plot in plots[1:]:
+                merged_plot_data = merged_plot_data + plot[0].data
+            fig = go.Figure(data=merged_plot_data, layout=plots[0][0].layout).to_html()
```
andrea-pasquale, line 130: Here there could be a problem because you are assuming that there is only one figure under each plot.
Here is an example where it fails:
image

image

andrea-pasquale (commented):
Thanks @GabrielePalazzo, I'm sorry for the very late review.
Overall, the code looks good. I've spotted a few bugs and I suggested some improvements.
Unfortunately after merging #909 there could be quite a lot to fix in terms of conflicts.
Regarding the improvements, I've played a bit with plotly and I found a few tricks to
make the legends appear in a better way in #937, feel free to have a look and let me know if you agree.

andrea-pasquale:
Thanks @GabrielePalazzo for https://github.com/qiboteam/qibocal/pull/714/commits/d19465755ae8e82fd4bea7149b07378802d666f1. As soon as  tests are passing we can merge :)

GabrielePalazzo:
> Thanks @GabrielePalazzo for [d194657](https://github.com/qiboteam/qibocal/commit/d19465755ae8e82fd4bea7149b07378802d666f1). As soon as tests are passing we can merge :)

I'd like to move the `History` too.

src/qibocal/web/compared_report.py
```diff
@@ -0,0 +1,227 @@
...
+        merged_figures = []
+        for fig0, fig1 in zip(plots[0], plots[1]):
+            for trace_0, trace_1 in zip(fig0.data, fig1.data):
+                trace_0.legendgroup = None
+                trace_1.legendgroup = None
+                trace_0.legend = "legend"
+                trace_1.legend = "legend2"
+
+            if any(isinstance(trace, go.Heatmap) for trace in fig0.data):
+                # TODO: check if this is valid for all protocols
```
andrea-pasquale, line 88: There are still problems with same 2D plots like Chevron, flux dependence or readout characterization. Of course we cannot have something that works for everything. It would be nice to support the readout characterization out of the box given that it would used for the monitoring most likely.
If it is too much of an issue we can postpone it for a later PR.

src/qibocal/cli/compare.py
```diff
@@ -0,0 +1,65 @@
...
+        css_styles=report_css_styles(STYLES),
+        path=folder,
+        title=combined_report.meta.title,
+        report=ComparedReport(
+            report_paths=paths,
+            folder=folder,
+            meta=combined_report.meta.dump(),
+        ),
+    )
+    combined_report.dump(folder)
```
andrea-pasquale, line 64: Here if the option `-o` is not specified it will fail.
At this point we can even turn into an argument.

andrea-pasquale (commented):
Thanks @GabrielePalazzo I have a few comments.
I start playing with `qq compare` and I have noticed that if I combined together two monitoring report it feels very slow: http://login.qrccluster.com:9000/nmg4AkeKRuCBKYp3M-NnrA==/
I think because of the fact that the HTML along is 20 MB. We don't have to address it here but I was wondering if there could be a way to produce less heavy reports. Any ideas @alecandido?

src/qibocal/cli/compare.py
GabrielePalazzo: It should work now. I think `-o` is preferrable, as we might want to change the argument of the command.

src/qibocal/web/compared_report.py
GabrielePalazzo: I'll have a look. In case it takes too long, we can do it later.

src/qibocal/cli/compare.py
andrea-pasquale: Thanks! Now it works!

src/qibocal/web/compared_report.py
andrea-pasquale: Ok, actually I would say that if it works just for `readout_characterization` it might be better because as we already discussed people will not compare much Heatmaps. I can also help with that.

alecandido:
> I think because of the fact that the HTML along is 20 MB. We don't have to address it here but I was wondering if there could be a way to produce less heavy reports. Any ideas @alecandido?

I'm pretty sure that the size of the HTML is all driven by plots (the rest of the page is incredibly minimal).
I still believe that the best is to decouple the frontend from Qibocal, and size optimization, lazy-loading of plots (or even multiple files splitting, even when they are loaded eagerly), and the HTML generation in general are all frontend features (well, frontend doesn't mean only the aesthetics).

To be fair, even `qq compare` is fully a frontend project, and I would have started from optimizing base reports before creating a more advanced infrastructure on top.
But since it was achievable (and maybe easier for someone else), that has not necessarily been a bad idea. At least you're getting the feature now.

However, regarding performances, I'd really advise putting here the threshold: this is definitely not the best way to optimize, and a black hole in terms of time required to improve.
Let's take it as it is, and schedule a proper overhauling of Qibocal's report interface from the basics, whenever we'll be more stable with other core features (or we'll have spare manpower, which is the same thing).

andrea-pasquale:
Another thing @GabrielePalazzo, you could add just a couple of sentences in the documentation that would be perfect.

andrea-pasquale (approved):

Edoardo-Pedicillo (approved):
LGTM, thanks !

Issue #712, opened 2024-02-14
RB data structure


andrea-pasquale:
Here is a  proposal for storing data form RB experiment.
After we merge #597 we will switch to `np.arrays`, however to be able to debug and reproduce the code some improvements could be performed.
In order to store both circuits and shots for each iteration and depth we could have something like this:

```py

@dataclass 
class RBData():

    data: dict[QubitId, Depth] = npt.NDArray[int]
    circuits: dict[QubitId, Depth]  npt.NDArray[int]

    ....

```
The idea is that for each qubit and depth we will store a matrix with dimension `Niterations` x `Shots` for `data` and
for `circuits` a matrix with dimension `Niterations` x `Depths`. For `circuits` each column will be filled with numbers in the range `0-23` that corresponds to the possible clifford single qubit gates, following the convention defined here: 
https://github.com/qiboteam/qibocal/blob/4576764a680ce23e208f647fcd8b5730f321d5bc/src/qibocal/protocols/characterization/randomized_benchmarking/utils.py#L11-L44
I believe that in this way it will be possible to easily run custom post-processing operations.

Jacfomg:
I think we can close this with the #718

Edoardo-Pedicillo:
:+1:

PR #717, opened 2024-02-16
quick patch to save circuits and numbers

Jacfomg:
@igres26 to save circs and number for the RB, this is a draft as I am not handling the file path and just saving it from the folder you ran it.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


andrea-pasquale:
At this point why don't we implement directly #712?

Jacfomg:
> At this point why don't we implement directly #712?

I would try to go over it once I fix some stuff #597, they can take a look at this if they really need to same the data now.

PR #718, opened 2024-02-16
Fix error in bootstrap.py and standard_rb.py

Jacfomg:
Fixes the gate dict and [wip] error bars

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/bootstrap.py
+++ b/src/qibocal/bootstrap.py
@@ -39,7 +39,7 @@ def data_uncertainties(data, method=None, data_median=None, homogeneous=True):
         )
         return uncertainties
 
-    return None
+    return None  # TODO: # Raise an error on invalid method
 
 
 def bootstrap(
--- a/src/qibocal/protocols/characterization/randomized_benchmarking/standard_rb.py
+++ b/src/qibocal/protocols/characterization/randomized_benchmarking/standard_rb.py
@@ -222,7 +222,7 @@ def _acquisition(
                 RBType,
                 (qubit_id, depth),
                 dict(
-                    samples=sample[nqubit],
+                    samples=1 - sample[nqubit],
                 ),
             )
     return data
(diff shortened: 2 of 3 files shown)


andrea-pasquale:
Is it ready to review?

Jacfomg:
> Is it ready to review?

Not yet, I would like to implement properly the #712 or someone could start it from here

Jacfomg:
I found another issue that for multi qubit we would generate one circuit per qubit as run them separately, we should create an option to choose between that or parallel circuits in another PR. From this, results seem to be more sensible now, without bootstrapping as it seems like I broke it.

src/qibocal/protocols/characterization/randomized_benchmarking/standard_rb.py
```diff
@@ -207,24 +209,23 @@ def _acquisition(
             for circuit in circuits
         ]
 
-    for i in executed_circuits:
-        samples.extend(i.samples())
-    nqubits = len(qubits_ids)
-    samples = np.reshape(samples, (-1, params.nshots, nqubits))
+    for circ in executed_circuits:
+        samples.extend(circ.samples())
+    samples = np.reshape(samples, (params.nshots, -1, nqubits))
```
Edoardo-Pedicillo, line 214: Are you sure ?

Jacfomg: I think so, as you are adding `circ.samples()` that would be an array with `nshots` element that would be the 1st axis, right ?

Jacfomg:
Looks good to me @Edoardo-Pedicillo

src/qibocal/bootstrap.py
```diff
@@ -22,9 +22,17 @@ def data_uncertainties(data, method=None, data_median=None, homogeneous=True):
...
-    if isinstance(method, (int, float)) and 0 <= method <= 100:
+
+    else:
+
+        if data_median is None:
+            data_median = (
+                np.median(data, axis=1)
+                if homogeneous
+                else [np.median(row) for row in data]
+            )
```
andrea-pasquale, line 35: Here we can just have two if statements I believe.

andrea-pasquale (approved):

Jacfomg:
Let's merge this into #597 and run it on hardware or ask for further reviews there

PR #720, opened 2024-02-19
Add support for sequence unrolling in flipping and flipping_signal

Jacfomg:
It adds support for sequence unrolling in flipping and flipping_signal as requested in https://github.com/qiboteam/qibolab_platforms_qrc/pull/119. It has only been tested with dummy.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


igres26 (approved):
Seems to work as expected in the tries made with quantum machines. Albeit the speed-up in that case is not very significant, it is still there.

andrea-pasquale (approved):
I've not tested the PR on hardware, but the code seems to be fine.

andrea-pasquale:
Maybe one quick thing, could you add also tests with the option `unrolling=True`?

Issue #696, opened 2024-01-23
Document Ramsey experiment


hay-k:
We need to add proper documentation to all experiments, but let's start with this one.

Yesterday I learned that the Ramsey experiment is not doing true frequency detuning, but rather an artificial/fake detuning based on manipulating the relative phase of a pulse. This is not documented, and people don't even clearly remember the motivation behind. Before this is completely forgotten, it should be documented. 
- There should be brief documentation explaining what is the experiment doing at all in high level. 
- There should be clear mentions of particular choices (like the handling of detuning mentioned above) done in the experiment, their motivation, and any references if available.

Edoardo-Pedicillo:
> - There should be brief documentation explaining what is the experiment doing in high level. 

We are working in this direction, all our progresses about docs are documented in #649

andrea-pasquale:
Thanks @hay-k for opening the issue. 
Here is a reference to the artificial detuning by changing the relative phase
![image](https://github.com/qiboteam/qibocal/assets/49183315/0abeab6b-9c2e-4041-8969-3aec51310ec9)
[Gao's paper](https://arxiv.org/pdf/2106.06173.pdf).
Theoretically the results should be the same but we should investigate more after running on hardware directly in both ways.
Moreover, this could be a good chance to use a proper notation instead of `n_osc` which corresponds to the number of oscillation that the user is expected to see. Perhaps indicating directly the detuning in Hz would be a better choice.

andrea-pasquale:
https://github.com/qiboteam/qibocal/pull/861#discussion_r1649014670

scarrazza:
Shall we close this?

andrea-pasquale:
I left this open on purpose because we want to expand a bit more the description of the Ramsey experiment.

PR #721, opened 2024-02-19
Refactor ramsey to use detuning instead of number of oscillations

andrea-pasquale:
Refactoring all Ramsey protocol to use directly a `detuning` parameter instead of the number of oscillations.
In this way the detuning is introduced directly and not artificially through the phase as already mentioned in #696.
I also tried to remove some code duplication.

I'm planning to do also the following:
- [x] Improve documentation
- [x] Check detuning sign

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`


src/qibocal/protocols/characterization/ramsey/ramsey.py
```diff
@@ -166,48 +152,18 @@ def _acquisition(
                     errors=errors,
                 ),
             )
-    if params.n_osc != 0:
+
+    if params.detuning != 0:
```
stavros11, line 156: Why do we still need a separate case for `detuning != 0`? If we are detuning the frequency, I think we can use the same sweeper in both cases. It is not like before that we had to sweep the phase together with the duration which was not supported.

src/qibocal/protocols/characterization/ramsey/utils.py
```diff
@@ -0,0 +1,37 @@
...
+    wait: Optional[int] = 0,
+    detuning: Optional[int] = 0,
+):
+
+    sequence = PulseSequence()
+    first_pi_half_pulse = platform.create_RX90_pulse(qubit, start=0)
+    second_pi_half_pulse = platform.create_RX90_pulse(
+        qubit, start=first_pi_half_pulse.finish + wait
+    )
+    second_pi_half_pulse.frequency += detuning
```
stavros11, line 21: I think here we need to detune both pulses. At least this is what would happen if you manually detune the RX frequency in the runcard, which I believe it is what @igres26 has been doing.

src/qibocal/protocols/characterization/ramsey/ramsey.py
andrea-pasquale: Indeed when I was refactoring I was thinking that we can avoid having the two different cases.
Perhaps it could be useful to keep also a version with unrolling given that some instruments (qblox) have memory issues when dealing with duration sweepers. Let me know what you think.

src/qibocal/protocols/characterization/ramsey/utils.py
andrea-pasquale: This is also something that I wanted to check. I think that before given that we were only modifying the relative phase of the second RX90 pulse probably we were detuning only the second one. In general you are right we should detune both.

andrea-pasquale: Implemented in https://github.com/qiboteam/qibocal/pull/721/commits/0b565ff417b263bdd582f419f6e3465eeb13c5f1

igres26:
Could you add some of the images that you get from running this one on the available chip. Detune around 1 MHz and with maximum duration of around 2000 ns. With this, we should also be able to detect the right direction of the detuning and correct it accordingly.

src/qibocal/protocols/characterization/ramsey/ramsey.py
andrea-pasquale: Now both ramsey and ramsey detuned work with sweepers.
I kept the option for `unrolling`: if the user set `unrolling=True` the platform will unroll the sequences instead of using the sweeper.

andrea-pasquale: I've also decided to drop `ramsey_sequences` since everything should be now included in `ramsey` and `ramsey_signal`.

andrea-pasquale:
The PR is now ready to review. I've performed some tests with both unrolling/sweepers and I've tested also that the correction for the drive frequency has the correct sign.

igres26 (approved):

stavros11 (approved):

PR #725, opened 2024-02-21
Adding unrolling to spin echo

### LINKED ISSUE #724: `Spin Echo` experiment slow
Currently `spin_echo` is quite slow given that the only implementation that we currently have relies on a `for` loop performed on software.
I believe that with the current version of qibolab it is not possible to perform this experiment with a sweeper given that the sequence is the following `RX90 - wait - RX - wait - RX90 - MZ` , correct @stavros11 @hay-k @alecandido?
For the time being I was thinking of providing the option to unroll the sequences.

If we all agree we could also decide to drop all protocols in qibocal that still perform this sort of for loop and just implement the unrolling. A possible idea would be to provide both unrolling and sweepers whenever possible like I'm doing in #721.

After benchmarking both unrolling and sweepers we could eventually decide to keep only one, but I believe that given that we support different drivers it is not a bad idea to keep both.

andrea-pasquale:
Closes #724.
I've added the option to unroll sequences when running spin echo which should improve performances.
There was also a typo since the protocol `spin_echo_sequences` was actually implementing `spin_echo_signal` which I fixed.

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/characterization/__init__.py
+++ b/src/qibocal/protocols/characterization/__init__.py
@@ -5,7 +5,7 @@ from .allxy.allxy_drag_pulse_tuning import allxy_drag_pulse_tuning
 from .allxy.drag_pulse_tuning import drag_pulse_tuning
 from .classification import single_shot_classification
 from .coherence.spin_echo import spin_echo
-from .coherence.spin_echo_sequence import spin_echo_sequence
+from .coherence.spin_echo_signal import spin_echo_signal
 from .coherence.t1 import t1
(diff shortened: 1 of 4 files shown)


Jacfomg (approved):
It runs in Zurich and the pulses are fine and everything, i did not run on calibrated qubit thought. The code also looks fine.

Edoardo-Pedicillo (approved):

PR #730, opened 2024-02-27
RB fix U3

Jacfomg:
Fixes an issue regarding how Qibo handles U3 parameters for the RB

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


igres26 (approved):
This should fix the issue. 

KEEP IN MIND that this is a small patch, and that we should also look into the real issue in the u3 decomposition formula so that it takes into account the determinant of the gate it is trying to decompose.

andrea-pasquale (approved):

PR #731, opened 2024-02-27
Nix

alecandido:
Add Nix development shell


alecandido:
@andrea-pasquale in 2aac948 the tests failed at random (since nothing changed from their point of view, but also because they worked for ubuntu-py3.11, and not for ubuntu-py3.10, with no known difference).

https://github.com/qiboteam/qibocal/actions/runs/8068203197/job/22040390172?pr=731

Any idea?

andrea-pasquale:
I think that it might be related to the fact that we are fitting a lorentzian over random numbers.
If the fitting fails currently we just keep the same value while the test if looking for a change in some platform' parameters if `update` is set to `true`.
It is weird that it failed just now... I was working on modifying the default values in case the fitting fails in #707.
We could fix the tests by asserting that the fitting didn't fail before checking if the platform has been updated.
Let me me open an issue.

andrea-pasquale (approved):

PR #735, opened 2024-02-28
Save rb circs to json

Jacfomg:
Draft to save `Circuit` to a json file. Right now we would need to change Qibo `Gate.raw` to serialize easily a gate to json.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


Jacfomg:
If I understood correctly, we won't save circuits for the RB so we can close it and we can work on saving the dictionary keys.

PR #742, opened 2024-03-03
Fix bug in lorentzian plot for spectroscopies

### LINKED ISSUE #741: Resonator spectroscopy fit fails
Hi,
The fit of resonator_spectroscopy no longer works as it used to:
image

I have run this with the main branches of qibolab and qibocal as of the 02/03/2024.
Attached is the report generated and the platform runcard used.
[resonator_spectroscopy_fit_fails.zip](https://github.com/qiboteam/qibocal/files/14471203/resonator_spectroscopy_fit_fails.zip)

andrea-pasquale:
Closes #741.
There was no error in the fit since also the table shows the correct fitted value, the problem was just a factor 10^9 in the plot.
It is possible to test the fit by simply running 
```sh
qq fit <output_folder>
```
 and to generate the new report
```sh
qq report <output_folder>
```

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/characterization/utils.py
+++ b/src/qibocal/protocols/characterization/utils.py
@@ -225,11 +225,10 @@ def spectroscopy_plot(data, qubit, fit: Results = None):
 
     if fit is not None:
         params = fit.fitted_parameters[qubit]
-
         fig.add_trace(
             go.Scatter(
                 x=freqrange,
-                y=lorentzian(freqrange * GHZ_TO_HZ, *params),
+                y=lorentzian(freqrange, *params),
                 name="Fit",
                 line=go.scatter.Line(dash="dot"),
             ),


aorgazf (approved):

igres26:
This makes the line appear correctly in qubit spectroscopy for me, but still for resonator spectroscopy at low power it did not appear. Could it be that it failed to fit? I don't even see the table with the fitted values when that is the case.

andrea-pasquale:
> This makes the line appear correctly in qubit spectroscopy for me, but still for resonator spectroscopy at low power it did not appear. Could it be that it failed to fit? I don't even see the table with the fitted values when that is the case.

When the fit fails you should not see both the table and the plot with the fit.
It seems weird to me, please share the report where the fitting does not appear.

andrea-pasquale:
> This makes the line appear correctly in qubit spectroscopy for me, but still for resonator spectroscopy at low power it did not appear. Could it be that it failed to fit? I don't even see the table with the fitted values when that is the case.

This has been addressed here https://github.com/qiboteam/qibocal/pull/744/commits/72873bb23badfbfc0f08aaab994fa7a9f1fe2066 in https://github.com/qiboteam/qibocal/pull/744 because the issue was similar.

stavros11 (approved):

PR #747, opened 2024-03-05
fix: Added json extension to results for singleshot classification

### LINKED ISSUE #746: Missing extension to the results filename for singleshot classification
The results file obtained by the singleshot classification routine is named `results` instead of `results.json`.

Runcard used:

``` yaml
actions:

  - id: t1
    priority: 0
    operation: t1
    main: single_shot_classification
    parameters:
      delay_before_readout_start: 16
      delay_before_readout_end: 2000
      delay_before_readout_step: 20

  - id: single_shot_classification
    priority: 0
    operation: single_shot_classification
    parameters:
      nshots: 1024
```

GabrielePalazzo:
Closes #746.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/auto/operation.py
+++ b/src/qibocal/auto/operation.py
@@ -27,7 +27,7 @@ QubitsPairs = dict[tuple[QubitId, QubitId], QubitPair]
 
 DATAFILE = "data"
 """Name of the file where data are dumped."""
-RESULTSFILE = "results"
+RESULTSFILE = "results.json"
 """Name of the file where results are dumped."""
 
 
@@ -133,7 +133,7 @@ class AbstractData:
     def _to_json(self, path: Path, filename: str):
         """Helper function to dump to json."""
         if self.params:
-            (path / f"{filename}.json").write_text(
+            (path / f"{filename}").write_text(
                 json.dumps(serialize(self.params), indent=4)
             )
 
--- a/src/qibocal/auto/operation.py
+++ b/src/qibocal/auto/operation.py
@@ -27,7 +27,7 @@ QubitsPairs = dict[tuple[QubitId, QubitId], QubitPair]
 
 DATAFILE = "data"
 """Name of the file where data are dumped."""
-RESULTSFILE = "results.json"
+RESULTSFILE = "results"
 """Name of the file where results are dumped."""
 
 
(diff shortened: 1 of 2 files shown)


andrea-pasquale (commented):
Thanks @GabrielePalazzo, do you know why tests are failing?

The problem was related to this line:
https://github.com/qiboteam/qibocal/blob/ddba2d3da995a308756ed25a1072b7e230a546ac/src/qibocal/protocols/characterization/classification.py#L127

An alternative solution would be to add `.json` directly in this line.
Maybe it is better to put the extension directly in the top-level-constant.

Let me know what you think.

GabrielePalazzo:
If I revert this line, tests are passing. 
https://github.com/qiboteam/qibocal/blob/e3737a37b9caa307b8042b8da9fd382087f7ed15/src/qibocal/auto/operation.py#L136
The cause is here (when we read json):
https://github.com/qiboteam/qibocal/blob/e3737a37b9caa307b8042b8da9fd382087f7ed15/src/qibocal/auto/operation.py#L169
I think the safest approach is what you suggested (explicitly adding `.json` in singleshot), considering that, for example, `DATAFILE` cannot have its extension (it can be both `json` and `npz`).
What do you think?

andrea-pasquale:
> If I revert this line, tests are passing.
> 
> https://github.com/qiboteam/qibocal/blob/e3737a37b9caa307b8042b8da9fd382087f7ed15/src/qibocal/auto/operation.py#L136
> 
> The cause is here (when we read json):
> 
> https://github.com/qiboteam/qibocal/blob/e3737a37b9caa307b8042b8da9fd382087f7ed15/src/qibocal/auto/operation.py#L169
> 
> I think the safest approach is what you suggested (explicitly adding `.json` in singleshot), considering that, for example, `DATAFILE` cannot have its extension (it can be both `json` and `npz`).
> What do you think?

Sure, let's add `.json` in singleshot I forgot to update it when we refactored the data structure in #690.

andrea-pasquale (approved):
Thanks @GabrielePalazzo

Edoardo-Pedicillo (approved):

Issue #748, opened 2024-03-11
Reactive programming for calibration


andrea-pasquale:
As we know, calibration is tricky and it requires several experiments that may be connected in a non-trivial way plus error handling.
Currently to support dependencies between experiments we can rely on a direct acyclic graph (DAG) implementation which has to be coded through `yaml` . Lately we also tried to implement error handling, the so called "exceptional flow", by defining validators https://github.com/qiboteam/qibocal/pull/668 such as $\chi^2$ and adding the possibility to change the flow of the DAG mid-execution. One of the limitation of this approach is that everything has to pass through `yaml` which is not ideal especially if we also need to include `if-else` statements.

The current layout is missing a mechanism to perform easily closed loop optimizations over one or custom protocols.

It would be helpful to have a higher-level API beyond the protocols execution through a DAG coded in yaml
We should be able to handle the following features:

- Change the "calibration flow" depending on the output of nodes
- Perform closed-loop optimization
- Handling errors at run-time
- Compose experiment with standard nodes and more complex structure (including DAG or closed optimization loops)

A possible idea that was discussed during a meeting was to use `reactive programming` to address both the dependencies and the error handling. This might be necessary given that the calibration appears to be an event driven algorithm. Personally I like this idea, I'm currently working on a prototype using https://github.com/ReactiveX/RxPY. 

Here is a first sketch of pseudo-code proposed by @alecandido
```py
from qibocal import Routines as r
from qibocal import reactive as rx
from qibocal import Subprograms as p
from qibocal import Subgraphs as g
from qibocal import Program as prog
                                          
                                          
def f(executor, *params):
    r1 = r.ResSpec()
    if r1 > 5:
        r2 = r.QubSpec()
                                          
        executor.schedule(p.RamseyLoop())
                                          
                                          
r3 = p.RamseyLoop()
                                          
r4 = rx(r.Rabi(), dep=r2)
                                          
myprog = prog(f, r3, r4)
myprog.execute()
```
As an initial step I want to remove the DAG from yaml and change it to a sequential execution, all the shenanigans about dependencies and error handling will be reintroduce later through reactive programming. As soon as I have something ready I will open a poc PR.

I'm happy to start/continue the discussion here about the possibility of using reactive programming. Suggestions or features needed are welcomed.

alecandido:
[...]

PR #749, opened 2024-03-11
Dropping DAG

andrea-pasquale:
Changing executor to deal only with sequential protocols execution, as discussed in #748.

Given that I'm removing a lot of things that we may want to use later like `Validators` perhaps it would be better to have a development branch... and merge everything to main once we have implemented some/all of the points raised in #748.

This first step was quite straightforwarded to do, this is why I'm already raising the PR.

- [x] update documentation
- [x] update runcards everywhere

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`


andrea-pasquale:
Let me know what you think @stavros11 @Edoardo-Pedicillo @alecandido @hay-k

alecandido (approved):
It is a breaking change, so all runcards will have to be updated (no problem for those manually written and throw-away, I'm more worried about the monitoring and similar use cases).

But since no one ever used the graph, if you allow for now random keys (`main:`, `next:`, `priority:`, ...) maybe you could preserve compatibility "in practice" (with existing runcards), and merge immediately.

andrea-pasquale:
> But since no one ever used the graph, if you allow for now random keys (`main:`, `next:`, `priority:`, ...) maybe you could preserve compatibility "in practice" (with existing runcards), and merge immediately.

Indeed, this is something that I can do. If we treat them as optional arguments I can also put warning if they are set telling the user that they will be dropped soon and that they are unused from now on.

andrea-pasquale:
@GabrielePalazzo although this PR should not break the monitoring, I updated the monitoring runcard. Let me know if you see any issues.

Edoardo-Pedicillo:
> But since no one ever used the graph, if you allow for now random keys (`main:`, `next:`, `priority:`, ...) maybe you could preserve compatibility "in practice" (with existing runcards), and merge immediately.

Are you sure we should rely on the order of the routines in the runcard, instead of keeping a `next` key? 
Since we are still supporting yaml, I think it is easier to change the value of a key instead of moving around bunch of lines,  to change the order of the routines. Maybe we could keep it for now and eventually drop it later, when we will use the API to calibrate.

alecandido:
> Maybe we could keep it for now and eventually drop it later, when we will use the API to calibrate.

It's defnitely an option. But most of the runcards I've seen are either sorted, and not frequently changing (like the monitoring), or running single routines.
I would also take into account that, unless you're doing it very frequently, moving lines around is not so painful (thanks clipboards), and having large YAML runcards is not adviced (nor happening?)

Edoardo-Pedicillo:
> unless you're doing it very frequently,

I think this is the case, for what I saw, if you calibrate a device and the results are not good :)

alecandido:
> I think this is the case, for what I saw, if you calibrate a device and the results are not good :)

Changing runcard frequently is ok, I understand it. But even changing order? (and are people calibrating with more than one routine at a time?)

Edoardo-Pedicillo:
> But even changing order?

For what I saw, generally a single runcard is used multiple times and to change calibration experiment some parts are commented and uncommented, doing this it happens frequently that the order of routines is not what desired.

>and are people calibrating with more than one routine at a time?

Yes, if the chip is calibrated but you want just to update the platform with the small parameters drifts.

alecandido:
> For what I saw, generally a single runcard is used multiple times and to change calibration experiment some parts are commented and uncommented, doing this it happens frequently that the order of routines is not what desired.

Yes, but if you uncomment a single routine at a time you don't care about the order (and moreover, you're already making changes to your runcard that are comparable to moving things around, the comment button is not much heavier than the cut one...).

> Yes, if the chip is calibrated but you want just to update the platform with the small parameters drifts.

This would have been my initial expectation (month ago), but:
- I've barely seen any report like this on the report server
- it is months since we had an actual calibrated chip...

Edoardo-Pedicillo:
> For what I saw, generally a single runcard is used multiple times and to change calibration experiment some parts are commented and uncommented, doing this it happens frequently that the order of routines is not what desired.

Sorry, here I was referring to the time when DAG was not present.

> * I've barely seen any report like this on the report server

I'm not sure how statistically significant the report server is.

By the way, I don't want to push too much the use of the key, I am just suggesting speaking to the people using Qibocal, because I don't see any particular preference from the software point of view.

alecandido:
> By the way, I don't want to push too much the use of the key, I am just suggesting asking to the people using Qibocal, because I don't see any particular preference from the software point of view.

Eventually, I'd keep the runcard mode for very straightforward usage. That's why I'd make it as simple as possible.
For a transient, we could do whatever. If you believe to be really useful we could keep it, it's just a bit of extra work to maintain it.

andrea-pasquale:
Following today's discussion I decided to drop the warnings, I'm planning to merge it as it is as soon as I get the second approval.

Edoardo-Pedicillo (approved):

andrea-pasquale:
For dropping `priority`, `main` and `next` from all runcards in a single go here is the regex:

```sh
^[# \t]*(priority|next|main).*?(\n|$)
```

PR #750, opened 2024-03-12
remove Tensorflow dependency

### LINKED ISSUE #734: Remove TensorFlow dependency
TensorFlow is only required for the classifiers, but it has been shown by @Edoardo-Pedicillo that the classifiers available in `scikit-learn` are already good enough, at least for the time being.

TensorFlow is a complex dependency, with multiple sub-packages, and not always promptly distributing wheels (and possibly not for all platforms), unlike NumPy, e.g..

Most likely, upgrading to the last release candidate could be enough, as it will already be in Qibo.
https://github.com/qiboteam/qibo/pull/1231#issuecomment-1968357835
But if it is not strictly needed, I'd make our lives easier and drop it, together with the related code.

I'm not suggesting to drop the classifiers and their infrastructure at all. The qubit fit and the `scikit-learn` ones are not causing any issue, and they could be very valuable.

Edoardo-Pedicillo:
This PR closes #734

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


pyproject.toml
```diff
@@ -32,13 +32,9 @@ skops = "^0.6.0"
...
-keras-tuner = { version = "^1.3.0,<1.3.1", optional = true, markers = "sys_platform == 'linux' or sys_platform == 'darwin'" }
 matplotlib = { version = "^3.7.0", optional = true }
 seaborn = { version = "^0.12.2", optional = true }
 pydot = { version = "^1.4.2", optional = true }
-tensorflow = { version = "^2.12.0", optional = true, markers = "sys_platform == 'linux' or sys_platform == 'darwin'" }
-# TODO: the marker is a temporary solution due to the lack of the tensorflow-io 0.32.0's wheels for Windows, this package is one of
-# the tensorflow requirements
 skl2onnx = { version = "^1.14.0", optional = true }
 onnxruntime = { version = "^1.14.1", optional = true }
 onnx = { version = "^1.13.1", optional = true }
```
alecandido, line 40: Do we still need `onnx`? Is there a different way of serializing Scikit-Learn objects? (maybe even just dictionaries for the hyperparams, and dump in json)

alecandido (approved):
Thanks @Edoardo-Pedicillo, I only had that comment, and it could be split to a separate issue/PR as well. The PR itself seems alright.

alecandido:
Consider that the `dump()` function, for which you need ONNX at the moment (even in SciKit you need the runtime), seems not to be used.

https://github.com/qiboteam/qibocal/blob/29cc2676010572071e617fdb4062e0fb69971ab3/src/qibocal/fitting/classifier/ada_boost.py#L41

At least not in [`run.py`](https://github.com/qiboteam/qibocal/blob/remove_nn/src/qibocal/fitting/classifier/run.py).

alecandido:
> The `dump()` function is used in the classification routine

Correct, that's why I was not finding it in classifiers. So, you need of course to dump the result of the fit. But is there no intrinsic way of Scikit-Learn to serialize this? (I mean, w/o ONNX)

Edoardo-Pedicillo:
Pikle ;) btw onnx is suggested by scikit itself (https://scikit-learn.org/stable/model_persistence.html) and even the other options require extra dependencies

andrea-pasquale (approved):

PR #757, opened 2024-03-19
Store mitigation matrix when running  CHSH

andrea-pasquale:
Now when running CHSH the mitigation matrix will be stored in a `mitgation_matrix.npz` file.
It should be useful for debugging but also in order to generate the CHSH value after performing the acquisition (`qq acquire`) with `qq fit`. 
For the future it might be interesting to have this behave like kernels if we want to support readout mitigation (of course also qibolab will need to be updated to support this feature).

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


igres26 (approved):
Looks good. The readout matrix can be reused for other routines so saving it separately is good.

stavros11 (approved):

PR #777, opened 2024-03-28
Tests with dummy_couplers

### LINKED ISSUE #776: Two runcards to test protocols
Some routines need to ran with dummy_couplers to test all the code or to be tested at all. While the RB routines need to be ran in dummy because of the noise_simulation backend change. That means we would need to split and test some protocols on one dummy and some others in dummy_couplers.

Jacfomg:
Closes #776. I can try to look for a smarter way to avoid duplicating. 

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/tests/runcards/protocols_couplers.yml
+++ b/tests/runcards/protocols_couplers.yml
@@ -73,6 +73,20 @@ actions:
       power_level: low
       nshots: 10
 
+  - id: coupler_resonator_spectroscopy
+    operation: coupler_resonator_spectroscopy
+    targets: [[1, 2], [0, 2]]
+    parameters:
+      bias_width: 1
+      bias_step: 0.1
+      freq_width: 10_000_000
+      freq_step: 1_000_000
+      measured_qubits: [1, 0]
+      amplitude: .3
+      nshots: 10
+      relaxation_time: 3_000
+
+
   - id: qubit spectroscopy
     operation: qubit_spectroscopy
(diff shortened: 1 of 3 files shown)


tests/runcards/protocols_couplers.yml
andrea-pasquale, line 1: Why does this runcard contains also non-couplers protocols?

tests/test_protocols.py
```diff
@@ -70,6 +82,40 @@ def test_auto_command(runcard, update, platform, backend, tmp_path):
         assert (tmp_path / "data" / f"{protocol}_0" / PLATFORM_DIR).is_dir()
 
 
+@pytest.mark.parametrize("platform", ["dummy_couplers"])
+@pytest.mark.parametrize("backend", ["qibolab"])
+@pytest.mark.parametrize("update", ["--update", "--no-update"])
+@pytest.mark.parametrize(
+    "runcard", generate_runcard_single_protocol_couplers(), ids=idfn
+)
```
andrea-pasquale, line 90: I would prefer another solution here. I need to think if there is a better way of doing it.

andrea-pasquale (changes requested):

andrea-pasquale (approved):

Edoardo-Pedicillo (approved):

PR #801, opened 2024-04-16
[poc] Attempt at simplifying `Action`->`Task`-> `Completed` nested dependency

andrea-pasquale:
This is just a proposal to simplify some of the code related to `Task`, `Action` and `Completed`.
I think that there is a bit of redundancy between this classes since `Completed` as a reference to `Task` which has a reference to `Action`.

- `Action` is mainly used to deserialize the protocol written in the `yaml`.
- `Task` "continues" the deserialization by mapping the operation to the corresponding `Routine` object, it is also responsible for handling the execution through the `run` method.
- `Completed` is responsible for storing stuff and updating the platform.

In this PR I propose a different approach which is based on the [`state` ](https://refactoring.guru/design-patterns/state) design pattern.
The idea is to treat an object `Experiment` as something that will change depending on its internal state, the interface exposes the following methods: `acquire`, `fit`, `update_platform` and `dump`. The possible states where an experiment can be are currently the following:

- `Pending`: deserialized experiment waiting to be executed. By performing `acquire` the internal state switches to an internal state `Acquired`.
- `Acquired` experiment where acquisition has been performed. By performing `fit` the internal state switches to `Fitted`
- `Fitted` experiment where fitting has been performed. By performing `update_platform` it switches to `Updated`.
- `Updated` experiment that has updated the platform

This new approach should also simplify the API:

```py
from qibocal.auto.experiment import Experiment
from pathlib import Path

PLATFORM = create_platform("dummy")
PATH = Path("test")

exp = Experiment("my_resonator_spectroscopy",
                 "resonator_spectroscopy",
                 parameters=dict(freq_step=1e5,
                                 freq_width=1e6,
                                 power_level="high"))

# run acquisition
exp.acquire(platform=PLATFORM, targets=[0])
# retrieve data
data = exp.data
# dumping data
exp.dump(PATH)


# fitting
exp.fit()

# retrieve results
results = exp.results

# dumping results
exp.dump(PATH)

# retrieve results
results = exp.results

# update platform
exp.update_platform(PLATFORM)

# dump updated platform
exp.dump(PATH)
```

For the time being this is just an idea that came to my mind. I'm not sure if this is the best approach to simplify the code.
After this simplification I think that it should be easier to transition to experiment composition or reactive programming. 

Thoughts @alecandido @hay-k @Edoardo-Pedicillo?

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


alecandido:
The reason why there were three objects, instead of one, was to avoid mutating a state. This is not directly a criticism to this proposal, but just an explanation of the rationale for the current behavior.

The idea is that mutations makes more complex to trace the issues in an application. If you have purely functional components, given a certain input you will always obtain the same output, and it helps to improve modularity and reproducibility.
The issue is especially relevant in the case of applications with a large UI, where the user interacts multiple times, changing the state over and over (and to tackle it, they made [entire](https://redux.js.org/) [libraries](https://zustand-demo.pmnd.rs/) [to avoid mutations](https://vuex.vuejs.org/) and keeping the state in a single place).

Qibocal is not an application with a large UI, but (in principle) a complex autocalibration could keep going for a long while, with a non-trivial flow.
For this reason, the design wanted to keep separate the original input (immutable `Action`s, reusing the name they had before, but not especially meaningful in this context), the final output of each step (immutable `Completed` records), from the only objects involved in the execution (the `Task`s), which should contain some more minimal information wrt the other two.

andrea-pasquale:
Thanks for having a quick look.

> The reason why there were three objects, instead of one, was to avoid mutating a state.

I see, I was not aware of the fact the intent was not have a mutating state. I also don't want to to enforce a mutating state. I felt that it was a bit convoluted to juggle between `Action`, `Task` and `Completed`. My aim here was just to propose a simplified API (without going through dictionaries or yaml). Because although there is already a way of doing [it](https://qibo.science/qibocal/stable/tutorials/advanced.html#how-to-use-qibocal-as-a-library) the code is mostly oriented at executing protocols through the CLI. We need to go beyond that in order to implement composable experiments #748.

Also, I understand the need for the 3 objects, but I think that the nested dependency is complicated to handle and it seems that we are storing redundant information, also because there is not a proper mechanism for serializing and deserializing each object. 

> Qibocal is not an application with a large UI, but (in principle) a complex autocalibration could keep going for a long while, with a non-trivial flow.

I know that this could be the case, in our case having a mutated state would be just of way of handling the three most basic operation such as data acquisition, fitting and updating the platform, which are fundamental part of each experiments. Also since there are only 4 possible states in the current implementation I believe that it could be quite easily kept under control. 

Anyway, this was just an experimental approach that I tried to follow while learning something new.

alecandido:
> I see, I was not aware of the fact the intent was not have a mutating state. I also don't want to to enforce a mutating state. I felt that it was a bit convoluted to juggle between `Action`, `Task` and `Completed`. My aim here was just to propose a simplified API (without going through dictionaries or yaml). Because although there is already a way of doing [it](https://qibo.science/qibocal/stable/tutorials/advanced.html#how-to-use-qibocal-as-a-library) the code is mostly oriented at executing protocols through the CLI. We need to go beyond that in order to implement composable experiments #748.

Agreed. I exposed the rationale above, and I would consider keeping some separation between the input/output and the object holding the execution.

> Also, I understand the need for the 3 objects, but I think that the nested dependency is complicated to handle and it seems that we are storing redundant information, also because there is not a proper mechanism for serializing and deserializing each object.

I have the feeling that the main problem could be not much in the `Action`/`Task`/`Completed`-split, but rather in the protocols interface.

So, I'd rather focus on the `Action` alone, and even better *how to create a node*, to expose a better interface for the library.

<details><summary>Concerning the other two...</summary>
<p>

`Completed` should be properly serialized (and deserialized, possibly even peekable - i.e. deserializing it partially, but giving optionally giving access to the full information), but that's fully internal (not something the user should really interact with, but rather used by intermediate tools).
`Task` should only be used during execution (and, in my mind, the further information wrt the action should be limited to the iteration and for which specific qubit), thus we don't even need to (de)serialize it.
In principle, we could drop the separation between `Task` and `Completed`, but I still believe to be better to split the `Completed` API from the one available during execution (and avoid defaulting to `None` some information only required after). However, all of this is debatable.

</p>
</details>

andrea-pasquale:
> `Completed` should be properly serialized (and deserialized, possibly even peekable - i.e. deserializing it partially, but giving optionally giving access to the full information), but that's fully internal (not something the user should really interact with, but rather used by intermediate tools).

About this, maybe this is a naive question, but how can you properly (de)serialize `Completed` which contains a reference to `Task` without (de)serializing `Task`?

alecandido:
> About this, maybe this is a naive question, but how can you properly (de)serialize `Completed` which contains a reference to `Task` without (de)serializing `Task`?

Well, maybe you can not automatically (de)serialize, but it can be done manually, even with a reference.

However, the reference is not required: it was an easy way to add the completed `Task` to the `History`, but we just need some information from that.
So, we could make a `Completed.from_task(task, ...)` and extract from the `Task` the information we want to preserve. At that point, it should be simple enough to serialize it.

In any case, this is a possible layout, not necessarily the only one. I find it convenient because you can serialize both the input and the output, with just the required info for each of them. But we can discuss if there are any drawbacks in this.

PR #803, opened 2024-04-18
fix frequency range type

### LINKED ISSUE #762: Frequency range type
Currently frequency range types in experiments are defined with integer division instead of standard division.

https://github.com/qiboteam/qibocal/blob/c171ad9cb6e95a37581b636108d5445d194b7a6f/src/qibocal/protocols/characterization/resonator_spectroscopy.py#L151-L153

Is there a specific reason to keep `int` as type? If we see that drivers can handle float frequencies we can covert everything to float. @stavros11 @hay-k @alecandido @Jacfomg

Edoardo-Pedicillo:
This PR closes #762 

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/characterization/coherence/spin_echo.py
+++ b/src/qibocal/protocols/characterization/coherence/spin_echo.py
@@ -87,8 +87,8 @@ def _acquisition(
         # save data as often as defined by points
 
         for qubit in targets:
-            RX_pulses[qubit].start = RX90_pulses1[qubit].finish + wait // 2
-            RX90_pulses2[qubit].start = RX_pulses[qubit].finish + wait // 2
+            RX_pulses[qubit].start = RX90_pulses1[qubit].finish + wait / 2
+            RX90_pulses2[qubit].start = RX_pulses[qubit].finish + wait / 2
             ro_pulses[qubit].start = RX90_pulses2[qubit].finish
 
         sequences.append(deepcopy(sequence))
--- a/src/qibocal/protocols/characterization/coherence/spin_echo_signal.py
+++ b/src/qibocal/protocols/characterization/coherence/spin_echo_signal.py
@@ -102,8 +102,8 @@ def _acquisition(
         # save data as often as defined by points
 
         for qubit in targets:
-            RX_pulses[qubit].start = RX90_pulses1[qubit].finish + wait // 2
-            RX90_pulses2[qubit].start = RX_pulses[qubit].finish + wait // 2
+            RX_pulses[qubit].start = RX90_pulses1[qubit].finish + wait / 2
+            RX90_pulses2[qubit].start = RX_pulses[qubit].finish + wait / 2
             ro_pulses[qubit].start = RX90_pulses2[qubit].finish
 
         sequences.append(deepcopy(sequence))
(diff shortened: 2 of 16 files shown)


GabrielePalazzo (approved):

andrea-pasquale (commented):
Thanks @Edoardo-Pedicillo.
Two points:

- Can we have a function to generate the `delta_frequency_range`? In this way we can avoid to change many times the same thing.
- Before merging we should check that all drivers work with float frequencies? Is this the case @stavros11 @hay-k @alecandido?

alecandido:
> Before merging we should check that all drivers work with float frequencies? Is this the case @stavros11 @hay-k @alecandido?

This was one (tiny) point in Qibolab 0.2: the goal is to always make use of floats in the interface for real quantities (not only for frequencies), and integers inside the drivers (when needed).

I would say that there is no electronics actually supporting floats, but they should all support fixed precision. So, the conversion should happen inside.
However, I still have it on my list, and the final proof is only to see them working, of course...

(the rfsoc is already using floats for frequencies, laboneq should always accept floats, and in QUA you declare them as fixed, but they are initialized with floats - for qblox I still have to find out...)

https://github.com/qiboteam/qibolab/blob/a514441f7c020b190c7871df90d6892392d0b4da/src/qibolab/instruments/rfsoc/convert.py#L117

stavros11:
> * Before merging we should check that all drivers work with float frequencies? Is this the case @stavros11 @hay-k @alecandido?

In QM frequency sweeps are casted to `int` by the driver (https://github.com/qiboteam/qibolab/blob/a514441f7c020b190c7871df90d6892392d0b4da/src/qibolab/instruments/qm/sweepers.py#L97) so this change shouldn't affect. I also checked the resonator spectroscopy and it works.

> and in QUA you declare them as fixed, but they are initialized with floats

That's correct, however for frequency sweeps which are relevant for this PR we have to use the `update_frequency` function which updates frequencies in real time, so it also matters what type this accepts. From their docs this is not clear, since `new_frequency` is of type `int` but in the example they pass floats... It seems to also support units, so most likely float is fine. However, sub-Hz resolution (if we ever need that), seems to need special treatment (see example).

In any case, I agree that these conversions should be handled by the drivers and we should just use floats externally.

andrea-pasquale:
Thanks for the feedback @stavros11 @alecandido!
After testing we should be able to merge the PR, then.
If we see that something is not working we should wait for qibolab `0.2`, correct?

Edoardo-Pedicillo:
The changes do not brake QM 
http://login.qrccluster.com:9000/Rke3NbOaRCKA_KK7EIvyKg==/

alecandido:
> If we see that something is not working we should wait for qibolab `0.2`, correct?

The moment we will get to the drivers' update, the goal is to make the required casts in the driver (this will go together a units update, since the goal is to use *GHz* and *ns* everywhere, and to avoid 1e9/1e6 conversions - confining all the required conversions to the drivers).

andrea-pasquale:
Also Zurich seems fine http://login.qrccluster.com:9000/vwrYTwXoTGe6Nu8u1KYkuQ==/

andrea-pasquale:
Qblox is also fine: http://login.qrccluster.com:9000/W97M2TjkRKCzAEh-BdPPNw==/
Let's merge!

andrea-pasquale (approved):

PR #805, opened 2024-04-18
Simplifying plot generation

### LINKED ISSUE #506: Create `dataclass` in `web.report`
_Originally posted by @AleCandido in https://github.com/qiboteam/qibocal/pull/406#discussion_r1310602115_

andrea-pasquale:
Closes #506.
The issue was quite old, if I misunderstood the task let me know @alecandido, currently I just tried to simplify a bit the report generation.

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`

### DIFF
--- a/src/qibocal/cli/report.py
+++ b/src/qibocal/cli/report.py
@@ -1,25 +1,24 @@
 import json
 import tempfile
-from functools import cached_property
-from pathlib import Path
+from typing import Union
 
 import yaml
+from jinja2 import Environment, FileSystemLoader
 from qibo.backends import GlobalBackend
-from qibolab.qubits import QubitId
+from qibolab.qubits import QubitId, QubitPairId
 
 from qibocal.auto.execute import Executor
 from qibocal.auto.mode import ExecutionMode
 from qibocal.auto.runcard import Runcard
-from qibocal.auto.task import TaskId
+from qibocal.auto.task import Completed
+from qibocal.cli.utils import META, RUNCARD
 from qibocal.config import log
-
-META = "meta.json"
-RUNCARD = "runcard.yml"
-UPDATED_PLATFORM = "new_platform.yml"
-PLATFORM = "platform.yml"
+from qibocal.web.report import STYLES, TEMPLATES, Report
 
 
-def generate_figures_and_report(node, target):
+def generate_figures_and_report(
+    node: Completed, target: Union[QubitId, QubitPairId, list[QubitId]]
+):
     """Returns figures and table for report."""
 
     if node.results is None:
@@ -31,6 +30,22 @@ def generate_figures_and_report(node, target):
         return figures, "An error occurred when performing the fit."
     # plot acquisition and fit
     return node.task.operation.report(data=node.data, fit=node.results, target=target)
+
+
+def plotter(node, target: Union[QubitId, QubitPairId, list[QubitId]]):
+    """Generate single target plot."""
+
+    figures, fitting_report = generate_figures_and_report(node, target)
+    with tempfile.NamedTemporaryFile(delete=False) as temp:
+        html_list = []
+        for figure in figures:
+            figure.write_html(temp.name, include_plotlyjs=False, full_html=False)
+            temp.seek(0)
+            fightml = temp.read().decode("utf-8")
+            html_list.append(fightml)
+
+    all_html = "".join(html_list)
+    return all_html, fitting_report
 
 
 def report(path):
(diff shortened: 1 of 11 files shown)


src/qibocal/web/report.py
```diff
@@ -1,29 +1,36 @@
...
-    html = template.render(
-        is_static=True,
-        css_styles=css_styles,
-        version=__version__,
-        report=report,
-    )
+    def routine_name(self, routine, iteration):
+        """Prettify routine's name for report headers."""
+        name = routine.replace("_", " ").title()
+        return f"{name} - {iteration}"
```
alecandido, line 31: `@staticmethod`

alecandido:
> The issue was quite old, if I misunderstood the task let me know @alecandido, currently I just tried to simplify a bit the report generation.

To be fair, I also had to take a look at the cited comments, because I didn't remember anything before...

But yes, I believe that was the idea.

andrea-pasquale:
> To be fair, I also had to take a look at the cited comments, because I didn't remember anything before...
> 
> But yes, I believe that was the idea.

Thanks! I converted it back to draft because apparently with more than one protocol somethings breaks...

andrea-pasquale:
Now everything seems fine.

src/qibocal/cli/report.py
```diff
@@ -18,8 +20,16 @@
...
+) -> tuple[str, list[go.Figure]]:
+    """Calling protocol plot by checking if fit has been performed.
+
+    Args:
+        node (Completed): completed node
+        target (Union[QubitId, QubitPairId, list[QubitId]]): protocol target
+
+    Returns:
+        tuple[str, list[go.Figure]]: report outcome
+    """
```
alecandido, line 32: This is a nice example, so I'm just saying for this reason.

I would make the docs way more succinct, e.g. by saving all the parameters' types, since already present in the signature.
Moreover, they could also be made much more human-readable, by avoiding the fixed template
```suggestion
    """Calling protocol plot by checking if fit has been performed.
    
    It operates on a completed `node` and a specific protocol `target`, generating
    a report outcome (cf. `ReportOutcome`).
    """
```

And describe what is a `ReportOutcome` in the docstring of such an alias (just define `ReportOutcome = tuple[str, list[go.Figure]]` above).

This is just a proposal, not necessarily a request for this PR. But the idea is to reduce the SNR even in our files :P
(so, avoid the parts that could be autogenerated, and let Sphinx do that work)

andrea-pasquale: I agree with the fact that we should not duplicate documentation.
In my case I just installed an extension that will generate the docstring for me that's why I put it there :)
Since it is already there I prefer to keep it, once we unify the notation among all libraries we can update all docstrings.

alecandido: > In my case I just installed an extension that will generate the docstring for me that's why I put it there :)

I had something similar. Then, I thought: "but if it's able to generate it, why should I save it to Git at all?", and started writing it more concise.

> Since it is already there I prefer to keep it, once we unify the notation among all libraries we can update all docstrings.

It was just a proposal, to start moving in that direction.

andrea-pasquale: > It was just a proposal, to start moving in that direction.

Actually having a `ReportOutcome` makes sense. I've also updated the docstrings :)

andrea-pasquale:
@alecandido @Edoardo-Pedicillo feel free to review this PR whenever you have time. It should be quite easy.
I've also implemented your suggestions in the end @alecandido.

src/qibocal/cli/report.py
```diff
@@ -33,14 +41,35 @@ def generate_figures_and_report(node, target):
...
+) -> tuple[str, str]:
+    """Run plotly pipeline for generating html.
 
-    Arguments:
+    Performs conversions of plotly figures in html rendered code for completed
+    node on specific target.
+
+    """
+    figures, fitting_report = generate_figures_and_report(node, target)
+    with tempfile.NamedTemporaryFile(delete=False) as temp:
```
alecandido, line 54: What's the purpose of `delete=False`?

src/qibocal/cli/report.py
```diff
@@ -33,14 +41,35 @@ def generate_figures_and_report(node, target):
...
+        html_list = []
```
alecandido, line 55: I wonder if it's worth to return just tables, instead of HTML.

I have the feeling this has already been discussed in the past, but I believe we're only using tables, and I don't remember why we went for HTML.

src/qibocal/web/report.py
```diff
@@ -1,29 +1,37 @@
...
-        is_static=True,
-        css_styles=css_styles,
-        version=__version__,
-        report=report,
-    )
+    @staticmethod
+    def routine_name(routine, iteration):
+        """Prettify routine's name for report headers."""
+        name = routine.replace("_", " ").title()
+        return f"{name} - {iteration}"
```
alecandido, line 32: Maybe we should get rid of `iteration`, since we got rid of the graph, and we're only supporting the sequential executions (it can come back whenever we'll do something more complex, but now it is kind of redundant... isn't it?)

alecandido (approved):
Just a few comments, mostly not for this PR

src/qibocal/cli/report.py
andrea-pasquale: https://github.com/qiboteam/qibocal/commit/91edadb9966e74cb1959fb2c13a5eece0ff38e7e
I remember that tests were failing without it, especially the ones related to RB.

src/qibocal/cli/report.py
andrea-pasquale: HTML will leave more freedom to the developer when writing the plot functions, it was requested by the RB team.

src/qibocal/web/report.py
andrea-pasquale: Iteration removed

src/qibocal/cli/report.py
alecandido: I believe that is not the best way to fix it.

Most likely, the temporary file is getting deleted too soon. But it is weird, since you're using `temp` just to write a file and then read it as a string...

Btw, an alternative solution would be to not use at all the file system: what you need is not necessarily a file, but a stream (i.e. an object with `.read()` and `.write()` methods) as an open file, but not necessarily that.
An equivalent object, with a buffer only operating in memory is provided by the standard library, [`StringIO`](https://docs.python.org/3/library/io.html#text-i-o)

andrea-pasquale: If you agree I would postpone this to a later PR, for sure it is not a priority.

alecandido: Fine by me, but leave an issue. This is potentially generating a lot of temporary files, with no cleanup...

andrea-pasquale: https://github.com/qiboteam/qibocal/pull/805/commits/09520b2a1ac07cf292bfb7fd2eb09acadd2af803
Now we are using a memory buffer :)

Edoardo-Pedicillo (approved):

PR #807, opened 2024-04-19
Use fft to enhance flipping fit

andrea-pasquale:
Similar approach already followed in other protocols.

http://login.qrccluster.com:9000/HWM0P_piSxy1gFm4Ce9RBg==/

The fit is still not perfect, as soon as we have #761 performances should improve even more.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/characterization/flipping.py
+++ b/src/qibocal/protocols/characterization/flipping.py
@@ -9,6 +9,7 @@ from qibolab.platform import Platform
 from qibolab.pulses import PulseSequence
 from qibolab.qubits import QubitId
 from scipy.optimize import curve_fit
+from scipy.signal import find_peaks
 
 from qibocal.auto.operation import Routine
 from qibocal.config import log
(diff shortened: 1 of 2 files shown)


src/qibocal/protocols/characterization/flipping.py
```diff
@@ -186,40 +200,38 @@ def _fit(data: FlippingData) -> FlippingResults:
...
+                float(perr[2] * pi_pulse_amplitude / np.pi / 2),
+            )
+            chi2[qubit] = (
+                chi2_reduced(
+                    y,
+                    flipping_fit(x, *popt),
+                    qubit_data.error,
+                ),
+                np.sqrt(2 / len(x)),
+            )
```
GabrielePalazzo, line 232: Is there a valid reason to extend the `try`/`except` block to this section?
The smaller it is the better. Additionally, if the code crashes some return variables might be badly defined (e.g.: empty dictionary) and could raise more exceptions later on.

andrea-pasquale: As you already said the idea is that this block will perform the fitting and fill the dictionaries, if it fails we return empty dictionaries. I agree that we should try to make it smaller. A possible idea would be to isolate all common fittings among protocol and have only one function in the try-except block.

GabrielePalazzo (approved):

PR #813, opened 2024-05-01
Single qubit state tomography

andrea-pasquale:
As title says.

TODO:
- [x] improve plotting


andrea-pasquale:
There are still a few things to be improved but it seems to work: http://login.qrccluster.com:9000/lIw8tz8rT1-ZfWeOKdmn6Q==

andrea-pasquale:
@stavros11 https://github.com/qiboteam/qibocal/pull/813/commits/a88257ad923d32c44d76782752597203cc9bc2a7 is a first attempt to pass circuits as json files.

andrea-pasquale:
Final report example: http://login.qrccluster.com:9000/XzGRWIMQRDG5sAC4BFMVHQ==/

src/qibocal/protocols/characterization/state_tomography.py
```diff
@@ -0,0 +1,285 @@
...
+
+    x_basis_state: QuantumState = None
+    y_basis_state: QuantumState = None
+    z_basis_state: QuantumState = None
+    data: dict[tuple[QubitId, str], np.float64] = field(default_factory=dict)
+
+    def save(self, path):
+        self.x_basis_state.dump(path / "x.json")
+        self.y_basis_state.dump(path / "y.json")
+        self.z_basis_state.dump(path / "z.json")
```
stavros11, line 60: Why is the theoretical expected state dumped to different jsons instead of included in the npz? This may be slightly inconvenient for 2+ qubit tomography where more measurement basis are needed (and if we follow the same approach, more jsons).

src/qibocal/protocols/characterization/state_tomography.py
```diff
@@ -0,0 +1,285 @@
...
+        params.circuit.add(gates.H(i) for i in range(len(targets)))
+
+    backend = GlobalBackend()
+    transpiler = dummy_transpiler(backend)
+
+    data = StateTomographyData()
+
+    for basis in ["X", "Y", "Z"]:
+        basis_circuit = deepcopy(params.circuit)
+        # FIXME: basis
```
stavros11, line 106: Is that still a `FIXME`?

src/qibocal/protocols/characterization/state_tomography.py
```diff
@@ -0,0 +1,285 @@
...
+        if basis != "Z":
+            for i in range(len(targets)):
+                basis_circuit.add(getattr(gates, basis)(i).basis_rotation())
+
+        # basis_circuit.add(gates.M(i, basis=getattr(gates, basis)) for i in targets)
+        for i in range(len(targets)):
+            basis_circuit.add(gates.M(i))
```
stavros11, line 113:
```suggestion
        basis_circuit.add(gates.M(i) for i in range(len(targets)))
```
should still work (following the comment), but regardless I would still drop the comment (dead code).

src/qibocal/protocols/characterization/state_tomography.py
```diff
@@ -0,0 +1,285 @@
...
+
+
+def _acquisition(
+    params: StateTomographyParameters, platform: Platform, targets: list[QubitId]
+) -> StateTomographyData:
+    """Acquisition protocol for single qubit state tomography experiment."""
+
+    if params.circuit is None:
+        params.circuit = Circuit(len(targets))
+        params.circuit.add(gates.H(i) for i in range(len(targets)))
```
stavros11, line 97: Personally, I wouldn't use any default circuit, I would just require the user to specify it, as I don't see anything special about the circuit full of H-gates that makes it a good default. However, this is not a big deal and mostly a matter of choice.

stavros11 (commented):
Thanks @andrea-pasquale, generally looks good to me. I will start working on the two-qubit case on top of this.

The only two issues I currently see as a bit annoying is the multiple jsons for data saving (mentioned in comment below) and the plotting. I would personally prefer bar plots, or at least print the numerical value for each matrix element in the color plot (I am not sure if this is possible with plotly). Currently we can see the values by hovering over the color plot, but it is a bit inconvenient: if I want to compare the experiment with theory I have to hover over 8 boxes. For two-qubit that would be 32 boxes. In the barplot you can overlap theory and experiment and see the difference immediately (even using multiple 2D bar plots, not the fancy 3D one).

For the two-qubit case, I was also plotting bar plots of the shots in each basis, before the matrix reconstruction. This was mainly because the matrix reconstruction is less trivial in that case, so most likely not relevant for the current PR (I might still do it for 2q though).

Jacfomg:
> Thanks @andrea-pasquale, generally looks good to me. I will start working on the two-qubit case on top of this.
> 
> The only two issues I currently see as a bit annoying is the multiple jsons for data saving (mentioned in comment below) and the plotting. I would personally prefer bar plots, or at least print the numerical value for each matrix element in the color plot (I am not sure if this is possible with plotly). Currently we can see the values by hovering over the color plot, but it is a bit inconvenient: if I want to compare the experiment with theory I have to hover over 8 boxes. For two-qubit that would be 32 boxes. In the barplot you can overlap theory and experiment and see the difference immediately (even using multiple 2D bar plots, not the fancy 3D one).
> 
> For the two-qubit case, I was also plotting bar plots of the shots in each basis, before the matrix reconstruction. This was mainly because the matrix reconstruction is less trivial in that case, so most likely not relevant for the current PR (I might still do it for 2q though).

I would love to have something like this https://community.plotly.com/t/is-it-possible-to-make-this-with-plotly/73696/2 , but seems too tricky honestly

src/qibocal/protocols/characterization/state_tomography.py
andrea-pasquale: This is related to https://github.com/qiboteam/qibo/issues/1318

src/qibocal/protocols/characterization/state_tomography.py
andrea-pasquale: This is still related to https://github.com/qiboteam/qibo/issues/1318, I would drop the dead code and provide a reference to the issue.

src/qibocal/protocols/characterization/state_tomography.py
andrea-pasquale: I forgot to remove this, I was just testing different gates.
My idea was just to have an empty circuit as default.

src/qibocal/protocols/characterization/state_tomography.py
andrea-pasquale: Actually no, my bad... I can implement it

src/qibocal/protocols/characterization/state_tomography.py
```diff
@@ -0,0 +1,285 @@
...
+        for i in range(len(targets)):
+            basis_circuit.add(gates.M(i))
+        for i, target in enumerate(targets):
+            _, results = execute_transpiled_circuit(
+                basis_circuit,
+                targets,
+                backend,
+                nshots=params.nshots,
+                transpiler=transpiler,
+            )
```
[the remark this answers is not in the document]
andrea-pasquale: Yes, I just noticed now. I'm trying to fix it. Thanks :)

andrea-pasquale:
I fixed all the points but the plotting so far, let me see if I can do something quickly.

andrea-pasquale:
Final version of the plot: http://login.qrccluster.com:9000/3T67I38gTlizhImJ6AkXbA==/

Jacfomg (approved):
Please fix the conflicts. Plotting seems nice.

stavros11 (approved):

PR #822, opened 2024-05-03
Resonator depletion measurement

Jacfomg:
Following https://arxiv.org/abs/1604.00916. Just to measure the resonator depletion time.
Another route would be a simpler Ramsey with a pre-measurament pulse as QM does https://github.com/qua-platform/qua-libs/blob/main/Quantum-Control-Applications/Superconducting/Single-Flux-Tunable-Transmon/09b_resonator_depletion_time.py#L124

For fitting reason I have the impression QM way would be easier to fit but the AllXY gives more information and possible ways of calibrating a depletion pulse I think.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


src/qibocal/protocols/characterization/allxy/allxy_resonator_depletion_tuning.py
```diff
@@ -0,0 +1,174 @@
...
+                sequence, ro_pulses[qubit] = allxy.add_gate_pair_pulses_to_sequence(
+                    platform,
+                    gates,
+                    qubit,
+                    sequence,
+                    sequence_delay=delay_param,
+                    readout_delay=996,  # becuase we already add a +4
+                )
+
+            print(sequence)
```
stavros11, line 88:
```suggestion
```

src/qibocal/protocols/characterization/allxy/allxy.py
```diff
@@ -134,13 +134,17 @@ def add_gate_pair_pulses_to_sequence(
...
+    sequence_delay=0,
+    readout_delay=0,
     beta_param=None,
 ):
     pulse_duration = platform.create_RX_pulse(qubit, start=0).duration
     # All gates have equal pulse duration
 
-    sequenceDuration = 0
-    pulse_start = 0
+    sequenceDuration = sequence.duration + sequence_delay
```
stavros11, line 144: I know this is a leftover from the past, but we shoud probably convert to the proper naming convention at some point
```suggestion
    sequence_duration = sequence.duration + sequence_delay
```

src/qibocal/protocols/characterization/allxy/allxy_resonator_depletion_tuning.py
```diff
@@ -0,0 +1,174 @@
...
+            for qubit in targets:
+                ro_pulse = platform.create_qubit_readout_pulse(qubit, start=0)
+                sequence.add(ro_pulse)
+                sequence, ro_pulses[qubit] = allxy.add_gate_pair_pulses_to_sequence(
+                    platform,
+                    gates,
+                    qubit,
+                    sequence,
+                    sequence_delay=delay_param,
+                    readout_delay=996,  # becuase we already add a +4
```
stavros11, line 85: Nothing to be done now, but these kind of manipulations will hopefully be easier using `Delay` in qibolab 0.2.

Also
```suggestion
                    sequence_delay=delay_param,
                    readout_delay=996,  # because we already add a +4
```

tests/runcards/protocols.yml
```diff
@@ -490,6 +490,13 @@ actions:
       unrolling: True
       nshots: 10
 
+  - id: resonator_depletion_tuning
+    operation: allxy_resonator_depletion_tuning
+    parameters:
+      delay_start: 1000.0
+      delay_end: 3000.0
+      delay_step: 1000.0
```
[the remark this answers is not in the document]
Jacfomg: Yes, so we either get an error coming from QM if we use floats or from json if we use ints.

src/qibocal/protocols/characterization/allxy/allxy.py
```diff
@@ -211,11 +208,14 @@ def add_gate_pair_pulses_to_sequence(
                 )
             sequence.add(RY90_pulse)
 
-        sequenceDuration = sequenceDuration + pulse_duration
-        pulse_start = pulse_duration
+        sequence_duration = sequence_duration + pulse_duration
```
andrea-pasquale, line 211:
```suggestion
        sequence_duration += pulse_duration
```

src/qibocal/protocols/characterization/allxy/allxy.py
```diff
@@ -211,11 +208,14 @@ def add_gate_pair_pulses_to_sequence(
...
+        pulse_start = sequence_duration
 
     # RO pulse starting just after pair of gates
-    ro_pulse = platform.create_qubit_readout_pulse(qubit, start=sequenceDuration + 4)
+    ro_pulse = platform.create_qubit_readout_pulse(
+        qubit, start=sequence_duration + 4 + readout_delay
```
andrea-pasquale, line 216: I know that the 4 was already there. Is it a remnant of some qblox shenanigans?

src/qibocal/protocols/characterization/allxy/allxy_resonator_depletion_tuning.py
```diff
@@ -0,0 +1,175 @@
...
+from qibolab.pulses import PulseSequence
+from qibolab.qubits import QubitId
+
+from qibocal.auto.operation import Data, Parameters, Results, Routine
+
+from . import allxy
+
+
+@dataclass
+class AllXYDragParameters(Parameters):
```
andrea-pasquale, line 18: Wrong class name

src/qibocal/protocols/characterization/allxy/allxy_resonator_depletion_tuning.py
```diff
@@ -0,0 +1,175 @@
...
+    delay_start: float
+    """Initial delay parameter for resonator depletion."""
+    delay_end: float
+    """Final delay parameter for resonator depletion."""
+    delay_step: float
+    """Step delay parameter for resonator depletion."""
+
+
+@dataclass
+class AllXYDragResults(Results):
```
andrea-pasquale, line 30: Wrong class name

src/qibocal/protocols/characterization/allxy/allxy_resonator_depletion_tuning.py
```diff
@@ -0,0 +1,175 @@
...
+    """AllXYDrag outputs."""
+
+
+@dataclass
+class AllXYDragData(Data):
```
andrea-pasquale, line 35: Wrong class name

src/qibocal/protocols/characterization/allxy/allxy_resonator_depletion_tuning.py
```diff
@@ -0,0 +1,175 @@
...
+) -> AllXYDragData:
+    r"""
+    Data acquisition for allXY experiment varying delay after a measurement.
+    The AllXY experiment is a simple test of the calibration of single qubit gatesThe qubit (initialized in the |0> state)
+    is subjected to two back-to-back single-qubit gates and measured. In each round, we run 21 different gate pairs:
+    ideally, the first 5 return the qubit to |0>, the next 12 drive it to superposition state, and the last 4 put the
+    qubit in |1> state.
+
+    The AllXY iteration method allows the user to execute iteratively the list of gates playing with the drag pulse shape
+    in order to find the optimal drag pulse coefficient for pi pulses.
```
andrea-pasquale, line 64: This docstring is copy and paste, you can write to refer to the allxy experiment.

src/qibocal/protocols/characterization/allxy/allxy_resonator_depletion_tuning.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+
+    data = AllXYDragData()
+
+    delays = np.arange(params.delay_start, params.delay_end, params.delay_step).round(4)
```
andrea-pasquale, line 69: At this point can we define this 4 as a top level variable (?)

src/qibocal/protocols/characterization/allxy/allxy_resonator_depletion_tuning.py
```diff
@@ -0,0 +1,175 @@
...
+                sequence.add(ro_pulse)
+                sequence, ro_pulses[qubit] = allxy.add_gate_pair_pulses_to_sequence(
+                    platform,
+                    gates,
+                    qubit,
+                    sequence,
+                    sequence_delay=int(
+                        delay_param
+                    ),  # We need conversion to int due to devices for now
+                    readout_delay=996,  # because we already add a +4
```
andrea-pasquale, line 87: I would avoid hardcoding numbers

src/qibocal/protocols/characterization/allxy/allxy_resonator_depletion_tuning.py
andrea-pasquale: Given that I am not familiar with the experiment an appropriate documentation should be provided, possibly referencing the paper cited in the PR.

andrea-pasquale (changes requested):

andrea-pasquale:
Shouldn't the protocol also have an `update` function where it writes the `depletion_time` on the platform?

src/qibocal/protocols/characterization/allxy/allxy.py
Jacfomg: I guess so. Can I extinguish it ?

src/qibocal/protocols/characterization/allxy/allxy_resonator_depletion_tuning.py
Jacfomg: It's what they do in the paper, they wait 1ms bfore reading

Jacfomg:
> Shouldn't the protocol also have an `update` function where it writes the `depletion_time` on the platform?

How would you fit it ?

src/qibocal/protocols/characterization/allxy/allxy_resonator_depletion_tuning.py
Jacfomg: I will remove it came from the drag

andrea-pasquale:
> > Shouldn't the protocol also have an `update` function where it writes the `depletion_time` on the platform?
> 
> How would you fit it ?

I am not familiar with the experiment. I was just asking if this protocol is supposed to measure the depletion time.
If not forget it.

Jacfomg:
> > > Shouldn't the protocol also have an `update` function where it writes the `depletion_time` on the platform?
> > 
> > 
> > How would you fit it ?
> 
> I am not familiar with the experiment. I was just asking if this protocol is supposed to measure the depletion time. If not forget it.

It would involve comparing two allxy plots, the one coming from this routines and the one coming from the regular allxy and getting the smaller delay when they look similar enough I guess.

andrea-pasquale:
> It would involve comparing two allxy plots, the one coming from this routines and the one coming from the regular allxy and getting the smaller delay when they look similar enough I guess.

Then we should implement a protocol that does exactly this, right? (Outside this PR)

src/qibocal/protocols/characterization/allxy/allxy_resonator_depletion_tuning.py
andrea-pasquale: In any case please remove the hardcoded number

Jacfomg:
> > It would involve comparing two allxy plots, the one coming from this routines and the one coming from the regular allxy and getting the smaller delay when they look similar enough I guess.
> 
> Then we should implement a protocol that does exactly this, right? (Outside this PR)

Either that or we go for the QM way of getting this for the fit.

sorewachigauyo:
The use of reset schemes should also reduce the delay between shots (`relaxation_time` runcard parameter), no? It would be good to also have this new time somewhere.

Jacfomg:
> The use of reset schemes should also reduce the delay between shots (`relaxation_time` runcard parameter), no? It would be good to also have this new time somewhere.

Yes, the active reset schemes will lower the relaxation time, but they will be somewhat limited by this resonator depletion time as you could not adress the qubit properly until the photons are depleted (They are also another schemes to active deplete the resonator to make all this faster). For now, I made this to store it in qibolab https://github.com/qiboteam/qibolab/pull/895.

sorewachigauyo:
I see. I thought this was a reset scheme, my bad.

@alecandido Should probably take note of this when conditional logic is implemented in Qibolab pulse sequencing, which is probably a necessary feature to add eventually, since QM and Zurich already supports this (and probably can have an initial implementation with `fast_reset` in `ExecutionParameters`).

@Jacfomg Have you tested the depletion time with the active reset, if it improves the fidelity?

alecandido:
Yes, conditioning pulses on measurement outcomes is currently unsupported.

Interface-wise, it shouldn't be a too complex addition. Though it is not well-fitting the idea of a pulse sequence, and it would make it closer to a program. So, it is simple only as a patch.
I have the feeling that going towards full-fledged support for flow control, it would make us rebuilding QUA (or anything like the other DSLs), since that will resemble more a program with a processor/interpreter (i.e. the sequencer), rather than a pre-compiled sequence.

Jacfomg:
> @alecandido Should probably take note of this when conditional logic is implemented in Qibolab pulse sequencing, which is probably a necessary feature to add eventually, since QM and Zurich already supports this (and probably can have an initial implementation with `fast_reset` in `ExecutionParameters`).
> 
> @Jacfomg Have you tested the depletion time with the active reset, if it improves the fidelity?

Let's just say that for now it is not supported anymore for the Zurich driver (I made a crappy implementation to try to overcome qibolab limitations a while ago, that we removed now). I tested active reset with Zurich before being aware of this so results were bad, I expect better performance after waiting the resonator relaxation time. Still, we would need to understand first how to optimize the QND of the readout as right now 1/10 states 1 measurered, when I tested this with QM, were then measured again and were 0s, so they didn't keep the state after measurement for a effective active reset scheme.

(Sorry I edited your comment instead of replying)

Jacfomg:
> @sorewachigauyo Yes, conditioning pulses on measurement outcomes is currently unsupported.
> 
> Interface-wise, it shouldn't be a too complex addition. Though it is not well-fitting the idea of a pulse sequence, and it would make it closer to a program. So, it is simple only as a patch. I have the feeling that going towards full-fledged support for flow control, it would make us rebuilding QUA (or anything like the other DSLs), since that will resemble more a program with a processor/interpreter (i.e. the sequencer), rather than a pre-compiled sequence.

At least Zurich has a way of checking if the measurement result was a 0 or a 1. So, having some pulse parameter to tell some of the pulses to be played in those situations would be enough for now.

alecandido:
> At least Zurich has a way of checking if the measurement result was a 0 or a 1. So, having some pulse parameter to tell some of the pulses to be played in those situations would be enough for now.

Yes, and that's why I said we could implement it as a patch. We could just label a pulse with the id of another measurement, and resolve the translation in the drivers.

> That would also allow for custom active reset sequences instead of just the pi_pulse I applied to the state 1 on my trials.

Point is that, for sequences, we need to label all the pulses. Feasible, but hits the boundary. If you wanted to nest two different `if`s, then we'd stop there, or we should start implementing boolean operations in the labeling ids, and that would be a mess...
At that point, much better to give up on the idea of the sequence, and switch paradigm for a program.

Jacfomg:
> > At least Zurich has a way of checking if the measurement result was a 0 or a 1. So, having some pulse parameter to tell some of the pulses to be played in those situations would be enough for now.
> 
> Yes, and that's why I said we could implement it as a patch. We could just label a pulse with the id of another measurement, and resolve the translation in the drivers.
> 
> > That would also allow for custom active reset sequences instead of just the pi_pulse I applied to the state 1 on my trials.
> 
> Point is that, for sequences, we need to label all the pulses. Feasible, but hits the boundary. If you wanted to nest two different `if`s, then we'd stop there, or we should start implementing boolean operations in the labeling ids, and that would be a mess... At that point, much better to give up on the idea of the sequence, and switch paradigm for a program.

Ah yes, but with the patch we could play with some simple active reset schemes that may cut down the relaxation time a lot or just some checks, even if it is just a conditional if you measure a 0 skip the relaxation time and just do resonator deplation and if you measure 1 wait the relaxation time. Just so maybe you could consider this for the qibolab 0.2. Also, the several mesurements seems a bit for more advanced active reset schemes or error correction.

alecandido:
Yes, definitely, I would do that (you could even add an issue in Qibolab, if not yet there, and add the milestone 0.2).

I hope the patch will be backward compatible, I have to think about that (I'm mainly worried about deserialization, since it will be an additional argument). If it is, we don't even need to do necessarily in 0.2.0 or postpone it forever, to avoid breaking, but rather release 0.2.0 asap, and then implement it in 0.2.1.

Jacfomg:
> Yes, definitely, I would do that (you could even add an issue in Qibolab, if not yet there, and add the milestone 0.2).
> 
> I hope the patch will be backward compatible, I have to think about that (I'm mainly worried about deserialization, since it will be an additional argument). If it is, we don't even need to do necessarily in 0.2.0 or postpone it forever, to avoid breaking, but rather release 0.2.0 asap, and then implement it in 0.2.1.
> 
> (The program approach is for sure breaking, and it would be 0.3 or further, no way it could land in current 0.2)
You have this already https://github.com/qiboteam/qibolab/issues/592 , so I won't create another active reset issue if this is okay for you

andrea-pasquale (approved):

PR #832, opened 2024-05-08
Add average option to spectroscopy experiments

### LINKED ISSUE #815: Error bars or no error bars: a case study
Recently (in some definition of the word :D), for some experiments (e.g. spectroscopies, Ramsey, Rabi, AllXY) ability to show error bars in plots has been implemented. This is done at the cost of running the experiments in singleshot mode (instead of hardware averaging), so that statistical moments other than average can be calculated. There is feedback from users that implementing this functionality made experiments slower, so I did a detailed study of one experiment to see how to proceed.

Let's look at the following experiment:
```yaml
  - id: resonator_spectroscopy_low_power
    priority: 0
    operation: resonator_spectroscopy
    parameters:
      freq_width: 100_000_000
      freq_step: 100_000
      power_level: low
      nshots: 8192
      relaxation_time: 5000
```
for a single qubit connected to a qblox cluster.

It takes ~212s to run, while the same experiment in hardware averaging mode takes only ~72s. The readout pulse used for this experiment was 2000ns long, hence the execution time calculated as `nshots * len(sweep) * (sequence.duration + relaxation_time)` is 57.344s. As you can see, executing in hardware averaging mode gets quite close to this figure, while in singleshot mode we are pretty far away. The expectation is that the experiment take execution time + some overhead time spent on preparing instruments for execution.

The readout instrument has a memory limitation, because of which it cannot store `nshots * len(sweep)` numbers in memory, hence qibolab does the execution in chunks. In this particular case it is done in 84 chunks. While in the hardware averaging mode the entire execution is done in one go.

There are two main reasons for the slowness in singleshot mode:

1. inefficiencies in the qibolab driver for qblox.
2. retrieving more data from instruments naturally takes more time.

I did some profiling to understand how much is the contribution of each. Here is a short summary:

- (1) For each chunk the driver processes and uploads the pulse sequence and instrument setting every time. While it could be done only once, since it is always the same. Doing this 84 times roughly amounts to 50s.
- (1) The driver retrieves unnecessary data (`scope_acquisition`) from instruments. Doing this once and for 1000 points usually goes unnoticed, but doing it 84 times for a total of 8192000 points amounts to ~40s.
- (2) Retrieving a total of 8192000 integration results in 84 rounds takes roughly ~50s, while doing only one retrieval of 1000 results is less than 0.05 seconds.

The first two bullet points shall be addressed in qibolab, which will take the experiment runtime from ~212s to ~122s. This is huge improvement, but still almost a minute away from the hardware averaged execution. So, this is definitely an improvement and will be done, but does not solve the problem (fully).

The third point may or may not be addressable. I did a benchmark experiment and confirmed that the time of getting results from instruments depends linearly on the number of points being retrieved. So there is at least no algorithmic bottleneck, however some constant improvements may be possible. However, assuming there is place for improvement, this is outside of the qibo ecosystem, and depends on discussions with hardware vendor(s), so the timeline is vague.

Users are hesitant to upgrade their installations because of this slowness, hence we should address it with high priority. On the qibocal side, for now I suggest to make the error bars optional so that users can choose whether they want to sacrifice runtime performance in favor of error bars or not.

andrea-pasquale:
Closes #815.
Now it would be possible to run spectroscopies experiments with the possibility to run in `AveragingMode.SINGLESHOT` or `AveragingMode.CYCLIC`:
Here is a runcard with an example:
```yaml
actions:
- id: resonator_spectroscopy hardware average
  operation: resonator_spectroscopy
  parameters:
    hardware_average: true
    freq_step: 100000
    freq_width: 5000000
    nshots: 1024
    power_level: low
    relaxation_time: 5000
```
http://login.qrccluster.com:9000/8uH-o4LqQSah8HK7sB7NLQ==/
Here is a runcard with error bars:
```yaml
actions:
- id: resonator_spectroscopy high power single shot
  operation: resonator_spectroscopy
  parameters:
    hardware_average: false
    freq_step: 100000
    freq_width: 5000000
    nshots: 1024
    power_level: low
    relaxation_time: 5000
```
http://login.qrccluster.com:9000/ghqpTrAFTNqmxJJovJgOLQ==/

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `fix_result_average`


pyproject.toml
```diff
@@ -17,8 +17,8 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = ">=3.9,<3.12"
-qibolab = { git = "https://github.com/qiboteam/qibolab.git"}
-qibo = "^0.2.7"
+qibolab = { git = "https://github.com/qiboteam/qibolab.git", branch="fix_result_average"}
```
andrea-pasquale, line 20: To be changed after merging the corresponding PR in qibolab.

src/qibocal/protocols/characterization/utils.py
```diff
@@ -250,39 +259,64 @@ def spectroscopy_plot(data, qubit, fit: Results = None):
...
+                        table_dict(
+                            qubit,
+                            [label, "amplitude", "attenuation"],
+                            [
+                                freq[qubit],
+                                data.amplitudes[qubit],
+                                data.attenuations[qubit],
+                            ],
+                            display_error=False,
+                        )
```
Edoardo-Pedicillo, line 318: Can you build the html table outside the if branches to increase readibility?

src/qibocal/protocols/characterization/qubit_spectroscopy_ef.py
```diff
@@ -156,25 +150,44 @@ def _plot(
...
+                        "Amplitude [a.u.]",
+                        "Anharmonicity [Hz]",
+                    ],
+                    [
+                        fit.frequency[target],
+                        fit.amplitude[target],
+                        fit.anharmonicity[target],
+                    ],
+                    display_error=False,
+                )
```
Edoardo-Pedicillo, line 189: Also here

doc/source/protocols/resonator_spectroscopy.rst
```diff
@@ -44,6 +44,16 @@ The bare resonator frequency can be found setting a large value for the amplitud
             nshots: 1024
             relaxation_time: 100000
 
+.. note::
+    The resonator spectroscopy experiment will be performed by computing
+    the average on hardware. If the user wants to retrieve all the shots
+    and perform the average afterwards it can be done by specifying the
+    entry `hardware_average: false` in the experiment parameters. In this
+    case the plot will include also error bands.
```
hay-k, line 52: Is it only the plot, or the fitting itself also takes error bars into account?

src/qibocal/auto/operation.py
```diff
@@ -71,6 +72,10 @@ class Parameters:
     """Number of executions on hardware"""
     relaxation_time: float
     """Wait time for the qubit to decohere back to the `gnd` state"""
+    hardware_average: bool = False
+    """By default hardware average will be performed."""
+    classify: bool = False
+    """By default shot classification will not be performed."""
```
hay-k, line 78:
```suggestion
    """By default qubit state classification will not be performed."""
```

src/qibocal/protocols/characterization/utils.py
```diff
@@ -133,22 +133,23 @@ def lorentzian_fit(data, resonator_type=None, fit=None):
...
+            if not np.isnan(data.error_signal).any():
+                fit_parameters, perr = curve_fit(
+                    lorentzian,
+                    frequencies,
+                    voltages,
+                    p0=model_parameters,
+                    sigma=data.error_signal,
+                )
+                perr = np.sqrt(np.diag(perr)).tolist()
+                return model_parameters[1] * GHZ_TO_HZ, list(model_parameters), perr
```
hay-k, line 145: Should the value of `fit_parameters` be used here somehow?

doc/source/protocols/resonator_spectroscopy.rst
andrea-pasquale: Also the fitting, I will modify the documentation.

src/qibocal/protocols/characterization/utils.py
andrea-pasquale: The fit_parameters are passed in `model_parameters`, they are later used in the plotting to plot the fit.
I think I am not understanding your point.

hay-k: In the `if` branch, the `fit_parameters` variable is not used anywhere before the function returns. I am wondering if this is expected behavior, in which case I would suggest to replace `fit_parameters` with an underscore to make it explicit that you didn't intend to use it, or if the code should be fixed to use the variable.

andrea-pasquale: Oh, I see it now. There was an error. `fit_parameters` should be used.

hay-k (approved):

Edoardo-Pedicillo (approved):

PR #839, opened 2024-05-13
added flux pulse to the sequence

DavidSarlle:
This PR allows the user to add a flux pulse into the qubit spectroscopy sequence (optional) in order to adjust/check the shift produced in the drive frequency when applying a flux pulse to the qubit. 

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/characterization/qubit_spectroscopy.py
+++ b/src/qibocal/protocols/characterization/qubit_spectroscopy.py
@@ -32,6 +32,7 @@ class QubitSpectroscopyParameters(Parameters):
     flux_duration: Optional[float] = None
     """Flux pulse duration (optional). Same for all qubits."""
 
+
 @dataclass
 class QubitSpectroscopyResults(Results):
     """QubitSpectroscopy outputs."""
@@ -85,8 +86,14 @@ def _acquisition(
 
         if params.flux_amplitude is not None or params.flux_duration is not None:
             # Define default values if parameters are None
-            flux_amplitude = params.flux_amplitude if params.flux_amplitude is not None else 0
-            flux_duration = params.flux_duration if params.flux_duration is not None else sequence.duration
+            flux_amplitude = (
+                params.flux_amplitude if params.flux_amplitude is not None else 0
+            )
+            flux_duration = (
+                params.flux_duration
+                if params.flux_duration is not None
+                else sequence.duration
+            )
 
             # Create the flux pulse with the specified or default parameters
             flux_pulses[qubit] = platform.create_qubit_flux_pulse(
@@ -96,7 +103,6 @@ def _acquisition(
                 amplitude=flux_amplitude,
             )
             sequence.add(flux_pulses[qubit])
-
 
     # define the parameter to sweep and its range:
     delta_frequency_range = np.arange(
--- a/src/qibocal/protocols/characterization/qubit_spectroscopy.py
+++ b/src/qibocal/protocols/characterization/qubit_spectroscopy.py
@@ -84,7 +84,10 @@ def _acquisition(
 
         if params.flux_amplitude is not None:
             flux_pulses[qubit] = platform.create_qubit_flux_pulse(
-                qubit, start=0, duration=sequence.duration, amplitude=params.flux_amplitude
+                qubit,
+                start=0,
+                duration=sequence.duration,
+                amplitude=params.flux_amplitude,
             )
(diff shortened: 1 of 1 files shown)


hay-k:
@DavidSarlle can you please explain in a bit more detail the rationale behind this proposal? What is the problem that this change is attempting to address?

DavidSarlle:
@hay-k there is not a problem behind. This is an optional feature that allows the users to add a flux pulse to a qubit spectroscopy, showing how a qubit shifts the drive frequency if a flux pulse is applied. Introducing it here, we can check if the guess parameter obtained from a qubit flux dependence that shifts a qubit to the frequency needed for other opertions like CZs is correct.

Doing it using this routine is faster and gives the user more precission to fine tune the flux amplitude needed for the shift.

andrea-pasquale:
I'm also confused because this change is basically injecting what we were doing before in the flux dependence here, since we are playing with a **long flux** pulse, which we are planning to drop #841.
In order to address the frequency detuning induced by short flux pulses a proper experiment should be implemented.

DavidSarlle:
Sorry @andrea-pasquale @hay-k, I think the latest code was not uploaded to github for some reason. Check the latest version. Here the user have the possibility of defining the duration and amplitude of the flux pulse applied

andrea-pasquale:
> Sorry @andrea-pasquale @hay-k, I think the latest code was not uploaded to github for some reason. Check the latest version. Here the user have the possibility of defining the duration and amplitude of the flux pulse applied

Thanks for the update @DavidSarlle.
I still don't think that those changes should be included in the qubit spectroscopy experiment. We should look in the literature how to measure the detuning generated by short flux pulses.

DavidSarlle:
@andrea-pasquale ok. No worries. I will keep this routine with me, because it has been very useful chracterizing the CZs. No problem

hay-k:
> there is not a problem behind. This is an optional feature that allows the users to add a flux pulse to a qubit spectroscopy

Well, I didn't mean "what is the problem with the qubit spectroscopy routine that you are trying to fix". What I mean is "what is the use case; what is the problem/situation that you are trying to handle that is not handled by any existing routine". Are you trying to calibrate something with this? Or are you trying to validate calibration that you did by other means? (If yes, what is the metric of validation?)

From the perspective of autocalibration, we want to have strictly defined small routines that are responsible for calibrating a specific thing, or validating specific aspect of calibration. We need to understand your problem better and come up with an optimal solution that is along the lines of this idea. Perhaps a separate dedicated experiment? But what exactly should that experiment do/calibrate/validate?

> Also, this routine is faster and gives the user more precision to fine tune the flux amplitude needed for the shift.

Are you using this experiment to manually sweep the flux pulse amplitude to find a suitable value?

> we can check if the guess parameter obtained from a qubit flux dependence that shifts a qubit to the frequency needed for other operations like CZs is correct.

This seems to be the problem that you are trying to solve, but can you please still elaborate a bit more? This PR uses a very long qubit flux pulse that spans the entire experiment. For CZ gates one does not need such pulses, but rather short pulses that are applied during the CZ gate only. How are these two related? At what stage of CZ calibration is this experiment useful? How do you intend to calibrate the actual short pulses that are needed for CZ gate?

DavidSarlle:
@hay-k as I said, when we are calibrating the flux pulse needed for shiting high freq qubits into the low freq qubit, we need to determine the am;kitude and duration of that pulse. For that, we use the qubit flux dependence, that gives use the possibilty to obtain these values. But qubit flux dependence routine is slower and less precise method (3D measurement) than a qubit spec + flux pulse (2D experiment). So, with that we can confirm and play easily with the amplitude and duration, andsee how it sets the high energy qubit at the desired freq. Also, we can automatize the iteration over a range of amps and durations to obtain the best value to shift the high energy qubit. These values must be very accurate in order to caracterize the couplers or the CZ gate in any chip.

But, again, I have been using this method to characterize the couplers and the CZs in iqm an it is very useful, at lest for me.
You can see how I use it in my notion reports:
https://www.notion.so/qcomp/e0c0346bf4d449c2b0c1e0785067e533?v=c5c2aa2c3fa1497faf893043b12cf3f4&p=9a44e3ae18924710905a1407330e1d14&pm=s

But again, is not a problem if you do not consider that should be the way to do it or that it is not useful.

alecandido:
Replaced by #839

PR #851, opened 2024-05-15
Add protocols to documentation

rodolfocarobene:
Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/doc/source/protocols/allxy_drag_pulse_tuning.rst
+++ b/doc/source/protocols/allxy_drag_pulse_tuning.rst
@@ -1,0 +1,30 @@
+AllXY_drag_pulse_tuning
+=======================
+
+Experiment Description
+----------------------
+
+The AllXY experiment is a simple test of the calibration of single qubit gatesThe qubit (initialized in the |0> state) is subjected to two back-to-back single-qubit gates and measured.
+
+In each round, we run 21 different gate pairs: ideally, the first 5 return the qubit to |0>, the next 12 drive it to superposition state, and the last 4 put the qubit in |1> state.
+
+The AllXY iteration method allows the user to execute iteratively the list of gates playing with the drag pulse shape in order to find the optimal drag pulse coefficient for pi pulses.
+
+Example Runcard
+---------------
+
+.. code-block::
+
+    - id: drag_pulse_tuning
+      operation: allxy_drag_pulse_tuning
+      parameters:
+        beta_start: -0.2
+        beta_end: 0.2
+        beta_step: 0.1
+
+Parameters
+----------
+
+- beta_start: start value for the beta parameter (for Drag pulses)
+- beta_end: end value for the beta parameter (for Drag pulses)
+- beta_step: step value for the beta parameter (for Drag pulses)
--- a/doc/source/protocols/allxy_drag_pulse_tuning.rst
+++ b/doc/source/protocols/allxy_drag_pulse_tuning.rst
@@ -4,9 +4,9 @@ AllXY_drag_pulse_tuning
 Experiment Description
 ----------------------
 
-The AllXY experiment is a simple test of the calibration of single qubit gatesThe qubit (initialized in the |0> state) is subjected to two back-to-back single-qubit gates and measured.
+The AllXY experiment is a simple test of the calibration of single qubit gatesThe qubit (initialized in the \|0> state) is subjected to two back-to-back single-qubit gates and measured.
 
-In each round, we run 21 different gate pairs: ideally, the first 5 return the qubit to |0>, the next 12 drive it to superposition state, and the last 4 put the qubit in |1> state.
+In each round, we run 21 different gate pairs: ideally, the first 5 return the qubit to \|0>, the next 12 drive it to superposition state, and the last 4 put the qubit in \|1> state.
 
 The AllXY iteration method allows the user to execute iteratively the list of gates playing with the drag pulse shape in order to find the optimal drag pulse coefficient for pi pulses.
 
--- a/doc/source/getting-started/protocols.rst
+++ b/doc/source/getting-started/protocols.rst
@@ -10,7 +10,7 @@ Here is a scheme with the protocols currently available in Qibocal.
 The source code for each protocol is available in the :ref:`Components`.
 
 .. toctree::
-    :maxdepth: 2
+    :maxdepth: 1
     :caption: Protocols
 
     ../protocols/index
--- a/doc/source/getting-started/runcard.rst
+++ b/doc/source/getting-started/runcard.rst
@@ -45,5 +45,5 @@ Under ``actions`` are listed the protocols that will be executed sequentially.
 
 For each protocol the user needs to specify the following:
     * ``id``: custom id chosen by the user.
-    * ``operation``: protocol available in ``Qibocal``. See :ref:`calibration_routines` for a complete list of the protocols available.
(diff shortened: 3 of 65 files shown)


rodolfocarobene:
Hi @andrea-pasquale , what is the difference between Spin-Echo, Ramsey and T2? 

I know the difference between ramsey and spin-echo, but I don't see where T2 fits :-)

andrea-pasquale:
> Hi @andrea-pasquale , what is the difference between Spin-Echo, Ramsey and T2?

T2 is basically a Ramsey experiment with two differences:
- there is no detuning
- the post-processing is just an exponential

We decided to split them given that to correct the qubit frequency through a Ramsey experiment if you apply enough detuning you should be able to do it using short scans, but the T2 estimate will probably be poor. While with the T2 experiment we aim at extracting T2 with longer scans assuming that possible detunings on the drive frequency have already been corrected, hence why the fit consists just of an exponential. This is just a possible approach... that should be documented properly :)

doc/source/protocols/allxy_drag_pulse_tuning.rst
```diff
@@ -0,0 +1,30 @@
...
+        beta_start: -0.2
+        beta_end: 0.2
+        beta_step: 0.1
+
+Parameters
+----------
+
+- beta_start: start value for the beta parameter (for Drag pulses)
+- beta_end: end value for the beta parameter (for Drag pulses)
+- beta_step: step value for the beta parameter (for Drag pulses)
```
Edoardo-Pedicillo, line 30: Hi @rodolfocarobene, thank you very much for the documentation. I have some doubts regarding the contents:
1. I will skip the `Parameters` section and eventually link the (updated) routine docs in the API reference, because this is just a duplication of it and could be difficult to maintain.
2. We could provide some pictures of the reports (I know it is still a draft; maybe you have already planned it).

rodolfocarobene: Hi @Edoardo-Pedicillo! Regarding (2) I fully agree. I think many of the routines could benefit by a better explanation (sometimes is non-existent at the moment) and of an example plot. But  anyway I think this could be done in the next weeks / months, without much hurry. The main point was to provide a complete list of the routine and this PR is doing that.

Regarding (1), on the other hand, I'm not fully sure. It is true that is duplication, but almost the same could be said of all the documentation. 
In this way the user has a single page with all the parameters and useful information. I personally don't think that the user should be required to look into the source.
Moreover, it happens sometimes that the parameters are not even in the same file as the routine (because they are define in the utils.py or in the signal version of the experiment) and this would mean asking the user to accept something like a treasure hunt...

I don't know if this is the right solution, I understand this put an additional burden on the developers. But I think taking care of the documentation should be part of the standard work.

andrea-pasquale: > Regarding (1), on the other hand, I'm not fully sure. It is true that is duplication, but almost the same could be said of all the documentation.
> In this way the user has a single page with all the parameters and useful information. I personally don't think that the user should be required to look into the source.
> Moreover, it happens sometimes that the parameters are not even in the same file as the routine (because they are define in the utils.py or in the signal version of the experiment) and this would mean asking the user to accept something like a treasure hunt...

I was able to generate correctly the API reference in #854.
Let me know what you think @rodolfocarobene, eventually we could start from there.

rodolfocarobene: Hi @andrea-pasquale , I just tried #854, but I'm not sure of exactly what it does... I cannot find in the API-reference the docomentation of each protocols.

In general, do you think a "protocol" list as I did in this PR could be useful or not much? I believe it could be important, but if you disagree it's fine and I will eventually close this PR. :-)

Edoardo-Pedicillo: > Regarding (1), on the other hand, I'm not fully sure. It is true that is duplication, but almost the same could be said of all the documentation.
> In this way the user has a single page with all the parameters and useful information. I personally don't think that the user should be required to look into the source.

The user is not supposed to read the source code, I was suggesting just to move all the documentation in the .py files that will be rendered by Sphinx on the website.

> Moreover, it happens sometimes that the parameters are not even in the same file as the routine (because they are define in the utils.py or in the signal version of the experiment) and this would mean asking the user to accept something like a treasure hunt...

About the signal, this could be solved easily.
https://stackoverflow.com/questions/30344289/how-can-i-show-inherited-members-of-a-class-in-my-sphinx-documentation
Concerning the utils, we could solve it with inheritance and forcing all routines to provide the Parameters (for the moment, it seems just the Couplers routines share the Parameters in the Utils).


> I don't know if this is the right solution, I understand this put an additional burden on the developers. But I think taking care of the documentation should be part of the standard work.

I am perfectly aware of it, but if some tasks could be automated, why should we not do it ?

rodolfocarobene: Ah sorry, I didn't remove some files so sphinx was not re-compiling them.

For sure it's not easy to write a proper qibocal documentation, but I think that we can do something (and maybe should before any paper). The problem with just having what you generated in @854 is that there is no reference on what the routines do (and sometime the name is not self-explanatory, for example "chevron" means something very unique in qibocal, in respect to others) and there is no mention of the parameters. 
If I was a new user of qibocal and wanted to do some experiment already existing, I would have little idea on how to use it.

Imo, asking a user to look at the code / code-documentation should be avoided. In particular if the user is just a user and does not need to contribute. It's making it harder for them to start to use qibocal.

I know it's a bit of additional work for you guys, but I don't think it would be particularly cumbersome... One would have to change / add a documentation file only in the case of modifying a routine and we already have `documentation is updated` checkbox in the PR for this reason.

Moreover, I see this PR as a "starting point", in the sense that we could merge it and than start adding more examples / images to the protocol descriptions. To me it seems much better than nothing, although not super useful if the description is just few lines.

rodolfocarobene:
I just recompiled #854 and I see that there is everything in the produced page (even the parameters).

Personally, I think it's lacking a bit of usability but I'll not insist. If you disagree with me, I'll close this PR :-)

doc/source/protocols/allxy_drag_pulse_tuning.rst
andrea-pasquale: I think we are confusing the API-reference with "tutorials", which is what you would like to have I think.

What you started is needed, but as we already said a tutorial should not repeat what it is already mentioned in docstrings.

> Imo, asking a user to look at the code / code-documentation should be avoided. In particular if the user is just a user and does not need to contribute. It's making it harder for them to start to use qibocal.

It depends which type of user, the user who doesn't know anything about the library will most likely go through tutorials first and afterwards for specific problems it will look at the API reference, this is why it is needed.

We should reproduce something similar to what we already have in resonator spectroscopy and resonator punchout. What you started is ok, however I believe that we should not replicate the documentation. For example for the parameters you shoud just **link** the corresponding module/class in the API reference. 

In the tutorials we should have:

- A proper description of the experiment
- Code examples (runcard)
- References
- Plots and post-processing explanation

If possible the tutorial should explain in more detail the experiment (not a single line explanation) and it should not replicate the documentation of the `_acquisition` function and the `Parameters` function.

If you want to start to add tutorials for some of the protocols as I described above you are welcome to contribute :)

PR #865, opened 2024-05-21
Expose single protocol run to executor

andrea-pasquale:
Adds possibility to run one protocol at a time using `Executor`.
Possible script:
```py
from qibocal.auto.execute import Executor
from qibocal.auto.mode import ExecutionMode
from qibocal.auto.runcard import Action
import pathlib

protocol = {"id": "flipping",
            "operation": "flipping",
            "targets": [0,1,2],
            "parameters": {
                "nflips_max": 20,
                "nflips_step": 2,
                "detuning": +0.1,
            }}

action = Action(**protocol)
executor = Executor([action], platform=create_platform("dummy"), output=pathlib.Path("FOLDER"))

completed = executor.run_protocol(ExecutionMode.acquire, action)
```
@alecandido I think this should be more or less what we discussed :)

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


alecandido:
To make it (slightly) simpler, I would:
- use a custom constructor, e.g. `Executor.create
- accept for the first parameter a `list[Union[Action, ActionLike]]`, where `ActionLike` is a dictionary that could be loaded into an action
- for the `platform` a `Union[Platform, str]` (make the call to `create_platform()` internally, if it's not already an instance of `Platform`)
- for the `output` a `PathLike` (same of the previous)
- re-export `Executor` and `ExecutionMode` top-level

turning your example into:
```py
from qibocal import Executor, ExecutionMode

protocol = {"id": "flipping",
            "operation": "flipping",
            "targets": [0,1,2],
            "parameters": {
                "nflips_max": 20,
                "nflips_step": 2,
                "detuning": +0.1,
            }}
executor = Executor.create([protocol], platform="dummy", output="FOLDER")

completed = executor.run_protocol(ExecutionMode.acquire, action)
```

This is just for interfacing purposes, saving the user some boilerplate and limiting the required understanding of Qibocal internals.

alecandido:
Concerning the mode, currently, Qibocal is using:
https://github.com/qiboteam/qibocal/blob/a668a08c9cfd118f185bcd17ab80acc00fbc27eb/src/qibocal/auto/mode.py#L4-L8

However, we could use an [`IntFlag`](https://docs.python.org/3/library/enum.html#enum.IntFlag) and define `auto` as:
```py
auto = ExecutionMode.acquire | ExecutionMode.fit | ExecutionMode.report
``` 
with all the bits set.

For `.run_protocol()` I'd just use `ExecutionMode.acquire | ExecutionMode.fit`. We could give it a name, or, to make it simpler, set the default to `None`, and set the mode internally to that, instead of retrieving it by name.

src/qibocal/auto/execute.py
```diff
@@ -18,23 +20,28 @@ class Executor:
...
+    platform: Optional[Platform] = None
     """Qubits' platform."""
     update: bool = True
     """Runcard update mechanism."""
 
+    def __post_init__(self):
+        # create default history
+        if self.history is None:
+            self.history = History({})
+
```
alecandido, line 38: The proposed usage of `Execute.create()` would save you the burden of maintaining a `__post_init__`, and it would allow you to keep all members required (such that you don't have to later check `if executor.kljlkjkl is not None`).

src/qibocal/auto/execute.py
```diff
@@ -50,25 +57,31 @@ def load(
...
+        )
+
+        if mode.name in ["autocalibration", "fit"] and self.platform is not None:
+            completed.update_platform(platform=self.platform, update=self.update)
+
+        self.history.push(completed)
+
+        return completed
+
+    def run_protocols(self, mode: ExecutionMode) -> Iterator[Id]:
```
alecandido, line 78: Ok, I agree we need two different objects, and one of them could largely consist in calling the other. But one of the two could be called just `.run()`, both for simplicity and to better differentiate them :P

alecandido:
Apart from names and other interface suggestions, the current proposal looks essentially good: it should be minimal, but it should allow running a fully flexible calibration program.

Of course, if you have many of these protocols (calibration programs, not individual routines) there will be repeated patterns (e.g. the dependencies) and we could abstract them in the framework.
But given that (for the time being) we do not even have a single one of them, I believe this could be an optimal starting point :)

andrea-pasquale:
Thanks for the review @alecandido.
I think I have addressed all comments.
There is only one think worth mentioning `ExecutionMode.REPORT` will not actually generate any report.
It is meant to indicate a mode where both acquisition and fit are skipped. The data is just loaded and then pass it to the `ReportBuilder` to perform the report generation. At this point we could even drop it or expose to the user a proper way to generate the report.

src/qibocal/__init__.py
```diff
@@ -2,6 +2,8 @@
 
 import importlib.metadata as im
 
+from qibocal.auto.execute import Executor
```
alecandido, line 5: I'm trying to propose this everywhere, though it is not yet my definitive suggestion for the public API definition, but only part of it: add below
```py
__all__ = ["Executor"]
```

src/qibocal/auto/execute.py
```diff
@@ -34,7 +38,7 @@ class Executor:
     def load(
         cls,
         card: Runcard,
-        output: Path,
+        output: Path = None,
```
alecandido, line 41: Either
```suggestion
        output: Path,
```
or
```suggestion
        output: Optional[Path] = None,
```
(and the same would be true for the following two arguments).

Why do you need the default now? Nothing against that, just to understand.

src/qibocal/auto/mode.py
```diff
@@ -1,8 +1,12 @@
-from enum import Enum
+from enum import IntFlag, auto
```
alecandido, line 1: Ok, I said myself `IntFlag`, but even just `Flag` could be enough (I got confused, and I forgot that there were two different ones).

It will work in the exact same way, simply you do not need that to be an integer.

src/qibocal/auto/task.py
```diff
@@ -102,7 +101,7 @@ def run(
                 completed.data, completed.data_time = operation.acquisition(
                     parameters, platform=platform
                 )
-        if mode.name in ["autocalibration", "fit"]:
+        if mode & ExecutionMode.FIT:
```
alecandido, line 104:
```suggestion
        if ExecutionMode.FIT in mode:
```

This is slightly more clear (the `__contains__` is offered by the `Flag`/`IntFlag` implementation, and this is the documented way of doing this check).

alecandido (approved):
(Very) minor further suggestions, but to me, it's perfect.

It should be non-breaking for the current way of using it (i.e. CLI + runcards), and it should allow running individual protocols.

The only thing missing is an example of how to use it, but given that is a non-breaking/minimal contribution you could even merge it like this, and contribute the example in a further PR, considering this just as a refactor that will allow that feature (and not a new feature to be documented).

andrea-pasquale:
> The only thing missing is an example of how to use it, but given that is a non-breaking/minimal contribution you could even merge it like this, and contribute the example in a further PR, considering this just as a refactor that will allow that feature (and not a new feature to be documented).

I can write an example on how to use it. We have already a section in documentation on how to use qibocal as a library https://qibo.science/qibocal/stable/tutorials/advanced.html#how-to-use-qibocal-as-a-library, which was doing similar things in a more complicated way. At this point shall we remove it and modify this section with the new layout proposed here?
Again the only difference concern the plotting, I believe that we the current layout it is not trivial to generate plots.
What do you think?

scarrazza:
In view of #866, I would suggest a minor API change (demo code):
```python
from qibocal import Executor, Routine

setup = {"id": "flipping",
            "operation": "flipping",
            "targets": [0,1,2],
            "parameters": {
                "nflips_max": 20,
                "nflips_step": 2,
                "detuning": +0.1,
            }}

def acquire(data, ...):
    """Custom routine"""
    return ...

my_routine = Routine(acquire)
executor = Executor.create(platform="dummy", output="FOLDER", optional_routines=[my_routine])

for i in range(10):
    completed = executor.run_protocol("flipping", mode="acquire", setup=setup)
    setup['detuning'] += i
    completed2 = executor.run_protocol("my_routine", mode="acquire", setup=setup)
```
Where:
- we do not have to specify in `Executor.create` the protocols that are already registered internally in qibocal.
- we have the flexibility to update the runcard on-the-fly and execute an arbitrary number of time the same protocol.
- we can create an `Executor` with custom plugin routines.

alecandido:
> We have already a section in documentation on how to use qibocal as a library [qibo.science/qibocal/stable/tutorials/advanced.html#how-to-use-qibocal-as-a-library](https://qibo.science/qibocal/stable/tutorials/advanced.html#how-to-use-qibocal-as-a-library), which was doing similar things in a more complicated way. At this point shall we remove it and modify this section with the new layout proposed here?

Yes, and I'd move it out from "Advanced Examples", and rather make its own page under tutorials (just to avoid passing the idea that this is something you should beware of).

> Again the only difference concern the plotting, I believe that we the current layout it is not trivial to generate plots.
> What do you think?

Currently, we're coupling report generation to the execution, though the two things are well separate.
What we could do is to expose the report function
https://github.com/qiboteam/qibocal/blob/77f7bd1364a19bb3b749460d63b337e4c6a9e2a1/src/qibocal/cli/report.py#L65
(that already has a pretty simple API), even dropping the `executor` argument, i.e. assuming that an executor already ran.
What we will need instead will be a `History`, but I would make that optional (if possible), and try to deserialize the `History` from the path itself (the purpose of having a `History` anyhow is to make the input path optional, i.e. you could generate everything in memory, before dumping to disk).
The only other `executor` attribute used is:
https://github.com/qiboteam/qibocal/blob/77f7bd1364a19bb3b749460d63b337e4c6a9e2a1/src/qibocal/cli/report.py#L99
but we can inherit that in the `history`.

This should be sufficient for `qq report` (it is not `qq auto`, so it's not supposed to run a calibration on its own), and `report(path)` or `report(path, executor.history)` should be a simple enough API even for an end user (at which point I'd take out the `report` function from the `cli` subpackage, and expose top-level).
We could even do `report(executor=executor)`, in such a way that the path and history are automatically inferred (or `report(executor)` and check the type, also using [`@overload`](https://docs.python.org/3/library/typing.html#typing.overload) for that).

In any case, I'd do this in a separate PR.

src/qibocal/auto/execute.py
```diff
@@ -50,25 +56,72 @@ def load(
...
+            platform if isinstance(platform, Platform) else create_platform(platform)
+        )
+        return cls(
+            actions=[],
+            history=History({}),
+            output=Path(output),
+            platform=platform,
+            targets=list(platform.qubits),
+            update=True,
+        )
```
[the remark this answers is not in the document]
alecandido: Ok, against my own forecast, the rant syntax is actually possible with little Python magic:
https://github.com/alecandido/py-import-from-object/blob/main/test.py

So, we could abuse the `import` mechanism to pre-register the `executor` in the routines, and truly avoid repeating the invocation of its methods (though that would remain an option).

Of course this is just syntactic sugar, but it would make the calibration program exactly like a normal Python program, where the routine *auto-register* results in the executor. With a thin and transparent layer of magic.

alecandido:
Since this is working, and #869's target is to provide a simpler syntax to use what's already here, I wonder whether we should aim to merge this now.

I will re-review, but I've already approved it, and this PR is just exposing something that is already more or less there in `main` (i.e. executing routines explicitly with the `Executor`, now possible individually).

Since it's at an internal stage, we can avoid documenting it as a public feature (the public interface will come with #869), but the improvement is simple and valid on its own.

src/qibocal/auto/execute.py
```diff
@@ -50,25 +56,72 @@ def load(
             update=update,
         )
 
-    def run(self, mode):
+    @property
+    def _actions_dict(self):
+        """Helper dict to find protocol."""
+        return {action.id: action for action in self.actions}
```
alecandido, line 62: The other thing we may consider is to kill the `.actions` attribute of the `Executor`, and essentially kill `Executor.load()` as well, just keeping `Executor.run()` for the runcards.

I.e., `.load()` is there to load a runcard, and `.run()` is there to execute it (since if you don't have a runcard you would use a script, not running all routines in one shot). So, we have two methods that are always going to be called one after the other (`Executor.load(runcard).run(mode)`), thus it could be just one (`Executor.run(runcard, mode)`).

src/qibocal/auto/execute.py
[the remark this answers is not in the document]
alecandido: Delegated to #869

src/qibocal/cli/report.py
```diff
@@ -81,9 +80,12 @@ def report(path: pathlib.Path, executor: Optional[Executor] = None):
...
     # load executor
     if executor is None:
-        executor = Executor.load(runcard, path, targets=runcard.targets)
+        executor = Executor.load(
+            runcard, path, targets=runcard.targets, platform=GlobalBackend().platform
+        )
         # produce html
-        list(executor.run(mode=ExecutionMode.report))
+        # TODO: skip report in a proper way
+        list(executor.run(mode=[]))
```
alecandido, line 88: However, to replace `Executor.load()` we'll need to be able to deserialize a `History` (i.e. let's use `History.load(path)`), and replace `executor` occurrences here with `history`, as proposed in https://github.com/qiboteam/qibocal/pull/865#issuecomment-2124015170.

But this starts being a separate PR.

runcards/actions_qq.yml
```diff
@@ -5,11 +5,11 @@ targets: [0,1]
...
-#   - id: resoantor_amplitude
-#     operation: resonator_amplitude
-#     parameters:
-#       amplitude_step: 0.1
-#       amplitude_stop: 0.5
+  - id: resoantor_amplitude
+    operation: resonator_amplitude
+    parameters:
+      amplitude_step: 0.1
+      amplitude_stop: 0.5
```
Edoardo-Pedicillo, line 12: comment the lines

Edoardo-Pedicillo (approved):

alecandido:
@andrea-pasquale @Edoardo-Pedicillo should we then solve the conflicts and merge this PR?

I know we are going to change it even further, in #870 and #869, but this PR is rather self-contained, despite introducing a temporary solution (in any case, `main` is not a release, and the other two PRs will come quite soon).

PR #866, opened 2024-05-21
Plugins

alecandido:
The gist of this proposal is to query further packages for `Routine`s, to allow extending Qibocal once installed (without patching it at runtime).

The core of the idea is to register plugins in the `Executor` (since we're already using it to supervise the execution), directly or automatically, and then use the executor to retrieve the operation.

Direct registration will be done passing a list of `Operation`-like objects.
Autodiscovery will inspect installed packages to match some pattern, automatically loading those objects.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,19 +3,25 @@
     "cachix": {
       "inputs": {
         "devenv": "devenv_2",
-        "flake-compat": "flake-compat_2",
+        "flake-compat": [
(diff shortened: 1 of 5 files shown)


alecandido:
Closed in favor of direct `Routine` object usage
https://github.com/qiboteam/qibocal/pull/865#issuecomment-2123923273

PR #869, opened 2024-05-24
Simplified calibration scripts

alecandido:
Here (there will be) the implementation of the *rant* syntax of #865.

It seems that it is actually possible, and it should result in simple and nice scripts.

Plugins (i.e. protocols set extensions) will also be supported, though we now have the option of avoiding auto-discovery, since there will be room for explicit registration (that seems more intuitive and transparent).

#### This PR

- [x] the routines call is not optimal for the new simplified layout
    - instead of requesting the full `Action` to be passed as input, we should spread its arguments over its `*args` and `**kwargs`, and later collect them into the `Action` (providing suitable defaults, whenever possible)
- ~~example using [`rxpy`](https://github.com/ReactiveX/RxPY)~~ cf. #916 
- ~~process output~~
  - for the time being, the full output is being returned (it may change in the near future)

#### Beyond

- ~~tests are creating files that are left in the repo: we should use the `tmp_path` fixture to avoid this (possibly changing working directory there)~~
    - ~~in particular, the folders: `cls_results`, `data`, `qubit0`..`qubit4`~~
    - done in #909 a6f26e73de3227f4f9ec44acdb83468a6a1d0cb5
- many `Data` and `Results` contain as attributes `dict[QubitId, ...]` - it would be nicer if this redundancy (the qubit dependence of these values) were managed at a higher level (not individually per-routine)
    - if we need to mark this per-routine or per-attribute, we could consider using the fields' metadata for the purpose


alecandido:
I added the call simplification: now it is possible to use the names of the parameters as keyword arguments.

There are three semi-reserved names: `id`, `mode`, and `parameters` itself.
These can not be used as keyword arguments for parameter names, as they are interpreted as other routine attributes.
However, the `parameters` argument can still be used to pass a dictionary, that will be updated with the further explicit keyword arguments. In that dictionary, there is no reserved key (but using `id`, `mode`, and `parameters` as names for routines' parameters is discouraged anyhow, as it would be confusing, despite being possible).

https://github.com/qiboteam/qibocal/blob/d63bc73efd189cf316c7bd31bde6341af6151a3c/src/qibocal/auto/execute.py#L46-L49

alecandido:
The main point of the PR is implemented, and vaguely tested.

However, I'm asking for review to have an actual feedback, not necessarily to merge immediately (and in any case, this is pointing #909, which is pointing #921).
Moreover, the script support is not completed, since, to fully reproduce runcards, we need other basic functionalities like those described in #922.

But with a bit of boilerplate scripts could be already used, reimplementing in each and every script the saving operations in https://github.com/qiboteam/qibocal/blob/manage-output/src/qibocal/cli/autocalibration.py and the other commands.
So, I'm inclined to merge this with this feature alone, and introduce further features in further PRs, to avoid super-huge PRs and consequently more complicated reviews.

alecandido:
In b03665e I added a new (small) feature: positional arguments. Now, it should be also possible to omit the name of the parameters, and rely on the order they are listed in the `Parameters` class.

I checked with the RX calibration script that nothing is broken.

I'm not dedicating a huge effort to properly test all these features, also because they are *experimental*. As soon as we will improve a bit these scripts (and start using them), I will also expand the testing infrastructure, to make them even more reliable.

src/qibocal/auto/execute.py
```diff
@@ -1,22 +1,67 @@
...
+        action = Action.cast(
+            source={
+                "id": id,
+                "operation": name,
+                "parameters": params | positional | kwargs,
+            }
+        )
+        return executor.run_protocol(protocol, parameters=action, mode=mode)
+
+    return wrapper
```
andrea-pasquale, line 62: Something useful here if we want to consider scripts for multi-qubit calibration could be to pass also `targets`.
In this way we could parse good/bad qubits when calibrating and decide to execute specific protocols on specific targets. We could keep this also for a future PR.

alecandido: We will need it to reproduce all runcards features. But since it's not much used, I'd postpone it for a moment.

alecandido: #929

src/qibocal/auto/execute.py
```diff
@@ -26,20 +63,62 @@ class Executor:
     """Qubits' platform."""
     update: bool = True
     """Runcard update mechanism."""
+    name: Optional[str] = None
+    """Name, used just as a label but also to register the module."""
```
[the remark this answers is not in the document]
alecandido: Improved in 38bb4c0

runcards/rx_calibration.py
andrea-pasquale, line 1: I might add some documentation here, I can do it directly in this PR.
If instead we will open a separate PR to update the documentation I can also do it later on.

src/qibocal/auto/execute.py
```diff
@@ -1,22 +1,67 @@
...
+        parent_module = importlib.import_module(parent_name)
+        setattr(parent_module, child_name, obj)
+
+    sys.modules[qualified] = obj
+    obj.name = obj.__name__ = qualified
+    obj.__spec__ = None
+
+
+def _wrapped_protocol(executor: "Executor", protocol: Routine, name: str):
+    def wrapper(*args, parameters=None, id=name, mode=AUTOCALIBRATION, **kwargs):
```
andrea-pasquale, line 50: Although these are internal functions/methods I would add a few comments to explain what is going on.

andrea-pasquale (commented):
Thanks @alecandido, I have only a few minor comments.
We should also write a proper example in the documentation about how to use the new syntax. For me we can also do it in a separate PR. Let me know what you think.

alecandido:
Sorry, I didn't see the review. I'll reply to your comments asap

runcards/rx_calibration.py
alecandido: I would say that this can be considered "experimental" for the time being, so we can try it internally, before advertising as a feature.

Since a few things may change, I would avoid documenting in details, and postpone the documentation for when we'll decide to make it the general way for running scripts.

In particular, we won't need to rewrite it from scratch, because this is just going to *replace* the example you wrote, and @rodolfocarobene updated in #903 (true that `run_protocol` will always be available, but I'd advertise only one way of doing things programmatically).
We will extend that, but over the existing one.

So, for the time being, let's keep developing the feature, and just test it among us.

src/qibocal/auto/execute.py
alecandido: This I can do :)

src/qibocal/cli/acquisition.py
```diff
@@ -29,6 +29,7 @@ def acquire(runcard: Runcard, folder: Path, force: bool):
 
     # generate meta
     meta = Metadata.generate(path.name, backend)
+    meta.targets = runcard.targets
```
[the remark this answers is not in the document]
alecandido: Even fine. It doesn't change that much for me, but I'm okay with that.

src/qibocal/auto/execute.py
alecandido: Done in 1eda60380125064d785ee09a0a99d0fb2ee565f1

andrea-pasquale (approved):

alecandido:
Nothing new wrt @andrea-pasquale review, I just rebased on top of the current `main`.

src/qibocal/auto/execute.py
```diff
@@ -29,24 +72,138 @@ class Executor:
...
+                # internal attributes should never be routines
+                raise AttributeError
+
+            protocol = getattr(protocols, name)
+            return self._wrapped_protocol(protocol, name)
+        except AttributeError:
+            # fall back on regular attributes
+            return super().__getattribute__(name)
+
+    def _wrapped_protocol(self, protocol: Routine, operation: str):
```
Edoardo-Pedicillo, line 118: Why is `operation` an input ? I expect to be able to run a protocol with just `protocol`.

Edoardo-Pedicillo (approved):
Thanks @alecandido, LGTM

PR #870, opened 2024-05-26
Add possibility to dump and load `History`

andrea-pasquale:
In this PR I address some of the comments raised in #865, given that the changes were substantial I decided to open a separate PR.
 
- [x] (de)serialize `History`
- [x] drop `load` from Executor
- [ ] restore iterative dumping when running autocalibration mode 

There might be a better way to (de)serialize `History`, the current implementation should be acceptable.
The only annoying object left to deal with seems to be `meta.json`, we could think about a way of moving its content somewhere else.

### DIFF
--- a/src/qibocal/auto/history.py
+++ b/src/qibocal/auto/history.py
@@ -1,5 +1,7 @@
 """Track execution history."""
 
+from pathlib import Path
+
 from .runcard import Id
 from .task import Completed
 
@@ -22,7 +24,7 @@ def add_timings_to_meta(meta, history):
     return meta
 
 
-class History(dict[tuple[Id, int], Completed]):
+class History(dict[Id, Completed]):
     """Execution history.
 
     This is not only used for logging and debugging, but it is an actual part
@@ -30,6 +32,14 @@ class History(dict[tuple[Id, int], Completed]):
     former ones from here.
 
     """
+
+    @classmethod
+    def load(cls, path: Path):
+        """To be defined"""
+        instance = cls()
+        for protocol in path.glob("data/*"):
+            instance.push(Completed.load(protocol))
+        return instance
 
     def push(self, completed: Completed):
         """Adding completed task to history."""
--- a/src/qibocal/auto/runcard.py
+++ b/src/qibocal/auto/runcard.py
@@ -1,8 +1,10 @@
 """Specify runcard layout, handles (de)serialization."""
 
 import os
+from dataclasses import asdict
 from typing import Any, NewType, Optional, Union
 
+import yaml
 from pydantic.dataclasses import dataclass
 from qibo.backends import Backend, GlobalBackend
 from qibolab.platform import Platform
@@ -15,6 +17,11 @@ Id = NewType("Id", str)
 
 Targets = Union[list[QubitId], list[QubitPairId], list[tuple[QubitId, ...]]]
 """Elements to be calibrated by a single protocol."""
+
+RUNCARD = "runcard.yml"
+"""Runcard filename."""
+
+SINGLE_ACTION = "action.yml"
 
 
 @dataclass(config=dict(smart_union=True))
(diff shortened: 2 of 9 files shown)


src/qibocal/auto/runcard.py
```diff
@@ -16,6 +18,11 @@
 Targets = Union[list[QubitId], list[QubitPairId], list[tuple[QubitId, ...]]]
 """Elements to be calibrated by a single protocol."""
 
+RUNCARD = "runcard.yml"
+"""Runcard filename."""
+
+SINGLE_ACTION = "action.yml"
```
alecandido, line 24: I'm tempted to tell you to use JSON as well xD

src/qibocal/auto/task.py
```diff
@@ -30,6 +32,14 @@ class Task:
     """Action object parsed from Runcard."""
     operation: Routine
 
+    @classmethod
+    def load(cls, path):
```
alecandido, line 36: Here and below
```suggestion
    def load(cls, path: Path):
```
or
```suggestion
    def load(cls, path: PathLike):
```

(irrelevant remark, but a few type hints are slightly helpful and almost for free)

src/qibocal/auto/runcard.py
```diff
@@ -36,6 +43,15 @@ def __hash__(self) -> int:
         """Each action is uniquely identified by its id."""
         return hash(self.id)
 
+    def dump(self, path):
```
alecandido, line 46: Here and below:
```suggestion
    def dump(self, path: Path):
```
as in the other file.

src/qibocal/auto/history.py
```diff
@@ -22,7 +24,7 @@ def add_timings_to_meta(meta, history):
     return meta
 
 
-class History(dict[tuple[Id, int], Completed]):
+class History(dict[Id, Completed]):
```
alecandido, line 27: > [!NOTE]
> not for this PR

On the one side, I perfectly agree that it has been a useless complication until now.

However, we may want to somehow reintroduce the iteration number for calibration scripts, since now loops will be possible. Though, we could even require the number to be added by the user as part of the identifier itself.

To be clear: this is just a comment, I'm not asking to change anything in this PR

src/qibocal/auto/history.py
```diff
@@ -31,6 +33,14 @@ class History(dict[tuple[Id, int], Completed]):
 
     """
 
+    @classmethod
+    def load(cls, path: Path):
+        """To be defined"""
+        instance = cls()
+        for protocol in (path / "data").glob("*"):
+            instance.push(Completed.load(protocol))
```
alecandido, line 41: > [!NOTE]
> not for this PR

I know I introduced `.push()` myself, I guess after [this](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.push).

I was now wondering whether it would be more Pythonic to call it `.append()`...

(I'd never propose such a vanity change for the public API, but this method should not be part of it, at least not yet, and most likely it won't be for a while)

src/qibocal/auto/task.py
```diff
@@ -162,6 +172,17 @@ def data(self, data: Data):
...
+    def dump(self, path):
+        """test"""
+        self.task.dump(self.datapath)
+
+    @classmethod
+    def load(cls, folder: Path):
+        """Loading completed from path."""
+
+        task = Task.load(folder)
+        return cls(task=task, folder=folder.parent.parent)
```
alecandido, line 184:
```suggestion
        return cls(task=task, folder=folder.parents[1])
```
just a shortcut

alecandido (approved):

src/qibocal/auto/history.py
andrea-pasquale: I would until we actually have for loops to discuss this. Shall we open an issue about this?

src/qibocal/auto/history.py
andrea-pasquale: Between `push` and `append` I don't have a strong preference.

src/qibocal/auto/history.py
alecandido: Not sure, because we may leave it to the user... just raise an error in case you reuse an id already present...

andrea-pasquale: Implemented in https://github.com/qiboteam/qibocal/pull/870/commits/726867e881b7ffead3a67ce6258b45ebd1fc627c

Edoardo-Pedicillo (approved):

PR #871, opened 2024-05-27
Documentation for tof and kernels

rodolfocarobene:
Hi, I'm trying to be useful.
I've added a simple documentation page for the two signal experiments.
The two plots are old ones (kernel took from @Jacfomg )... I don't know if any improvement has been done regarding this experiments and if the plots are good enough!

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


Jacfomg:
Fitting aside which I think is right for that plot, since that increasing slope you may to remove since it is related to the 3D acvity getting filled with photons, and it is not how it is seem in 2D devices as I could check with Zurich and QM. Maybe we should include this in the docs, and somewhere in qibolab that the rfsoc raw acquisition is doing some postprocessing (demodulation at least) that QM and Zurich do not include as part of a RAW acquisition.
image

doc/source/protocols/signal/calibrate_kernel.rst
```diff
@@ -0,0 +1,31 @@
+Calibrate Discrimination Kernels
+================================
+
+This experiment is quite similar to the time-of-flight experiment but has a different objective.
+
+In the time-of-flight experiment, we were only interested in the timing of the pulse, regardless of the current qubit state. Here, on the other hand, we want to build a "vector of weights," with a value for each point in the measurement, to maximize the differences between :math:`\ket{0}` and :math:`\ket{1}`. Points that are further away, considering the two states, are given a higher weight since they are "more interesting" for discrimination purposes.
```
Jacfomg, line 6: Include ref: https://arxiv.org/pdf/2106.06173 fig 12 for example ?

doc/source/protocols/signal/calibrate_kernel.rst
```diff
@@ -0,0 +1,31 @@
...
+    - id: calibrate kernels                     # custom ID of the experiment
+      operation: calibrate_state_discrimination # unique name of the routine
+      parameters:
+        nshots: 1024
+        relaxation_time: 20_000
+
+Acquisition
+^^^^^^^^^^^
+
+The acquisition procedure is described in :func:`qibocal.protocols.signal_experiments.calibrate_state_discrimination._acquisition`. It is important to note that this experiment makes use of the RAW acquisition mode, which may require some specific care depending on the instrument employed.
```
Jacfomg, line 22: So, what I mentioned about some devices doing demodulaiton as part of a RAW acquisition ?

doc/source/protocols/signal/calibrate_kernel.rst
```diff
@@ -0,0 +1,31 @@
...
+
+.. image:: kernel.png
```
Jacfomg, line 24: We should try to get a better plot where we say the the decay as well, @hay-k @stavros11 would Zurich or QM be able to acquire for different duration than the readout pulse lenght ?

doc/source/protocols/signal/time_of_flight.rst
```diff
@@ -0,0 +1,39 @@
+Time Of Flight (Readout)
+========================
+
+In this section, we present the time-of-flight experiment for qibocal.
+
+In the `time of flight` experiment, we measure the time it takes for a readout pulse to travel to the qubit and back, being acquired by the control instrument ADCs.
+
+Carefully calibrating this delay time is important to optimize the readout length. In particular, it is useful to acquire for a duration equal to that of the pulse, where it is possible to see differences (in amplitude and phase) between the two states of a qubit.
```
Jacfomg, line 8:
```
Carefully calibrating this delay time is important to optimize the readout length.
```

I don't understand this part, you will not change you readout duration based on the time of flight measure, right ?

doc/source/protocols/signal/time_of_flight.rst
```diff
@@ -0,0 +1,39 @@
...
+        relaxation_time: 20_000
+
+Although it is possible to avoid setting a specific readout amplitude, it is generally useful to set a high value here. Indeed, we are not looking for the optimal amplitude but want to have a signal with enough power so that it is clear when it starts.
+
+Acquisition
+^^^^^^^^^^^
+
+The acquisition procedure is described in :func:`qibocal.protocols.signal_experiments.time_of_flight_readout._acquisition`. It is important to note that this experiment makes use of the RAW acquisition mode, which may require some specific care depending on the instrument employed.
+
+.. image:: time_of_flight.png
```
Jacfomg, line 32: I think we due to this slope we may need to provide two plots, 3D and 2D. I would like to verify this slope for 3D, but I don't know if we have 3D cavities right now here.

doc/source/protocols/signal/time_of_flight.rst
```diff
@@ -0,0 +1,39 @@
...
+
+In this image, for example, we can see that the fit procedure did not find the correct pulse starting point (that is around 200). This can be improved increasing the window size.
```
Jacfomg, line 34:
```suggestion
In this image, for example, we can see that the fit procedure did not find the correct pulse starting point (that is around 200).
```

Are you really sure we should not discard this points as there changes to the rectangular pulse sent due to the 3d cavity ?

Jacfomg: Also the fit may not work as well for signals we you are demodulating, we then would need different fits or at least moving averages depeding of what each devices calls RAW.

doc/source/protocols/signal/time_of_flight.rst
```diff
@@ -0,0 +1,39 @@
...
+        nshots: 1024
+        relaxation_time: 20_000
+
+Although it is possible to avoid setting a specific readout amplitude, it is generally useful to set a high value here. Indeed, we are not looking for the optimal amplitude but want to have a signal with enough power so that it is clear when it starts.
+
+Acquisition
+^^^^^^^^^^^
+
+The acquisition procedure is described in :func:`qibocal.protocols.signal_experiments.time_of_flight_readout._acquisition`. It is important to note that this experiment makes use of the RAW acquisition mode, which may require some specific care depending on the instrument employed.
+
```
Edoardo-Pedicillo, line 31: Hi @rodolfocarobene the documentation is good. While reading it, I was wondering some questions about this routine ( for example why @Juan has added random pulses). 
Can you explain a little better the meaning of this sentence?
 With ZI or QM (http://login.qrccluster.com:9000/csF2J5KXQ4Oh1KczbMKUGg==/) the result of this routine is different from the picture, is it related to the quantum device or the instrument acquisition? In yours I don't see the readout oscillations.

Also the fit it's quite strange I will expect the tof to be here, what am I missing?
image

Btw, the doc is fine I am just using this occasion to stress our routines.

doc/source/protocols/signal/calibrate_kernel.rst
```diff
@@ -0,0 +1,31 @@
...
+
+Fit
+^^^
+
+The fit procedure (:func:`qibocal.protocols.signal_experiments.calibrate_state_discrimination._fit`) uses the numpy conjugate function, as it takes the conjugate of the differences of the two RAW traces acquired (one for the ground state and one for the excited state) as the kernel, after normalization.
+
+The kernel is also updated in the platform runcard after successful fitting.
```
Edoardo-Pedicillo, line 31: With @andrea-pasquale we agree to add a small section at the end, where we link the other protocols docs that are required to run this one (not all of them just the nearest ones in terms of execution). I proposed a title like "Requirements", but if you have better ideas please share them with us.

doc/source/protocols/signal/time_of_flight.rst
rodolfocarobene: I fixed it, I wrote it without thinking too much about it... but of course that wasn't making much sense

doc/source/protocols/signal/time_of_flight.rst
rodolfocarobene: I added a plot for the 2D case (still with demodulation that I think provides a better plot)

doc/source/protocols/signal/time_of_flight.rst
rodolfocarobene: I am not sure, but I think that the starting point of the signal, that is what the experiment should find, is indeed at 200... I think that this point would not change even if we employed more optimized pulses, so I think it's fine.

I've added a disclaimer regarding the fit procedure, since it is true that this plot could change a lot

doc/source/protocols/signal/time_of_flight.rst
Jacfomg: Should we also include demodulation and not demodulation plots ?

doc/source/protocols/signal/time_of_flight.rst
Jacfomg: Yeah, I agree the point would not change, but is that part of the signal considered effectively as time of flight for cavities since you would not want to acquire that part of the signal ?

doc/source/protocols/signal/calibrate_kernel.rst
rodolfocarobene: What should I add in this case? Like, we need to have fully calibrated pi-pulses so resonator-punchout, resonator, qubit, (fluxes?), rabi? 
I understand the idea, but I think that what is "required" may be a bit arbitrary in some cases.

Maybe we could do something similar by ordering in a sensible manner the routines (?)

doc/source/protocols/signal/time_of_flight.rst
rodolfocarobene: Regarding the fit, I think the problem was just the use of a window size too small

rodolfocarobene:
Thanks for the reviews!

doc/source/protocols/signal/time_of_flight.rst
rodolfocarobene: We could, but I think we will kinda never stop to have different plots... If you have a nice example we can add it

doc/source/protocols/signal/time_of_flight.rst
rodolfocarobene: I think that the "growing part" of the signal should be considered useful, while the "dacaying" part not (with a CLEAR pulse you could almost completly remove it). It is true that is less informative than the rest, but has still some differences right?

immagine

It should be the first part of the trajectory

Jacfomg: I would take the plot right to this one, it seems both might be slighlty relevant when if you forget about the very start and ending of the signal ? One thing for sure from what I read in several paper is that the acquisition time is longer than the readout pulse, so for sure some parts of this decay (up or down) are relevant.

image

Jacfomg: https://arxiv.org/pdf/2112.03708 in appendix F for example

doc/source/protocols/signal/time_of_flight.rst
Jacfomg: Yeah @Edoardo-Pedicillo has one for the 2D demodulation case.

doc/source/protocols/signal/calibrate_kernel.rst
Edoardo-Pedicillo: > What should I add in this case? Like, we need to have fully calibrated pi-pulses so resonator-punchout, resonator, qubit, (fluxes?), rabi?
> I understand the idea, but I think that what is "required" may be a bit arbitrary in some cases.

I would say just the bare minimum requirements to see something, in this case, since you need a pi-pulse, just a Rabi, I know the choice is not unique, but I think this could help some users to troubleshooting.

> Maybe we could do something similar by ordering in a sensible manner the routines (?)

This could be an option, but it will not consider the case of routines with the same requirements.

doc/source/protocols/signal/calibrate_kernel.rst
stavros11: For QM, I believe it is possible to acquire for more time using smearing. I have not tried, but quoting from the docs:
* https://docs.quantum-machines.co/1.1.7/qm-qua-sdk/docs/Introduction/qua_overview/?h=smearing#measure-statement
"The sampling time window will be of a duration that is the duration of the pulse plus twice the smearing (given in ns)."
* https://docs.quantum-machines.co/1.1.7/qm-qua-sdk/docs/Guides/demod/?h=measureme#timing-of-the-measurement-operation
"The length of the raw ADC data recording window is set by the length of the readout pulse as well as the smearing parameter."

From that it seems that acquisiton can only be longer than the readout pulse. I guess that if you want to acquire shorter than the pulse you can just slice the acquired waveform when using RAW, or integrate a smaller part, by passing less or zero integration weights when using INTEGRATION.

For Zurich, @hay-k was telling me that acquisition can be a seperate command from playing the readout pulse, so I am guessing it is more flexible (but I am not sure). I haven't been able to find something similar for QM. `measure` plays and acquires, `play` only plays (without acquiring), but I cannot find any command that only acquires.

doc/source/protocols/signal/calibrate_kernel.rst
Jacfomg: Think about the qibocal users, they might be a bit lost without a simple guide.

rodolfocarobene:
I think now the PR itself is fairly good. In the requirements section it is missing the link to the rabi documentation (since this it does not exist for the moment).

 How do we want to proceed? Do we want to merge to main all these PR separately or to merge them all to the add_doc branch and all at once to main?

doc/source/protocols/signal/time_of_flight.rst
```diff
@@ -0,0 +1,65 @@
...
+In this section, we present the time-of-flight experiment for qibocal (see Fig.12 :cite:p:`gao2021practical`).
+
+In the `time of flight` experiment, we measure the time it takes for a readout pulse to travel to the qubit and back, being acquired by the control instrument ADCs.
+
+Carefully calibrating this delay time is important to optimize the readout. In particular, it is useful to acquire just for the duration of the readout pulse, where differences between the two states really appear (both in amplitude and phase).
+
+The parameters for the experiment are :class:`qibocal.protocols.signal_experiments.time_of_flight_readout.TimeOfFlightReadoutParameters`, namely the amplitude of the readout pulse (if it is not set, the runcard one is used) and an integer that is used as the window size for a moving average.
+
+How to Execute the experiment
```
andrea-pasquale, line 12:
```suggestion
How to execute the experiment
```

doc/source/protocols/signal/time_of_flight.rst
```diff
@@ -0,0 +1,65 @@
...
+As we can see the plot is very different, but the starting point of the signal is still clearly visible.
+
+.. image:: time_of_flight_2d_nodem.png
+
+Fit
+^^^
+
+The fit procedure (:func:`qibocal.protocols.signal_experiments.time_of_flight_readout._fit`) employs a moving average, returning the time when it is maximum, namely when the signal starts being acquired.
+
+Since these plots could vary a lot, depending on the controller instrument specifics and on the design of the QPU under analysis, we suggest to verify each time the correctness of the fit procedure.
```
andrea-pasquale, line 52: The last sentence doesn't make much sense to me because it could be applied to all protocols. Can we do something better than a moving average?

doc/source/protocols/signal/time_of_flight.rst
```diff
@@ -0,0 +1,65 @@
...
+Acquisition
+^^^^^^^^^^^
+
+The acquisition procedure is described in :func:`qibocal.protocols.signal_experiments.time_of_flight_readout._acquisition`. It is important to note that this experiment makes use of the RAW acquisition mode, which may require some specific care depending on the instrument employed (for some devices demodulation could be used, or this mode could be available for just a single qubit at a time).
+
+For a 3D cavity we expect a plot (considering demodulation) as the following:
+
+.. image:: time_of_flight.png
+
+In this image, for example, we can see that the fit procedure did not find the correct pulse starting point (that is around 200). This could be improved increasing the window size.
```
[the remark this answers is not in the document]
rodolfocarobene: I am not sure of what the situation at TII is (if it is possible to produce a better plot), but at the moment this is the only plot I have that clearly shows both exponential parts

doc/source/protocols/signal/time_of_flight.rst
```diff
@@ -0,0 +1,65 @@
...
+
+For a 2D resonator, some more oscillation could appear:
```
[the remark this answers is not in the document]
rodolfocarobene: I personally couldn't find a nice reference on readout

doc/source/protocols/signal/time_of_flight.rst
rodolfocarobene: I removed it, since it was indeed probably useless. I think the moving average is fine

doc/source/protocols/signal/time_of_flight.rst
```diff
@@ -0,0 +1,65 @@
...
+        window_size: 10
+        nshots: 1024
+        relaxation_time: 20_000
+
+Although it is possible to avoid setting a specific readout amplitude, it is generally useful to set a high value here. Indeed, we are not looking for the optimal amplitude but want to have a signal with enough power so that it is clear when it starts.
+
+Acquisition
+^^^^^^^^^^^
+
+The acquisition procedure is described in :func:`qibocal.protocols.signal_experiments.time_of_flight_readout._acquisition`. It is important to note that this experiment makes use of the RAW acquisition mode, which may require some specific care depending on the instrument employed (for some devices demodulation could be used, or this mode could be available for just a single qubit at a time).
```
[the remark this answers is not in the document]
rodolfocarobene: I've added the explicit link to the qibo.science documentation and a doc-test to verify that the acquisition mode is still there... Not sure if this is the best way of doing this, but it's what I could think of

doc/source/protocols/signal/time_of_flight.rst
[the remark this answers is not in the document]
andrea-pasquale: We could try to run something on the platinum chip @Edoardo-Pedicillo @Jacfomg @stavros11?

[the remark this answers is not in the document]
andrea-pasquale: Window size needs to be reduced?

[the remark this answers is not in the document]
Edoardo-Pedicillo: I think the fit of the tof is not made for this particular case, because when the instrument is acquiring pure noise the offset is the same of the signal, maybe @Jacfomg picture is better.

Jacfomg (approved):
Thanks @rodolfocarobene !

Edoardo-Pedicillo (approved):
Thanks @rodolfocarobene !

rodolfocarobene:
Thanks for the reviews! Should I merge this, wait to merge `add_doc`, try to rebase to main (I have no real idea if this is reasonable)?

andrea-pasquale:
Feel free to merge in `add_doc`, thanks!

PR #884, opened 2024-06-17
Fix Rabi length signal fit

### LINKED ISSUE #879: Rabi length signal fit no longer works
While debugging the emulator I noticed that the fit for rabi length doesn't work anymore.
The bug was added by #857, the initial guess now for all rabi experiments is

https://github.com/qiboteam/qibocal/blob/a67eeaddf0d98ee35633d6e0be2232d05a0e382f/src/qibocal/protocols/rabi/utils.py#L294-L295

while before when running rabi length signal it was 
https://github.com/qiboteam/qibocal/blob/37d060599c739b3b70602505d29509af295aedb0/src/qibocal/protocols/rabi/length_signal.py#L171-L172

I believe that the -1 on the second term is triggering the error. Probably we will need to investigate more, for the being it would be enough if we are able to revert some of changes to the previous version of the code.
Here are the reports for [main](http://login.qrccluster.com:9000/YTs6fM2yTK-5lnx_A1tkxQ==/) and [before #857](http://login.qrccluster.com:9000/FL1XMvYrSS-8SVtjQkEUXQ==/) 
Could you take care of it @rodolfocarobene?

rodolfocarobene:
Closes #879.

I guess that the "problem" arises from the fact that in rabi length we expect to see oscillations given by detuning (potential at least) spanning along a large duration. While for rabi amplitude the length is usually kept small -> we do not fit oscillation and we don't have the same problem.

I'm not sure that I like the asymmetry, but it doesn't probably matter

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/rabi/utils.py
+++ b/src/qibocal/protocols/rabi/utils.py
@@ -284,6 +284,7 @@ def guess_frequency(x, y):
 def fit_length_function(
     x, y, guess, sigma=None, signal=True, x_limits=(None, None), y_limits=(None, None)
 ):
+    inf_bounds = [0, -1, 0, -np.pi, 0] if signal else [0, 0, 0, -np.pi, 0]
     popt, perr = curve_fit(
         rabi_length_function,
         x,
@@ -291,7 +292,7 @@ def fit_length_function(
         p0=guess,
         maxfev=100000,
         bounds=(
-            [0, 0, 0, -np.pi, 0],
+            inf_bounds,
             [1, 1, np.inf, np.pi, np.inf],
         ),
         sigma=sigma,


src/qibocal/protocols/rabi/utils.py
```diff
@@ -284,14 +284,15 @@ def guess_frequency(x, y):
 def fit_length_function(
     x, y, guess, sigma=None, signal=True, x_limits=(None, None), y_limits=(None, None)
 ):
+    inf_bounds = [0, -1, 0, -np.pi, 0] if signal else [0, 0, 0, -np.pi, 0]
```
Edoardo-Pedicillo, line 287: We could normalize between 0 and 1 in the signal case, since this PR is an hot fix, I would merge it as it is.

Edoardo-Pedicillo (approved):

andrea-pasquale (approved):
Thanks @rodolfocarobene, it seems to work as expected now.

src/qibocal/protocols/rabi/utils.py
andrea-pasquale: @Edoardo-Pedicillo feel free to open an issue about this.

Issue #890, opened 2024-06-19
Execution parameters in protocols


alecandido:
As pointed out in #887, sometimes the routines might want to specify the `ExecutionParameters` by themselves.

However, direct usage of `params.execution_parameters` would be advantageous for all the routines, since it would leave the freedom to the caller to always override the parameters.
Still, when parameters are not specified, some default values should be clearly routine-dependent (in particular, `acquisition_type` and `averaging_mode`).

Thus, we should enforce the internal usage of `params.execution_parameters` everywhere, but leave some space for the routine implementor to specify a default value, to be carefully used.
If there is no special default value to recommend (like it will be always, or almost always, the case for `nshots` and `relaxation_time`), it should be left empty in the routine, and just use the one determined globally.

PR #892, opened 2024-06-19
unify execution pars for dispersive shift

### LINKED ISSUE #887: Dispersive shift not working as expected
There is an asymmetry between the `ExecutionParameters` of the sequence to generate $\ket{0}$ and the one to generate $\ket{1}$. We spotted this issue with @rodolfocarobene yesterday.

https://github.com/qiboteam/qibocal/blob/3765c14f59774162543d24822ccecc20d9dc6501/src/qibocal/protocols/dispersive_shift.py#L128-L143

The error was most likely introduced by #832.

@rodolfocarobene I would like to have the fix before friday for the (possible) release.
I can take care of it, if someone from your team would like to give it a try let me know.

rodolfocarobene:
Closes #887, but does not take care at all of #890.

How could we use `params.execution_parameters` and also pass standard values?

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/dispersive_shift.py
+++ b/src/qibocal/protocols/dispersive_shift.py
@@ -125,20 +125,21 @@ def _acquisition(
         type=SweeperType.OFFSET,
     )
 
+    execution_pars = ExecutionParameters(
+        nshots=params.nshots,
+        relaxation_time=params.relaxation_time,
+        acquisition_type=AcquisitionType.INTEGRATION,
+        averaging_mode=AveragingMode.CYCLIC,
+    )
     results_0 = platform.sweep(
         sequence_0,
-        ExecutionParameters(
-            nshots=params.nshots,
-            relaxation_time=params.relaxation_time,
-            acquisition_type=AcquisitionType.INTEGRATION,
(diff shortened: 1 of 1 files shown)


alecandido:
> Closes #887, but does not take care at all of #890.

That's fine, one issue at a time. #890 is for the longer term

> How could we use `params.execution_parameters` and also pass standard values?

We extend the routine object to contain an `ExecutionParameters` object, possibly with some fields set to `None` (that might require a mirror object, to possibly fix the attribute types with `Optional`), and the executor will inspect the routine attribute when determining which are the `params.execution_parameters`, setting them according to the following priority:
- explicit routine parameters
- explicit global parameters
- routines default
- global default

and that's what the routine should get access to in the `params.execution_parameters`. But it has to be implemented at the level of the execution, not the individual routine.

andrea-pasquale:
> How could we use `params.execution_parameters` and also pass standard values?

As @alecandido suggested, we can take care of `ExecutionParameters` in a separate PR.

Edoardo-Pedicillo (approved):

Issue #564, opened 2023-10-17
Criterion for TWPA optimization


Edoardo-Pedicillo:
When a TWPA optimization routine is launched, the twpa  frequency (power) is updated with the best one of the last qubit in the qubit list of the runcard. We should find a better method to evaluate the best values.

PR #899, opened 2024-06-24
Small fixes for TWPA calibration routines

stavros11:
Some issues I realized while trying to use these routines:

1. `frequency_power` was giving some `KeyError`s with `qubit`s missing from the `initial_twpa_freq` and `initial_twpa_power` dictionaries. I believe these are because the execution part of the routine is nested in the loop that initializates these dictionaries, while I think it should be a separate loop.
2. It seems that `relaxation_time` is ignored for both SNR routines as it is not passed in the `resonator_spectroscopy` call. This ends up using the default `relaxation_time` from the platform runcard, which is usually much higher than what we need for spectroscopy, thus making execution slower.

Checklist:
- [x] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/readout_optimization/twpa_calibration/frequency_power.py
+++ b/src/qibocal/protocols/readout_optimization/twpa_calibration/frequency_power.py
@@ -100,46 +100,46 @@ def _acquisition(
         ].twpa.local_oscillator.frequency
         initial_twpa_power[qubit] = platform.qubits[qubit].twpa.local_oscillator.power
 
-    for freq in freq_range:
-        platform.qubits[qubit].twpa.local_oscillator.frequency = (
-            initial_twpa_freq[qubit] + freq
-        )
+        for freq in freq_range:
+            platform.qubits[qubit].twpa.local_oscillator.frequency = (
+                initial_twpa_freq[qubit] + freq
+            )
 
-        for power in power_range:
-            for qubit in targets:
-                platform.qubits[qubit].twpa.local_oscillator.power = (
-                    initial_twpa_power[qubit] + power
+            for power in power_range:
+                for qubit in targets:
+                    platform.qubits[qubit].twpa.local_oscillator.power = (
+                        initial_twpa_power[qubit] + power
+                    )
+
+                classification_data = classification._acquisition(
+                    classification.SingleShotClassificationParameters.load(
+                        {"nshots": params.nshots}
+                    ),
+                    platform,
+                    targets,
                 )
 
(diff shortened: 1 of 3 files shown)


andrea-pasquale (approved):
Thanks @stavros11, I agree with all the fixes.
Regarding the relaxation I already had a feeling that it was not used correctly in all protocols (#829).

src/qibocal/protocols/readout_optimization/twpa_calibration/frequency_power.py
```diff
@@ -100,46 +100,46 @@ def _acquisition(
         ].twpa.local_oscillator.frequency
         initial_twpa_power[qubit] = platform.qubits[qubit].twpa.local_oscillator.power
 
-        for freq in freq_range:
-            platform.qubits[qubit].twpa.local_oscillator.frequency = (
-                initial_twpa_freq[qubit] + freq
+    for freq in freq_range:
+        platform.qubits[qubit].twpa.local_oscillator.frequency = (
+            initial_twpa_freq[qubit] + freq
+        )
```
Edoardo-Pedicillo, line 106: Here `qubit` is not assigned

src/qibocal/protocols/readout_optimization/twpa_calibration/frequency_power.py
```diff
@@ -100,46 +100,46 @@ def _acquisition(
...
+
+        for power in power_range:
+            for qubit in targets:
```
Edoardo-Pedicillo, line 109:
```suggestion
```
At this point, I suggest removing this line and leaving the indentation as in main.

stavros11: Thanks, both comments are right. I reverted the indentation and removed this in 79cd1cb47ed516abfcabcedfe4051ab31ecf3377.

The routine should now work, but we should keep in mind that it is suboptimal. The acquisition is repeated in multiplexed fashion for all qubits multiple times and in each case we drop the results for all qubits except of one. We should only need one acquisition per TWPA pump, not per qubit. For 5-qubits that share a readout line, which is the case for most of our current chips, we should only need one acquisition, not five.

Edoardo-Pedicillo: We have already opened #564 , for sure what we are doing is wrong, but we could collect all the results and find a better way

Edoardo-Pedicillo (approved):

PR #905, opened 2024-06-27
Adding `qq update` command

### LINKED ISSUE #833: Add command to perform platform update
It would be interesting to have a command in qibocal to copy directly the update platform in the directory where the user contains the platforms.
Something like:
```sh
qq update <qibocal_output_folder> <platform_directory>
```
Originally suggested by @sorewachigauyo

andrea-pasquale:
Addition of a `qq update` program to generate the updated platform from CLI after the qibocal execution. Closes #833.

Checklist:
- [x] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/auto/execute.py
+++ b/src/qibocal/auto/execute.py
@@ -7,6 +7,7 @@ from typing import Union
 
 from qibolab import create_platform
 from qibolab.platform import Platform
+from qibolab.serialize import dump_platform
 
 from qibocal import protocols
 from qibocal.config import log, raise_error
@@ -16,6 +17,9 @@ from .mode import ExecutionMode
 from .operation import Routine
 from .runcard import Action, Runcard, Targets
 from .task import Completed, Task
+
+PLATFORM_DIR = "platform"
+"""Folder where platform will be dumped."""
 
 
 @dataclass
@@ -83,6 +87,9 @@ class Executor:
         )
         if ExecutionMode.FIT in mode and self.platform is not None and update:
             completed.update_platform(platform=self.platform, update=self.update)
+
+        (completed.datapath / PLATFORM_DIR).mkdir(parents=True, exist_ok=True)
+        dump_platform(self.platform, completed.datapath / PLATFORM_DIR)
 
         self.history.push(completed)
         completed.dump(self.output)
--- a/src/qibocal/auto/execute.py
+++ b/src/qibocal/auto/execute.py
@@ -88,8 +88,9 @@ class Executor:
         if ExecutionMode.FIT in mode and self.platform is not None and update:
             completed.update_platform(platform=self.platform, update=self.update)
 
-        (completed.datapath / PLATFORM_DIR).mkdir(parents=True, exist_ok=True)
-        dump_platform(self.platform, completed.datapath / PLATFORM_DIR)
+        if self.platform is not None:
+            (completed.datapath / PLATFORM_DIR).mkdir(parents=True, exist_ok=True)
+            dump_platform(self.platform, completed.datapath / PLATFORM_DIR)
 
         self.history.push(completed)
         completed.dump(self.output)
--- a/src/qibocal/auto/execute.py
+++ b/src/qibocal/auto/execute.py
@@ -56,6 +56,7 @@ class Executor:
         protocol: Routine,
         parameters: Union[dict, Action],
         mode: ExecutionMode = ExecutionMode.ACQUIRE | ExecutionMode.FIT,
+        update: bool = True,
     ) -> Completed:
         """Run single protocol in ExecutionMode mode."""
         if isinstance(parameters, dict):
@@ -80,6 +81,8 @@ class Executor:
             folder=self.output,
             mode=mode,
         )
+        if ExecutionMode.FIT in mode and self.platform is not None and update:
+            completed.update_platform(platform=self.platform, update=self.update)
 
         self.history.push(completed)
         completed.dump(self.output)
@@ -87,7 +90,7 @@ class Executor:
         return completed
 
 
-def run(runcard: Runcard, output: Path, mode: ExecutionMode):
+def run(runcard: Runcard, output: Path, mode: ExecutionMode, update: bool = True):
     """Run runcard and dump to output."""
     platform = runcard.platform_obj
     targets = runcard.targets if runcard.targets is not None else list(platform.qubits)
@@ -104,5 +107,6 @@ def run(runcard: Runcard, output: Path, mode: ExecutionMode):
             protocol=getattr(protocols, action.operation),
             parameters=action,
             mode=mode,
+            update=update,
         )
     return instance.history
--- a/src/qibocal/auto/execute.py
+++ b/src/qibocal/auto/execute.py
@@ -81,8 +81,8 @@ class Executor:
             mode=mode,
         )
 
-        if ExecutionMode.FIT in mode and self.platform is not None:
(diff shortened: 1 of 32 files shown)


src/qibocal/cli/update.py
```diff
@@ -0,0 +1,67 @@
...
+    meta = json.loads((path / META).read_text())
+    # load runcard
+    runcard = Runcard.load(yaml.safe_load((path / RUNCARD).read_text()))
+    set_backend(backend=runcard.backend, platform=runcard.platform)
+    # run
+    history = run(
+        output=path,
+        runcard=runcard,
+        mode=ExecutionMode.UPDATE,
+    )
```
alecandido, line 54: We should already have the most updated runcard in the output folder, when you invoke `qq update`.

Maybe it is in a subfolder, corresponding to the latest routine (which is perfectly fine), but I'd just maintain a link to the latest one in the top-level folder, and copy that one to the original platform.

alecandido: In this way, you won't even need to track the timing. It's just a copy, it will take almost nothing.

src/qibocal/update.py
```diff
@@ -42,25 +42,29 @@ def readout_attenuation(att: int, platform: Platform, qubit: QubitId):
     platform.qubits[qubit].readout.attenuation = int(att)
 
 
-def drive_frequency(freq: Union[float, tuple], platform: Platform, qubit: QubitId):
+def drive_frequency(
+    freq: Union[float, tuple, list], platform: Platform, qubit: QubitId
+):
     """Update drive frequency value in platform for specific qubit."""
-    if isinstance(freq, tuple):
+    if isinstance(freq, (tuple, list)):
```
alecandido, line 49: What about just using one of the two?

andrea-pasquale: This is just an ugly patch. I will try to improve it. The error was triggered when I tested with `qq update`, I think that there are some attributes in `Results` which are tuple since they contain also the error and when they are deserialized they are converted into lists.

andrea-pasquale: Implemented in https://github.com/qiboteam/qibocal/pull/905/commits/cf006ca0a4553e57a3296609aa80fb177f081dc1

src/qibocal/cli/update.py
andrea-pasquale: > We should already have the most updated runcard in the output folder, when you invoke `qq update`.
> 
> Maybe it is in a subfolder, corresponding to the latest routine (which is perfectly fine), but I'd just maintain a link to the latest one in the top-level folder, and copy that one to the original platform.

Actually this might not be the case. This is true only when running with `AUTOCALIBRATION`, but in this execution mode all updates will be performed in any case (local one after each protocol and the global one at the end).
I know that we still need to decide how to handle local updates. If by default we decide to keep local updates what you are suggesting makes sense. What do you think @alecandido?

src/qibocal/update.py
alecandido: I've been through cf006ca, and I marked a few tuples I've found, but there are many more.

Maybe I missed where the tuples had to be dropped from. Sorry...

alecandido:
@andrea-pasquale I'm sorry for the tuple business. As soon as I will finish #909 I have other few simpler improvements in mind, including a consistent migration to Pydantic. Whenever that will be completed, tuples will be available again, since the serialization will come out of the box.

Unfortunately, I will have to balance those with #869 (but if I will be good enough with #909, a first version of that should be pretty simple to achieve).

andrea-pasquale:
> @andrea-pasquale I'm sorry for the tuple business. As soon as I will finish #909 I have other few simpler improvements in mind, including a consistent migration to Pydantic. Whenever that will be completed, tuples will be available again, since the serialization will come out of the box.
> 
> Unfortunately, I will have to balance those with #869 (but if I will be good enough with #909, a first version of that should be pretty simple to achieve).

No problem @alecandido.
Now I believe that it should be fixed. I will try to take care of tests now.
In which order do you recommend that we merge this alongside #909 and #869?

alecandido:
> In which order do you recommend that we merge this alongside #909 and #869?

Do not bother about #909 and #869: whenever you're ready, we review and merge. If it comes before the others, I will rebase myself ( ╥ ᴗ ╥)

andrea-pasquale:
Now it should be ready to review. There are a lot of file changes just for the tuple issue

src/qibocal/cli/autocalibration.py
```diff
@@ -63,7 +63,10 @@ def autocalibrate(runcard, folder, force, update):
 
     report(path, history)
 
+    # checking if at least on the task had local update
+    local_update = any(completed.task.update for completed in list(history.values()))
+
     # dump updated runcard
-    if platform is not None:
+    if platform is not None and (update or local_update):
```
alecandido, line 70: This I already changed in #909, but I would suggest the former behavior: just dump it, always (if it's the same, you could always find with a simple `diff output/platform.yml output/new_platform.yml).

src/qibocal/cli/fit.py
```diff
@@ -60,8 +60,13 @@ def fit(input_path, update, output_path, force):
...
     meta["end-time"] = e.strftime("%H:%M:%S")
 
+    # checking if at least on the task had local update
+    local_update = any(completed.task.update for completed in list(history.values()))
+
     # dump updated runcard
-    if runcard.platform_obj is not None and update:  # pragma: no cover
+    if runcard.platform_obj is not None and (
+        update or local_update
+    ):  # pragma: no cover
```
alecandido, line 69: Same of above

alecandido (commented):
I have just a comment about `local_update`, but other than that it should be fine.

I'm also glad that the superposition with #909 is very small, it should be pretty simple to rebase :D

alecandido (approved):

src/qibocal/protocols/coherence/t1.py
```diff
@@ -28,9 +28,7 @@ class T1Parameters(t1_signal.T1SignalParameters):
 class T1Results(t1_signal.T1SignalResults):
     """T1 outputs."""
 
-    chi2: Optional[dict[QubitId, tuple[float, Optional[float]]]] = field(
-        default_factory=dict
-    )
+    chi2: Optional[dict[QubitId, list[float, float]]] = field(default_factory=dict)
```
Edoardo-Pedicillo, line 31: I have not found any example of `list[float, float]`, could you explaining the what does it means ? I think `list[float]` is enough

andrea-pasquale: Yes, indeed it should be `list[float]`, thanks
```suggestion
    chi2: Optional[dict[QubitId, list[float]]] = field(default_factory=dict)
```

src/qibocal/auto/execute.py
```diff
@@ -80,17 +85,20 @@ def run_protocol(
             folder=self.output,
             mode=mode,
         )
-
-        if ExecutionMode.FIT in mode and self.platform is not None:
+        if ExecutionMode.FIT in mode and self.platform is not None and update:
             completed.update_platform(platform=self.platform, update=self.update)
 
+        if self.platform is not None:
```
Edoardo-Pedicillo, line 91: Maybe you can move this line before `if ExecutionMode.FIT in mode and self.platform is not None and update:`

There is a small inconsistency with the typing, according to the Executor definition a platform is always defined even in the `create` method, for the moment could you change it ?

Edoardo-Pedicillo (approved):

Issue #908, opened 2024-06-27
Drop `Action`


alecandido:
While it makes sense to keep the distinction between a `Task` to be executed, and a `Completed` one, there is essentially no motivation to keep the `Task` specification separate.

So, I will follow up on https://github.com/qiboteam/qibocal/pull/801, and merge `Action` into the `Task` object, that already contains only that and the resolved `Routine`.

alecandido:
The original idea was correct, but the object that is not useful is the `Task` not `Action`, since we need to keep recording and validating the input parameters.

The only thing we need from `Task` is `Task.run`, which could be just:
```py
def run(
    routine: Routine,
    parameters: Parameters,
    platform: Platform,
    targets: Targets,
    mode: ExecutionMode,
):
    ...
```

PR #909, opened 2024-06-27
Move output folder management

alecandido:
The original goal was to make `History` the manager of the whole output folder.

It is almost already happening, but there are a few main exceptions:
- the metadata in `meta.json`
- the original runcard dump in `runcard.yml` (which will be optional, since the execution could be initiated by an arbitrary Python call)
- the old and new platforms, respectively in `platform/` and `new_platform/`
- the report in `index.html`

To avoid overcomplicating the `History` object, I will create a new `Output` one, and make such that everything could be retrieved just with a handle on an instance of this object.

The main motivation is to expose in a more structured way some operations that are now done internally by the CLI functions, e.g.
https://github.com/qiboteam/qibocal/blob/4b8d407b11ebce72544821b361ea22df2f61c73a/src/qibocal/cli/fit.py#L58-L61
in such a way that it will be simpler to expose them for scripts, as those in #869


alecandido:
@andre-pasquale I should have made enough mess, now I will try to put the pieces back together.

However, I (more or less) reached the goal I had in mind, i.e. simplify the `autocalibration` command implementation, by delegating most of the operations to the new `Output` and friends.
https://github.com/qiboteam/qibocal/blob/c7b3da326c6a86c49947c4a43ac3ba2acb90b46f/src/qibocal/cli/autocalibration.py#L13-L52

The main reason is that for #869 most of the operations will be done by the executor.
We can avoid dumping the runcard, since there won't be a runcard, and some operations might be left to the user.
But I wanted to avoid leaving *too many* operations to the user, because it would just become boilerplate, to copy-paste in each and every script...

alecandido:
By now, few operations are left:
- creating the platform and connecting: this will not be specified in the runcard, but the platform can be specified in the same way in the `Executor`, so I'm planning to move it there
- there are other two symmetric operations to be done before and after running
    - dump the folder (metadata, tasks known till that point, platform serialization)
    - tap the timer
  
  these I will delegate to the `Executor` connection and disconnection, that will perform these operations and handle the platform connection, respectively upon creation, and with an explicit call to `Executor.close()`
- the `report()` instead I will keep separate: it is a single function, so, if desired, it could be called at the end of the descript (otherwise just run `qq report ...` from the CLI later on)

alecandido:
The former means that the `Executor` will be in charge of everything, including dumping.

I wanted to isolate the disk operations from the execution, with two separate domains. But the trade-off is to handle two different objects, the `Executor` and the `Output`, with an explicit information flow from one to the other.

I'm not a huge fun of having everything happening under a single object, i.e. the `Executor`, so later on I might put some more effort to make it more modular.
However, for the time being that's a simple enough solution, and it will simplify the UX.

src/qibocal/cli/autocalibration.py
```diff
@@ -1,69 +1,52 @@
...
-    set_backend(backend=runcard.backend, platform=runcard.platform)
     # rename for brevity
-    backend = runcard.backend_obj
-    platform = runcard.platform_obj
-    # generate output folder
-    path = generate_output_folder(folder, force)
+    backend = construct_backend(backend=runcard.backend, platform=runcard.platform)
+    platform = backend.platform
+    if platform is None:
+        raise ValueError("Qibocal requires a Qibolab platform to run.")
```
alecandido, line 24: Here, I decided that there will be always a non-`None` platform in Qibocal.

In general, it makes no sense at the moment to execute a protocol on the `NumpyBackend`, even if it's a circuit.

However, it might make sense to execute with some kind of *noise model*.
We could consider creating something like a `qibocal.Platform`, that could be either a Qibolab one, or encoding a noise model on top of a plain circuit simulator. In the limit case, just pure circuit execution, with no noise.

The current Qibocal is tuned to work with a Qibolab's platform, including the parameters' update and dump.
But if that makes sense, we could open an issue to wrap the `qibolab.Platform` into a `qibocal.Platform` object. It could be even a `Union`, but it should respect a common interface for the two cases.

src/qibocal/auto/task.py
```diff
@@ -137,61 +139,59 @@ def __post_init__(self):
...
-        if self.task.update and update:
-            for qubit in self.task.targets:
-                try:
-                    self.task.operation.update(self.results, platform, qubit)
-                except KeyError:
-                    log.warning(
-                        f"Skipping update of qubit {qubit} due to error in fit."
-                    )
-            (self.datapath / PLATFORM_DIR).mkdir(parents=True, exist_ok=True)
-            dump_platform(platform, self.datapath / PLATFORM_DIR)
```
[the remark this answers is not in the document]
andrea-pasquale: Thanks @alecandido, in general I agree with having disk operation separated.
Maybe to ease the transition until we have #910 I would suggest to dump the platform in `run_protocol` as you were suggesting.

src/qibocal/cli/autocalibration.py
andrea-pasquale: I mean I know that qibocal is now coded to work primarily with a `qibolab.Platform`, however, as you said, running circuits with noise models if for sure useful, also for debugging. It is true that now with the emulator having noisy circuits might be redundant given we could encode the noise using the emulator.
There were some discussions in the past about the possibility to expose `backend` alongside `platform` during the acquisition.
I know that it is not a priority, but let's just say that I would prefer not to drop the feature. We could discuss about alternatives solution, if they are easy to implement.

andrea-pasquale (commented):
Thanks for the refactor @alecandido, I had a quick look at generally the code looks more elegant now. 
I agree with all the minor changes, regarding some of the breaking changes that you highlighted I replied directly in the comments. Let me know when this is ready to review and I will have another look.

alecandido:
> I agree with all the minor changes, regarding some of the breaking changes that you highlighted I replied directly in the comments. Let me know when this is ready to review and I will have another look.

Unfortunately, I would have liked to postpone some changes, to immediately review & merge a smaller unit. That's why I split this before #869, despite that being the goal.
I opened some issues for that purpose, but I'm not sure how many things I will be able to keep separate (e.g. I ended up implementing #911 here, at least preliminarily).

src/qibocal/cli/autocalibration.py
alecandido: Yes, the alternative I had in mind was to do something like:
```py
class CircuitPlatform:
    ...

Platform = Union[qibolab.Platform, CircuitPlatform]
```
and use `Platform` in place of `qibolab.Platform`.

Simply, I was proposing to postpone this to the moment when we will want to actively make use of circuit simulation. There will be other things to be adjusted as well, so this should be pretty simple.

I'm ok to keep discussing the best trade-off. I'd just prefer to have *another kind of `Platform`*, instead of having `None`, because it becomes less special, and you'd need less dedicated handling (that will be localized in the new platform class, instead).

src/qibocal/auto/task.py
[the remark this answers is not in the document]
alecandido: That's definitely an option. However, at the moment most of those platforms are never used, since you'd need to copy the parameters manually in `$QIBOLAB_PLATFORMS/my-platform` to have it used even during `qq fit` (I was planning to reload the platform in the folder, but to do that I need to be sure that is a Qibolab platform...).

If you know that you need this intermediate savings for something specific, we could add the dump in `run_protocol`, indeed.
But otherwise, we could even skip the save, until we'll need them (and with #910 they will arrive anyhow at some point, though I'm not sure when).

src/qibocal/auto/output.py
```diff
@@ -0,0 +1,175 @@
...
+
+        # dump tasks
+        self.history.flush(path)
+
+        # update platform
+        if self.platform is not None:
+            self.update_platform(self.platform, path)
+
+    @staticmethod
+    def update_platform(platform: Platform, path: Path):
```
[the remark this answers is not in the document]
alecandido: Because it is overwriting `UPDATED_PLATFORM` in-place. That's why I called it *update*.

src/qibocal/auto/output.py
```diff
@@ -0,0 +1,175 @@
...
+    versions: Versions
+    more: Optional[dict] = None
+
+    @classmethod
+    def generate(cls, name: str, backend, platform: str):
+        """Methods that takes care of:
+        - dumping original platform
+        - storing qq runcard
+        - generating meta.yml
+        """
```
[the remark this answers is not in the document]
alecandido: Ah, nono, that's perfectly fine.

I should even rewrite both from scratch, that's just the copy-paste of the old `utils`'s docstring.

src/qibocal/auto/history.py
```diff
@@ -1,54 +1,53 @@
...
+    @staticmethod
+    def route(completed: Completed, folder: Path):
+        """Determine the path related to a completed task.
+
+        `folder` should be ussually the general output folder, used by Qibocal to store
+        all the execution results. Cf. :cls:`qibocal.auto.output.Output`.
+        """
+        return folder / "data" / f"{completed.task.id}"
+
+    def flush(self, output: Optional[Path] = None):
```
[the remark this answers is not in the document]
alecandido: I just missed one step, I truly wanted to release memory, using `Completed.flush()`:
https://github.com/qiboteam/qibocal/blob/86a9890869b33fc22b8dd8178d6f01349fd5c7dc/src/qibocal/auto/task.py#L210-L214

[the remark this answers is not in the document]
alecandido: Updated in f0b4e8c

src/qibocal/auto/output.py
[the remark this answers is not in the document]
alecandido: Done in bf219d0

alecandido:
@andrea-pasquale @Edoardo-Pedicillo I still need to finish providing tests, to improve coverage and test the new structures in isolation.

However, tests are already passing, and since this review might take a while I already switched to "ready for review". This is in order to accelerate the process, since this could be useful, and maybe enough to complete #869 as well.

Nevertheless, since I changed a few things, even in high-level behavior, I would ask you to run something to test that Qibocal is not broken in any major and fully undesirable way (not detected by tests).
Since I'm mostly playing with the output, and I almost didn't touch the routines (other than minimal support for lists), this should not require actual execution, but `"dummy"` should be more than enough.

src/qibocal/auto/runcard.py
```diff
@@ -3,89 +3,67 @@
...
-                yaml.safe_load((runcard / RUNCARD).read_text(encoding="utf-8"))
-            )
+            return cls(yaml.safe_load((runcard / RUNCARD).read_text(encoding="utf-8")))
         return cls(**runcard)
 
     def dump(self, path):
         """Dump runcard object to yaml."""
         (path / RUNCARD).write_text(yaml.safe_dump(asdict(self)))
+
+    def run(
```
andrea-pasquale, line 51: I guess that the reason why you put `run` in `Runcard`, is because at the end we ere always doing something like
```py
from ... import Runcard
from .. import run 
runcard = Runcard.load(...)
history = run(runcard, ...)
```
I am fine with this.

andrea-pasquale (commented):
Thanks @alecandido, generally it looks good to me.
While testing https://github.com/qiboteam/qibolab_platforms_qrc/pull/148
I got the following error which is not present in main.
Most likely it is due to the fact that this particular platform is using strings (which I know you don't like as QubitId :) ). I don't know if we can force them to use `int`, in any case if you could make the refactor works also with strings that would be perfect.

aceback (most recent call last):
  File "/nfs/users/andrea.pasquale/default/bin/qq", line 8, in <module>
    sys.exit(command())
  File "/nfs/users/andrea.pasquale/default/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
  File "/nfs/users/andrea.pasquale/default/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/nfs/users/andrea.pasquale/default/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
  File "/nfs/users/andrea.pasquale/default/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
  File "/nfs/users/andrea.pasquale/default/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/nfs/users/andrea.pasquale/qibocal/src/qibocal/cli/_base.py", line 73, in auto
    autocalibrate(runcard, folder, force, update)
  File "/nfs/users/andrea.pasquale/qibocal/src/qibocal/cli/autocalibration.py", line 41, in autocalibrate
    history = runcard.run(
  File "/nfs/users/andrea.pasquale/qibocal/src/qibocal/auto/runcard.py", line 63, in run
    instance.run_protocol(
  File "/nfs/users/andrea.pasquale/qibocal/src/qibocal/auto/execute.py", line 49, in run_protocol
    completed = task.run(platform=self.platform, targets=self.targets, mode=mode)
  File "/nfs/users/andrea.pasquale/qibocal/src/qibocal/auto/task.py", line 160, in run
    completed.results, completed.results_time = operation.fit(completed.data)
  File "/nfs/users/andrea.pasquale/qibocal/src/qibocal/auto/operation.py", line 43, in wrapper
    out = func(*args, **kwds)
  File "/nfs/users/andrea.pasquale/qibocal/src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py", line 180, in _fit
    qubit_data = data[qubit]
  File "/nfs/users/andrea.pasquale/qibocal/src/qibocal/auto/operation.py", line 136, in __getitem__
    return self.data[qubit]
KeyError: ('D', '1')

src/qibocal/auto/operation.py
```diff
@@ -130,6 +131,8 @@ def __init__(
 
     def __getitem__(self, qubit: Union[QubitId, tuple[QubitId, int]]):
         """Access data attribute member."""
+        if isinstance(qubit, Iterable):
+            qubit = tuple(qubit)
```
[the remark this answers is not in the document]
alecandido: It's written in the commit message 877a240da800797ab84c3b899d9449abcebb052c.

The issue is that we are using qubit pairs as indices in a dictionary, but we can not index with lists. And we can not serialize tuples.

alecandido:
> While testing [qiboteam/qibolab_platforms_qrc#148](https://github.com/qiboteam/qibolab_platforms_qrc/pull/148)
> I got the following error which is not present in main.
> Most likely it is due to the fact that this particular platform is using strings (which I know you don't like as QubitId :) ). I don't know if we can force them to use `int`, in any case if you could make the refactor works also with strings that would be perfect.

You're right I don't like, but I didn't anything specific to break it.

Can you tell me how to reproduce the error?

src/qibocal/auto/operation.py
```diff
@@ -249,9 +253,11 @@ class Results(AbstractData):
     def __contains__(self, key: Union[QubitId, QubitPairId, tuple[QubitId, ...]]):
         """Checking if qubit is in Results.
 
-        If key is not present means that fitting failed or
-        was not performed.
+        If key is not present means that fitting failed or was not
+        performed.
         """
+        if isinstance(key, Iterable):
+            key = tuple(key)
```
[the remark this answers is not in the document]
alecandido: Same of above: it could be a deserialized qubit pair, and the `key in` would fail...

src/qibocal/auto/runcard.py
alecandido: Yes, and also because this will be related only to the "runcard mode", but there will be scripts which won't need the `run()` at all.

That's why I decided it was not an executor feature, but built *on top* of the executor.

andrea-pasquale:
> > While testing [qiboteam/qibolab_platforms_qrc#148](https://github.com/qiboteam/qibolab_platforms_qrc/pull/148)
> > I got the following error which is not present in main.
> > Most likely it is due to the fact that this particular platform is using strings (which I know you don't like as QubitId :) ). I don't know if we can force them to use `int`, in any case if you could make the refactor works also with strings that would be perfect.
> 
> You're right I don't like, but I didn't anything specific to break it.
> 
> Can you tell me how to reproduce the error?

Sure, sorry for not specifying it before.
I am running with this runcard
```yaml
actions:
- id: qubit flux dependence
  operation: qubit_flux
  parameters:
    bias_step: 0.004
    bias_width: 0.2
    drive_amplitude: 0.007
    drive_duration: 1000
    freq_step: 250000
    freq_width: 20000000
    nshots: 1000
    relaxation_time: 10000
  targets: null
  update: true
backend: qibolab
platform: qw11q
targets:
- D1
- D2
- D3
- D4
- D5
update: true
```
On the branch which I pointed out before in qibolab_platforms_qrc

alecandido:
With QM I've not been able to reproduce the issue, because of the following error:
```py
  File "/home/users/alessandro.candido/.cache/pypoetry/virtualenvs/qrcli-S0iYXEb--py3.10/lib/python3.10/site-packages/qm/octave/calibration_db.py", line 98, in __del__
    self._db.close()
AttributeError: 'CalibrationDB' object has no attribute '_db'
```
(maybe I should fiddle a bit with the versions)

However, I already get an error using `dummy`, and there is a different error I obtained by not specifying the output folder (despite a default is available)
<details><summary>The errors</summary>

```py
  File "/home/alessandro/Projects/quantum/qrc/.venv/bin/qq", line 8, in <module>
    sys.exit(command())
             ^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/.venv/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/.venv/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/.venv/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/.venv/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/.venv/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/qibocal/src/qibocal/cli/_base.py", line 72, in auto
    autocalibrate(runcard, folder, force, update)
  File "/home/alessandro/Projects/quantum/qrc/qibocal/src/qibocal/cli/autocalibration.py", line 51, in autocalibrate
    history = run(output=path, runcard=runcard, mode=AUTOCALIBRATION)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/qibocal/src/qibocal/auto/execute.py", line 106, in run
    instance.run_protocol(
  File "/home/alessandro/Projects/quantum/qrc/qibocal/src/qibocal/auto/execute.py", line 77, in run_protocol
    completed = task.run(
                ^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/qibocal/src/qibocal/auto/task.py", line 97, in run
                                          ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/qibocal/src/qibocal/auto/operation.py", line 42, in wrapper
    out = func(*args, **kwds)
          ^^^^^^^^^^^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/qibocal/src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py", line 94, in _acquisition
    qd_pulses[qubit] = platform.create_qubit_drive_pulse(
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/qibolab/src/qibolab/platform/platform.py", line 379, in create_qubit_drive_pulse
    qubit = self.get_qubit(qubit)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/qibolab/src/qibolab/platform/platform.py", line 321, in get_qubit
    return list(self.qubits.keys())[qubit]
           ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
TypeError: list indices must be integers or slices, not str
```

```py
  File "/home/alessandro/Projects/quantum/qrc/.venv/bin/qq", line 8, in <module>
    sys.exit(command())
             ^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/.venv/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/.venv/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/.venv/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/.venv/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/.venv/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alessandro/Projects/quantum/qrc/qibocal/src/qibocal/cli/_base.py", line 72, in auto
    autocalibrate(runcard, folder, force, update)
  File "/home/alessandro/Projects/quantum/qrc/qibocal/src/qibocal/cli/autocalibration.py", line 42, in autocalibrate
    runcard.dump(folder)
  File "/home/alessandro/Projects/quantum/qrc/qibocal/src/qibocal/auto/runcard.py", line 91, in dump
    (path / RUNCARD).write_text(yaml.safe_dump(asdict(self)))
     ~~~~~^~~~~~~~~
TypeError: unsupported operand type(s) for /: 'NoneType' and 'str'
```

</details>

I will now focus on those. Then, I will ask you again for the environment. Mine is the following:
https://github.com/alecandido/qrc/blob/ffc94221b866f6c5be538deac688a335c479b396/pyproject.toml#L17-L18
https://github.com/alecandido/qrc/blob/qibocal-909/.gitmodules

andrea-pasquale:
> With QM I've not been able to reproduce the issue, because of the following error:
> 
> ```python
> Traceback (most recent call last):
>   File "/home/users/alessandro.candido/.cache/pypoetry/virtualenvs/qrcli-S0iYXEb--py3.10/lib/python3.10/site-packages/qm/octave/calibration_db.py", line 98, in __del__
>     self._db.close()
> AttributeError: 'CalibrationDB' object has no attribute '_db'
> ```
> 
> (maybe I should fiddle a bit with the versions)
> 
> However, I already get an error using `dummy`, and there is a different error I obtained by not specifying the output folder (despite a default is available)
> The errors
> 
> I will now focus on those. Then, I will ask you again for the environment. Mine is the following: https://github.com/alecandido/qrc/blob/ffc94221b866f6c5be538deac688a335c479b396/pyproject.toml#L17-L18 https://github.com/alecandido/qrc/blob/qibocal-909/.gitmodules

Uhm, that is weird... Hopefully I didn't break anything when I pushed my changes in https://github.com/qiboteam/qibolab_platforms_qrc/pull/148
I'll try again

alecandido:
Ah, apparently the two issues I described above are there even with `main`...
(switching to dummy)

And my bad: the error above was caused by me not exporting the environment variable `QIBOLAB_PLATFORMS` (on my machine I have it set by Nix, but there is no Nix on the QRC cluster, and I didn't study yet how to properly use it without admin privileges, so I'm definitely not proposing to use it there... yet...).

Now I've been able to run with https://github.com/qiboteam/qibocal/tree/main and `qw11q`.

alecandido:
Ok, I tested again, and now I'm not getting any error at all...

I'm now using https://github.com/alecandido/qrc/commit/bda9c728a26d995d9d7cada38d50c6deac7ab267

Here is the command and the printed output:
```py
alessandro.candido@saadiyat:~/qrc$ srun -p qw5q_platinum poetry run qq auto runcards/string-targets.yml -o var/ciao -f
2024-07-03 14:19:43,502 - qm - INFO     - Starting session: 62b8d03d-b91d-4e14-bdb9-c2ee14973ce6
2024-07-03 14:19:46,470 - qm - INFO     - Octave "octave4" Health check passed, current temperature 56
2024-07-03 14:19:47,389 - qm - INFO     - Octave "octave5" Health check passed, current temperature 59
2024-07-03 14:19:48,405 - qm - INFO     - Octave "octave6" Health check passed, current temperature 57
2024-07-03 14:19:48,405 - qm - INFO     - Performing health check
2024-07-03 14:19:48,424 - qm - WARNING  - Health check warning: Inter-OPX connectivity issues in OPX: con1. Missing ports are: 12, 11, 10, 9. See QM-App for more info.
2024-07-03 14:19:48,424 - qm - WARNING  - Health check warning: Inter-OPX connectivity issues in OPX: con2. Missing ports are: 12, 11, 10, 9. See QM-App for more info.
2024-07-03 14:19:48,424 - qm - WARNING  - Health check warning: Inter-OPX connectivity issues in OPX: con3. Missing ports are: 12, 11, 10, 9. See QM-App for more info.
2024-07-03 14:19:48,424 - qm - WARNING  - Health check warning: Inter-OPX connectivity issues in OPX: con4. Missing ports are: 12, 11, 10, 9. See QM-App for more info.
2024-07-03 14:19:48,425 - qm - WARNING  - Health check warning: Inter-OPX connectivity issues in OPX: con5. Missing ports are: 12, 11, 10, 9. See QM-App for more info.
2024-07-03 14:19:48,425 - qm - WARNING  - Health check warning: Inter-OPX connectivity issues in OPX: con6. Missing ports are: 12, 11, 10, 9. See QM-App for more info.
2024-07-03 14:19:48,425 - qm - WARNING  - Health check warning: Inter-OPX connectivity issues in OPX: con7. Missing ports are: 12, 11, 10, 9. See QM-App for more info.
2024-07-03 14:19:48,425 - qm - WARNING  - Health check warning: Inter-OPX connectivity issues in OPX: con8. Missing ports are: 12, 11, 10, 9. See QM-App for more info.
2024-07-03 14:19:48,425 - qm - WARNING  - Health check warning: Inter-OPX connectivity issues in OPX: con9. Missing ports are: 12, 11, 10, 9. See QM-App for more info.
2024-07-03 14:19:48,425 - qm - INFO     - Health check passed
2024-07-03 14:19:50,190 - qm - INFO     - Sending program to QOP for compilation
2024-07-03 14:19:50,681 - qm - WARNING  - pulse 'drive(1000, 0.007, Gaussian(5))' used in play is not part of element 'driveD2' operations
2024-07-03 14:19:50,681 - qm - WARNING  - pulse 'drive(1000, 0.007, Gaussian(5))' used in play is not part of element 'driveD3' operations
2024-07-03 14:19:50,682 - qm - WARNING  - pulse 'drive(1000, 0.007, Gaussian(5))' used in play is not part of element 'driveD4' operations
2024-07-03 14:19:50,682 - qm - WARNING  - pulse 'drive(1000, 0.007, Gaussian(5))' used in play is not part of element 'driveD5' operations
2024-07-03 14:19:50,682 - qm - WARNING  - pulse 'readout(2000, 0.002, Rectangular())' used in measure is not part of element 'readoutD5' operations
2024-07-03 14:19:50,917 - qm - INFO     - Executing program
```
and here is a zip with all the output, including the report.
[test-qibocal-909.zip](https://github.com/user-attachments/files/16086389/test-qibocal-909.zip)

Ok, the fits are weird, and I should check that's also the case with `main`, but I got no error...

alecandido:
> Ok, the fits are weird, and I should check that's also the case with `main`, but I got no error...

Ok, apparently they're weird in the same way also with `main`. But it would be nice if you could reproduce it (either the behavior I'm observing or the error).

src/qibocal/auto/execute.py
```diff
@@ -58,54 +42,17 @@ def run_protocol(
...
-    targets = runcard.targets if runcard.targets is not None else list(platform.qubits)
-    instance = Executor(
-        history=History.load(output),
-        platform=platform,
-        targets=targets,
-        output=output,
-        update=runcard.update,
-    )
+        # TODO: drop, as the conditions won't be necessary any longer, and then it could
+        # be performed as part of `task.run` https://github.com/qiboteam/qibocal/issues/910
```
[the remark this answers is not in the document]
alecandido: Because you will write the update anyhow, as an update dictionary.

Then, you may decide to apply it or not, but I have a different plan also for that: we can return the update as part of the `Completed` object, and call `completed.update()` to update it.

Thus, I'd rather move the conditional to `Runcard.run()`. Which I could even do immediately, to be fair (too many things to refactor, I'm always missing something).

andrea-pasquale:
> test-qibocal-909.zip

I think that zip is empty @alecandido

alecandido:
> I think that zip is empty @alecandido

Correct, it was empty, my bad. I replaced it in the same comment.

alecandido:
Steps to reproduce:
```sh
git clone git@github.com:alecandido/qrc
cd qrc/
git switch qibocal-909
git submodule init
git submodule update
poetry install --with dev
export QIBOLAB_PLATFORMS=$(realpath ./qibolab_platforms_qrc)
srun -p qw5q_platinum poetry run qq auto runcards/string-targets.yml -o var/test-qibocal-909
```

But this time I also failed with:
```py
  File "/nfs/users/alessandro.candido/qrc/qibocal/src/qibocal/auto/operation.py", line 136, in __getitem__
    return self.data[qubit]
KeyError: ('D', '1')
```

I can now investigate!

alecandido:
The problem is happening during the fit. And I know what is causing it:
https://github.com/qiboteam/qibocal/blob/7e57ee60575226a5a8b8049a9a72feb60e6ec23f/src/qibocal/auto/operation.py#L134-L135
i.e. while `tuple` and `list` are iterable, and `int` is not, a `str` is iterable as well.

In the process, I found out that, while the problem is occurring during the fit, and data have already been acquired, there are no data in the output folder, since the dump only happens at the end of `Task.run()`.

This is very suboptimal, because in case of exceptions in the fit data are lost. I won't fix this immediately, to focus on closing this and #869.
But I will try to refactor the `Task` right after, tackling #908 and this is issue as well (which I'm about to open).

alecandido:
@andrea-pasquale now it seems to work, though the fits are still failing...

[test-qibocal-909.zip](https://github.com/user-attachments/files/16108484/test-qibocal-909.zip)
https://github.com/alecandido/qrc/commit/06a734b2498f6c678efad0f9ac583bd8875e0e13

andrea-pasquale:
> @andrea-pasquale now it seems to work, though the fits are still failing...
> 
> [test-qibocal-909.zip](https://github.com/user-attachments/files/16108484/test-qibocal-909.zip) [alecandido/qrc@06a734b](https://github.com/alecandido/qrc/commit/06a734b2498f6c678efad0f9ac583bd8875e0e13)

Thanks @alecandido, indeed now the acquisition is working.
I'm already aware of the fits failing, I'm trying to improve them in #914

alecandido:
I rebased this on #921, since I believe it will be merged first (and in the hope it will make fewer workflows crashing...)

src/qibocal/update.py
```diff
@@ -42,25 +43,25 @@ def readout_attenuation(att: int, platform: Platform, qubit: QubitId):
     platform.qubits[qubit].readout.attenuation = int(att)
 
 
-def drive_frequency(freq: Union[float, tuple], platform: Platform, qubit: QubitId):
+def drive_frequency(freq: Union[float, list], platform: Platform, qubit: QubitId):
     """Update drive frequency value in platform for specific qubit."""
-    if isinstance(freq, tuple):
+    if isinstance(freq, Iterable):
```
andrea-pasquale, line 48: As soon as we merge #905, here it should be enough to put `list`.

src/qibocal/auto/history.py
```diff
@@ -1,54 +1,105 @@
...
-                raise_error(KeyError, f"{completed.task.id} already in History.")
-        self[completed.task.id] = completed
+        id = completed.task.id
+        self._tasks[id].append(completed)
+        task_id = TaskId(id=id, iteration=len(self._tasks[id]) - 1)
+        self._order.append(task_id)
+        return task_id
+
+    @staticmethod
+    def route(completed: Completed, folder: Path):
```
andrea-pasquale, line 86:
```suggestion
    def route(completed: Completed, folder: Path) -> Path:
```

src/qibocal/auto/output.py
```diff
@@ -0,0 +1,252 @@
...
+    @classmethod
+    def load(cls, path: Path):
+        """Load output from existing folder."""
+        return cls(
+            history=History.load(path),
+            meta=Metadata.load(path),
+        )
+
+    @staticmethod
+    def mkdir(path: Optional[Path] = None, force: bool = False):
```
andrea-pasquale, line 162:
```suggestion
    def mkdir(path: Optional[Path] = None, force: bool = False) -> Path:
```

src/qibocal/auto/task.py
```diff
@@ -137,61 +188,59 @@ def __post_init__(self):
...
+        task = Task.load(path)
+        return cls(path=path, task=task)
+
+    def update_platform(self, platform: Platform):
+        """Perform update on platform' parameters by looping over qubits or
+        pairs."""
```
andrea-pasquale, line 241: Although this is not related to this PR, here it would be interesting to pass also targets as argument.
The idea is that especially if you are using a python script you might want to update only specific targets, for example those that satisfied a specific condition.
Feel free to open an issue about this.

andrea-pasquale (approved):
Thanks @alecandido, LGTM.
I have some suggestions in the comments, but no major changes.

andrea-pasquale:
@alecandido when you will fix conflicts we could try to address also #925.
Given that you removed the possibility to run on simulation we can now remove the check `if platform is not None` when running `qq update`.

src/qibocal/update.py
alecandido: I kept `Iterable`, but if you're sure it is never a `tuple` I can restore this.

runcards/actions_qq.yml
```diff
@@ -5,11 +5,11 @@ targets: [0,1]
...
-#   - id: resoantor_amplitude
-#     operation: resonator_amplitude
-#     parameters:
-#       amplitude_step: 0.1
-#       amplitude_stop: 0.5
+  - id: resoantor_amplitude
+    operation: resonator_amplitude
+    parameters:
+      amplitude_step: 0.1
+      amplitude_stop: 0.5
```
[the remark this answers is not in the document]
alecandido: Done in a37ee849cbdb2c02519fbf784bd2c0eb6481158a

runcards/script.py
```diff
@@ -0,0 +1,20 @@
...
+    "parameters": {
+        "delay_before_readout_start": 0,
+        "delay_before_readout_end": 20_000,
+        "delay_before_readout_step": 50,
+    },
+}
+
+with TemporaryDirectory() as tmp:
+    exec = Executor.create(output=Path(tmp), platform="dummy")
+    out = exec.run_protocol(t1_signal, parameters=t1_params)
```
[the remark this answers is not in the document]
alecandido: Definitely, but now you provided it yourself :)

I actually removed this one, since it would miss the many pieces we are maintaining in the new scripts.

src/qibocal/protocols/two_qubit_state_tomography.py
```diff
@@ -279,7 +279,9 @@ def plot_measurements(data: StateTomographyData, target: QubitPairId):
 
 
 def _plot(data: StateTomographyData, fit: StateTomographyResults, target: QubitPairId):
-    """Plotting for two qubit state tomography"""
+    """Plotting for two qubit state tomography."""
+    if isinstance(target, list):
+        target = tuple(target)
```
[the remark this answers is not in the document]
alecandido: I would not worry too much about this in this PR, this is just a temporary patch.
We need to face the whole serialization issue and the targets handling. I would dedicate a specific PR to each of them.

Since it's not extremely urgent (it's already affecting many routines, and they are not going to grow much more in the near future), let's keep it for later.

src/qibocal/auto/task.py
alecandido: I would deal with this in #910, when we will decide how to apply the generated updates.

src/qibocal/cli/autocalibration.py
alecandido: Now tracked by #913

src/qibocal/auto/output.py
```diff
@@ -0,0 +1,256 @@
...
+    """Execution metadata."""
+
+    title: str
+    backend: str
+    platform: str
+    start_time: Optional[datetime]
+    end_time: Optional[datetime]
+    stats: dict[str, TaskStats]
+    versions: Versions
+    more: Optional[dict] = None
```
Edoardo-Pedicillo, line 53: I understand the point of this, but it is not used, maybe we could drop it

src/qibocal/auto/output.py
```diff
@@ -0,0 +1,256 @@
...
+        d["versions"] = versions["other"] | {"qibocal": versions["qibocal"]}
+        d["date"] = str(self.start_time.date()) if self.start_time is not None else None
+        d["start-time"] = (
+            self.start_time.strftime("%H:%M:%S")
+            if self.start_time is not None
+            else None
+        )
+        d["end-time"] = (
+            self.end_time.strftime("%H:%M:%S") if self.end_time is not None else None
+        )
```
Edoardo-Pedicillo, line 116: There are both "start-time" and "start_time", the same for "end_time" and "end-time". Is this intended?

alecandido: Yes, this is intended, since `start_time` is used to dump the full timestamp (and it's the one that will be automatically provided later on), but `start-time` and friends are used for the report.
Notice that `start-time`, `end-time`, and `date`, only exists in the dump, and they are ignored in the load.

I will make everything consistent later on, but for this PR it was just a patch.

src/qibocal/auto/output.py
alecandido: Yes, you're right: it was a temporary solution, to make sure I was not losing anything from the previous one. But yes, I ended up not using it, so it has to be dropped.

alecandido: Done in 7560bc49a2495b082a6965d0456cfd3dc4fada04

src/qibocal/cli/update.py
```diff
@@ -3,12 +3,12 @@
...
 import shutil
 
+from ..auto.output import META, UPDATED_PLATFORM
 from ..config import log, raise_error
-from .utils import META, UPDATED_PLATFORM
 
 
 def update(path: pathlib.Path):
-    """Perform copy of updated platform in QIBOLAB_PLATFORM."
+    """Perform copy of updated platform in QIBOLAB_PLATFORM.".
```
andrea-pasquale, line 11:
```suggestion
    """Perform copy of updated platform in QIBOLAB_PLATFORM
```

andrea-pasquale (commented):
Thanks for the updates @alecandido, I have only a few minor comments.

src/qibocal/cli/update.py
alecandido: Sorry, this was most likely something automated.

src/qibocal/update.py
alecandido: I trust you, and restored the file from `main`.

Done in 2c6cad860f91d76e300659fefca86f235c91e3d7

src/qibocal/protocols/two_qubit_state_tomography.py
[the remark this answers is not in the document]
alecandido: Postponed to #932

src/qibocal/auto/output.py
alecandido: Since it is intended, I'm closing this conversation. If you still have doubts, feel free to reopen it.

Edoardo-Pedicillo: Understood, can you open an issue ? (or are you changing it in another PR?)

andrea-pasquale:
> @andrea-pasquale it seems we needed `Iterable`, for some reason, since tests were passing before 2c6cad8, and failing on it:
> 
> image
> 
> 
> 
> Should I restore?
> 
> 
> 
> (most likely the problem could be solved even differently, if you do it that's perfect, but otherwise I would go for the simplest)

Yeah, let's restore `Iterable` then.

src/qibocal/auto/output.py
alecandido: It will be addressed in #932

alecandido: Ah no, that's not enough, because I will need to fix the reports as well. The issue is now #933

andrea-pasquale:
@alecandido I'm testing this PR and I noticed the following error:
```sh
[Qibo 0.2.7|INFO|2024-07-10 18:54:09]: Loading platform dummy
[Qibocal 0.0.11|WARNING|2024-07-10 18:54:09]: Deleting previous directory test_dummy_readout.
[Qibocal 0.0.11|INFO|2024-07-10 18:54:09]: Creating directory test_dummy_readout.
[Qibo 0.2.7|INFO|2024-07-10 18:54:09]: Connecting to instrument <qibolab.instruments.dummy.DummyInstrument object at 0x7fc59c907eb0>.
[Qibo 0.2.7|INFO|2024-07-10 18:54:09]: Connecting to dummy instrument.
[Qibo 0.2.7|INFO|2024-07-10 18:54:09]: Connecting to instrument <qibolab.instruments.dummy.DummyLocalOscillator object at 0x7fc59c9075e0>.
[Qibocal 0.0.11|INFO|2024-07-10 18:54:09]: Executing mode AUTOCALIBRATION on readout characterization.
  File "/nfs/users/andrea.pasquale/default/bin/qq", line 8, in <module>
    sys.exit(command())
  File "/nfs/users/andrea.pasquale/default/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
  File "/nfs/users/andrea.pasquale/default/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/nfs/users/andrea.pasquale/default/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
  File "/nfs/users/andrea.pasquale/default/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
  File "/nfs/users/andrea.pasquale/default/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/nfs/users/andrea.pasquale/qibocal/src/qibocal/cli/_base.py", line 74, in auto
    autocalibrate(runcard, folder, force, update)
  File "/nfs/users/andrea.pasquale/qibocal/src/qibocal/cli/autocalibration.py", line 41, in autocalibrate
    history = runcard.run(
  File "/nfs/users/andrea.pasquale/qibocal/src/qibocal/auto/runcard.py", line 63, in run
    instance.run_protocol(
  File "/nfs/users/andrea.pasquale/qibocal/src/qibocal/auto/execute.py", line 52, in run_protocol
    completed = task.run(platform=self.platform, targets=self.targets, mode=mode)
  File "/nfs/users/andrea.pasquale/qibocal/src/qibocal/auto/task.py", line 149, in run
  File "/nfs/users/andrea.pasquale/qibocal/src/qibocal/auto/operation.py", line 42, in wrapper
    out = func(*args, **kwds)
  File "/nfs/users/andrea.pasquale/qibocal/src/qibocal/protocols/readout_characterization.py", line 83, in _acquisition
    qubit_frequencies={
  File "/nfs/users/andrea.pasquale/qibocal/src/qibocal/protocols/readout_characterization.py", line 84, in <dictcomp>
    qubit: platform.qubits[qubit].drive_frequency for qubit in targets
KeyError: '0'
srun: error: yas: task 0: Exited with exit code 1
```
Here is the runcard:
```yaml
targets: [0, 1]

actions:

  - id: readout characterization
    operation: readout_characterization
    parameters:
      nshots: 5000
      delay: 1000
```
and I launched the command with `srun -p sim qq auto runcards/readout.yml -o test_dummy_readout -f --platform dummy`. The same error appear if I try to run with `qw11q`.
Any ideas?

alecandido:
> Any ideas?

It seems it is attempting to index with a string, instead of an integer. I will investigate why.

src/qibocal/auto/mode.py
```diff
@@ -2,12 +2,18 @@
...
+    """Perform fitting only."""
+
+    def __str__(self):
+        if self is AUTOCALIBRATION:
+            return "AUTOCALIBRATION"
+
+        return self.name
 
 
 AUTOCALIBRATION = ExecutionMode.ACQUIRE | ExecutionMode.FIT
```
Edoardo-Pedicillo, line 19: Probably the name is misleading: For sure doing an Acquisition and a fit is not autocalibration

andrea-pasquale: I mean, fine. What name would you suggest instead?

alecandido: I agree we could change the name. But as you can see, it is not introduced by this PR, so we can do it even somewhere else.

If not `AUTOCALIBRATION`, we could limit to `AUTO`, and otherwise we should even change `qq auto`. But it's fine for me, we can change, and we can even change both.

src/qibocal/auto/operation.py
```diff
@@ -130,6 +130,8 @@ def __init__(
 
     def __getitem__(self, qubit: Union[QubitId, tuple[QubitId, int]]):
         """Access data attribute member."""
+        if isinstance(qubit, list):
```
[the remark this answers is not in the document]
alecandido: I would not worry too much, not here. This is really going to improve with #932

src/qibocal/auto/history.py
```diff
@@ -1,54 +1,105 @@
...
+    def push(self, completed: Completed) -> TaskId:
         """Adding completed task to history."""
-        if completed.task.id in self:
-            # patch to make sure that calling fit after acquire works
-            if self[completed.task.id]._results is not None:
-                raise_error(KeyError, f"{completed.task.id} already in History.")
-        self[completed.task.id] = completed
+        id = completed.task.id
+        self._tasks[id].append(completed)
+        task_id = TaskId(id=id, iteration=len(self._tasks[id]) - 1)
```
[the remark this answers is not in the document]
alecandido: But it is never empty, because you just appended. We could compute the length before appending, but it's the exact same.

src/qibocal/auto/history.py
```diff
@@ -1,54 +1,105 @@
...
+        Representing the object as a map of sequences makes it smoother to identify the
+        iterations of a given task, since they are already grouped together.
+    """
+    _order: list[TaskId] = field(default_factory=list)
+    """Record of the execution order."""
 
-class History(dict[Id, Completed]):
-    """Execution history.
+    @singledispatchmethod
+    def __contains__(self, elem: Id):
```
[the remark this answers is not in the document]
alecandido: True, the check is done, but type hints are not for runtime.
Here I'm not claiming that is not working if you pass something else, but that if you use something different from `Id` that's wrong a priori (and since the analysis is embedded in IDEs, you will get the message in the editor - or in general without running tests).

I believe the confusion is coming from the use of type hints at runtime (here, but also in dataclasses and so on). You can, because this information is there, but type hints are mainly intended for static analysis.

alecandido:
@andrea-pasquale I'm still investigating the issue above. The weirdest thing is that I can reproduce the issue on the cluster, even running IPython on saadiyat (Linux py3.10). But I can not on my laptop (MacOS py3.11).

The issue is incredibly simple, and it doesn't even require loading a runcard. Just instantiating a `Runcard` object:
```py
In [1]: from qibocal.auto.runcard import Runcard

In [2]: Runcard(targets=[0, 1], actions=[])
Out[2]: Runcard(actions=[], targets=['0', '1'], backend='qibolab', platform='dummy', update=True)
```

alecandido:
Ok, it is not related to the Python version or OS, but rather to the Pydantic version.

This branch upgraded to Pydantic v2
https://github.com/qiboteam/qibocal/blob/a26ad0542ea50ddf2626f86e077dc5f50752c253/pyproject.toml#L25
and thus smart unions are already the default. But this is not the case for v1.

(Un)fortunately, my repo is also yielding that version, as a result of something else (not sure why), but as you can see the new test is not failing.
(it's not that bad that I also had Pydantic v1 somewhere, at least I've been able to find the issue)

Edoardo-Pedicillo (approved):

Issue #897, opened 2024-06-21
Rabi fit failing in some instances


andrea-pasquale:
While testing the emulator I saw that the rabi fit tends to fail even if the output is quite well defined.

![image_2024-06-19_20-28-00](https://github.com/qiboteam/qibocal/assets/49183315/9e15f21a-1a7b-475b-8e14-832ca8e77572)

I think that the error is related to the fft guess, we need to investigate this more.

alecandido:
Here we can see multiple oscillations, and the emulator is returning a fully imbalanced signal, since it's always positive (at the end of the day, it is always a probability, despite working as a signal).

Thus, we have to subtract from the FFT the 0th moment, i.e. the integral itself, otherwise it will dominate the transformed function, and it may lead to a guess for initial parameters that is very far from the desired global minimum.

Most likely, this was already done in a former version of Qibocal, but it got lost in the current one.

sorewachigauyo:
On this topic, it might be nice to add tests for individual routines using the emulator (single qubit for now), since the results are deterministic, but that might increase the CI/CD cost.

andrea-pasquale:
Thanks @sorewachigauyo, indeed we could consider doing something like this.
After discussing with @alecandido, a possible way to speed up the execution time with the emulator could be to run using `AveragingMode.SEQUENTIAL`. We just need to come up with a mechanism to plug this information into tests.

alecandido:
To be more precise: that's a possible way, provided that the emulator is instructed to use it properly (which I believe is not, right now).

`AveragingMode.SEQUENTIAL` makes the shots the innermost loop. If this information is taken into account, since the evolution is deterministic, the emulator should not re-evolve for each and every shot, just sampling the final state as many times as required.
This would make the shots almost priceless (i.e. for any reasonable amount of shots, the price should be pretty cheap, in comparison to the evolution).
But you will pay for each and every parameters' combination.

In any case, I fully support using the emulator for testing the routines in the CI, as many as possible (even when the results are not fully meaningful - but it will always be incredibly better than with dummy). But I would dedicate an issue on its own ^^

Issue #863, opened 2024-05-21
Plot excited state probability in Ramsey


andrea-pasquale:
Currently the Ramsey experiment is showing on the `y` axis the ground state probability.
Usually it should be plotted to excited state probability to see a proper decay when measuring `T2`.

PR #857, opened 2024-05-16
Rabi with frequency detuning

rodolfocarobene:
I'm trying to close #608. I've currently tested it with dummy only, obtaining fairly reasonable results (I changed the fake values generator):

immagine

It should be tested with real hardware. I'm not sure, in particular, of the fit procedure so at the moment I have not add any update mechanism.

I have also a question regarding "duplication": I have now written the "rabi amplitude frequency signal" routine, I was thinking to write also the one with length. Should I do something regarding the probability routines or the sequences routines?

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.
- [ ] Compatibility with Qibo modules (Please edit this section if the current pull request is not compatible with the following branches).
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`
    - [ ] Qibolab_platforms_qrc: `main`

PR #651, opened 2023-12-04
Fix fit Rabi

Edoardo-Pedicillo:
This PR fix Rabi fit and refactor Rabi utils.
Checklist:
- [x] Reviewers confirm new code works as expected.
- [x] Tests are passing.
- [x] Coverage does not decrease.
- [x] Documentation is updated.
- [x] Compatibility with Qibo modules (Please edit this section if the current pull request is not compatible with the following branches).
    - [x] Qibo: `master`
    - [x] Qibolab: `main`
    - [x] Qibolab_platforms_qrc: `main`

PR #915, opened 2024-07-03
Fixing fft guess 

andrea-pasquale:
Closes https://github.com/qiboteam/qibocal/issues/897.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/flipping_signal.py
+++ b/src/qibocal/protocols/flipping_signal.py
@@ -9,12 +9,11 @@ from qibolab.platform import Platform
 from qibolab.pulses import PulseSequence
 from qibolab.qubits import QubitId
 from scipy.optimize import curve_fit
-from scipy.signal import find_peaks
 
 from qibocal import update
 from qibocal.auto.operation import Data, Parameters, Results, Routine
 from qibocal.config import log
-from qibocal.protocols.utils import table_dict, table_html
+from qibocal.protocols.utils import guess_period, table_dict, table_html
 
 
 @dataclass
@@ -28,7 +27,7 @@ class FlippingSignalParameters(Parameters):
     unrolling: bool = False
     """If ``True`` it uses sequence unrolling to deploy multiple sequences in a single instrument call.
     Defaults to ``False``."""
-    detuning: float = 0
+    delta_amplitude: float = 0
     """Amplitude detuning."""
 
 
@@ -55,7 +54,7 @@ class FlippingSignalData(Data):
 
     resonator_type: str
     """Resonator type."""
-    detuning: float
+    delta_amplitude: float
     """Amplitude detuning."""
     pi_pulse_amplitudes: dict[QubitId, float]
     """Pi pulse amplitudes for each qubit."""
@@ -63,20 +62,22 @@ class FlippingSignalData(Data):
     """Raw data acquired."""
 
 
-def flipping_sequence(platform: Platform, qubit: QubitId, detuning: float, flips: int):
+def flipping_sequence(
+    platform: Platform, qubit: QubitId, delta_amplitude: float, flips: int
+):
 
     sequence = PulseSequence()
     RX90_pulse = platform.create_RX90_pulse(qubit, start=0)
     sequence.add(RX90_pulse)
     # execute sequence RX(pi/2) - [RX(pi) - RX(pi)] from 0...flips times - RO
     start1 = RX90_pulse.duration
-    drive_frequency = platform.qubits[qubit].native_gates.RX.frequency
+    drive_amplitude = platform.qubits[qubit].native_gates.RX.amplitude
     for _ in range(flips):
         RX_pulse1 = platform.create_RX_pulse(qubit, start=start1)
-        RX_pulse1.frequency = drive_frequency + detuning
+        RX_pulse1.amplitude = drive_amplitude + delta_amplitude
         start2 = start1 + RX_pulse1.duration
         RX_pulse2 = platform.create_RX_pulse(qubit, start=start2)
-        RX_pulse2.frequency = drive_frequency + detuning
+        RX_pulse2.amplitude = drive_amplitude + delta_amplitude
 
         sequence.add(RX_pulse1)
         sequence.add(RX_pulse2)
@@ -110,7 +111,7 @@ def _acquisition(
 
     data = FlippingSignalData(
         resonator_type=platform.resonator_type,
-        detuning=params.detuning,
+        delta_amplitude=params.delta_amplitude,
         pi_pulse_amplitudes={
             qubit: platform.qubits[qubit].native_gates.RX.amplitude for qubit in targets
         },
@@ -131,7 +132,10 @@ def _acquisition(
         sequence = PulseSequence()
         for qubit in targets:
             sequence += flipping_sequence(
-                platform=platform, qubit=qubit, detuning=params.detuning, flips=flips
+                platform=platform,
+                qubit=qubit,
+                delta_amplitude=params.delta_amplitude,
+                flips=flips,
             )
 
         sequences.append(sequence)
@@ -166,7 +170,7 @@ def _acquisition(
 
 
 def flipping_fit(x, offset, amplitude, omega, phase, gamma):
-    return np.sin(x * omega + phase) * amplitude * np.exp(-x * gamma) + offset
+    return np.cos(x * omega + phase) * amplitude * np.exp(-x * gamma) + offset
 
 
 def _fit(data: FlippingSignalData) -> FlippingSignalResults:
@@ -185,30 +189,27 @@ def _fit(data: FlippingSignalData) -> FlippingSignalResults:
     delta_amplitude_detuned = {}
     for qubit in qubits:
         qubit_data = data[qubit]
-        detuned_pi_pulse_amplitude = data.pi_pulse_amplitudes[qubit] + data.detuning
+        detuned_pi_pulse_amplitude = (
+            data.pi_pulse_amplitudes[qubit] + data.delta_amplitude
+        )
         voltages = qubit_data.signal
         flips = qubit_data.flips
+
+        period = guess_period(flips, voltages)
+
         y_min = np.min(voltages)
-        # Guessing period using Fourier transform
-        ft = np.fft.rfft(voltages)
-        # Remove the zero frequency mode
-        mags = abs(ft)[1:]
-        local_maxima = find_peaks(mags, height=0)
-        peak_heights = local_maxima[1]["peak_heights"]
-        # Select the frequency with the highest peak
-        index = (
-            int(local_maxima[0][np.argmax(peak_heights)] + 1)
-            if len(local_maxima[0]) > 0
-            else None
-        )
-        f = flips[index] / (flips[1] - flips[0]) if index is not None else 1
         y_max = np.max(voltages)
         x_min = np.min(flips)
         x_max = np.max(flips)
-        x = (flips - x_min) * 2 * np.pi * f / (x_max - x_min)
+        x = (flips - x_min) / (x_max - x_min)
         y = (voltages - y_min) / (y_max - y_min)
 
+        period = guess_period(x, y)
+        f = 1 / period
         pguess = [0.5, 0.5, 1, np.pi, 0]
+
+        x = (flips - x_min) * 2 * np.pi / period / (x_max - x_min)
+        y = (voltages - y_min) / (y_max - y_min)
         try:
             popt, _ = curve_fit(
                 flipping_fit,
@@ -246,7 +247,9 @@ def _fit(data: FlippingSignalData) -> FlippingSignalResults:
                 * detuned_pi_pulse_amplitude
                 / (np.pi + signed_correction)
             )
-            delta_amplitude_detuned[qubit] = delta_amplitude[qubit] - data.detuning
+            delta_amplitude_detuned[qubit] = (
+                delta_amplitude[qubit] - data.delta_amplitude
+            )
         except Exception as e:
             log.warning(f"Error in flipping fit for qubit {qubit} due to {e}.")
 
--- a/src/qibocal/protocols/flipping_signal.py
+++ b/src/qibocal/protocols/flipping_signal.py
@@ -198,13 +198,10 @@ def _fit(data: FlippingSignalData) -> FlippingSignalResults:
         x_min = np.min(flips)
         x_max = np.max(flips)
         x = (flips - x_min) / (x_max - x_min)
-        # shift so that first point is in 0.5
-        b = voltages[0] - 0.5
-        translated_voltages = voltages - b
-        y_max = np.max(translated_voltages)
-        y_min = np.min(translated_voltages)
+        y_max = np.max(voltages)
+        y_min = np.min(voltages)
         # normalize between 0 and 1
-        y = (translated_voltages - y_min) / (y_max - y_min)
+        y = (voltages - y_min) / (y_max - y_min)
 
         period = guess_period(x, y)
         pguess = [0.5, 0.5, 2 * np.pi / period, 0, 0]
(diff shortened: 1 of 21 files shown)


rodolfocarobene:
Sorry, this PR closes which issue? #651 seems just an old PR.
Anyway, it looks good to me, but I don't understand exactly what it is doing, was the fft not working before?

andrea-pasquale:
> Sorry, this PR closes which issue? #651 seems just an old PR. Anyway, it looks good to me, but I don't understand exactly what it is doing, was the fft not working before?

Thanks @rodolfocarobene I linked the wrong issue.
It was working before, now it seems that during some refactor something went wrong...
Now it should work properly.

src/qibocal/protocols/rabi/utils.py
```diff
@@ -270,15 +270,20 @@ def sequence_length(
...
-    mags = abs(ft)
-    local_maxima = find_peaks(mags, threshold=10)[0]
-    index = local_maxima[0] if len(local_maxima) > 0 else None
+def guess_period(x, y):
+    """Return fft period estimation given a sinusoidal plot."""
+
+    fft = np.fft.rfft(y)
+    fft_freqs = np.fft.rfftfreq(len(y), d=(x[1] - x[0]))
+    mags = abs(fft)
+    mags[0] = 0
```
rodolfocarobene, line 279: Could you explain this?

rodolfocarobene (approved):
Apart to the line I don't really understand, it looks good to me

src/qibocal/protocols/rabi/utils.py
andrea-pasquale: Yes, when you run fft you expect to see a peak corresponding to the frequency of your signal.
However, you will always find a peak in the first element of the array which corresponds to the `dc` component, which can be linked to the [sample average](https://dsp.stackexchange.com/questions/12972/discrete-fourier-transform-what-is-the-dc-term-really).
`main` was currently broken because we were not removing this first point (in this case I'm just setting in to zero).

rodolfocarobene: I see, thanks for the explanation!

andrea-pasquale: Thanks for reviewing!
I requested you for a review since you recently edited the Rabi protocols https://github.com/qiboteam/qibocal/pull/857. I just wanted someone to double check that I didn't break anything :)

alecandido: Why not `fft[1:]`, instead of setting it to `0`?

Ok, it's stupid, but it's just more readable (for me):
```py
mags = abs(fft[1:])
```

It's not truly changing anything, so it definitely personal preference...

src/qibocal/protocols/rabi/utils.py
```diff
@@ -270,15 +270,20 @@ def sequence_length(
...
+    local_maxima, _ = find_peaks(mags)
+    if len(local_maxima) > 0:
+        dominant_freq = fft_freqs[np.argmax(mags)]
+    else:
+        dominant_freq = 0  # Default if no peaks are found
     # 0.5 hardcoded guess for less than one oscillation
-    return x[index] / (x[1] - x[0]) if index is not None else 0.5
+    return 1 / dominant_freq if dominant_freq != 0 else 0.5
```
Edoardo-Pedicillo, line 286: I think it is better to return 1 in stead of 0.5 when you acquire less than a period

tests/test_rabi.py
```diff
@@ -0,0 +1,14 @@
+import numpy as np
+import pytest
+
+from qibocal.protocols.rabi.utils import guess_period
+
+
+@pytest.mark.parametrize("period", [0.2, 0.5, 1])
```
Edoardo-Pedicillo, line 7: test the case with `period`>1

Edoardo-Pedicillo (commented):
Thank @andrea-pasquale ! You should propagate it also in other protocols like Ramsey

src/qibocal/protocols/rabi/utils.py
alecandido: 0.5 is the end of the first Nyquist zone, so it's the maximum value returned by [`np.fft.rfftfreq`](https://numpy.org/doc/stable/reference/generated/numpy.fft.rfftfreq.html)

Edoardo-Pedicillo: First of all it is returning a period so it should be 2. I am not fully familiar with this concepts, but the numpy function is returning that maximum value because it is expecting you chose the parameters wisely to completely reconstruct the function but in general this is not the case. Of course, if the frequency is zero and we are expecting an oscillation we are out of the Nisquit theorem, so we can return whatever value (or raise an error).

alecandido: To be fair: this is fully redundant, because we could deal with this in the above conditional, instead of reflecting it here.
I.e. instead of
https://github.com/qiboteam/qibocal/blob/26a43b53c1c0c4ecffbd5c47051d98c092717ca6/src/qibocal/protocols/rabi/utils.py#L281-L286
use:
```py
 if len(local_maxima) > 0: 
     return 1 / fft_freqs[np.argmax(mags)]  
 return 2.
```
Even returning that value when no maximum is found is quite arbitrary. In principle, we could just fail. But in this way you leave your optimizer a chance of finding on its own (in principle, we could even return `4.`, instead, as it would be at half the frequency range `[0, 0.5]`, but it's just a random guess)

> First of all, it is returning a period so it should be 2.

Correct.

> I am not fully familiar with this concepts, but the numpy function is returning that maximum value

The frequency is half the sampling frequency, because you want to make sure to have more than one point per period (approximate explanation). If you take only one point per period, exactly spaced by one period, it would appear a constant. If your frequency is slightly less, you might see as a much wider oscillation (because if you move by `2pi - delta`, it would appear as if you were moving by just `-delta`, think about when you look at the fan or helicopter blades, and they seem to move slowly in the opposite direction).
The Nyquist-Shannon theorem proves formally that this "stops" when you take at least two points per period, because you realize there has to be an oscillations in that time range.

So, yes, the period saturates to 2.

> it is expecting you chose the parameters wisely to completely reconstruct the function but in general this is not the case.

The problem is that, above the Nyquist frequency, you simply don't know the frequency, if you're sampling.

You can be use a *response function* to modulate your signal, and move to a different Nyquist zone. But this is inductive bias about you knowing already something about your frequency, so you can distinguish aliases (@rodolfocarobene has also a nice plot to show how this work when you used during synthesis)

Edoardo-Pedicillo: > Even returning that value when no maximum is found is quite arbitrary. In principle, we could just fail. But in this way you leave your optimizer a chance of finding on its own (in principle, we could even return `4.`, instead, as it would be at half the frequency range `[0, 0.5]`, but it's just a random guess)

Ok, but I don't think we can trust the parameters of the fit (otherwise this is against the Nyquist theorem).

>(because if you move by 2pi - delta, it would appear as if you were moving by just -delta, think about when you look at the fan or helicopter blades, and they seem to move slowly in the opposite direction)

Or the Ramsey :)

>The problem is that, above the Nyquist frequency, you simply don't know the frequency, if you're sampling.

Yes, so if the `dominant frequency` is zero, this means that your real frequency is above the Nyquist one, but you cannot easily infer which one, so 2 is good as 3,4,5. For this reason I will suggest to raise an error.

alecandido: Well, while I mostly agree with you, this is not entirely correct because of the way the extraction is performed.

Since you're essentially fitting a sine, this sine should stand out of noise. Indeed, the FFT is decomposing the signal as:

$$
A_1 e^{i\omega_1 t} + A_2 e^{i\omega_2 t} + ... + A_n e^{i\omega_n t}
$$

If your frequency is among them, it should definitely stand out.

However, you will have a finite set of frequencies, and in most of the cases the frequency you're after will not be part of this set. For this reason, it is correct to use `find_peak()`, and not just `max(fourier)`.

However, noise can blur a lot the situation: if you have some noise, peaked at one of your frequencies, that would be a spike in the transformed. And it could even deform and destroy the main peak, if it's not standing that much.
Of course, there is a level of noise where it is almost impossible to make a fit. But the fit can resolve a float as a frequency (below the Nyquist), while the FFT can not. So, it could be worth to try anyhow.

Moreover, I'm not sure if `find_peak()` is optimal (considering noise), or we could just take the average of two quantiles, that in well-resolved cases should behave in the same way.
The problem is that we'd like to identify the "bulk" of the FFT, and take the middle point. But this would be something like:

$$
\mathrm{argmax}_I \left[ \int_I \mathcal{F}(f) - L(\mu(I)) \right]
$$

where you try to identify the interval that contains most of the weight of the transformed, with a small as possible interval (a constraint represented by the loss $L$, only depending on the size of the interval). But this is a fit on its own, and since there will be another one, most likely it's not so needed. That's why the idea of the fixed quantiles, and maybe `find_peak` is already good enough, though not necessarily the most stable. And that's why we could allow even for the random guess, if `find_peak` is failing.

Edoardo-Pedicillo: I was not taking into account the noisy case, usually (especially with probabilities) it is not so common, but also the case with frequency zero is not, so it is better to take it into account and return a guess. Maybe I suggest to have a warning just to be aware of the situation.

alecandido: As usual, I'm not a huge fun of warnings: they are ignored in programs (since side-effects), and then ignored also by users.

In case, much better to return `None`, and leave the choice to the caller. If this information should be called, it can be further propagated, and then made part of the `Results`.

(i.e. always prefer returning over warnings - otherwise use exceptions)

andrea-pasquale: Actually by putting `None` I believe scipy will be put `1` instead.
Shall we stick to 2 or 4 as an initial guess?

src/qibocal/protocols/flipping_signal.py
```diff
@@ -185,30 +189,26 @@ def _fit(data: FlippingSignalData) -> FlippingSignalResults:
...
-        x = (flips - x_min) * 2 * np.pi * f / (x_max - x_min)
-        y = (voltages - y_min) / (y_max - y_min)
+        x = (flips - x_min) / (x_max - x_min)
+        # shift so that first point is in 0.5
+        b = voltages[0] - 0.5
+        translated_voltages = voltages - b
+        y_max = np.max(translated_voltages)
+        y_min = np.min(translated_voltages)
+        # normalize between 0 and 1
+        y = (translated_voltages - y_min) / (y_max - y_min)
```
andrea-pasquale, line 207: Here to be able to make the fit more symmetric between signal and probability I decided to rescale the first point to 0.5, which should corresponds to no flips, afterwards I rescale everything consistently between 0 and 1.
Up to know it seems to work as expected:
http://login.qrccluster.com:9000/JzSuDWcOSLOpU5Pd3TDJKQ==

src/qibocal/protocols/ramsey/ramsey.py
```diff
@@ -307,7 +307,7 @@ def _plot(data: RamseyData, target: QubitId, fit: RamseyResults = None):
     fig.update_layout(
         showlegend=True,
         xaxis_title="Time [ns]",
-        yaxis_title="Ground state probability",
+        yaxis_title="Excited state probability",
```
andrea-pasquale, line 310: Here I'm addressing #863 
After the changes Ramsey seems to work as expected.

src/qibocal/protocols/rabi/utils.py
alecandido: Well, there are two considerations here:
1. if you return 2 or 4, it is indistinguishable from an actual outcome, so you'll never know (out of this function) whether the maximum was found or not - moreover, this function's sake is to give the best estimate of the sine frequency, not to provide an initial value for a parameter in a fit (a function should never know how it is used, it's definitely anti-modular)
2. if you return `None`, you will have to provide a fallback in every place the function is used - it's not terrible, since the "optimal" fallback value may depend on the usage, but if it's happening a lot could be pretty redundant

A modular design could be something like:
```py
fallback_period(guess_period())
```
with the outer function to be something like `lambda x: x if x is not None else 4`. If the function is not going to be used many times, you can just inline it.

src/qibocal/protocols/flipping_signal.py
alecandido: @andrea-pasquale if I'm not missing something, the `b` translation should be redundant, since it will be absorbed by the `y_min` translation.

Indeed:
```py
voltages - np.min(voltages) == translated_voltages - np.min(translated_voltages)
```

alecandido: Thanks!

src/qibocal/protocols/rabi/utils.py
andrea-pasquale: Thanks I implemented the `fallback` function in  https://github.com/qiboteam/qibocal/pull/915/commits/b3891dc33a4044e52be41410895fe48dd3f66329

src/qibocal/protocols/utils.py
```diff
@@ -772,3 +773,21 @@ def extract_feature(
...
+    mags[0] = 0
+    local_maxima, _ = find_peaks(mags)
+    if len(local_maxima) > 0:
+        dominant_freq = fft_freqs[np.argmax(mags)]
+        return 1 / dominant_freq
+    return None
+
+
+fallback_period = lambda period: period if period is not None else 4
+"""Function to estimate period if guess_period fails."""
```
alecandido, line 793: Pylint will prefer:
```suggestion
def fallback_period(period):
    """Function to estimate period if guess_period fails."""
    return period if period is not None else 4
```
and we're not paying per line, so we could make him happy :P

src/qibocal/protocols/rabi/amplitude_frequency.py
```diff
@@ -160,8 +161,8 @@ def _fit(data: RabiAmplitudeFreqData) -> RabiAmplitudeFrequencyResults:
         x = (amps - x_min) / (x_max - x_min)
         y = (y - y_min) / (y_max - y_min)
 
-        f = guess_frequency(amps, y)
-        pguess = [0.5, 0.5, 1 / f, 0]
+        f = fallback_period(guess_period(amps, y))
```
alecandido, line 164: Maybe `f` is not the best name for a period :P

(also present in other places)

alecandido (approved):
I didn't test, but it should be mostly fine.

*This PR is already doing more than one thing, since it started by fixing the FFT guess, and then deduplicated this guess in many places. Which is great, but it could be another PR, before or after this.
I believe there are more things related to sinusoidal fits to be deduplicated, but let's keep them for another PR.*

src/qibocal/protocols/rabi/amplitude_frequency.py
andrea-pasquale: Fixed https://github.com/qiboteam/qibocal/pull/915/commits/b0def1b718ff549f540246e7d2a59b6e40db0a18

src/qibocal/protocols/utils.py
```diff
@@ -772,3 +773,22 @@ def extract_feature(
...
+    """Return fft period estimation given a sinusoidal plot."""
+
+    fft = np.fft.rfft(y)
+    fft_freqs = np.fft.rfftfreq(len(y), d=(x[1] - x[0]))
+    mags = abs(fft)
+    mags[0] = 0
+    local_maxima, _ = find_peaks(mags)
+    if len(local_maxima) > 0:
+        dominant_freq = fft_freqs[np.argmax(mags)]
+        return 1 / dominant_freq
```
Edoardo-Pedicillo, line 788:
```suggestion
    if len(local_maxima) > 0:
        return 1 / fft_freqs[np.argmax(mags)]
```

Issue #932, opened 2024-07-10
Automate (de)serialization


alecandido:
The idea is to drop as many `load()`/`dump()` pairs as possible using Pydantic models.

PR #935, opened 2024-07-11
Fixing report generation after upload

### LINKED ISSUE #934: Running `qq report` after `qq upload` fails
I think that the error is related to the fact that after upload the `meta.json` will contain also an `author` field.
It should be easy to fix. To reproduce the error it should be sufficient to run `qq report <output>` when `<output>` was previously uploaded using `qq upload <output>`.

andrea-pasquale:
Closes #934 

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


src/qibocal/auto/output.py
```diff
@@ -50,6 +50,8 @@ class Metadata:
     end_time: Optional[datetime]
     stats: dict[str, TaskStats]
     versions: Versions
+    author: str = None
+    tag: str = None
```
alecandido, line 54:
```suggestion
    author: Optional[str] = None
    tag: Optional[str] = None
```

src/qibocal/auto/output.py
```diff
@@ -69,6 +71,12 @@ def generate(cls, name: str, backend):
             versions=versions,
         )
 
+    def _prepare_upload(self, author: str, tag: str):
+        """Adding author and tag attributes."""
+        self.author = author
+        if tag is not None:
```
alecandido, line 77: Why do we need this?

src/qibocal/cli/upload.py
```diff
@@ -26,11 +25,9 @@
...
 def upload_report(path: pathlib.Path, tag: str, author: str):
     # load meta and update tag
-    meta = yaml.safe_load((path / META).read_text())
-    meta["author"] = author
-    if tag is not None:
-        meta["tag"] = tag
-    (path / META).write_text(json.dumps(meta, indent=4))
+    meta = Metadata.load(path)
+    meta._prepare_upload(author=author, tag=tag)
+    (path / META).write_text(json.dumps(meta.dump(), indent=4))
```
alecandido, line 30: I just realized that `load` and `dump` are asymmetric: the second works with dictionaries, the other with paths.

Maybe we should make them symmetric, and go for dictionaries (or just wait for #932).

src/qibocal/auto/output.py
andrea-pasquale: Given that in order to upload we just need to perform this dummy operation I guess that it can be removed.
https://github.com/qiboteam/qibocal/pull/935/commits/f934aa6e58dbe91aa8bc884a73b21e037cf66186

src/qibocal/cli/upload.py
andrea-pasquale: I would wait for #932

alecandido (approved):

Edoardo-Pedicillo (approved):

PR #937, opened 2024-07-14
Few improvements to `qq compare`

andrea-pasquale:
Some improvements to `qq compare`:

- Handling multiple figures
- Legend representation

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


src/qibocal/cli/compare.py
```diff
@@ -118,34 +118,119 @@ def merge_plots(self, plots: list[list[go.Figure]]) -> list[go.Figure]:
...
+                trace_1.legend = "legend2"
+
+            if any(isinstance(trace, go.Heatmap) for trace in fig0.data):
+                # TODO: check if this is valid for all protocols
+                fig = make_subplots(rows=2, cols=2)
+                for j, trace in enumerate(fig0.data):
+                    # TODO: remove harcoded
+                    # usually with heatmap first should be signal col = 1
+                    # then phase (col = 2)
+                    # if there is another element it should be a fit done on the first column
```
GabrielePalazzo, line 137: Shall we raise a warning if `j>2`?

andrea-pasquale: Yes, we could. Actually here the comments are wrong because it should be (0) signal, (1) phase, (2) fit.
So yes, we could raise a warning.

GabrielePalazzo: Ok. Let's leave the comment, just to remember it.

src/qibocal/cli/compare.py
```diff
@@ -118,34 +118,119 @@ def merge_plots(self, plots: list[list[go.Figure]]) -> list[go.Figure]:
...
-                    row=i + 1,
-                    col=1,
-                )
-                fig.append_trace(
-                    plot[0].data[1],
-                    row=i + 1,
-                    col=2,
+        for i in range(len(plots[0])):
+            fig0 = plots[0][i]
+            fig1 = plots[1][i]
```
GabrielePalazzo, line 123: Minor remark.
Can we zip `plots[0]` and `plots[1]`? It removes a variable and it's safer (even though we expect `plots[0]` and `plots[1]` to have the same length). We can either save the combined plot into a new list or we can reuse `fig0`.
```suggestion
        for fig0, fig1 in zip(plots[0], plots[1]):
```

GabrielePalazzo (approved):

src/qibocal/cli/compare.py
andrea-pasquale: Implemented in https://github.com/qiboteam/qibocal/pull/937/commits/366b6327d9195a4f63ea6e493ef9e001a06a160b
I've also changed from `append_trace` to `add_trace` since I believe that for what we are doing the result should be the same.
If it looks good to you feel free to merge @GabrielePalazzo

PR #927, opened 2024-07-09
Minor correction to qubit spectroscopy ef and drag calibration

andrea-pasquale:
For qubit spectroscopy ef the correction is just a minus sign.
For drag, given that in the literature the parameter reported is usually not beta but beta/anharmonicity.
I've decided to rescale the plot accordingly, in the platform update I made sure to use the correct beta.
Here is an example where I put on purpose a wrong value for beta: http://login.qrccluster.com:9000/vErrCeyfS1KihyZYqviaig==/

Checklist:
- [x] Reviewers confirm new code works as expected.
- [x] Tests are passing.
- [x] Coverage does not decrease.
- [x] Documentation is updated.
- [x] Compatibility with Qibo modules (Please edit this section if the current pull request is not compatible with the following branches).
    - [x] Qibo: `master`
    - [x] Qibolab: `main`
    - [x] Qibolab_platforms_qrc: `main`

PR #939, opened 2024-07-17
Hotfix for drag protocol

andrea-pasquale:
I missed a `HZ_TO_GHZ` conversion in #927.

Here is an example: http://login.qrccluster.com:9000/2_PXMY2xQBSV1w9CC1yezQ==

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/drag.py
+++ b/src/qibocal/protocols/drag.py
@@ -283,7 +283,7 @@ def _plot(data: DragTuningData, target: QubitId, fit: DragTuningResults):
 def _update(results: DragTuningResults, platform: Platform, target: QubitId):
     try:
         update.drag_pulse_beta(
-            results.betas[target] / platform.qubits[target].anharmonicity,
+            results.betas[target] / platform.qubits[target].anharmonicity / HZ_TO_GHZ,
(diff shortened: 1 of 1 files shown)


Edoardo-Pedicillo (approved):

alecandido (approved):
Fine, but why in the first place one is in Hz and the other in GHz?

andrea-pasquale:
> Fine. But actually why one is in Hz and the other in GHz?

In `parameters.json` all frequencies, including the anharmonicity are stored in `Hz`.
For the DRAG pulse in  #927 I replaced $\beta \rightarrow \beta / \alpha$ where $\alpha$ is the anharmonicity. I believe that in the literature people are using GHz since the optimal values for $\beta$ are between 0 and 1. Therefore, let's say
that for the normalization that it is currently used in the literature the anharmoncity has to be converted to GHz.

alecandido:
Ok, let's wait for Qibolab to standardize all units to GHz and ns, using floats. For the time being, keep going with conversions.

PR #942, opened 2024-07-17
Deduplicate script operations

### LINKED ISSUE #922: Executor operations
In a sense, this is an extension of #878.

It could be helpful to expose more operations from the executor, other than just the routines.
An example are functions to save or close the execution:
```py
from myexecutor import ramsey, flipping, save, close

ramsey(...)
save()
flipping(...)
close()
```

Below a list of interesting operations:
- [ ] `save()`: dump the state of the execution as it is
- [ ] `flush()`: dump the state and free memory
- [ ] `close(save=True)`: disconnect the platform, and do the other operation to gracefully close the session (e.g. timing, and a default save, that could be disabled with a parameter)

alecandido:
~~C.loses #922~~


alecandido:
Currently, I just moved the non-routine scripts' operations
https://github.com/qiboteam/qibocal/blob/93d89083b31e2f5d8d17d9ad8fff6bd4698d2b8d/runcards/single_shot.py#L10-L48
to a couple of executor methods:
https://github.com/qiboteam/qibocal/blob/b1416669a8b74c88a13d8c68fdaf1d13b9040fb7/src/qibocal/auto/execute.py#L246-L292

However, I'm not fully happy to have
https://github.com/qiboteam/qibocal/blob/b1416669a8b74c88a13d8c68fdaf1d13b9040fb7/src/qibocal/auto/execute.py#L92-L93
but I also want to limit the overhead in the scripts. 
So, I'm trying to figure out whether and how it is possible to rearrange the execution and dump in a sensible way, in order to delegate the dump as much as possible to the user (limiting the executor state to the history, excluding the output), but without being too much annoying.

alecandido:
@Jacfomg your commits are now available in
https://github.com/qiboteam/qibocal/tree/flux-script

Jacfomg:
> @Jacfomg your commits are now available in https://github.com/qiboteam/qibocal/tree/flux-script

Thanks and sorry for the mess

alecandido:
Let's close this here, and continue in further PRs.

src/qibocal/auto/execute.py
```diff
@@ -237,3 +242,56 @@ def __del__(self):
         except KeyError:
             # it has been explicitly unloaded, no need to do it again
             pass
+
+    def init(
+        self,
+        path: os.PathLike,
+        force: bool = False,
+        platform: Union[Platform, str, None] = None,
+        targets: Optional[Targets] = None,
```
andrea-pasquale, line 251: Here we should also pass `update`. I realize yesterday that by default is `True`.
Therefore, all the `if-else` condition related to platform update in the example scripts are useless...
Perhaps, we could put it as `False` when using execution scripts while we keep True for CLI execution.

runcards/rx_calibration.py
```diff
@@ -1,73 +1,41 @@
...
-# connect and initialize platform
-platform.connect()
-
-# run
-meta.start()
+from myexec import close, drag_tuning, init, rabi_amplitude, ramsey
 
+target = 0
+platform = executor.platform
+platform.settings.nshots = 2048
```
andrea-pasquale, line 10: I think that if you don't put it we will `nshots` specified in `parameters.json`, correct?

andrea-pasquale (approved):
Thanks @alecandido, LGTM.
I have one relevant comment related to `update`, the rest seems ready to merge.

Edoardo-Pedicillo (commented):
LGTM, maybe instead of `init` and `close`, I would prefer two names that are more related to each other, like `connect` and `disconnect`.

Edoardo-Pedicillo (approved):

alecandido:
> LGTM, maybe instead of `init` and `close`, I would prefer two names that are more related to each other, like `connect` and `disconnect`.

There is already an `open`, to match Python usual `open`/`close` pair (it will be `Executor.open()`, to avoid shadowing the built-in `open()`).
However, I'm open to change names. Scripts should be still considered "experimental", we're trying to define the best API. But I'd avoid `connect`/`disconnect` specifically, because it's doing something more than just connecting to the instruments.

alecandido:
I'm now merging, but let's keep discussing improvements, and implement them in new PRs.

PR #949, opened 2024-07-28
Executor context manager

alecandido:
This is more or less complete as it is.

I will test, but it is already doing little to nothing.

Since this was originally an idea by @andrea-pasquale, I'd say it is ready for *your* review in the first place, and I'd like to collect your feedback.

### DIFF
--- a/runcards/rx_calibration.py
+++ b/runcards/rx_calibration.py
@@ -1,77 +1,73 @@
 from qibocal.auto.execute import Executor
 from qibocal.cli.report import report
 
-executor = Executor.create(name="myexec", platform="dummy")
-
-from myexec import close, drag_tuning, init, rabi_amplitude, ramsey
-
 target = 0
-platform = executor.platform
-platform.settings.nshots = 2048
-init("test_rx_calibration", force=True, targets=[target])
+with Executor.open(
+    "myexec", path="test_rx_calibration", platform="dummy", targets=[target], force=True
+) as e:
+    e.platform.settings.nshots = 2000
 
-rabi_output = rabi_amplitude(
-    min_amp_factor=0.5,
-    max_amp_factor=1.5,
-    step_amp_factor=0.01,
-    pulse_length=platform.qubits[target].native_gates.RX.duration,
-)
-# update only if chi2 is satisfied
-if rabi_output.results.chi2[target][0] > 2:
-    raise RuntimeError(
-        f"Rabi fit has chi2 {rabi_output.results.chi2[target][0]} greater than 2. Stopping."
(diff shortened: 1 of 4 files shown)


andrea-pasquale (commented):
Thanks @alecandido, LGTM.
I will test it shortly.
There is one thing still to be addressed after #942:
https://github.com/qiboteam/qibocal/pull/942#discussion_r1694648818.
I think that we should expose also the `update` attribute. If we don't it will be set to `True` and all checks that we are performing in https://github.com/qiboteam/qibocal/blob/exec-ctx-manager/runcards/rx_calibration.py to assess when to update the platform don't make much sense. Let me know what you think.

alecandido:
> There is one thing still to be addressed after #942:
> [#942 (comment)](https://github.com/qiboteam/qibocal/pull/942#discussion_r1694648818).

Sorry, I've been too fast to merge that. But they are all additions, I will provide them here.

runcards/rx_calibration.py
```diff
@@ -1,77 +1,75 @@
...
+    if ramsey_output.results.chi2[target][0] > 2:
+        raise RuntimeError(
+            f"Ramsey fit has chi2 {ramsey_output.results.chi2[target][0]} greater than 2. Stopping."
+        )
+    if ramsey_output.results.delta_phys[target][0] < 1e4:
+        print(
+            f"Ramsey frequency not updated, correction too small { ramsey_output.results.delta_phys[target][0]}"
+        )
+    else:
+        ramsey_output.update_platform(e.platform)
```
andrea-pasquale, line 43: If we keep `update=True` the update will always happen.
So probably we can remove some of those lines.

andrea-pasquale (approved):
Thanks @alecandido, works for me. Just a tiny comment.

runcards/rx_calibration.py
alecandido: Yes, they are all removed, but this one.

Indeed, if you check above the Ramsey call, it is `e.ramsey(..., update=False)`, since, even in the original example, the update was not always happening, but not even failing when it's not happening (only failing for large chi2, but not for small chi2 small corrections)

andrea-pasquale: Yes, you are right. I missed that in `ramsey` you already put `update=False`.

alecandido:
Ok, tests are definitely not extensive, and they should be improved.

However, the feature is there, and minimally tested. Since there are already other PRs relying on this, let's merge as soon as @Jacfomg confirms it is fine :)

Edoardo-Pedicillo (approved):

PR #950, opened 2024-07-29
Fix default executor platform

alecandido:
Initially setting it to dummy, to avoid loading attempts from other platforms, through undesired env vars

- [ ] add a test failing in `main`, passing now
- ~~convert one of the two example scripts to `qibocal.routines`~~ done in #949

### DIFF
--- a/src/qibocal/__init__.py
+++ b/src/qibocal/__init__.py
@@ -7,7 +7,7 @@ from .version import __version__
 
 __all__ = ["Executor", "protocols", "command", "__version__"]
 
-DEFAULT_EXECUTOR = Executor.create(".routines")
+DEFAULT_EXECUTOR = Executor.create(".routines", platform="dummy")
 """Default executor, registered as a qibocal submodule.
 
 It is defined for streamlined usage of qibocal protocols in simple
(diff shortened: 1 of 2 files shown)


andrea-pasquale:
@alecandido shall we keep this PR even if there is https://github.com/qiboteam/qrccluster/issues/202?
I don't remember what was the agreement at the end of the meeting.
I guess that we can always merge this to have a safer option.

alecandido:
Still needed: this should be considered a bug everywhere, because it is trying to look for a platform at `import` time (no one asked for).

This is just a small simplification for some scripts, and if unused should have no impact.
Loading dummy should be negligible, so I consider it fine. But here it was trying to may try to load a platform no one asked for, just because you're importing it (and maybe failing on `import qibocal` because you didn't install `laboneq`...).

I've been busy with Qibosoq, but I hope to close this asap. The fix should be good on its own, I just wanted a test.

alecandido:
@stavros11, in case you're wondering what this is, it should fix the bug that required `unset QIBO_PLATFORM` in the queue when using Qibocal (plus the test ensuring this won't happen again).

alecandido:
Here I should have done a better job with testing (compared to #949)

stavros11 (approved):
Thanks @alecandido. It fixes the issue with `$QIBO_PLATFORM`.

andrea-pasquale (approved):

PR #953, opened 2024-07-30
Example of script syntax to run RB

andrea-pasquale:
Here is an example on how to run RB with the script syntax.
I'm running a super short RB changing the amplitude of the pi pulse.
Here are the two reports that are generated after the execution:
http://login.qrccluster.com:9000/nstuLkrSTXe_INZTUnaIjA==
http://login.qrccluster.com:9000/otZWGds9Qi6coougeOVEVQ==

It seems to work as expected.
I believe that you can already start playing with this @wilkensJ.

For any plots or post-processing which involve multiple protocols for time being feel free to dump custom data and use custom plotting functions on top of that. 

Let me know if anything is not clear.

### DIFF
--- a/runcards/rb_script.py
+++ b/runcards/rb_script.py
@@ -1,11 +1,12 @@
 from qibolab import create_platform
 
 from qibocal.auto.execute import Executor
+from qibocal.auto.history import History
 from qibocal.cli.report import report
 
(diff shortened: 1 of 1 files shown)


andrea-pasquale:
Closing in favor of #975

PR #955, opened 2024-07-31
Dispersive shift restless

Jacfomg:
Draft for the Dispersive shift restless. It can help to track the resonator behavior with restless operation of the device.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


alecandido:
Couldn't it be an option, instead of another routine?

Jacfomg:
> Couldn't it be an option, instead of another routine?

Not sure, can't run in the QPU it gets free then I can see if they are usefull and I can merge them with existant routines if so

PR #960, opened 2024-08-01
rb_correction_qua

Jacfomg:
I dont know how to point this to two branches or if it's even possible so I merge rb_qua branch here

It needs #959 to be merged

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


runcards/rb_correction_qua.py
```diff
@@ -0,0 +1,199 @@
...
+delta_clifford: int = 50
+"Play each sequence with a depth step equals to delta_clifford"
+seed: Optional[int] = 1234
+"Pseudo-random number generator seed"
+n_avg: int = 128
+"Number of averaging loops for each random sequence"
+save_sequences: bool = True
+apply_inverse: bool = True
+state_discrimination: bool = True
+"Flag to enable state discrimination if the readout has been calibrated (rotated blobs and threshold)"
```
andrea-pasquale, line 60: If those parameters are only used once perhaps I would just used directly during the protocol calls instead of adding 30 lines of code.

runcards/rb_correction_qua.py
```diff
@@ -0,0 +1,199 @@
...
+for target in [args.target]:
+
+    # NOTE: Center around the sweetspot [Optional]
+    centered_biases = biases + platform.qubits[target].sweetspot
+
+    for i, bias in enumerate(centered_biases):
+        with Executor.open(
+            f"myexec_{i}",
+            path=args.path / Path(f"flux_{i}"),
+            platform=args.platform,
```
andrea-pasquale, line 100:
```suggestion
            platform=platform,
```
Otherwise the executor will create a different platform.

runcards/rb_correction_qua.py
```diff
@@ -0,0 +1,199 @@
...
+                    pulse_fidelity_uncorrected=[
+                        rb_output_uncorrected.results.pulse_fidelity[target]
+                    ],
+                    pulse_fidelity_corrected=[
+                        rb_output_corrected.results.pulse_fidelity[target]
+                    ],
+                ),
+            )
+
+        # close()
```
andrea-pasquale, line 156:
```suggestion
```

andrea-pasquale (commented):
I've added some comments just for the script

andrea-pasquale:
> I dont know how to point this to two branches or if it's even possible so I merge rb_qua branch here

Perhaps the easiest solution is to point to #959 and merge #949 afterwards.
I'm saying this also because if #959 doesn't get merged also this one cannot be merged.

runcards/rb_correction_qua.py
Jacfomg: it was cleaner to change them there than to go aorund looking for them

alecandido:
> I dont know how to point this to two branches or if it's even possible so I merge rb_qua branch here

@Jacfomg I can tell you, it is not xD

Jacfomg:
> > I dont know how to point this to two branches or if it's even possible so I merge rb_qua branch here
> 
> 
> 
> Perhaps the easiest solution is to point to #959 and merge #949 afterwards.
> 
> I'm saying this also because if #959 doesn't get merged also this one cannot be merged.

I did it the other way around should be fine too

alecandido:
> I did it the other way around should be fine too

As a temporary solution, I'd say it's definitely fine. I will try to merge #949 asap, at which point I will rebase #959, and you could rebase your branch there.

> Someone should make an issue for this XD

Unfortunately, they do not accept issues :P
https://github.com/git/git
But you could write them an email :)

Jacfomg:
@wilkensJ @ingoroth  This would be the code to run some kind of RB corrections when simulating flux changes in the qubit. Feel free to play with it to generate the data you need. Also, you can add other routines to have a longer correction or a smaller one.

andrea-pasquale:
Closing in favor of #975

PR #968, opened 2024-08-12
Hotfix for sweetspot calculation

andrea-pasquale:
The error comes from the fact there was a redundant factor pi which is already included in 
https://github.com/qiboteam/qibocal/blob/78d4938552e46b678e4b6f10acc2bea64e039cc1/src/qibocal/protocols/flux_dependence/utils.py#L220-L221
Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
+++ b/src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
@@ -218,12 +218,10 @@ def _fit(data: QubitFluxData) -> QubitFluxResults:
                 "charging_energy": data.charging_energy[qubit] * HZ_TO_GHZ,
             }
             frequency[qubit] = popt[0] * GHZ_TO_HZ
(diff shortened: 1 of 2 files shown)


alecandido (approved):

alecandido:
Even in this case, it would be important to provide closure tests and similar, to individually (and effectively) test the various fits even in the CI.

andrea-pasquale:
> Even in this case, it would be important to provide closure tests and similar, to individually (and effectively) test the various fits even in the CI.

Definitely, as soon as the emulator also supports flux tunable qubits we can give it a try.

alecandido:
> Definitely, as soon as the emulator also supports flux tunable qubits we can give it a try.

In this case, I'd even do a non-emulator test. Generate data from the function you want to fit, add a bit of noise (even just Gaussian noise), fit.

Edoardo-Pedicillo (approved):

Issue #910, opened 2024-06-28
Save updates


alecandido:
This is something I said many times, but it seems I never wrote it.

The idea is to create updates in routines as objects (most likely, as dictionaries), instead of modifying dynamically an instance (the `Platform`).

This will have several benefits:
- no need to propagate the `update` flag within the routine: you can always run the update, and always save it
- you can later decide (at any time) which subset of updates you want to apply
- it will finally allow the time travel, resuming the execution from a certain state of the `Platform`, with different routines (without re-executing the previous ones)
- part of these features may even be exposed to `qq update` (cf. #833 #905)

The idea is that applying updates will be supported directly by Qibolab. But, in case Qibocal moves faster in this direction, this can also be implemented in here, and migrated later on.

andrea-pasquale:
> The idea is to create updates in routines as objects (most likely, as dictionaries), instead of modifying dynamically an instance (the `Platform`).

Here I'm a bit confused. In the past Qibocal was already generating updates in the form of dictionaries, which were passed to the platform that was updating itself based on those dictionaries. I believe that since the update in qibolab was a bit messy we decided to let qibocal handle the update of the platform https://github.com/qiboteam/qibolab/pull/596.
Moreover, I think that this will also generate more entanglement between qibolab and qibocal, and I believe that we were going into the opposite direction.

andrea-pasquale:
What would make sense is to generate a serialization for the updates, but I still think that the update should be performed by qibocal.

alecandido:
The idea, yet missing in Qibolab, is to be able to apply an update with a shape like:
```py
{
  "single_qubit[1].RX.frequency": 5e7,
  "single_qubit[5].MZ.duration": 1.5e3,
  ...
}
```
These are the kind of updates I have in mind.

Then there will be a function, reading the keys into an `AttributePath` object (possibly a `list[str]`), that through `__setattr__` and `__setitem__` (or `dataclasses.replace`), will apply this update to an existing platform object (or just to its serialization, slightly more inconvenient).

The overall feature will be simple: parse the path splitting on `{.[]}` characters, and then iteratively access the inner object, to eventually mutate atomic attributes (the equivalent with immutable objects might be just more annoying - but if Qibolab makes use of immutable objects, which I support, we will need to address even that, recursively replacing in the innermost object, and from there on).

The `update(platform, flattened_update)` function could be implemented both in Qibolab and Qibocal, but it is general enough that making it part of Qibolab it's not going to hurt.
[...]

PR #976, opened 2024-09-04
Fix ramsey update

andrea-pasquale:
The drive frequency is now updated only when a detuning is provided, which is one of the assumption of the fit. When detuning is 0 we update T2 since we expect it to be more reliable.

Now Ramsey can be used to monitor T2 in a more stable way given that the fit will include both oscillations and an exponential decay.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/doc/source/protocols/ramsey/ramsey.rst
+++ b/doc/source/protocols/ramsey/ramsey.rst
@@ -56,8 +56,8 @@ The expected output is the following:
 :math:`T_2` and :math:`\Delta \omega` are determined by fitting the output signal using
 the formula presented above.
 
-If the protocols is successful the drive frequency will be updated. For updating :math:`T_2`
-the user is invited to run ``T2 experiment``.
+If the protocols is successful the drive frequency will be updated only if a non-zero
+detuning is provided. :math:`T_2` is updated only in the case where detuning is 0.
 
 Requirements
 ^^^^^^^^^^^^
--- a/doc/source/protocols/ramsey/ramsey.rst
+++ b/doc/source/protocols/ramsey/ramsey.rst
@@ -57,7 +57,7 @@ The expected output is the following:
 the formula presented above.
 
 If the protocols is successful the drive frequency will be updated only if a non-zero
-detuning is provided. :math:`T_2` is updated only in the case where detuning is 0.
+detuning is provided. :math:`T_2` is updated only in the case where detuning is not specified.
 
 Requirements
 ^^^^^^^^^^^^
--- a/src/qibocal/protocols/ramsey/ramsey_signal.py
+++ b/src/qibocal/protocols/ramsey/ramsey_signal.py
@@ -14,8 +14,8 @@ from qibocal import update
 from qibocal.auto.operation import Data, Parameters, Results, Routine
 from qibocal.config import log
 
-from ..utils import GHZ_TO_HZ, table_dict, table_html
-from .utils import fitting, ramsey_fit, ramsey_sequence
+from ..utils import table_dict, table_html
+from .utils import fitting, process_fit, ramsey_fit, ramsey_sequence
 
 
 @dataclass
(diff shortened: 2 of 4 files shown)


alecandido:
In general, it would be better to use `None` as default (and type is already `Optional` which is the union of `None` with the thing its parametrized, not their to signal it has a default).

In this way, it would be smooth, and working in the same way for a `0.` or very small detuning.

All the behavior currently attributed to `0.` should be moved to `None`, including the update (with `float` is also possible to use `nan`, but I wouldn't recommend it: `nan` is not always a missing value, but possibly the result of an undefined float operation).

src/qibocal/protocols/ramsey/ramsey_signal.py
```diff
@@ -295,7 +295,10 @@ def _plot(data: RamseySignalData, target: QubitId, fit: RamseySignalResults = No
 
 
 def _update(results: RamseySignalResults, platform: Platform, target: QubitId):
-    update.drive_frequency(results.frequency[target][0], platform, target)
+    if int(results.delta_phys[target][0]) == int(results.delta_fitting[target][0]):
+        update.t2(results.t2[target][0], platform, target)
```
GabrielePalazzo, line 299: Shall we log what we are updating in the runcard in case of conditional updates (that depend on fitting)?

andrea-pasquale: This on the roadmap https://github.com/qiboteam/qibocal/issues/910
For the time being I would keep those manipulations directly in the python file where we are handling scripts.

GabrielePalazzo (approved):

andrea-pasquale:
> In general, it would be better to use `None` as default (and type is already `Optional` which is the union of `None` with the thing being parametrized, not there to signal it has a default).
> 
> In this way, it would be smooth, and working in the same way for a `0.` or very small detuning.
> 
> All the behavior currently attributed to `0.` should be moved to `None`, including the update (with `float` is also possible to use `nan`, but I wouldn't recommend it: `nan` is not always an intentionally missing value, but possibly the result of an undefined float operation - it's also sometimes non-trivial to specify, since it lacks a literal, and identify, because of "weird" comparison).

Thanks @alecandido in the last commits I've changed the default behavior to `None`

doc/source/protocols/ramsey/ramsey.rst
```diff
@@ -56,8 +56,8 @@ The expected output is the following:
 :math:`T_2` and :math:`\Delta \omega` are determined by fitting the output signal using
 the formula presented above.
 
-If the protocols is successful the drive frequency will be updated. For updating :math:`T_2`
-the user is invited to run ``T2 experiment``.
+If the protocols is successful the drive frequency will be updated only if a non-zero
+detuning is provided. :math:`T_2` is updated only in the case where detuning is 0.
```
alecandido, line 60:
```suggestion
detuning is provided. :math:`T_2` is updated only in the case where detuning is not specified.
```

src/qibocal/protocols/ramsey/ramsey_signal.py
```diff
@@ -52,6 +52,9 @@ class RamseySignalResults(Results):
     fitted_parameters: dict[QubitId, list[float]]
     """Raw fitting output."""
 
+    def is_detuned(self, qubit: QubitId):
+        return int(self.delta_phys[qubit][0]) != int(self.delta_fitting[qubit][0])
```
alecandido, line 56: I'd rather rely on `detuning` being `None`, instead of check the `int` (that in this case is almost a `float`, given the accuracy...)

andrea-pasquale: Yes, I know. The only problem is that the class results doesn't have access to `detuning`...
This is a workaround for the time being.

alecandido: Well, you could apply the usual workaround: propagate `detuning` through data and results

alecandido (approved):

PR #989, opened 2024-10-01
Update coupling in dispersive shift protocol

Edoardo-Pedicillo:
This PR updates the coupling after the execution of the dispersive shift.

http://login.qrccluster.com:9000/uAibm0TISF-mW_mtaUkRFQ==/ (not the best parameters, but it is just a test)

Checklist:
- [ ] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `0.1`

### DIFF
--- a/src/qibocal/protocols/dispersive_shift.py
+++ b/src/qibocal/protocols/dispersive_shift.py
@@ -59,7 +59,7 @@ class DispersiveShiftResults(Results):
         "Evaluate the dispersive shift"
         return {
             key: (self.frequency_state_zero[key] - self.frequency_state_one[key]) / 2
-            for key in self.frequency_state_zero.keys()
+            for key in self.frequency_state_zero
         }
 
 
--- a/src/qibocal/protocols/dispersive_shift.py
+++ b/src/qibocal/protocols/dispersive_shift.py
@@ -190,7 +190,6 @@ def _fit(data: DispersiveShiftData) -> DispersiveShiftResults:
             iq_couples[i].append(np.stack((i_measures, q_measures), axis=-1))
 
         frequencies = data[qubit, 0].freq
-
         max_index = np.argmax(
             np.linalg.norm(iq_couples[0][idx] - iq_couples[1][idx], axis=-1)
         )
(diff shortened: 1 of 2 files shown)


src/qibocal/protocols/dispersive_shift.py
```diff
@@ -352,6 +348,22 @@ def _plot(data: DispersiveShiftData, target: QubitId, fit: DispersiveShiftResult
...
+    g = np.sqrt(chi(fit_data_0, fit_data_1, target) * delta)
+    update.coupling(g, platform, target)
+
+
+def chi(fit_data_0, fit_data_1, target):
+    "Evaluate the dispersive shift"
+    return (
+        fit_data_0["frequency_state_zero"][target]
+        - fit_data_1["frequency_state_one"][target]
+    ) / 2
```
andrea-pasquale, line 366: This function is ok, but I would rather have a method in `DispersiveShiftResults` that can computes chi. The reason is that it is quite difficult to understand what are `fit_data_0` and `fit_data_1` if you consider the function alone.
Adding a method like this for `DispersiveShiftResults` could improve readability.
```py
def chi(self, qubit: QubitId):
    """Computing chi for target qubit."""
    return (
        self.frequency_state_zero[target]
        - self.frequency_state_one[target]
    ) / 2
```

andrea-pasquale (commented):
Thanks @Edoardo-Pedicillo, just a minor suggestions and we should be ready to merge.

Edoardo-Pedicillo:
http://login.qrccluster.com:9000/qQTGtwvxROSOB4tHiEcjVg==/

src/qibocal/protocols/dispersive_shift.py
```diff
@@ -54,6 +54,14 @@ def state_zero(self):
     def state_one(self):
         return {key: value for key, value in asdict(self).items() if "one" in key}
 
+    @property
+    def chi(self):
+        "Evaluate the dispersive shift"
+        return {
+            key: (self.frequency_state_zero[key] - self.frequency_state_one[key]) / 2
+            for key in self.frequency_state_zero.keys()
```
andrea-pasquale, line 62:
```suggestion
            for key in self.frequency_state_zero
```

src/qibocal/protocols/dispersive_shift.py
```diff
@@ -289,54 +295,51 @@ def _plot(data: DispersiveShiftData, target: QubitId, fit: DispersiveShiftResult
...
-                ],
-                mode="lines",
-                line=go.scatter.Line(color="orange", width=3, dash="dash"),
-                name="Best frequency",
-            ),
-            row=1,
-            col=1,
-        )
+    if fit_data_0 != None or fit_data_1 != None:
+        if fit_data_0.keys() == fit_data_1.keys():
```
andrea-pasquale, line 299: I don't understand what is the problem here.
I could try to test it myself when I have time.

andrea-pasquale (commented):
Thanks @Edoardo-Pedicillo, just a couple of comments again.

src/qibocal/protocols/dispersive_shift.py
Edoardo-Pedicillo: About the first if, the problem is that we were checking if the `Result` object is `None`, but looking at https://github.com/qiboteam/qibocal/blob/f54851a497aa251e7c270156e9e142e1c5367cfe/src/qibocal/protocols/dispersive_shift.py#L190-L194
only `frequency_state_zero` and `frequency_state_one` are None.

For the second one, if just one fit fails, the `fit_data_i`s will contain a different number of keys. 

We could even change the `fit` such that when one fit fails it breaks and returns None.

andrea-pasquale, line 299:
```suggestion
    if fit is not None:
        if target in frequency_state_zero and target in frequency_state_one:
```

src/qibocal/protocols/dispersive_shift.py
```diff
@@ -54,6 +54,14 @@ def state_zero(self):
     def state_one(self):
         return {key: value for key, value in asdict(self).items() if "one" in key}
 
+    @property
+    def chi(self):
+        "Evaluate the dispersive shift"
+        return {
+            key: (self.frequency_state_zero[key] - self.frequency_state_one[key]) / 2
+            for key in self.frequency_state_zero
+        }
```
alecandido, line 63: Maybe `qubit`, or even just `q`, would be a more telling name than `key` (even in `state_zero`/`_one` methods)

src/qibocal/protocols/dispersive_shift.py
```diff
@@ -289,54 +295,51 @@ def _plot(data: DispersiveShiftData, target: QubitId, fit: DispersiveShiftResult
...
-                    np.max(np.concatenate((data_0.signal, data_1.signal))),
-                ],
-                mode="lines",
-                line=go.scatter.Line(color="orange", width=3, dash="dash"),
-                name="Best frequency",
-            ),
-            row=1,
-            col=1,
-        )
+    if fit_data_0 != None or fit_data_1 != None:
```
alecandido, line 298:
```suggestion
    if fit_data_0 is not None or fit_data_1 is not None:
```

`None` is a unique value in its type, so you could check for identity, rather than equality (it doesn't matter that match, but it's more idiomatic).

src/qibocal/protocols/dispersive_shift.py
alecandido: > We could even change the `fit` such that when one fit fails it breaks and returns None.

On the one side, you should retain as much information as possible. But if there is no way this information will be consumed, then it may be even worth to simplify

src/qibocal/protocols/dispersive_shift.py
```diff
@@ -352,6 +355,13 @@ def _plot(data: DispersiveShiftData, target: QubitId, fit: DispersiveShiftResult
 
 def _update(results: DispersiveShiftResults, platform: Platform, target: QubitId):
     update.readout_frequency(results.best_freq[target], platform, target)
+    fit_data_0 = results.state_zero
+    delta = np.abs(
+        platform.qubits[target].drive_frequency
+        - fit_data_0["frequency_state_zero"][target]
```
alecandido, line 361: Actually, I'm not sure why (here and everywhere else) you go through the double step of accessing first `.state_zero`, to then look into the `"frequency_state_zero"` key...

Isn't it a bit redundant?
Since
```py
results.state_zero["frequency_state_zero"] == results.frequencey_state_zero
```
why going through the double step?

alecandido (commented):
It's the first time I look in this protocol's code, so some of my doubts may be a bit naive.
However, I tried to comment just on the diff.

In any case, the update itself is pretty much fine (though I believe both $\chi$ and $\Delta$ have a sign, but it doesn't matter for $g$, so here is correct).
Maybe it could have been worth to proliferate PRs, and split the plot refactor from the update itself. But definitely not a big deal.

Edoardo-Pedicillo:
In the Qibocal implementation of the dispersive shift two actions happen:
- evaluating the best ro frequency
- fit the signals and evaluate the coupling.

I have decided to make a refactoring to disentangle the two actions and to solve the failing issue more cleanly.

In the new implementation, there is no difference between the results of state zero and one, the frequencies and the fit parameters are saved in dictionaries with qubit as a key and the values of the two states saved in a list.

Every time the fit of one Lorentzian fails, the corresponding dictionary value associated with that qubit is `None`, and in this case, the coupling is not evaluated and only the best frequency is displayed in the report.

http://login.qrccluster.com:9000/Y-N5Wo-nQYm9C5nsBct_jw==/

src/qibocal/protocols/dispersive_shift.py
```diff
@@ -35,24 +35,17 @@ class DispersiveShiftParameters(Parameters):
...
     """Readout frequency that maximizes the distance of ground and excited states in iq-plane"""
 
-    @property
-    def state_zero(self):
-        return {key: value for key, value in asdict(self).items() if "zero" in key}
-
-    @property
-    def state_one(self):
-        return {key: value for key, value in asdict(self).items() if "one" in key}
+    def chi(self, target):
```
andrea-pasquale, line 45:
```suggestion
    def chi(self, target: QubitId) -> float:
```

alecandido (approved):
Thanks @Edoardo-Pedicillo!

PR #990, opened 2024-10-01
Update to qibolab 0.2

stavros11:
I am starting this by squashing the commits of #809, because many of them was keeping up-to-date with the many API changes during the qibolab 0.2 development and these are not very relevant for the final commit history.

The update requires roughly the following steps:
- [x] Update the `_acquisition` in all routines (list of routines below).
- [x] Update the methods that update the `Platform` after each routine. This should mainly affect the update.py file and not the routines themselves.
- [x] Deal with parameters that are no longer in the qibolab `Platform`. These are not required for execution, but usually calculated/updated by qibocal (T1, T2, fidelities, etc.). #1021 
- [x] Fix pylint and tests. Most like this has to come after the first two (or at least the first), because currently even just the `PulseSequence` manipulation is broken almost everywhere.
- [x] Fix any direct references to qibolab, which are based on the 0.1 API, in the docs.

This update should not affect the qibocal interface. I called the branch `0.2` because I assumed that we will eventually release it as such, in order to maintain agreement between qibolab-qibocal versions, but this is not strictly required if we don't break the qibocal interface (we could keep going with qibocal 0.1.*). I could change the branch name if it is confusing.

@Edoardo-Pedicillo mentioned that a potential qibocal refactoring was being discussed, however I would suggest to keep the two independent unless there is an important reason not to. If you want the "refactored qibocal" to support qibolab 0.1, you need to do this refactor first, however based on the discussed priorities, I think we should go ahead with 0.2 and forget 0.1 as soon as possible.

To elaborate a bit more on the above points:
(1) requires the most work, however it is also the tedious kind of work, so there should be no significant blocking points. I will add below a list of the routines, to split them among the people involved and keep track what is updated. In every routine we need to update the `PulseSequence` creation, how the native pulses are obtained from the `Platform`, how `Sweeper`s are defined and potentially how results are obtained and post-processed. Qibolab devs could potentially give a tutorial with more details. Some routines are already updated (see below) so they could serve as template.

(2) may be a bit trickier because, unlike (1), we don't have a POC yet and qibolab 0.2 is not very user-friendly when it comes to `Platform` parameter updates. I am planning to open another PR (on top of this) with a POC to discuss and once this is finalized we can propagate to all updates. The good news is that this can also progress independently of point 1.

(3) is not strictly required for things to work and the simplest solution is to just drop all the updates of such parameters. If users want to report them, they can refer to the corresponding reports, until we come up with a better way to store them. However, alternative solutions are welcome (@andrea-pasquale @alecandido ).

Points 4 and 5 should come naturally as 1 and 2 are completed.

<details>

<summary>List of routines to be updated (taken from `__init__` ):</summary>

- [x] "allxy",
<del> - [ ] "allxy_drag_pulse_tuning", </del> (because it could be just a qibocal script)
- [x] "single_shot_classification",
- [x] "spin_echo" (saving single shots not tested),
- [x] "spin_echo_signal" (saving single shots not tested),
- [x] "t1" (saving single shots not tested),
- [x] <del> "t1_sequences", https://github.com/qiboteam/qibocal/pull/1036 </del>(we can drop it since all drivers in 0.2 should support sweepers over duration)
- [x] "t1_signal" (saving single shots not tested),
- [x] "t2", https://github.com/qiboteam/qibocal/pull/1036
- [x] <del> "t2_sequences", https://github.com/qiboteam/qibocal/pull/1036 </del>(we can drop it since all drivers in 0.2 should support sweepers over duration)
- [x] "t2_signal", https://github.com/qiboteam/qibocal/pull/1036
- [x] "zeno", https://github.com/qiboteam/qibocal/pull/1036
- [x] <del>"zeno_signal", https://github.com/qiboteam/qibocal/pull/1036 </del> (we can drop it since the version with probability should be enough) 
- [x] <del> "coupler_chevron" </del> #1039 ,
- [x] <del>  "coupler_qubit_spectroscopy"</del> #1039 ,
- [x] <del> "coupler_resonator_spectroscopy"</del> #1039 ,
- [x] "dispersive_shift", https://github.com/qiboteam/qibocal/pull/1036
- [x] "dispersive_shift_qutrit", https://github.com/qiboteam/qibocal/pull/1036
- [x] "drag_tuning", https://github.com/qiboteam/qibocal/pull/1036
- [x] <del> "fast_reset", </del>(we can drop it since no driver in 0.2 supports fast reset)
- [x] "flipping", https://github.com/qiboteam/qibocal/pull/1036
- [x] <del>"flipping_signal", https://github.com/qiboteam/qibocal/pull/1036 </del>
- [x] <del>"avoided_crossing", </del> (we can drop it since it will be replaced #1037)
- [x] "qubit_crosstalk",https://github.com/qiboteam/qibocal/pull/1036
- [x] "qubit_flux" (see https://github.com/qiboteam/qibocal/pull/990#issuecomment-2407367608),
- [x] <del>"qubit_flux_tracking", </del>(we can drop it temporarily since it performs correction of the order of kHz)
- [x] <del>"resonator_crosstalk", </del> (we can drop it since we can evaluate the crosstalk matrix just with qubit crosstalk)
- [x] "resonator_flux" (see https://github.com/qiboteam/qibocal/pull/990#issuecomment-2407367608),
- [x] "qubit_spectroscopy",
- [x] "qubit_spectroscopy_ef", https://github.com/qiboteam/qibocal/pull/1036
- [x] "qutrit_classification", https://github.com/qiboteam/qibocal/pull/1036
- [x] "rabi_amplitude",
- [x] "rabi_amplitude_signal",
- [x] "rabi_length",
- [x] "rabi_amplitude_ef", https://github.com/qiboteam/qibocal/pull/1036
- [x] <del>"rabi_length_sequences", </del>(to be dropped see other sequences experiments)
- [x] "rabi_length_signal",
- [x] "ramsey" (unrolling not tested),
- [x] "ramsey_signal" (unrolling not tested),
- [x] "filtered_rb", https://github.com/qiboteam/qibocal/pull/1036
- [x] "standard_rb", https://github.com/qiboteam/qibocal/pull/1036
- [x] "readout_characterization", https://github.com/qiboteam/qibocal/pull/1036
- [x] "readout_mitigation_matrix", https://github.com/qiboteam/qibocal/pull/1036
- [x] "resonator_amplitude", https://github.com/qiboteam/qibocal/pull/1036
- [x] "resonator_frequency", https://github.com/qiboteam/qibocal/pull/1036 (to be dropped since to evaluate the best frequency the dispersive shift protocol should enough)
- [x] <del>"twpa_frequency", </del>(to be converted into a script) 
- [x] <del> "twpa_frequency_power", </del>(to be converted into a script) 
- [x] <del>"twpa_frequency_snr", </del>(to be converted into a script)
- [x] <del>"twpa_power", </del>(to be converted into a script) 
- [x] <del>"twpa_power_snr", </del>(to be converted into a script)
- [x] "resonator_punchout",
- [x] <del> "resonator_punchout_attenuation", </del> (to be reintroduced after qblox driver is ready for 0.2)
- [x] "resonator_spectroscopy",
- [x] "calibrate_state_discrimination", https://github.com/qiboteam/qibocal/pull/1036
- [x] "time_of_flight_readout", https://github.com/qiboteam/qibocal/pull/1036
- [x] "chevron",
- [x] "chevron_signal",
- [x] <del>"chsh_circuits", </del>  (we will add it later)
- [x] <del>"chsh_pulses",  </del> (we will add it later)
- [x] "correct_virtual_z_phases",
- [x] <del>"correct_virtual_z_phases_signal"</del>, (to be removed)
- [x] "state_tomography", https://github.com/qiboteam/qibocal/pull/1036
- [x] "allxy_resonator_depletion_tuning", (to be discussed)
- [x] "two_qubit_state_tomography", https://github.com/qiboteam/qibocal/pull/1036
- [x] "qubit_power_spectroscopy", https://github.com/qiboteam/qibocal/pull/1036
- [x] "rabi_amplitude_frequency",  https://github.com/qiboteam/qibocal/pull/1036
- [x] "rabi_amplitude_frequency_signal", https://github.com/qiboteam/qibocal/pull/1036
- [x] "rabi_length_frequency", https://github.com/qiboteam/qibocal/pull/1036
- [x] "rabi_length_frequency_signal", https://github.com/qiboteam/qibocal/pull/1036
- [x] "standard_rb_2q", https://github.com/qiboteam/qibocal/pull/1036
- [x] "standard_rb_2q_inter", https://github.com/qiboteam/qibocal/pull/1036
- [x] "optimize_two_qubit_gate", https://github.com/qiboteam/qibocal/pull/1036

</details>


stavros11:
Note that testing updated routines on hardware also requires a working platform that complies with the qibolab 0.2 interface. As of now the only available is https://github.com/qiboteam/qibolab_platforms_qrc/pull/153.

src/qibocal/protocols/resonator_spectroscopy.py
```diff
@@ -184,35 +189,42 @@ def _acquisition(
...
-    for qubit in targets:
-        ro_pulses[qubit] = platform.create_qubit_readout_pulse(qubit, start=0)
+    for q in targets:
+        natives = platform.natives.single_qubit[q]
+        ro_sequence = natives.MZ.create_sequence()
+        ro_pulses[q] = ro_sequence[0][1]
 
         if params.amplitude is not None:
-            ro_pulses[qubit].amplitude = params.amplitude
+            ro_pulses[q].amplitude = params.amplitude
```
sorewachigauyo, line 198: I was asked to try out the 0.2 branch, and this routine was marked as updated.
This currently fails because of the frozen pulse parameters.

alecandido: I guess this has not been tested setting the `amplitude` parameter. But thanks for checking it, most likely the best course of action is just unmarking it, and check all parameters options.

stavros11: Indeed, thanks for pointing this out @sorewachigauyo (and sorry for not mentioning earlier). I have not used the option to overwrite default pulse parameters through the routine while testing. The same issue should exist when trying to overwrite duration/amplitude in Rabi amplitude/length respectively. You should still be able to use these routines if you don't ovewrite (just put `amplitude: null` in the action runcard), which will use the default amplitudes in your parameters.json.

I will soon fix this using the proper pydantic `replace`.

stavros11: Now this functionality should be fixed in 25dddc171c76128361d294059bbd8189f6fec8c9 and 47d97daa5eaf4544d57744382dda0f580e6724a9. However, I was only able to run using the dummy, because our QM does not seem online right now, so I am not 100% sure that everything works properly.

sorewachigauyo: I tested this to be working and sent you the results on matrix.

stavros11:
In 4154ea914fe8a6928f86b965cc6ff6076d5d6e35 I updated the resonator and qubit flux routines. Here are some examples on 

* qw11q: http://login.qrccluster.com:9000/zCADyOH9REGlihNzbs5wVQ==, http://login.qrccluster.com:9000/thAfEvfURuK21lagxVlkMQ==
* qw5q_platinum: http://login.qrccluster.com:9000/wf3OKDQuTDiOxUVCoOrNag==, http://login.qrccluster.com:9000/-dW4I4e6SgeToDRh1IxNUw==

As mentioned elsewhere, these routines are making use of some parameters that are no longer in the qibocal platform, such as the crosstalk matrix and bare resonator frequency for the fit of resonator flux and the anharmonicity for the qubit flux of the 02 state. I have temporarily disabled/set dummy values in the related points, however we should probably find a solution on how to deal with these parameters before merging this.

stavros11:
I have updated most routines that are needed for basic calibration of a chip from scratch. See the spoiler in the first post for the list of updated routines. I have tested these on qw11q and they seem to work. If anyone tries to use this and encounters any issues or if some important routine is missing let us know. For TII, it should work with the platforms on qibolab_platforms_qrc main and the latest qibolab main (qw11q should also work with 0.2.0 release).

For the rest of routines I will probably need some help. I will now shift focus on #996 to implement the updates using the method discussed.

andrea-pasquale:
> I am starting this by squashing the commits of #809, because many of them was keeping up-to-date with the many API changes during the qibolab 0.2 development and these are not very relevant for the final commit history.
> 
> The update requires roughly the following steps:
> 
>     * [ ]  Update the `_acquisition` in all routines (list of routines below).
> 
>     * [ ]  Update the methods that update the `Platform` after each routine. This should mainly affect the update.py file and not the routines themselves.
> 
>     * [ ]  Deal with parameters that are no longer in the qibolab `Platform`. These are not required for execution, but usually calculated/updated by qibocal (T1, T2, fidelities, etc.). #1021 
> 
>     * [ ]  Fix pylint and tests. Most like this has to come after the first two (or at least the first), because currently even just the `PulseSequence` manipulation is broken almost everywhere.
> 
>     * [ ]  Fix any direct references to qibolab, which are based on the 0.1 API, in the docs.
> 
> 
> This update should not affect the qibocal interface. I called the branch `0.2` because I assumed that we will eventually release it as such, in order to maintain agreement between qibolab-qibocal versions, but this is not strictly required if we don't break the qibocal interface (we could keep going with qibocal 0.1.*). I could change the branch name if it is confusing.
> 
> @Edoardo-Pedicillo mentioned that a potential qibocal refactoring was being discussed, however I would suggest to keep the two independent unless there is an important reason not to. If you want the "refactored qibocal" to support qibolab 0.1, you need to do this refactor first, however based on the discussed priorities, I think we should go ahead with 0.2 and forget 0.1 as soon as possible.
> 
> To elaborate a bit more on the above points: (1) requires the most work, however it is also the tedious kind of work, so there should be no significant blocking points. I will add below a list of the routines, to split them among the people involved and keep track what is updated. In every routine we need to update the `PulseSequence` creation, how the native pulses are obtained from the `Platform`, how `Sweeper`s are defined and potentially how results are obtained and post-processed. Qibolab devs could potentially give a tutorial with more details. Some routines are already updated (see below) so they could serve as template.
> 
> (2) may be a bit trickier because, unlike (1), we don't have a POC yet and qibolab 0.2 is not very user-friendly when it comes to `Platform` parameter updates. I am planning to open another PR (on top of this) with a POC to discuss and once this is finalized we can propagate to all updates. The good news is that this can also progress independently of point 1.
> 
> (3) is not strictly required for things to work and the simplest solution is to just drop all the updates of such parameters. If users want to report them, they can refer to the corresponding reports, until we come up with a better way to store them. However, alternative solutions are welcome (@andrea-pasquale @alecandido ).
> 
> Points 4 and 5 should come naturally as 1 and 2 are completed.
> List of routines to be updated (taken from `__init__` ):

alecandido:
Whenever someone can take a look, please consider merging/rebasing on `main` to solve the conflicts (even discarding everything, if they do not apply to the new version).

andrea-pasquale:
@alecandido @Edoardo-Pedicillo @stavros11 I think we should be more or less ready to merge. Only coverage is failing which at this point we could fix in another PR. I know that it will be pretty difficult too review the PR given the amount of files changes. I tested most of the things on hardware so I am not expecting to find huge problems.

@hay-k @HishamKazim eventually you could also start testing this (if you have time) given that you guys are mostly working on calibration.

andrea-pasquale (approved):

stavros11:
I have done some comparisons between 0.1 and 0.2 (using this branch) on qw11q:
* Single shot: http://login.qrccluster.com:9000/6Iisf1W1T1KrStCypqIR-Q==, http://login.qrccluster.com:9000/iQCjykMTSOOHd8Jq7knHtA==
* Rabi amplitude: http://login.qrccluster.com:9000/wAeqh9o8Q0ebEfbTslCzZQ==/, http://login.qrccluster.com:9000/2P5wVGx6TV6k0Aw9kAcfUQ==/
* T1: http://login.qrccluster.com:9000/fJ6Qi3ABTruVIOgCe4d78w==, http://login.qrccluster.com:9000/3Sd_pNOwSCes-qIuwOQgHA==
* Ramsey detuned: http://login.qrccluster.com:9000/OR3tXHH_T62xSM6orw8AaQ==, http://login.qrccluster.com:9000/1-QRamA7TJSiMTpqaIbYGg==
* AllXY: http://login.qrccluster.com:9000/C9djroZ8QJWrDQOLtftbuQ==/, http://login.qrccluster.com:9000/TyMHnbHoRpSsY2Iic3n_zw==
* Chevron: http://login.qrccluster.com:9000/uxAuiz8wTzC7b0W1D9fS1Q==, http://login.qrccluster.com:9000/aMzhVDeoQ722XyU9jbthZw==
* CZ phases: http://login.qrccluster.com:9000/jgnsuJs9Sp-Bh8fwPVWbeQ==, http://login.qrccluster.com:9000/jvzUSBsXRaOk0tuxTbM7dg==
* CHSH: http://login.qrccluster.com:9000/jCZJNfrHQiCrlrOC8n342Q==, http://login.qrccluster.com:9000/tjSYbmTSS6O4d7rl3TtlNg==/

These tests were done using the qw11q platform from the `0.1` branch of qibolab and the corresponding converted platform for 0.2 (which can be found in https://github.com/qiboteam/qibolab_platforms_qrc/pull/205).

Most routines I tried look similar among the two versions. There are known discrepancies on
* Rabi length with interpolated sweepers (to be fixed in driver),
* Ramsey detuned (see above - not investigated yet),
* CHSH (partly known issue that will be fixed in the driver). Also, this routine is not part of this PR so I would postpone the discussion for #1049.

Note that these tests are an in a small subset of routines. However, I would still ask others to try using 0.2 so that we can identify other problems, if there are any. Maybe merging this could be a good step towards that direction.

src/qibocal/protocols/two_qubit_interaction/chevron/chevron.py
```diff
@@ -169,41 +143,46 @@ def _aquisition(
...
         )
-        data.sweetspot[ordered_pair] = platform.qubits[ordered_pair[1]].sweetspot
         sweeper_duration = Sweeper(
-            Parameter.duration,
-            params.duration_range,
-            pulses=[sequence.get_qubit_pulses(ordered_pair[1]).qf_pulses[0]],
-            type=SweeperType.ABSOLUTE,
+            parameter=Parameter.duration,
+            range=(params.duration_min, params.duration_max, params.duration_step),
+            pulses=flux_pulses + [delay_low, delay_high],
```
stavros11, line 164: After a discussion with @aorgazf, I believe there is a mistake here (most likely from my side) because the `delay_low` and `delay_high` should also contain the delays for the initial drive pulses so we cannot sweep them together with the flux pulse. The proper solution would be to add separate constant delays on the acquisition channel to compenstate for the drives and then another delay that is swept here together with the flux pulse.

I can fix it (it may require small refactor), but I would also like to investigate in more detail what happens, because I am pretty sure that I ran this routine before on hardware and got similar results to 0.1. Has anyone else (@andrea-pasquale ) tried that routine?

andrea-pasquale: > After a discussion with @aorgazf, I believe there is a mistake here (most likely from my side) because the `delay_low` and `delay_high` should also contain the delays for the initial drive pulses so we cannot sweep them together with the flux pulse. The proper solution would be to add separate constant delays on the acquisition channel to compenstate for the drives and then another delay that is swept here together with the flux pulse.

I agree with the correction.



> I can fix it (it may require small refactor), but I would also like to investigate in more detail what happens, because I am pretty sure that I ran this routine before on hardware and got similar results to 0.1. Has anyone else (@andrea-pasquale ) tried that routine?

I've not tried it recently and I also didn't notice a significant difference. I can test it again.

andrea-pasquale: I've implemented the correct pulse sequence for the chevrons in https://github.com/qiboteam/qibocal/pull/990/commits/3ef4f92a6c3de8d37c6fde1b6699377e5d1f3cc5

src/qibocal/protocols/two_qubit_interaction/chevron/utils.py
```diff
@@ -14,68 +16,70 @@
...
+        (drive_channel, second_rx),
+    ]
 
-    sequence.add(measure_lowfreq)
-    sequence.add(measure_highfreq)
+    # add readout
+    sequence += low_natives.MZ() + high_natives.MZ()
 
-    return sequence
+    return sequence, flux_pulse, [ro_low_delay, ro_high_delay]
```
stavros11, line 82: Thanks @andrea-pasquale for updating the Chevron and sorry I almost forgot I was supposed to fix that (and the delay in reviewing).

Shouldn't we also return `drive_delay` here, as we need to sweep this together with the readout delays, right? This is for the drive pulse that plays after the flux pulses (if I am reading this right).

Also, I am not fully convinced that parking will work properly. In particular, I see that the parking pulses don't participate in the duration sweeper, however, I would expect that if we are stretching the CZ flux pulse, we also need to stretch any parking pulses accordingly. In any case, I think my original implementation was ignoring parking altogether, possibly because it was not very trivial to do and I didn't want to spend too much time, however we should probably have a look if we are actually interested in using it.

I did not test all this on hardware, I just looked over the code. I am also a bit concerned on how these `dt` delays will be handled by the QM driver, but if there are issues there they should be addressed in qibolab, not here.

andrea-pasquale: @stavros11 yes, you are correct we should return also the `drive_delay`. The motivation for this extra pulse is that the high frequency qubit is oscillating between $\ket{1}$ and $\ket{2}$, however since our classification doesn't take into account state $\ket{2}$ I don't know if we can trust exactly the probability measured. Some references (see section 5.8 [here](https://web.physics.ucsb.edu/~martinisgroup/theses/Chen2018.pdf)) suggest to play an extra $RX(\pi)$ pulse such that we can map $\ket{1} \rightarrow \ket{0}$ to get a better estimate for the probability of $\ket{0}$ (given that applying a $RX(\pi)$ should not modify the qubit).


> Also, I am not fully convinced that parking will work properly. In particular, I see that the parking pulses don't participate in the duration sweeper, however, I would expect that if we are stretching the CZ flux pulse, we also need to stretch any parking pulses accordingly. In any case, I think my original implementation was ignoring parking altogether, possibly because it was not very trivial to do and I didn't want to spend too much time, however we should probably have a look if we are actually interested in using it.

Since I was modifying the code I took a look also at the parking quickly (particulary because we will need it for the B3-B4 pair). I believe that if I stretch the parking pulse as you suggested it should work as expected.

andrea-pasquale: I think parking now is "working" here is an example with http://login.qrccluster.com:9000/DUbmlfLBQXO6GbTxfVrP3A==/ and without http://login.qrccluster.com:9000/xJivo9O1S1eRf_cwyzFrwQ==/.
I'm still not sure about the other signal which I am seeing above the resonance...

stavros11: Thanks for the updates and testing. Now it looks good to me, assuming that the parking pulses in `raw_flux_sequence` only play on channels of other qubits, not the ones already participating in the chevron (as low/high frequency, because then there may be issues with the pulse scheduling), but I think that's a fair assumption.

Edoardo-Pedicillo (approved):

PR #993, opened 2024-10-02
Replace qq auto with qq run

Edoardo-Pedicillo:
This choice comes from some internal discussions, the most recent one is https://github.com/qiboteam/xmind-diagrams/pull/31#discussion_r1773509461

Checklist:
- [ ] Tests are passing.
    - [x] Qibo: `master`
    - [x] Qibolab: `0.1`


Edoardo-Pedicillo:
@scarrazza do you think we can go on with this PR and update the Qibocal-paper before the submission?

scarrazza:
Yes, I agree.

alecandido (approved):

andrea-pasquale (approved):
Thanks @Edoardo-Pedicillo.
In theory we need to update the diagram that we have in the qibocal paper also here in the documenation, correct? Maybe it is worth waiting until the diagram is ready to merge the PR.

Also, perhaps we could add a log message when importing the library saying that `qq auto` has been renamed `qq run`? It could also makes sense to do it once we have a release I guess. I'm suggesting this to lower the confusion among users :)

alecandido:
> Also, perhaps we could add a log message when importing the library saying that `qq auto` has been renamed `qq run`? It could also makes sense to do it once we have a release I guess. I'm suggesting this to lower the confusion among users :)

I would exclude doing that at import time (it would appear also in any script not using the CLI), and I'm generally against warnings.

What I may support is to keep the `qq auto` subcommand, just to print a message telling the user to use `qq run` instead.

andrea-pasquale:
> What I may support is to keep the `qq auto` subcommand, just to print a message telling the user to use `qq run` instead.

This could work.

andrea-pasquale (approved):

PR #1002, opened 2024-10-08
Circuit serialization as parameters of the tomography

Edoardo-Pedicillo:
This PR is a patch to solve the serialization of Circuit parameters, like in state tomographies (this happens with the scripts). Maybe a better solution would be to define a dump method specific for each parameter class and call it directly in the action dump method.

Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [x] Qibolab: `0.1.9`

### DIFF
--- a/src/qibocal/auto/task.py
+++ b/src/qibocal/auto/task.py
@@ -46,9 +46,10 @@ class Action:
         if self.parameters is not None:
             for param, value in self.parameters.items():
                 if type(value) is Circuit:
-                    circuit_file = open(CIRCUIT, "w")
+                    circuit_path = path / CIRCUIT
+                    circuit_file = open(circuit_path, "w")
                     json.dump(value.raw, circuit_file)
-                    self.parameters[param] = str(CIRCUIT)
+                    self.parameters[param] = str(circuit_path)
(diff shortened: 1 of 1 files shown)


alecandido:
@Edoardo-Pedicillo we will eventually move to Pydantic serialization even in Qibocal (it's a waste of time to maintain our own, when there is a high-quality maintained library doing that).

If your solution is working, for the time being is more than enough.

src/qibocal/auto/task.py
```diff
@@ -40,6 +43,13 @@ class Action:
 
     def dump(self, path: Path):
         """Dump single action to yaml."""
+        if self.parameters is not None:
+            for param, value in self.parameters.items():
+                if type(value) is Circuit:
+                    circuit_path = path / CIRCUIT
+                    circuit_file = open(circuit_path, "w")
+                    json.dump(value.raw, circuit_file)
```
alecandido, line 51: Either use `with open(...) as fd:` or use `circuit_file.write_text(json.dumps(...))`. You should close file descriptors, and better doing it automatically than manually.

hay-k (approved):

stavros11 (approved):

PR #770, opened 2024-03-25
Mermin experiment

andrea-pasquale:
This PR adds a protocol for computing mermin inequality for 3 qubits (single value without sweep).

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.
- [ ] Compatibility with Qibo modules (Please edit this section if the current pull request is not compatible with the following branches).
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`
    - [ ] Qibolab_platforms_qrc: `main`

PR #1005, opened 2024-10-09
Generalizing Mermin to n qubits and adding theta angles for plot

igres26:
Push for the generalization of Mermin to any number of qubits and adding an angle theta that controls the violation.

This will override the other Mermin pull request (#770) when it is completed.

The model for this experiment is the CHSH protocol, as the output should be very similar: a plot with the Mermin value at different angles to track the violation.

I have based myself on those functions for this Mermin generalization, but I am sure there are things to be polished regarding pure qibocal code. Please @Edoardo-Pedicillo and @andrea-pasquale lend me a hand with that.


igres26:
I have fixed some of the issues, but I have not been able to iron out everything to conform with Qibocal. I think at this point I might need a hand cleaning things up properly.

Edoardo-Pedicillo:
After we have merged #1038, we can close this PR.

PR #1012, opened 2024-10-16
Protocols list

Edoardo-Pedicillo:
Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/doc/dev/README.md
+++ b/doc/dev/README.md
@@ -1,7 +1,7 @@
 # Qibocal Roadmap
 
 ## Protocols
-### <font color="green">High Priority</font>
+### High Priority
 #### Single qubit
 
 - Rabi and flipping to calibrate pi/2 rotations (https://web.physics.ucsb.edu/~martinisgroup/theses/Chen2018.pdf par. 5.9)
@@ -23,7 +23,7 @@
 - Measure ZZ coupling in couplers (flux amplitude vs coupling) (Manenti Motta, par. 14.8.4)
 - Improve and test iSWAP implementation
 
-### <font color="green">Low Priority</font>
+### Low Priority
 #### Single Qubit
 
 - Calibrate the other qubit states
--- a/doc/dev/README.md
+++ b/doc/dev/README.md
@@ -1,0 +1,38 @@
+# Qibocal Roadmap
+
+## Protocols
+### <font color="green">High Priority</font>
+#### Single qubit
+
+- Rabi and flipping to calibrate pi/2 rotations (https://web.physics.ucsb.edu/~martinisgroup/theses/Chen2018.pdf par. 5.9)
+- Kernels for integration
+- Drag with detuning  (https://arxiv.org/pdf/1904.06560)
+- Cross-entropy benchmarking
+- SNR in spectroscopies and SNR in the IQ plane
+  https://dsp.stackexchange.com/questions/24372/what-is-the-connection-between-analog-signal-to-noise-ratio-and-signal-to-noise.
+  https://arxiv.org/pdf/2106.06173.pdf
+- Improve readout amplitude optimization with the outliers probability
+(https://escholarship.org/content/qt0g29b4p0/qt0g29b4p0.pdf?t=prk0gj)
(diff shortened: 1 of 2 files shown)


doc/dev/README.md
```diff
@@ -0,0 +1,38 @@
+# Qibocal Roadmap
+
+## Protocols
```
alecandido, line 3: Are you planning to include other sections in the roadmap? (other than *protocols*)

doc/dev/README.md
```diff
@@ -0,0 +1,38 @@
...
+- Calibrate the other qubit states
+- Carr-Purcell-Meiboom_Gill sequence
+- Explore cosine pulse as X pulse (https://web.physics.ucsb.edu/~martinisgroup/theses/Chen2018.pdf par. 5.4)
+- Active reset
+- Measurement tomography https://arxiv.org/pdf/1310.6448.pdf
+- XY-Z timing
+(https://escholarship.org/content/qt0g29b4p0/qt0g29b4p0.pdf?t=prk0gj par. 5.10)
+- Optimal control with randomize benchmarking (https://arxiv.org/pdf/1403.0035)
+- Quantum volume
+- Gate Set Tomography
```
alecandido, line 38: Speaking of MM. Should we also add other possible implementations of Z rotations? (14.6.4)

As low priority, of course.

Not much because they may be useful as gates, but just because they are possible, and they may be used as further checks on the qubit, since they are playing with the same parameters.

alecandido, line 38: Is there a reason to completely exclude the process tomography?

doc/dev/README.md
```diff
@@ -0,0 +1,38 @@
...
+#### Single Qubit
+
+- Calibrate the other qubit states
+- Carr-Purcell-Meiboom_Gill sequence
+- Explore cosine pulse as X pulse (https://web.physics.ucsb.edu/~martinisgroup/theses/Chen2018.pdf par. 5.4)
+- Active reset
+- Measurement tomography https://arxiv.org/pdf/1310.6448.pdf
+- XY-Z timing
+(https://escholarship.org/content/qt0g29b4p0/qt0g29b4p0.pdf?t=prk0gj par. 5.10)
+- Optimal control with randomize benchmarking (https://arxiv.org/pdf/1403.0035)
```
andrea-pasquale, line 36: Perhaps this one can be moved to high priority given that we are more or less working on it already?

andrea-pasquale (approved):

doc/dev/README.md
alecandido: Agreed.

Though, I'm not sure I would classify that as a protocol at all...
Instead, I would build as a protocols' composition with scripts. And, as part of this effort, I could finally work out a way to meaningfully distribute even this kind of pre-compiled applications with Qibocal, or in parallel to it (but within the repo).

In any case, this is just an opinion, and up for discussion.

(there will be an acquisition, a fit, and a report - but the acquisition and the fit won't be really two separate phases, that's why I'd not make it a protocol itself)

doc/dev/README.md
Edoardo-Pedicillo: I would peronally avoid them since they don't add information, maybe the only one to take into consideration is the one using the AC Stark effect to study the phenomenon itself.

Edoardo-Pedicillo: No, just forgotten

alecandido: To be fair, that's the one I mostly had in mind...
(the virtual Z we already had, and the composition of other X and Y gates requires a calibration of the Y for arbitrary angles...)

Maybe, the other one I would have considered is fluxing the qubit to measure the Z rotation. As a (redundant?) way to show control over the flux.

In any case, if interesting, we could even reconsider at a later stage.

alecandido (approved):

doc/dev/README.md
alecandido: Of course, we could even try to schedule some more relevant software feature for the future.

However, software-wise we're really moving on a need-based approach (mainly because of the lack of person power), so I'd not be so systematic in this respect, since it may result completely useless...

PR #1017, opened 2024-10-18
In VZ signal experiment sweep phase as absolute, instead of relative to match the non-signal experiment

### LINKED ISSUE #1004: Results from virtual_z_phases and virtual_z_phases_signal do not match
virtual_z_phases: http://login.qrccluster.com:9000/3dRNAIU0Tgy5ebT4-7Bvog==
virtual_z_phases_signal: http://login.qrccluster.com:9000/oh4sU7f5QvapoYaqGjzoHg==/

Note that the calculated virtual phases are very different.

I believe the difference stems from the difference in these two lines:

https://github.com/qiboteam/qibocal/blob/6789ac85b096310087415369f04cad9b88eac717/src/qibocal/protocols/two_qubit_interaction/virtual_z_phases_signal.py#L102

https://github.com/qiboteam/qibocal/blob/6789ac85b096310087415369f04cad9b88eac717/src/qibocal/protocols/two_qubit_interaction/virtual_z_phases.py#L249

Changing the line in _signal to match the other one, I get matching results indeed: http://login.qrccluster.com:9000/iSWSGTFcSAKlvWGOrDFA1g==

hay-k:
resolves #1004

### DIFF
--- a/src/qibocal/protocols/two_qubit_interaction/virtual_z_phases_signal.py
+++ b/src/qibocal/protocols/two_qubit_interaction/virtual_z_phases_signal.py
@@ -99,7 +99,7 @@ def _acquisition(
                 )
                 sweeper = Sweeper(
                     Parameter.relative_phase,
-                    theta - data.vphases[ord_pair][target_q],
+                    theta,
                     pulses=[theta_pulse],
                     type=SweeperType.ABSOLUTE,
                 )


igres26 (approved):
Looks good to me. Both should be the same. Thanks!

Edoardo-Pedicillo (dismissed):

igres26:
Actually, could we do it the other way around? Change the the other routine to match with the signal one? The point of this is that after setting the virtual phase, you want to see both signals starting at the same point.

Edoardo-Pedicillo:
> Actually, could we do it the other way around? Change the the other routine to match with the signal one? The point of this is that after setting the virtual phase, you want to see both signals starting at the same point.

I prefer the current implementation (just `theta`) because it does not hide non-documented post-processing. 
I am also not sure the sign is correct: if theta = 0, I am expecting the correction phase to be the one in the platform and
not the opposite.

Edoardo-Pedicillo:
> Actually, could we do it the other way around? Change the the other routine to match with the signal one? The point of this is that after setting the virtual phase, you want to see both signals starting at the same point.

I prefer the current implementation (just `theta`) because it does not hide non-documented post-processing. 
I am also not sure the sign is correct: if theta = 0, I am expecting the correction phase to be the one in the platform and
not the opposite.

andrea-pasquale:
> > Actually, could we do it the other way around? Change the the other routine to match with the signal one? The point of this is that after setting the virtual phase, you want to see both signals starting at the same point.
> 
> I prefer the current implementation (just `theta`) because it does not hide non-documented post-processing. I am also not sure the sign is correct: if theta = 0, I am expecting the correction phase to be the one in the platform and not the opposite.

Indeed, to be more clear we can draw a vertical line where the virtual phase is supposed to be.

igres26:
Okay, the vertical line might add clarity. I would also encourage to keep the virtual phases modulo 2\pi then, as otherwise they might not appear on the graph.

hay-k:
Somehow visualizing the calculated angles (be it a vertical line, or other means) sounds good. Using angles modulo 2*pi if it is not the case already - completely agree. Since these are irrelevant to the problem that this PR is addressing, let's create separate issue(s) for these requests and address in them in a separate PR.

Edoardo-Pedicillo (approved):

PR #1022, opened 2024-10-23
Compatibility with the new Qibo global backend handling

alecandido:
https://github.com/qiboteam/qibo/pull/1440#issuecomment-2431599631

Qibocal is not directly broken, until the next Qibo release.

Still, we want to have a strategy ready for when this release will happen (soon), and it will break Qibolab itself.
Qibo has no well-defined public API, and so this breaking change will happen in a minor version - thus leading to breakage if installed with the latest release.

In any case, all Qibolab users are invited to keep using Qibo v0.2.12 until the next Qibocal release, if they use a release. Otherwise, hopefully this will be soon ready to be merged in https://github.com/qiboteam/qibocal/tree/main


alecandido:
@igres26, if you already tested and got an error, try to take a look at this branch, and please if there is any further error being raised.

I'm also scanning and checking other possible usages that are prone to fail because of https://github.com/qiboteam/qibo/pull/1440.
I will even properly test later on, but for the time being, it was just a quick fix in case someone needs to use Qibocal's and Qibolab's development versions (which I hope no one is doing, but I don't know).

andrea-pasquale:
@alecandido I believe that we can close this since this problem got fixed by #1076

PR #1025, opened 2024-10-24
Remove onnx dep

Edoardo-Pedicillo:
Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -323,23 +323,6 @@ files = [
 ]
 
 [[package]]
-name = "coloredlogs"
-version = "15.0.1"
-description = "Colored terminal output for Python's logging module"
-optional = true
-python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
-files = [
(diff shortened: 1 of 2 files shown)


alecandido (commented):
https://github.com/qiboteam/qibocal/blob/6122a55a6ce77644bdf4274fbced510fc64c1486/src/qibocal/fitting/classifier/scikit_utils.py#L4

?

alecandido (approved):
Do you remember why we have both a qubit and Qblox fit?

The Qblox fit seems the simpler version of the qubit one, which is also doing fewer things (but most likely could be simplified a bit anyhow).

I'm not asking to do anything in this PR, or urgently. But if it's redundant we may have an issue for their deduplication.

Edoardo-Pedicillo:
> Do you remember why we have both a qubit and Qblox fit?

It came out after a discussion with the lab people to benchmark with the qubit fit method. For sure we can remove it, I will open an issue

hay-k (approved):

PR #1035, opened 2024-10-30
Fix 2q tomography

### LINKED ISSUE #1032: Two qubit state tomography reporting incorrect density matrix
In the report of this protocol, this matrix is showed

https://github.com/qiboteam/qibocal/blob/fa0467c73b7d91e23c88b944ba5e651393c3a0da/src/qibocal/protocols/two_qubit_state_tomography.py#L298-L300

After discussing with @stavros11, these lines should be replaced by
```py

 measured = np.array(fit.measured_density_matrix_real[target]) + 1j * np.array( 
     fit.measured_density_matrix_imag[target] 
 ) 

```

stavros11:
Fixes #1032. I have not tested how this affect results from hardware, but it is quite clear that the fix is needed (otherwise the real and imaginary part are not compatible).

I also updated the docstring of the `_fit` method with more clear descriptions of the process steps, instead of the wikipedia link, which I think was more confusing than helping.

### DIFF
--- a/src/qibocal/protocols/two_qubit_state_tomography.py
+++ b/src/qibocal/protocols/two_qubit_state_tomography.py
@@ -179,11 +179,18 @@ def project_psd(matrix):
 def _fit(data: StateTomographyData) -> StateTomographyResults:
     """Post-processing for two qubit state tomography.
 
-    Uses the standard linear inversion algorithm described in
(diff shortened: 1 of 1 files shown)


andrea-pasquale (approved):
Thanks @stavros11 , LGTM.

Edoardo-Pedicillo (approved):

PR #1041, opened 2024-11-15
Cryoscope for QM

stavros11:
Since it could be useful for the two-qubit calibration of qw11q (@hay-k ), this ports the cryoscope routine from qua-libs https://github.com/qua-platform/qua-libs/tree/main/Quantum-Control-Applications/Superconducting/Two-Flux-Tunable-Coupled-Transmons/Use%20Case%201%20-%20Two%20qubit%20gate%20optimization%20with%20cryoscope 
similarly to the RB of #959, so that it can be executed as a qibocal routine. 

Here is an example action runcard:
```yaml
platform: qw11q

  - id: cryoscope
    operation: cryoscope_qua
    targets: ["D2"]
    parameters:
      flux_amplitude: 0.1
      const_flux_len: 160
      zeros_before_pulse: 20
      zeros_after_pulse: 20
      other_qubits: ["D1"]  # will set the bias of these qubits to 0
```

and example execution: http://login.qrccluster.com:9000/ZSRrpJzVRK6w-ZGnOFat1A==/
The FIR is quite close to the filter we are actually using for D2: https://github.com/qiboteam/qibolab_platforms_qrc/blob/de4b358b79470d4efdff956e1d25fa9b7dd506c3/qw11q/parameters.json#L181-L189, however the IIR has opposite sign, so I am not sure if it is actually a good fit.

Note that this routine should preferably be executed without any previous filters in the runcard. Also, remember to recalibrate the classification parameters after removing the filters.


src/qibocal/protocols/qua/cryoscope.py
```diff
@@ -0,0 +1,398 @@
...
+    :param tau: decay time of the exponential decay
+    :param Ts: sampling period. Default is 1e-9
+    :return: FIR and IIR taps
+    """
+    tau = tau * Ts
+    k1 = Ts + 2 * tau * (A + 1)
+    k2 = Ts - 2 * tau * (A + 1)
+    c1 = Ts + 2 * tau
+    c2 = Ts - 2 * tau
+    feedback_tap = k2 / k1
```
andrea-pasquale, line 248: The minus sign in the IIR is triggered here. I believe that it is outdated in the qua-libs example.
In another repository [py-qua-tools](https://github.com/qua-platform/py-qua-tools) where there are some utilies regarding filters there is a minus sign:
https://github.com/qua-platform/py-qua-tools/blob/862bb161ec435dc0e4fe4025b7f4296c1e8b4153/qualang_tools/digital_filters/filters.py#L167

stavros11: Thanks! Then I guess that solves this issue (minus added 0b4b54286226ecaab466eb90764182857ea722bc) and the produced filters are close to the ones we were using. I am not sure if the difference is significant. Let's see if this produces any good filters for the B line.

Edoardo-Pedicillo:
Since we have implemented the cryoscope in Qibocal, I will close this PR

PR #1044, opened 2024-11-20
RX$\pi/2$ calibration implementation

ElStabilini:
Checklist:
- [ ] Tests are passing.
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`


andrea-pasquale:
Thanks @ElStabilini, the results seem to make sense since for both amplitude and duration there is a factor 2. I think that rather than changing the protocol completly we could consider adding a flag to specify whether we want to calibrate the RX or RX90 pulse. We should also check the corresponding [update](https://github.com/qiboteam/qibocal/blob/ec4b7ad5a3492605f50779740412af239d62cdaf/src/qibocal/protocols/rabi/amplitude.py#L142) function in the protocols to be consistent.
Eventually we can do the same for both the flipping and the drag calibration. For now let's finish first rabi.

src/qibocal/update.py
```diff
@@ -54,11 +54,16 @@ def drive_frequency(
...
-def drive_amplitude(amp: Union[float, tuple, list], platform: Platform, qubit: QubitId):
+def drive_amplitude(
+    amp: Union[float, tuple, list], pi_half: bool, platform: Platform, qubit: QubitId
+):
     """Update drive frequency value in platform for specific qubit."""
     if isinstance(amp, Iterable):
         amp = amp[0]
-    platform.update({f"native_gates.single_qubit.{qubit}.RX.0.1.amplitude": amp})
+    if pi_half:
+        platform.update({f"native_gates.single_qubit.{qubit}.RX90.0.1.amplitude": amp})
```
andrea-pasquale, line 64: I am not sure about this since in qibolab by default the RX90 is not supported as native I think. Perhaps a better solution is to just update the RX amplitude as twice the RX90 amplitude.

src/qibocal/protocols/rabi/length_signal.py
```diff
@@ -153,13 +160,17 @@ def _fit(data: RabiLengthSignalData) -> RabiLengthSignalResults:
...
+    )
 
 
 def _update(
     results: RabiLengthSignalResults, platform: CalibrationPlatform, target: QubitId
 ):
-    update.drive_duration(results.length[target], platform, target)
+    update.drive_duration(
+        results.length[target], results.pihalf_pulse, platform, target
+    )
```
andrea-pasquale, line 173: Also this function needs to be updated.

src/qibocal/protocols/rabi/length_frequency_signal.py
```diff
@@ -39,6 +39,8 @@ class RabiLengthFrequencySignalParameters(Parameters):
     """Frequency to use as step for the scan."""
     pulse_amplitude: Optional[float] = None
     """Pi pulse amplitude. Same for all qubits."""
+    pihalf_pulse: Optional[bool] = True
```
andrea-pasquale, line 42: Here I would rename it to something else, but I'm not sure how. Perhaps `calibrate_rx90`?

src/qibocal/update.py
ElStabilini: I should have just opened a PR in [qibolab](https://github.com/qiboteam/qibolab/pull/1103), I was waiting for access to the repo.

src/qibocal/protocols/rabi/length_signal.py
ElStabilini: I sould have fixed this in the new commit, thank you @andrea-pasquale

src/qibocal/protocols/rabi/length_frequency_signal.py
alecandido: In case, `rx90` should be sufficient :)

ElStabilini:
> Thanks @ElStabilini, the results seem to make sense since for both amplitude and duration there is a factor 2. I think that rather than changing the protocol completly we could consider adding a flag to specify whether we want to calibrate the RX or RX90 pulse. We should also check the corresponding [update](https://github.com/qiboteam/qibocal/blob/ec4b7ad5a3492605f50779740412af239d62cdaf/src/qibocal/protocols/rabi/amplitude.py#L142) function in the protocols to be consistent. Eventually we can do the same for both the flipping and the drag calibration. For now let's finish first rabi.

I see this only now, I should have done this by adding the pihalf_pulse (soon rx90) option, correct?
Regarding routines, should I change them even though I added a new RX90 native gate and left untouched the RX?

doc/source/protocols/rabi/rabi.rst
```diff
@@ -29,6 +29,9 @@ Rabi rate is larger than the decay and the pure dephasing rate,
 
 where :math:`\Omega_R` is the Rabi frequency and :math:`\tau` the decay time.
 
+In qibocal we implemented also another version of the Rabi experiment which can be used to tune the amplitude (duration) of the drive pulse in order
+to excite the qubit from the ground state up to state :math:`\frac{\ket{0}+\ket{1}}{\sqrt{2}}`.
```
andrea-pasquale, line 33: Given the definition of [RX](https://qibo.science/qibo/stable/api-reference/qibo.html#rotation-x-axis-rx) we are actually exciting to the state $\ket{-} = \frac{\ket{0}-i \ket{1}}{\sqrt{2}}$

src/qibocal/protocols/rabi/ef.py
```diff
@@ -124,8 +124,12 @@ def _update(
     results: RabiAmplitudeEFResults, platform: CalibrationPlatform, target: QubitId
 ):
     """Update RX2 amplitude_signal"""
-    update.drive_12_amplitude(results.amplitude[target], platform, target)
-    update.drive_12_duration(results.length[target], platform, target)
+    update.drive_12_amplitude(
+        results.amplitude[target], results.pihalf_pulse, platform, target
```
andrea-pasquale, line 128: Here we should force the user to calibrate just with RX, given that we have no RX1290 pulse.

doc/source/protocols/rabi/rabi.rst
```diff
@@ -103,6 +106,30 @@ It follows an example runcard and plot for the signal exepriment
...
+    - id: Rabi signal
+      operation: rabi_amplitude_signal
+      parameters:
+        min_amp: 0.2
+        max_amp: 1.
+        step_amp: 0.01
+        pulse_length: 40
+        nshots: 3000
+        relaxation_time: 50000
+	RX90: False
```
andrea-pasquale, line 128: This line needs to be fixed.

andrea-pasquale (commented):
Thanks for the updates @ElStabilini.
At this point I would rename everywhere `pihalf_pulse` to `rx90`.
Everything else looks good to me.

andrea-pasquale:
> Regarding routines, should I change them even though I added a new RX90 native gate and left untouched the RX?

Just to avoid too many things at the same time I suggest to focus first on rabi experiments. In another PR we can propagate these changes to the other protocols.

src/qibocal/protocols/rabi/amplitude_frequency_signal.py
```diff
@@ -52,6 +54,8 @@ class RabiAmplitudeFrequencySignalResults(RabiAmplitudeSignalResults):
 
     frequency: dict[QubitId, Union[float, list[float]]]
     """Drive frequency for each qubit."""
+    pihalf_pulse: bool
```
alecandido, line 57: Here and everywhere: I'd suggest adopting `rx90` as identifier, to be consistent with the parameter, since it will always contain exactly the same value. This will make more intuitive to understand the connection :)

doc/source/protocols/rabi/rabi.rst
ElStabilini: True, thank you for noticing, I should have fixed it.

doc/source/protocols/rabi/rabi.rst
ElStabilini: I should have fixed it and also added the figure

src/qibocal/protocols/rabi/amplitude_frequency_signal.py
ElStabilini: I changed it everywhere. Also changed `pulse : bool` in `rabi/utils.py` to `rx90: bool`

alecandido:
Most likely, the commit 3164fd8 is not required at all, for the reason mentioned in a [comment that should be published in https://github.com/qiboteam/qibocal/pull/990].

Feel free to delete or revert it :)

src/qibocal/protocols/rabi/length.py
```diff
@@ -31,14 +31,20 @@ class RabiLengthParameters(Parameters):
     """Step pi pulse duration [ns]."""
     pulse_amplitude: Optional[float] = None
     """Pi pulse amplitude. Same for all qubits."""
+    rx90: bool = False
+    """Calibration of native pi pulse, if true calibrates pi/2 pulse"""
     interpolated_sweeper: bool = False
     """Use real-time interpolation if supported by instruments."""
+    rx90: bool = False
+    """Calibration of native pi pulse, if true calibrates pi/2 pulse"""
```
andrea-pasquale, line 39:
```suggestion
```
`rx90` is already defined above.

src/qibocal/protocols/rabi/length.py
```diff
@@ -31,14 +31,20 @@ class RabiLengthParameters(Parameters):
...
 
 
 @dataclass
 class RabiLengthResults(RabiLengthSignalResults):
     """RabiLength outputs."""
 
+    rx90: bool
+    """Pi or Pi_half calibration"""
```
andrea-pasquale, line 47:
```suggestion
```
Already defined in `RabiLengthSignalResults`.

src/qibocal/protocols/rabi/length_frequency_signal.py
```diff
@@ -47,6 +49,8 @@ class RabiLengthFrequencySignalParameters(Parameters):
 class RabiLengthFrequencySignalResults(RabiLengthSignalResults):
     """RabiLengthFrequency outputs."""
 
+    rx90: bool
+    """Pi or Pi_half calibration"""
```
andrea-pasquale, line 53:
```suggestion
```
Already defined in `RabiLengthSignalResults`.

src/qibocal/protocols/rabi/length_frequency.py
```diff
@@ -33,6 +33,8 @@ class RabiLengthFrequencyParameters(RabiLengthFrequencySignalParameters):
 class RabiLengthFrequencyResults(RabiLengthFrequencySignalResults):
     """RabiLengthFrequency outputs."""
 
+    rx90: bool
+    """Pi or Pi_half calibration"""
```
andrea-pasquale, line 37:
```suggestion
```
Already defined in `RabiLengthFrequencySignalResults`

src/qibocal/protocols/rabi/length_frequency.py
```diff
@@ -51,6 +53,9 @@ class RabiLengthFrequencyResults(RabiLengthFrequencySignalResults):
 class RabiLengthFreqData(RabiLengthFreqSignalData):
     """RabiLengthFreq data acquisition."""
 
+    rx90: bool
+    """Pi or Pi_half calibration"""
+
```
andrea-pasquale, line 58:
```suggestion
```
Already defined in `RabiLengthFreqSignalData`

src/qibocal/update.py
```diff
@@ -54,22 +54,32 @@ def drive_frequency(
...
 
-def drive_amplitude(amp: Union[float, tuple, list], platform: Platform, qubit: QubitId):
+def drive_amplitude(
+    amp: Union[float, tuple, list], pi_half: bool, platform: Platform, qubit: QubitId
+):
     """Update drive frequency value in platform for specific qubit."""
     if isinstance(amp, Iterable):
         amp = amp[0]
-    platform.update({f"native_gates.single_qubit.{qubit}.RX.0.1.amplitude": amp})
+    if pi_half:
```
andrea-pasquale, line 63: At this point we can use `rx90` also here.

andrea-pasquale (commented):
Thanks @ElStabilini, I have a few comments (mainly some redundant `rx90` attributes).

We need to do a few more things before merging this PR.

1. Check why pylint is [failing](https://github.com/qiboteam/qibocal/actions/runs/11974556505/job/33385894293?pr=1044).
2. Add tests. In this case it should be sufficient to add a few entries [here](https://github.com/qiboteam/qibocal/blob/pi_half/tests/runcards/protocols.yml) with `rx90: True` for all protocols that you modified.
3. Tests might fail because the CI is installing the `main` branch of qibolab which doesn't contain your fix for RX90 https://github.com/qiboteam/qibolab/pull/1103. To test that everything works correctly you can modify the [`pyproject.toml` ](https://github.com/qiboteam/qibocal/blob/c28039847aad2869956f22ed785517188eba3bab/pyproject.toml#L20) by specifying your branch.

ElStabilini:
> Thanks @ElStabilini, I have a few comments (mainly some redundant `rx90` attributes).
> 
> We need to do a few more things before merging this PR.
> 
> 1. Check why pylint is [failing](https://github.com/qiboteam/qibocal/actions/runs/11974556505/job/33385894293?pr=1044).
> 2. Add tests. In this case it should be sufficient to add a few entries [here](https://github.com/qiboteam/qibocal/blob/pi_half/tests/runcards/protocols.yml) with `rx90: True` for all protocols that you modified.
> 3. Tests might fail because the CI is installing the `main` branch of qibolab which doesn't contain your fix for RX90 [added RX90 native gate and modified R method qibolab#1103](https://github.com/qiboteam/qibolab/pull/1103). To test that everything works correctly you can modify the [`pyproject.toml` ](https://github.com/qiboteam/qibocal/blob/c28039847aad2869956f22ed785517188eba3bab/pyproject.toml#L20) by specifying your branch.


I should have solved all the all the open reviews, tomorrow I'll try to work on these points, thank you fo r the notes and the suggestions @andrea-pasquale

ElStabilini:
I copy here a commit message I left in [c579cbb](https://github.com/qiboteam/qibocal/pull/1044/commits/c579cbb909a65410092b5f6ec36387e229575ac4)

The problem here is that right now we have the possibility to calibrate two different gates RX and RX90 so the `_update` function is defined accordingly taking `rx90: bool` as input.

Right now I simply added the input `rx90 = False` but maybe in the future can be useful to modify the flipping protocol in order to be able to run it for both RX and RX90 gates.

andrea-pasquale:
> The problem here is that right now we have the possibility to calibrate two different gates RX and RX90 so the `_update` function is defined accordingly taking `rx90: bool` as input.
> 
> Right now I simply added the input `rx90 = False` but maybe in the future can be useful to modify the flipping protocol in order to be able to run it for both RX and RX90 gates.

This is fine for now, eventually we could even put as default `rx90=True` in the update functions. In this way all other protocols that were using these functions will remain compatible.

src/qibocal/protocols/rabi/amplitude_frequency.py
```diff
@@ -234,10 +235,16 @@ def _plot(
             row=1,
             col=1,
         )
+
+        if data.rx90:
+            pulse_name = "Pi-half pulse"
+        else:
+            pulse_name = "Pi pulse"
```
alecandido, line 242: In these cases, it is especially convenient to use the ternary `if` operator (exact same of the `... ? ... : ...` construct in C)
```suggestion
        pulse_name = "Pi-half pulse" if data.rx90 else "Pi pulse"
```

alecandido:
@ElStabilini just to provide you some context about https://github.com/qiboteam/qibocal/pull/1044#discussion_r1863565286, the clear advantage of the proposed syntax is neither saving characters nor lines.

The most prominent difference between the equivalent expressions is that one is imperative, the other functional.
In the first one, you or conditioning the entire assignment operation, as the process of defining a variable name and its associated value. In principle, you need to analyze both branches of the conditional to ensure that the variable (`pulse_name` in this case) is even defined.
Instead, when using the ternary operation you are not conditioning the assignment as an operation, which becomes just an alias for a given computation. The computation is an expression, that always has a value. But it's the value itself the one being conditioned.

The functional approach makes the computation flow clearer, since it is somehow more rigid (you're not even thinking about operations, and possibly memory, but just values manipulation), and makes the analysis simpler and more effective. In languages which are actually functional, this may empower the compiler to act more aggressively. In Python, it makes the linters more effective[*], leading to fewer errors and more helpful messages :) 

[*]: since it optimizes the simplicity of the code, it makes it also more human-readable in general

ElStabilini:
> @ElStabilini just to provide you some context about [#1044 (comment)](https://github.com/qiboteam/qibocal/pull/1044#discussion_r1863565286), the clear advantage of the proposed syntax is neither saving characters nor lines.
> 
> The most prominent difference between the equivalent expressions is that one is imperative, the other functional. In the first one, you or conditioning the entire assignment operation, as the process of defining a variable name and its associated value. In principle, you need to analyze both branches of the conditional to ensure that the variable (`pulse_name` in this case) is even defined. Instead, when using the ternary operation you are not conditioning the assignment as an operation, which becomes just an alias for a given computation. The computation is an expression, that always has a value. But it's the value itself the one being conditioned.
> 
> The functional approach makes the computation flow clearer, since it is somehow more rigid (you're not even thinking about operations, and possibly memory, but just values manipulation), and makes the analysis simpler and more effective. In languages which are actually functional, this may empower the compiler to act more aggressively. In Python, it makes the linters more effective[*], leading to fewer errors and more helpful messages :)
> 
> [*]: since it optimizes the simplicity of the code, it makes it also more human-readable in general

It definitely makes sense, thank you!

doc/source/protocols/rabi/rabi.rst
```diff
@@ -29,6 +29,17 @@ Rabi rate is larger than the decay and the pure dephasing rate,
...
+
+Since many routines and protocols in quantum computing are based on `R_x(\pi/2)` rotations, in qibocal we implemented
+also another version of the Rabi experiment which can be used to tune the amplitude (duration) of the drive pulse in order
+to excite the qubit from the ground state up to state :math:`\frac{\ket{0}-i\ket{1}}{\sqrt{2}}`.
+
+The possibility to calibrate an `R_x(\pi/2)` rotation as native gate allows us to remove the errors that could arise from assuming that the `R_x(\pi/2)` amplitude (duration)
+is exactly half that of the `R_x(\pi)` amplitude (duration). This assumption presumes a perfectly linear response of the qubit to the drive pulse, which is
+often not the case due to nonlinearities in the qubit's response or imperfections in the pulse shaping :cite:p:`Chen2018MetrologyOQ`.
+
+In this case the pulse sequence is the same as before with the only difference that instad of a single `R_x(\pi)` pulse we use two concatenated `R_x(\pi/2)` pulses.
```
alecandido, line 41: I know that is not the most common tool for developing, but... sometimes even a spell-checker doesn't hurt :P
```suggestion
In this case the pulse sequence is the same as before with the only difference that instead of a single `R_x(\pi)` pulse we use two concatenated `R_x(\pi/2)` pulses.
```

doc/source/protocols/flipping.rst
```diff
@@ -35,6 +39,18 @@ The expected output is the following:
...
+If the same experiment is run setting the `rx90: True` the flipping is performed to calibrate the amplitude of the :math:`R_x(\pi/2)` rotation
+
+.. code-block:: yaml
+
+	- id: flipping
+	  operation: flipping
+	  parameters:
+	    delta_amplitude: 0.05
+	    nflips_max: 30
+	    nflips_step: 1
```
andrea-pasquale, line 51: Tab to be fixed

src/qibocal/protocols/flipping.py
```diff
@@ -23,23 +23,38 @@
...
     """Pulse sequence for flipping experiment."""
 
     sequence = PulseSequence()
     natives = platform.natives.single_qubit[qubit]
-    sequence |= natives.R(theta=np.pi / 2)
+
+    if rx90:
+        sequence |= natives.RX90()
+    else:
+        sequence |= natives.R(theta=np.pi / 2)
```
andrea-pasquale, line 40: Here it is not clear to me why we need to differentiate these two lines. I think we can just add the second version, since the fit is not "considering" the first RX90 since we extract the amplitude from the number of flips (which doesn't include the first one).

src/qibocal/protocols/flipping.py
```diff
@@ -23,23 +23,38 @@
...
-        qd_channel, rx_pulse = natives.RX()[0]
+        if rx90:
+            qd_channel, qd_pulse = natives.RX90()[0]
+        else:
+            qd_channel, qd_pulse = natives.RX()[0]
 
-        rx_detuned = update.replace(
-            rx_pulse, amplitude=rx_pulse.amplitude + delta_amplitude
+        qd_detuned = update.replace(
+            qd_pulse, delta_amplitude=qd_pulse.amplitude + delta_amplitude
```
andrea-pasquale, line 50:
```suggestion
            qd_pulse, amplitude=qd_pulse.amplitude + delta_amplitude
```
Shouldn't this be `amplitude`?

doc/source/protocols/flipping.rst
alecandido, line 51: Still plenty of tabs, instead of spaces...

image

doc/source/protocols/rabi/rabi.rst
```diff
@@ -103,6 +106,29 @@ It follows an example runcard and plot for the signal exepriment
...
+    - id: Rabi signal
+      operation: rabi_amplitude_signal
+      parameters:
+        min_amp: 0.01
+        max_amp: 0.16
+        step_amp: 0.002
+        pulse_length: 40
+        nshots: 1024
+        relaxation_time: 50000
+	rx90: True
```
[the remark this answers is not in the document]
alecandido: Still present...

alecandido (approved):
I did not test the current status, but I trust you did.

Other than some formatting, and tiny (and irrelevant) potential deduplication, this PR should be ready to go :)

andrea-pasquale (commented):
Thanks @ElStabilini for all the updates. I just have a few comments.
I still need to run the code on hardware myself but everything looks good so far.

andrea-pasquale (approved):

PR #1046, opened 2024-11-20
Process tomography

stavros11:
Allows to reconstruct the channel corresponding to the application of an arbitrary single or two-qubit gate (or gates) on hardware, following https://arxiv.org/abs/quant-ph/9610001.

Some example results:
```yaml
  - id: Process tomography
    operation: process_tomography
    targets: ["D2"]
    parameters:
      circuit: [["y90"]]
      nshots: 5000

  - id: Process tomography Two-Qubit
    operation: process_tomography
    targets: [["D2", "D1"]]
    parameters:
      circuit: [["cz", "cz"]]
      nshots: 5000
```

* Single-qubit (Y90 gate): http://login.qrccluster.com:9000/s9QyqzV7RASUWLq-oBqAkg==
* Two-qubit (CZ gate): http://login.qrccluster.com:9000/Vzx-IK8ERZy4LAEnF53TVQ==

The plot corresponds to the chi matrix introduced in Eq. (3. 2) in the operator basis from (4.1)-(4.4). This fully describes the channel. Note that the gate set is currently limited to `["x180", "y180", "x90", "y90", "-x90", "-y90"]` for simplification.

Draft mainly becasue it is lacking tests and documentation. Also, ideally this should be written using qibo circuits, similarly to the state tomographies, and potentially also reuse code from there. This would also fix the gate limitations, but I am not doing it because I want to keep it independent from transpiler/compiler until these are debugged.

### DIFF
--- a/src/qibocal/protocols/__init__.py
+++ b/src/qibocal/protocols/__init__.py
@@ -27,6 +27,7 @@ from .flux_dependence.qubit_flux_dependence import qubit_flux
 from .flux_dependence.qubit_flux_tracking import qubit_flux_tracking
 from .flux_dependence.resonator_crosstalk import resonator_crosstalk
 from .flux_dependence.resonator_flux_dependence import resonator_flux
+from .process_tomography import process_tomography
 from .qubit_power_spectroscopy import qubit_power_spectroscopy
 from .qubit_spectroscopy import qubit_spectroscopy
 from .qubit_spectroscopy_ef import qubit_spectroscopy_ef
@@ -152,4 +153,5 @@ __all__ = [
     "optimize_two_qubit_gate",
     "mermin",
     "ramsey_zz",
+    "process_tomography",
 ]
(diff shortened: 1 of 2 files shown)


igres26:
Great! Could we add a figure of merit to this? Some distance between the experiment data and the noiseless simulation.

alecandido:
> Great! Could we add a figure of merit to this? Some distance between the experiment data and the noiseless simulation.

This is something that also @stavros11 was looking for.

What I suggested (was the most trivial attempt) was just to use the $\mathbb{L}^2$ norm on the displayed matrix.
However, this might not be meaningful at all as a distance in the space of channels.

Instead, what could have been a meaningful attempt, is to have an $\sup$ norm, maximizing over the space of states' fidelities (i.e. computing the superior of the fidelities of the results of channels applications, simulated and measured, over the possible $\rho$ initial states).
Since with the tomography you have reconstructed the whole channel, no further measure is needed. And since the space is finite-dimensional, the inferior should just be a maximum.
I'm not sure whether this is an actual distance over the space of channels: it is symmetric, and 0 iff the channels are the exact same. Not sure if it's triangular...

In any case, just plotting the difference matrix (instead of the two $\chi_{mn} matrices) it would already help me interpreting the result...

PR #1049, opened 2024-11-26
Update CHSH pulses routine to 0.2

stavros11:
Still need to confirm it works properly on hardware.

### DIFF
--- a/src/qibocal/protocols/two_qubit_interaction/chsh/utils.py
+++ b/src/qibocal/protocols/two_qubit_interaction/chsh/utils.py
@@ -1,0 +1,26 @@
+"""Auxiliary functions to run CHSH protocol."""
+
+from qibo.config import log
+
+READOUT_BASIS = ["ZZ", "ZX", "XZ", "XX"]
+
+
+def compute_chsh(frequencies, basis, i):
+    """Computes the chsh inequality out of the frequencies of the 4 circuits executed."""
+    chsh = 0
+    aux = 0
+    for freq in frequencies:
+        for outcome in freq:
+            if aux == 1 + 2 * (
+                basis % 2
+            ):  # This value sets where the minus sign is in the CHSH inequality
+                chsh -= (-1) ** (int(outcome[0]) + int(outcome[1])) * freq[outcome][i]
+            else:
+                chsh += (-1) ** (int(outcome[0]) + int(outcome[1])) * freq[outcome][i]
+        aux += 1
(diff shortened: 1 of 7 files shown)


src/qibocal/protocols/two_qubit_interaction/chsh/utils.py
```diff
@@ -0,0 +1,26 @@
...
+            ):  # This value sets where the minus sign is in the CHSH inequality
+                chsh -= (-1) ** (int(outcome[0]) + int(outcome[1])) * freq[outcome][i]
+            else:
+                chsh += (-1) ** (int(outcome[0]) + int(outcome[1])) * freq[outcome][i]
+        aux += 1
+    nshots = sum(freq[x][i] for x in freq)
+    try:
+        return chsh / nshots
+    except ZeroDivisionError:
+        log.warning("Zero number of shots, returning zero.")
```
alecandido, line 25: Drop the warning, just noise :P

alecandido: If 0 is not a consistent limit value, consider returning `nan`. If it is consistent, no need to do anything :)

stavros11: I removed the warning in 549b504f80f29a7176634bf3c18afaba46e4940d, even though this code was just copy pasted from the `main` branch and not originally authored by me. Regarding `0`, I am not sure if it is consistent, however it seems to me that the `except` branch is never going to be executed in practice, as I don't see why someone would launch this routine (or anything on hardware) with `nshots == 0`.

stavros11:
Marking this as ready as I managed to get consistent results with 0.1 for B1-B3 pair:

0.1: http://login.qrccluster.com:9000/Pk4B8i3mRoaHOIWMDX7AWg==
0.2: http://login.qrccluster.com:9000/K4uyV9oBRxetDVfVp-a97w==

Note that I used https://github.com/qiboteam/qibolab/pull/1116 for this.

src/qibocal/protocols/two_qubit_interaction/chsh/utils.py
alecandido: We could even consider `nshots == 0`, and just leave the original exception.

alecandido (commented):
I believe I never finished this review, but since these comments has been here pending since I don't remember when, I decided to just publish them, and in case to continue later on...

PR #1057, opened 2024-12-11
Adding CPMG protocol

### LINKED ISSUE #896: CPMG protocol
Hi,
It would be nice to have routines that help us characterise the noise. CPMG is one such routine.
https://www.nature.com/articles/nphys1994
The implementation of this routine (in RT) requires advanced sweepers in qibolab.

andrea-pasquale:
Closes #896.
Here is an example: http://login.qrccluster.com:9000/M0By3hjUTAakgT8Jg1VlXA==
I've also fixed a factor 2 in the spin_echo protocol.


ElStabilini (approved):

src/qibocal/protocols/coherence/utils.py
```diff
@@ -30,43 +30,53 @@ def average_single_shots(data_type, single_shots):
...
+            sequence.append((qd_channel, drive_delays[2 * i + 1]))
+            sequence.append((ro_channel, ro_delays[2 * i + 1]))
+
+        sequence.append((qd_channel, rx90_pulse))
+
+        sequence.append(
+            (
+                ro_channel,
+                Delay(duration=2 * rx90_pulse.duration + n * pulse.duration),
+            )
```
Edoardo-Pedicillo, line 74: The sequence should be the following

 
I don't see the t/2N delays after the first and last pulse.

doc/source/protocols/cpmg/cpmg.rst
```diff
@@ -0,0 +1,50 @@
+CPMG sequence
+=============
+
+In this section we show how to run the dynamical decoupling sequence CPMG.
+
+The CPMG sequence consists in applying N equally spaced :math:`\pi` pulses
+within two :math:`\pi` pulses. By increasing the number of :math:`\pi` pulses :math:`T_2`
+should increase until reaching the :math:`2 T_1` limit.
```
Edoardo-Pedicillo, line 8: This behaviour depends on the kind of noise.

src/qibocal/protocols/coherence/cpmg.py
```diff
@@ -0,0 +1,172 @@
...
+        xaxis_title="Time [ns]",
+        yaxis_title="Probability of State 1",
+    )
+
+    figures.append(fig)
+
+    return figures, fitting_report
+
+
+cpmg = Routine(_acquisition, _fit, _plot, _update)
```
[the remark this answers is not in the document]
andrea-pasquale: Yes, I was thinking about this. Perhaps at this point it is easier to just remove the update for this protocol. What do you think?

src/qibocal/protocols/coherence/spin_echo.py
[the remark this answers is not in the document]
andrea-pasquale: Actually, the spin-echo protocol doesn't implement the CPMG sequence but the CP sequence, which is slightly different.
https://github.com/qiboteam/qibocal/blob/db00d755cbcbf119e672aba9525ed43468673443/src/qibocal/protocols/coherence/spin_echo.py#L44
Screenshot 2024-12-12 at 14 24 01

But yes, generally I agree the two protocols are exactly the same apart from the sequence type.

src/qibocal/protocols/coherence/utils.py
andrea-pasquale: The sequence can be re-written as 
RX90 - (wait(t/2N) RY wait(t/2N))^N RX90
which is exactly what I am doing in the for loop.

I've specified that the delay should have also a factor 1/2N in the acquisition https://github.com/qiboteam/qibocal/blob/db00d755cbcbf119e672aba9525ed43468673443/src/qibocal/protocols/coherence/cpmg.py#L61

doc/source/protocols/cpmg/cpmg.rst
andrea-pasquale: I am not sure about this. Generally by increasing the number of pi pulses according to [here](https://arxiv.org/pdf/1101.4707) I believe that we should be able to mitigate noises of the type 1/f. Then I don't know whether we could have other noises which are not 1/f but I don't know if we can detect them through a CPMG sequence.
Did you have something else in mind?

andrea-pasquale (commented):
Thanks for the comments @Edoardo-Pedicillo.
I will try to reduce code duplication :)

src/qibocal/protocols/coherence/utils.py
Edoardo-Pedicillo: Smart !

doc/source/protocols/cpmg/cpmg.rst
Edoardo-Pedicillo: > Did you have something else in mind?

You could have a white noise but I think the conclusions are the same. I just suggest to add what you have written in this comment in the docs.

> Then I don't know whether we could have other noises which are not 1/f but I don't know if we can detect them through a CPMG sequence.

By changing the number of pulses increase the sensitivity to higher frequencies, so I would say yes

src/qibocal/protocols/coherence/cpmg.py
```diff
@@ -0,0 +1,100 @@
...
+        params.delay_between_pulses_step,
+    )
+    for q in targets:
+        # this is assuming that RX and RX90 have the same duration
+        assert (
+            params.delay_between_pulses_start / 2 / params.n
+            > platform.natives.single_qubit[q].RX()[0][1].duration
+        ), (
+            f"Initial delay too short for qubit {q}, "
+            f"minimum delay should be {platform.natives.single_qubit[q].RX()[0][1].duration * 2 * params.n}"
```
andrea-pasquale, line 57: @Edoardo-Pedicillo I believe that this should fix the issue that appeared with QM. Eventually we could propagate this also to t1, t2 and ramsey protocols.

src/qibocal/protocols/coherence/cpmg.py
```diff
@@ -0,0 +1,97 @@
...
+    wait_range = np.arange(
+        params.delay_between_pulses_start,
+        params.delay_between_pulses_end,
+        params.delay_between_pulses_step,
+    )
+    for q in targets:
+        # this is assuming that RX and RX90 have the same duration
+        assert params.delay_between_pulses_start / 2 / params.n >= 1, (
+            f"Initial delay too short for qubit {q}, "
+            f"minimum delay should be {params.delay_between_pulses_start * 2 * params.n}"
```
andrea-pasquale, line 54: Here is the condition that we should be careful about (the previous one was wrong).
With this we should make sure that there is at least 1 ns between the drive pulses.

andrea-pasquale: Here are the latest results: http://login.qrccluster.com:9000/2JRKa7cnSnK_Av7faBQMrw==/

andrea-pasquale: In order to have a consisten x axis I've removed the duration of the pi-pulses from the: https://github.com/qiboteam/qibocal/pull/1057/commits/db904286cd80014c227e2d8c66ad56b346f0f818
I've also changed  the assert condition accordingly.

andrea-pasquale:
@stavros11 whenever you have time could you please have a look at the CPMG protocol? Given that it caused a few issues with QM I was wondering whether something could be done in a better way in Qibocal. Thanks!

stavros11:
> @stavros11 whenever you have time could you please have a look at the CPMG protocol? Given that it caused a few issues with QM I was wondering whether something could be done in a better way in Qibocal. Thanks!

I had a quick look and also tried some quick runs on qw11q and I do not see anything that could cause issues, or QM failure. Did you find any strong correlation between running this and QM breaking? If it happened only a few times, maybe the instrument had some other issue unrelated to this routine. I tried a few runs and it did not cause any problems, however I only used small values for `n`. Due to the way the QM compiler works, I would expect this being slow or even not working for large enough `n` (possibly `n > 100` but this is mostly random estimation), due to trying to compile many `play` commands. It's basically similar issue we had with unrolling and a potential solution would be to move the
https://github.com/qiboteam/qibocal/blob/526993b80e2eb469e6ccffbcf8c7b826cb03e3d4/src/qibocal/protocols/coherence/utils.py#L65
loop inside the instrument, however this is not possible with the qibolab 0.2 API and is one of the reasons for the 0.3 planning. Even in the high `n` case now, I would not expect the instruments to break completely, this routine will probably hang, but if you cancel you should be able to submit other jobs without restarting.

Another thing that could be problematic (but still not QM breaking), is that we are sweeping delay duration over non-trivial values: https://github.com/qiboteam/qibocal/blob/526993b80e2eb469e6ccffbcf8c7b826cb03e3d4/src/qibocal/protocols/coherence/cpmg.py#L68
Due to the QM `wait` command supporting only multiples of 4ns, these values will be rounded to the nearest integer multiple of 4ns, which may not be exactly what you expect. I noticed that depending on the choices of parameters sometimes you may even get an error
```py
Exception: The spacing of the input array must be even in linear or logarithmic scales. Please use `for_each_()` for arbitrary scans.
```
This is normal and there are workarounds, such as implementign the delay with baking, instead of using `wait`, which should at least allow 1ns resolution. I have not invested time in implementing these in the driver yet, but I could put more priority if these cases are relevant (basically if the 4ns step is not sufficient in practice for this routine).

stavros11:
> Did you also try to run the experiment for `n=10`?

No, before I tried with `n=2` and `n=4` and did not have any issues, in particular the following:
```yaml
delay_between_pulses_start: 1000
delay_between_pulses_end: 50000
delay_between_pulses_step: 120
n: 4
nshots: 1000
```
I just tried with `n=10` and I also got the issue you described. Hanging during compilation is kind of normal, I have seen that before when playing with unrolling, however usually there are no other side effects if you cancel. Now, QM seems to break and then restart itself, which is weird. Then after the restart the instrument runs fine again, but this whole process takes 10-15 minutes. I am not sure why this happens, but I would guess it has to do with the current firmware. Thanks for pointing it out.

> Regarding this, even if this a driver issue I could force this condition as I was doing before this commit [526993b](https://github.com/qiboteam/qibocal/commit/526993b80e2eb469e6ccffbcf8c7b826cb03e3d4) If you are telling me that this is not breaking QM I would keep it as it is.

I would not expect this to break QM completely because the driver converts what goes to `wait` commands to clock cycles: https://github.com/qiboteam/qibolab/blob/cbaf2c62c965393c645b5c73336d00628e9e1271/src/qibolab/_core/instruments/qm/program/instructions.py#L18. However, in that case the delays will not be exactly the ones you set (for example instead of 22ns it will wait 20ns etc.), so the final plot of the results will not be accurate. I am not sure if we would like to raise an error on the qibocal side, because what I am saying may not be true for other drivers. Maybe it makes more sense to check and raise this error in qibolab QM driver.

alecandido:
> I just tried with `n=10` and I also got the issue you described. Hanging during compilation is kind of normal, I have seen that before when playing with unrolling, however usually there are no other side effects if you cancel. Now, QM seems to break and then restart itself, which is weird. Then after the restart the instrument runs fine again, but this whole process takes 10-15 minutes. I am not sure why this happens, but I would guess it has to do with the current firmware. Thanks for pointing it out.

I expected the compiler to take forever (which is everything but a pleasure), but not to actually break. I wonder if it's even managing to saturate the memory, and just start swapping... (possibly with no swap available - and thus breaking)

However, I expect this to be an issue in the compilation, and not in the program itself: if the issue were in the execution, the compiler should have been able to detect it. But if the issue is the compilation itself, it may crash as a program because of insufficient resources.

> However, in that case the delays will not be exactly the ones you set (for example instead of 22ns it will wait 20ns etc.), so the final plot of the results will not be accurate. I am not sure if we would like to raise an error on the qibocal side, because what I am saying may not be true for other drivers. Maybe it makes more sense to check and raise this error in qibolab QM driver.

I can confirm there is no similar limitation in Qblox. In its case, it is possible to wait any multiple of 1 ns.

Edoardo-Pedicillo (approved):
LGTM, thanks @andrea-pasquale

PR #974, opened 2024-09-02
Cryoscope 

andrea-pasquale:
I'm currently using this protocol to calibrate the filters of QM.
It is not precisely a cryoscope but I'm just fitting with an exponential the time of rise and then I apply the corresponding FIR and IIR filters.

- [ ] Improve documentation
- [ ]  Move post-processing in fit

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.
- [ ] Compatibility with Qibo modules (Please edit this section if the current pull request is not compatible with the following branches).
    - [ ] Qibo: `master`
    - [ ] Qibolab: `main`
    - [ ] Qibolab_platforms_qrc: `main`

PR #1061, opened 2025-01-13
Flux crosstalk computation with flux pulse amplitude

andrea-pasquale:
In this PR I'm modifying the `flux_amplitude_frequency` protocol intially proposed in #974 to evaluate also flux crosstalk. This evaluation should be "more useful" compared to the one already available in main given that in this case I'm using flux_pulses.
Here are a few examples:
http://login.qrccluster.com:9000/PrWiWFppRtiyaTHy3jRocA==/
http://login.qrccluster.com:9000/h2SDjCTvTj66-XdkfUr1tg==/

Here is an example of the crosstalk matrix computed using qubits B3 and B4: 
test

Currently the crosstalk is almost 3 orders of magnitude lower which is more or less expected. See for example https://arxiv.org/pdf/2303.03347 and https://arxiv.org/pdf/2112.03708.

According to the qua documentation https://docs.quantum-machines.co/latest/docs/Guides/features/?h=crosstalk#crosstalk-correction-matrix it seems that it should be possible to just provide this crosstalk matrix to correct the flux pulses.

TODO:
- [ ] update documentation


andrea-pasquale:
I think the PR should be now ready to review: http://login.qrccluster.com:9000/Jnsoo476Q36AJjFw85nFLw==
I'm only missing documentation which I will do as soon as we finalize the PR.

tests/test_cryoscope.py
ElStabilini, line 1: I'm not sure if the fact that the tests are commented out is a choice or there's an issue that needs to be fixed

andrea-pasquale: Thanks for pointing it out. I will fix it now.

src/qibocal/protocols/flux_amplitude_frequency.py
```diff
@@ -33,18 +38,36 @@ class FluxAmplitudeFrequencyParameters(Parameters):
...
 class FluxAmplitudeFrequencyResults(Results):
     """FluxAmplitudeFrequency outputs."""
 
+    crosstalk: bool = False
+    """Check if this is crosstalk protocol."""
     detuning: dict[QubitId, float] = field(default_factory=dict)
     """Frequency detuning."""
-    fitted_parameters: dict[tuple[QubitId, str], list[float]] = field(
+    flux: dict[QubitId, float] = field(default_factory=dict)
+    """Frequency detuning."""
```
ElStabilini, line 63:
```suggestion
    """Derived flux  """
```

src/qibocal/protocols/two_qubit_interaction/cryoscope.py
```diff
@@ -347,25 +347,24 @@ def _fit(data: CryoscopeData) -> CryoscopeResults:
...
         derivative_window_size += (derivative_window_size + 1) % 2
 
         # find demodulatation frequency
-        demod_data = np.exp(2 * np.pi * 1j * x * demod_freq) * (norm_data)
+        demod_data = np.exp(2 * np.pi * 1j * x * np.abs(demod_freq)) * (norm_data)
 
         # compute phase
         phase = np.unwrap(np.angle(demod_data))
-
+        phase -= phase[0]
```
ElStabilini, line 367: I've seen this also in `flux_amplitude_frequency` routine but I didn't understand why we need it :)

andrea-pasquale: For sure it should be documented better. The reason is that I'm trying to impose a condition such that if the flux_pulse_amplitude is 0 the detuning should be zero. If I don't normalize the data I might obtain a few values of the order of MHz which are not expected. Moreover, in this new routine where I am computing the expecting flux the computation fails because we might compute the arccos of a value larger than 1.

ElStabilini (approved):

Edoardo-Pedicillo (approved):

ElStabilini (approved):

Issue #913, opened 2024-07-02
Restore support for simulation


alecandido:
Current routines are tuned to work with a `Platform`, that is ending up in many places.

There is no doubt that Qibolab is the main target of Qibocal, and to move faster I'd accept to support only that for a while.
That's why I imposed the existence of a `Platform` in #909.

However, running circuits routines in simulation, especially with a specified noise model (but sometimes even noiseless, for debugging) could be very useful.
The proposed way to recover this feature is to wrap a Qibo backend in a `qibocal.SimulationBackend`, which will hold a `qibocal.Platform` in the same way as the `qibolab.QibolabBackend` (and it will respect the `qibolab.Platform` interface - at least the part used within Qibocal, which we could define somewhere).

This will allow applying all the routines on a simulation backend, and fail if and only if pulses are received.

alecandido:
Instead of a platform, we could make a Qibocal backend wrapper.

This object will contain the Qibo backend for simulation, and execute circuits, rejecting pulses (but implementing the interface), while for Qibolab will contain `(platform, compiler, transpiler)`, and it will execute pulses (just passing the call to the platform) and circuits (by transpiling, compiling, and eventually passing the call to the platform).

At least, there will be a single object around. And, hopefully, we won't need to access the platform that often (in the future, now need)

PR #1063, opened 2025-01-13
Temporary fix of error mitigation routine if `QIBO_PLATFORM` is not set

stavros11:
The error mitigation routine relies on the `GlobalBackend` in a few places. This causes issues when the `QIBO_PLATFORM` environment variable is not properly set. Usually we do not see these issues because this environment variable is set in the slurm prologue, however this is not the case for the qw5q_platinum node.

### DIFF
--- a/src/qibocal/protocols/readout_mitigation_matrix.py
+++ b/src/qibocal/protocols/readout_mitigation_matrix.py
@@ -1,3 +1,4 @@
+import os
 from dataclasses import dataclass, field
 from typing import Optional
 
@@ -5,8 +6,8 @@ import numpy as np
 import numpy.typing as npt
 import plotly.express as px
 from qibo import gates
-from qibo.backends import get_backend
 from qibo.models import Circuit
+from qibolab.backends import QibolabBackend
 from qibolab.platform import Platform
 from qibolab.qubits import QubitId
 
@@ -61,8 +62,9 @@ def _acquisition(
     data = ReadoutMitigationMatrixData(
         nshots=params.nshots, qubit_list=[list(qq) for qq in targets]
     )
-    backend = get_backend()
-    backend.platform = platform
+    os.environ["QIBO_PLATFORM"] = platform.name
+    backend = QibolabBackend(platform)
+
     transpiler = dummy_transpiler(backend)
     qubit_map = [i for i in range(platform.nqubits)]
     for qubits in targets:


src/qibocal/protocols/readout_mitigation_matrix.py
```diff
@@ -61,8 +62,9 @@ def _acquisition(
     data = ReadoutMitigationMatrixData(
         nshots=params.nshots, qubit_list=[list(qq) for qq in targets]
     )
-    backend = get_backend()
-    backend.platform = platform
+    os.environ["QIBO_PLATFORM"] = platform.name
+    backend = QibolabBackend(platform)
```
alecandido, line 66: Wait: if you directly use `QibolabBackend` [*], then `$QIBO_PLAFORM` should be irrelevant. Isn't it?

If that's the case, this should be sufficient
```suggestion
    backend = QibolabBackend(platform)
```

[*]: which is a great idea - until we'll reintroduce support in Qibocal for backends other than Qibolab #913 (but that's a separate issue, to tackle later on)

stavros11: > Wait: if you directly use `QibolabBackend` [*], then `$QIBO_PLAFORM` should be irrelevant. Isn't it?

That is also what I initially expected, however it seems that another instance of `GlobalBackend` is instantiated within the https://github.com/qiboteam/qibocal/blob/53a94cadfd254c9fec7be7e34250751a64fb6248/src/qibocal/protocols/readout_mitigation_matrix.py#L81
call which causes a failure if the environment variable is not properly set (which is the case if the slurm prologue is wrong). From the error trace, it seemed that this second `GlobalBackend` is related to the transpilation process, but I did not investigate in detail.

In any case, this PR was just a quick hotfix for something that @Luca-Ben-Herrmann was running, caused by the wrong slurm prologue of `qw5q_platinum` partition (which has also been fixed). Ideally, I would like to make all routines independent of the `GlobalBackend` to avoid such issues, but I am not sure how much of rework that requires.

> [*]: which is a great idea - until we'll reintroduce support in Qibocal for backends other than Qibolab #913 (but that's a separate issue, to tackle later on)

I guess hard-coding `QibolabBackend` here is also not optimal as it would not work for simulation, which I am not sure if is relevant for that particular routine. Anyway, this PR is mostly to raise these issues, maybe we don't even need to merge as it is, unless you are all fine with restricting with qibolab for now and introducing other backends later, as you are proposing.

alecandido: > I guess hard-coding `QibolabBackend` here is also not optimal as it would not work for simulation,

As said, I would not worry too much about this. Not yet.

> call which causes a failure if the environment variable is not properly set (which is the case if the slurm prologue is wrong). From the error trace, it seemed that this second `GlobalBackend` is related to the transpilation process, but I did not investigate in detail.

Actually, I failed to find it, since in that call the `backend` you're instantiating is passed in (and transitively passed down into the called `transpile_circuits()` function as well).

andrea-pasquale:
Closing this in favor of #1076 which just got merged.

PR #1066, opened 2025-01-15
conditional and virtual phase experiment improvements

hay-k:
- Remove parameters `flux_pulse_amplitude` and `flux_pulse_duration` from the experiment `correct_virtual_z_phases`. I think they are not usefull - nobody is going to manually try out various combinations of these parameters to find a good one. There is the `optimize_two_qubit_gate` for this type of task.
  - BTW, `flux_pulse_duration` was not even used, it was ignored and always the flux pulse duration from platform was used.
- Get rid of the `amplitude` argument of the `create_sequence` function (as a consequence of the change in the above point). Rename the `duration` argument to `flux_pulse_max_duration` - this is needed for the optimize routine to make sure there is enough room for the duration sweep, and the flux pulse does not overlap with readout (this is different from `dt`, which is just a constant offset, not dependent on values used for sweep). 
- Change the return content of the `create_sequence` function. Now it return the sequence, reference to the flux pulse, reference to the theta pulse.
- Move the code for fitting a sinusoid to a separate function, so that it is reused instead of duplicated.
- And, the reason all of this started, the phase difference between two sinusoids is now calculated in a manner, that it is normalized in the [0, pi] range, and does not cause large jumps because of angle fluctuations close to boundary values (0 and pi).:
  - before:  image
  - after: image

### DIFF
--- a/src/qibocal/protocols/two_qubit_interaction/optimize.py
+++ b/src/qibocal/protocols/two_qubit_interaction/optimize.py
@@ -409,7 +409,7 @@ def _plot(
                     y=amps,
                     z=cz,
                     zmin=np.pi / 2,
-                    zmax=3 * np.pi / 2,
+                    zmax=np.pi,
                     name="{fit.native} angle",
                     colorbar_x=-0.1,
                     colorscale="RdBu",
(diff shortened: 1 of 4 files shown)


Edoardo-Pedicillo (approved):

andrea-pasquale (approved):
Thank you @hay-k

PR #1076, opened 2025-02-05
Handling global backend for protocols involving circuits

### LINKED ISSUE #1065: Circuit based protocols run only on simulation
Whenever we run a protocol containing circuit execution, it seems that the `platform` is ignored and the execution is launched on the default global backend which is most likely going to be `numpy`. I remembered that we already dealt with this issue in the past https://github.com/qiboteam/qibocal/pull/941/. I don't know why this is happening again, perhaps it could be due to some recent changes in Qibo. I need to investigate more.
To reproduce the issue it is sufficient to run any protocol involving circuits on the dummy platform.
Thanks @jykhoo1987 for spotting this issue.

andrea-pasquale:
Closes #1065, by directly instatiating a local backend in each acquisition function.

Other changes implemented in this PR:

- Drop noise models for RB
- Adding to pytest the env variable `QIBO_PLATFORM` to properly run tests on the cluster
- Small fixes related to other tests


src/qibocal/auto/transpile.py
```diff
@@ -1,16 +1,32 @@
...
+from qibolab import MetaBackend
+from qibolab.platform import Platform
 from qibolab.qubits import QubitId
 
 
+def _get_platforms():
+    """Qibolab platforms."""
+    try:
+        platforms = list(MetaBackend().list_available())
+    except RuntimeError:
```
alecandido, line 16: Not a comment on this PR, but just a point raised as a consequence.

Maybe we should actually drop this exception: if there is no path, we should already return an empty list.
It should be sufficient to do it here:
https://github.com/qiboteam/qibolab/blob/6b2a7b790e8d58d52be937c2f41b8606459a3bf1/src/qibolab/_core/platform/load.py#L23-L26

This should result in `.list_available()` returning an empty list, and `.load()` raising a different error.

@stavros11

src/qibocal/auto/transpile.py
```diff
@@ -28,14 +44,13 @@ def transpile_circuits(
...
     transpiled_circuits = []
-
-    qubits = list(backend.platform.qubits)
-    if isinstance(qubit_maps[0][0], str):
-        for i, qubit_map in enumerate(qubit_maps):
-            qubit_map = map(lambda x: qubits.index(x), qubit_map)
-            qubit_maps[i] = list(qubit_map)
-    if backend.name == "qibolab":
-        platform_nqubits = backend.platform.nqubits
+    if platform.name in AVAILABLE_PLATFORMS:
```
alecandido, line 47: Maybe I just forgot, but in which case is this condition supposed to be false?

alecandido (commented):
Just a couple of comments, but in general I agree with the overall idea (since we're now focusing on the platform, assuming the backend to be Qibolab, there is no need to require everywhere to create the entire backend)

src/qibocal/auto/transpile.py
andrea-pasquale: I think it should be False when we transpile circuits that are executed on simultation so no transpilation at all.

alecandido: > I think it should be False when we transpile circuits that are executed on simultation so no transpilation at all.

Ah, I see. But in that case there should be no `Platform` at all, since it is the `qibolab.Platform`.

My advice would be to drop any simulation reference, and assume it is always Qibolab (possibly the emulator, but still Qibolab).
I would reintroduce the option for simulation as part of a larger refactor, taking into account #913. I'd expect there should be other as well not supporting a backend other than Qibolab.

src/qibocal/auto/transpile.py
```diff
@@ -99,20 +97,19 @@ def execute_transpiled_circuit(
...
         transpiler,
     )[0]
+
     return transpiled_circ, backend.execute_circuit(
         transpiled_circ, initial_state=initial_state, nshots=nshots
     )
 
 
-def dummy_transpiler(backend) -> Optional[Passes]:
+def dummy_transpiler(backend: Backend) -> Optional[Passes]:
```
alecandido, line 106:
```suggestion
def dummy_transpiler(backend: Backend) -> Passes:
```

alecandido (approved):

src/qibocal/auto/transpile.py
```diff
@@ -28,20 +28,18 @@ def transpile_circuits(
         are all string or all integers.
     """
     transpiled_circuits = []
-
-    qubits = list(backend.platform.qubits)
+    platform = backend.platform
```
stavros11, line 31: In principle here you could add something like
```py

if not isinstance(platform, Platform):
    transpiled_circuits = circuits
```
to retain the previous behavior when the backend is not qibolab.

stavros11 (approved):
> @stavros11 I believe that with this approach we don't need anymore #1063, feel free to have a look and eventually we can just merge this PR if you agree.

Thanks @andrea-pasquale. Indeed #1063 was just a temporary solution, not intended to be merged, so feel free to close it after this is merged.

src/qibocal/auto/transpile.py
alecandido: I would keep assuming it is Qibolab, to make everything simpler. Even though in some places may be simpler to deal with other backends, it may fail in other places. We will overhaul this consistently later on.

andrea-pasquale: Indeed, the expected behavior would be to have a qibolab platform https://github.com/qiboteam/qibocal/blob/59f95e330c98d6aa274c536bdf174165cb2e2b0e/src/qibocal/cli/run.py#L22-L23

tests/conftest.py
```diff
@@ -4,6 +4,10 @@
 import pytest
 
 
+def pytest_configure():
+    os.environ["QIBO_PLATFORM"] = "dummy"
```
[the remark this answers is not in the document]
alecandido: Sorry, my bad.

I introduced that test when I implemented the scripting mechanism, because that was the context at the time. However, there is no special reason to keep it, so you could just drop the test.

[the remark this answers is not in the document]
andrea-pasquale: To avoid trigerring the error we can also decide to not rely on `QIBO_PLATFORM` anymore and just use `dummy` as default platform given that the behavior of `QIBO_PLATFORM` is mostly undocumented.

PR #1077, opened 2025-02-11
Native gates directly defined by the platform

### LINKED ISSUE #1075: Circuit based protocols use default native gate set instead of platform native gates
Circuit based protocols (e.g. 1 and 2 qubit rb) uses the default native gate set when calling the transpiler:

cls.CZ | cls.GPI2 | cls.I | cls.Z | cls.RZ | cls.M

When the platform has a different native gate set, these protocols cannot execute.

These protocols should allow for native gates to be declared in the qibocal runcard as a protocol parameter (such as how it is done in chevron), or use the native gates declared in the qibolab runcard instead of the above native gate set.

Edoardo-Pedicillo:
This PR fixes #1075. Now the `dummy_transpiler` takes the native gates directly from the platform (assuming all qubits and pairs to have the same ones) and also the compiler is updated taking into account the new rules. This PR doesn't solve the problem of define any possible native gates set, indeed Qibo requires them to be between the ones defined here https://github.com/qiboteam/qibo/blob/312efb9d6934e85cfdaafcdb7027cdfe46bf87c4/src/qibo/transpiler/unroller.py#L51-L59

Otherwise it will raise an error
https://github.com/qiboteam/qibo/blob/312efb9d6934e85cfdaafcdb7027cdfe46bf87c4/src/qibo/transpiler/unroller.py#L82-L83

### DIFF
--- a/src/qibocal/auto/transpile.py
+++ b/src/qibocal/auto/transpile.py
@@ -157,9 +157,11 @@ def set_compiler(backend):
         if gate not in compiler.rules:
             # TODO: this code is working only with pairs
             def rule(qubits_ids, platform, parameters=None):
-                pulses = getattr(
-                    platform.pairs[tuple(qubits_ids[1])].native_gates, native
-                ).pulses
+                if len(qubits_ids[1]) == 1:
+                    native_gate = platform.qubits[tuple(qubits_ids[1])].native_gates
+                else:
+                    native_gate = platform.pairs[tuple(qubits_ids[1])].native_gates
+                pulses = getattr(native_gate, native).pulses
                 return PulseSequence(pulses), {}
 
             backend.compiler[gate] = rule
--- a/src/qibocal/auto/transpile.py
+++ b/src/qibocal/auto/transpile.py
@@ -127,15 +127,16 @@ def execute_transpiled_circuit(
     )
 
 
-def dummy_transpiler(platform) -> Optional[Passes]:
+def dummy_transpiler(platform, native_gates: Optional[list] = None) -> Optional[Passes]:
     """
     If the backend is `qibolab`, a transpiler with just an unroller is returned,
     otherwise None.
     """
     if platform.name in AVAILABLE_PLATFORMS:
-        unroller = Unroller(NativeGates.default())
+        if native_gates is None:
+            native_gates = NativeGates.default()
+        unroller = Unroller(NativeGates.from_gatelist(native_gates))
         return Passes(connectivity=platform.topology, passes=[unroller])
-    return None
 
 
 def pad_circuit(nqubits, circuit: Circuit, qubit_map: list[int]) -> Circuit:
--- a/src/qibocal/auto/transpile.py
+++ b/src/qibocal/auto/transpile.py
@@ -1,16 +1,32 @@
 from typing import Optional
 
 from qibo import Circuit
-from qibo.backends.abstract import Backend
+from qibo.backends import construct_backend, get_backend
 from qibo.transpiler.pipeline import Passes
 from qibo.transpiler.unroller import NativeGates, Unroller
+from qibolab import MetaBackend
+from qibolab.platform import Platform
 from qibolab.qubits import QubitId
+
+
+def _get_platforms():
+    """Qibolab platforms."""
+    try:
+        platforms = list(MetaBackend().list_available())
+    except RuntimeError:
+        platforms = []
+
+    return platforms + ["dummy"]
+
+
+AVAILABLE_PLATFORMS = _get_platforms()
+"""Available qibolab platforms."""
 
 
 def transpile_circuits(
     circuits: list[Circuit],
     qubit_maps: list[list[QubitId]],
-    backend: Backend,
+    platform: Platform,
     transpiler: Optional[Passes],
 ):
     """Transpile and pad `circuits` according to the platform.
@@ -28,14 +44,13 @@ def transpile_circuits(
         are all string or all integers.
     """
     transpiled_circuits = []
-
-    qubits = list(backend.platform.qubits)
-    if isinstance(qubit_maps[0][0], str):
-        for i, qubit_map in enumerate(qubit_maps):
-            qubit_map = map(lambda x: qubits.index(x), qubit_map)
-            qubit_maps[i] = list(qubit_map)
-    if backend.name == "qibolab":
-        platform_nqubits = backend.platform.nqubits
+    if platform.name in AVAILABLE_PLATFORMS:
+        qubits = list(platform.qubits)
+        if isinstance(qubit_maps[0][0], str):
+            for i, qubit_map in enumerate(qubit_maps):
+                qubit_map = map(lambda x: qubits.index(x), qubit_map)
+                qubit_maps[i] = list(qubit_map)
+        platform_nqubits = platform.nqubits
         for circuit, qubit_map in zip(circuits, qubit_maps):
             new_circuit = pad_circuit(platform_nqubits, circuit, qubit_map)
             transpiled_circ, _ = transpiler(new_circuit)
@@ -48,7 +63,7 @@ def transpile_circuits(
 def execute_transpiled_circuits(
     circuits: list[Circuit],
     qubit_maps: list[list[QubitId]],
-    backend: Backend,
+    platform: Platform,
     transpiler: Optional[Passes],
     initial_states=None,
     nshots=1000,
@@ -66,9 +81,13 @@ def execute_transpiled_circuits(
     transpiled_circuits = transpile_circuits(
         circuits,
         qubit_maps,
-        backend,
+        platform,
         transpiler,
     )
+    if platform.name in AVAILABLE_PLATFORMS:
+        backend = construct_backend(backend="qibolab", platform=platform)
+    else:
+        backend = get_backend()
     return transpiled_circuits, backend.execute_circuits(
         transpiled_circuits, initial_states=initial_states, nshots=nshots
     )
@@ -77,7 +96,7 @@ def execute_transpiled_circuits(
 def execute_transpiled_circuit(
     circuit: Circuit,
     qubit_map: list[QubitId],
-    backend: Backend,
+    platform: Platform,
     transpiler: Optional[Passes],
     initial_state=None,
     nshots=1000,
@@ -96,22 +115,26 @@ def execute_transpiled_circuit(
     transpiled_circ = transpile_circuits(
         [circuit],
         [qubit_map],
-        backend,
+        platform,
         transpiler,
     )[0]
+    if platform.name in AVAILABLE_PLATFORMS:
+        backend = construct_backend(backend="qibolab", platform=platform)
+    else:
+        backend = get_backend()
     return transpiled_circ, backend.execute_circuit(
         transpiled_circ, initial_state=initial_state, nshots=nshots
     )
 
 
-def dummy_transpiler(backend) -> Optional[Passes]:
+def dummy_transpiler(platform) -> Optional[Passes]:
     """
     If the backend is `qibolab`, a transpiler with just an unroller is returned,
     otherwise None.
     """
-    if backend.name == "qibolab":
+    if platform.name in AVAILABLE_PLATFORMS:
         unroller = Unroller(NativeGates.default())
-        return Passes(connectivity=backend.platform.topology, passes=[unroller])
+        return Passes(connectivity=platform.topology, passes=[unroller])
     return None
 
 
--- a/src/qibocal/auto/transpile.py
+++ b/src/qibocal/auto/transpile.py
@@ -1,11 +1,12 @@
 from typing import Optional
 
-from qibo import Circuit
+from qibo import Circuit, gates
 from qibo.backends import construct_backend, get_backend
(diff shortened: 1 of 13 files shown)


src/qibocal/protocols/randomized_benchmarking/standard_rb_2q.py
```diff
@@ -31,7 +31,6 @@ def _acquisition(
     targets: list[QubitPairId],
 ) -> RB2QData:
     """Data acquisition for two qubit Standard Randomized Benchmarking."""
-
```
andrea-pasquale, line 34: Probably this was unintentional.

src/qibocal/auto/transpile.py
```diff
@@ -97,19 +98,65 @@ def execute_transpiled_circuit(
...
+    pairs = list(platform.pairs.values())[0]
+    qubit = list(platform.qubits.values())[0]
+    two_qubit_natives = list(pairs.native_gates.raw.keys())
+    single_qubit_natives = list(qubit.native_gates.raw.keys())
+    # Solve Qibo-Qibolab mismatch
+    single_qubit_natives.append("RZ")
+    single_qubit_natives.append("Z")
+    single_qubit_natives.remove("RX12")
+    replacements = {
+        "RX": "GPI2",
```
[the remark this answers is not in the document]
Edoardo-Pedicillo: This dictionary is used to extract the proper native gates (the values) from the pulses defined in the platform (the keys). So when the RX pulse is in the platform, it means that the GPI2 gate is well defined and we can use it as native gates. This code is not covering the case when just the RX90 is defined (I will fix it).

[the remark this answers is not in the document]
andrea-pasquale: Understood, thanks!

src/qibocal/auto/transpile.py
```diff
@@ -97,19 +98,65 @@ def execute_transpiled_circuit(
...
 
 
+def get_natives(platform):
+    """
+    Return the list of native gates defined in the `platform`.
+    This function assumes the native gates to be the same for each
+    qubit and pair.
+    """
+    pairs = list(platform.pairs.values())[0]
+    qubit = list(platform.qubits.values())[0]
```
alecandido, line 113: It is often annoying that Python has no function to get the first element of an iterable (it is not LISP...), but in general, you should prefer using:
```suggestion
    pairs = next(iter(platform.pairs.values()))
    qubit = next(iter(platform.qubits.values()))
```
(if you have a list of 1M elements, you want to take the first one without creating a full duplicate...)

src/qibocal/auto/transpile.py
```diff
@@ -97,19 +98,65 @@ def execute_transpiled_circuit(
...
     )
 
 
+def get_natives(platform):
+    """
+    Return the list of native gates defined in the `platform`.
+    This function assumes the native gates to be the same for each
+    qubit and pair.
+    """
+    pairs = list(platform.pairs.values())[0]
```
alecandido, line 112: Maybe singular is more meaningful...
```suggestion
    pair = list(platform.pairs.values())[0]
```

src/qibocal/auto/transpile.py
```diff
@@ -97,19 +98,65 @@ def execute_transpiled_circuit(
...
+    two_qubit_natives = list(pairs.native_gates.raw)
+    single_qubit_natives = list(qubit.native_gates.raw)
+    # Solve Qibo-Qibolab mismatch
+    single_qubit_natives.append("RZ")
+    single_qubit_natives.append("Z")
+    single_qubit_natives.remove("RX12")
+    replacements = {
+        "RX": "GPI2",
+        "MZ": "M",
+    }
```
alecandido, line 123: This could be a top-level constant

src/qibocal/auto/transpile.py
```diff
@@ -97,19 +98,65 @@ def execute_transpiled_circuit(
...
+    new_single_natives = [replacements.get(i, i) for i in single_qubit_natives]
+    natives = new_single_natives + two_qubit_natives
+    return natives
```
alecandido, line 126: No need to give it a name, if you immediately return it.
```suggestion
    return new_single_natives + two_qubit_natives
```

src/qibocal/auto/transpile.py
```diff
@@ -97,19 +98,65 @@ def execute_transpiled_circuit(
...
+    This function assumes the native gates to be the same for each
+    qubit and pair.
+    """
+    pairs = list(platform.pairs.values())[0]
+    qubit = list(platform.qubits.values())[0]
+    two_qubit_natives = list(pairs.native_gates.raw)
+    single_qubit_natives = list(qubit.native_gates.raw)
+    # Solve Qibo-Qibolab mismatch
+    single_qubit_natives.append("RZ")
+    single_qubit_natives.append("Z")
```
alecandido, line 118: I wonder if `Z` should be somehow directly inferred from `RZ` by the transpiler itself...

(implying that is not platform's/Qibocal's business)

src/qibocal/auto/transpile.py
```diff
@@ -97,19 +98,65 @@ def execute_transpiled_circuit(
         backend,
         transpiler,
     )[0]
-
     return transpiled_circ, backend.execute_circuit(
         transpiled_circ, initial_state=initial_state, nshots=nshots
     )
 
 
+def get_natives(platform):
```
alecandido, line 106: Style proposal:
```suggestion
def natives(platform):
```
`get_` usually adds nothing...

src/qibocal/auto/transpile.py
alecandido, line 113: Is there a reason to start from parameters, instead of starting directly from the natives listing exposed by the platform?
https://github.com/qiboteam/qibolab/blob/bfb48cc2d9ec5e563110720889dc74acca6e604c/src/qibolab/backends.py#L64-L79

src/qibocal/auto/transpile.py
```diff
@@ -97,19 +98,65 @@ def execute_transpiled_circuit(
...
+        gate = getattr(gates, native)
+        if gate not in compiler.rules:
+
+            def rule(qubits_ids, platform, parameters=None):
+                if len(qubits_ids[1]) == 1:
+                    native_gate = platform.qubits[tuple(qubits_ids[1])].native_gates
+                else:
+                    native_gate = platform.pairs[tuple(qubits_ids[1])].native_gates
+                pulses = getattr(native_gate, native).pulses
+                return PulseSequence(pulses), {}
```
alecandido, line 144: Using closures is always delicate, so, I'd suggest moving it in a function on its own, at least will make it clear which are the captured variables

```py
def create_rule(native):
    def rule(qubits_ids, platform, parameters=None):
        if len(qubits_ids[1]) == 1:
            native_gate = platform.qubits[tuple(qubits_ids[1])].native_gates
        else:
            native_gate = platform.pairs[tuple(qubits_ids[1])].native_gates
        pulses = getattr(native_gate, native).pulses
        return PulseSequence(pulses), {}

    return rule
```

alecandido: P.S.: type hints would also help

src/qibocal/auto/transpile.py
```diff
@@ -97,19 +98,65 @@ def execute_transpiled_circuit(
...
+
+            backend.compiler[gate] = rule
```
alecandido, line 146: Also, usually it is better to avoid mutating objects, especially nested ones.
I could lift the suggestion by one level, and suggest creating the entire backend (here just returning the compiler, instead of modifying the backend). But let's limit to the compiler itself.

You could create an empty dictionary `rules`, and here do
```suggestion
            rules[gate] = rule
```
eventually setting the compiler out of the loops as:
```py
    backend.compiler = Compiler(rules=rules)
```
(in principle, you can avoid even modifying the dictionary `rules`, by defining it immediately through comprehension - but I get this is a bit unfamiliar).

src/qibocal/auto/transpile.py
```diff
@@ -97,19 +98,65 @@ def execute_transpiled_circuit(
...
     """
     If the backend is `qibolab`, a transpiler with just an unroller is returned,
     otherwise None.
     """
-    unroller = Unroller(NativeGates.default())
-    return Passes(connectivity=backend.platform.topology, passes=[unroller])
+    platform = backend.platform
+    native_gates = get_natives(platform)
+    set_compiler(backend, native_gates)
+    native_gates = list(map(lambda x: getattr(gates, x), native_gates))
```
alecandido, line 157: Comprehension is doing the same, and it is much clearer to read:
```suggestion
    native_gates = [getattr(gates, x) for x in native_gates]
```

*(those languages heavily focusing on `map` and `lambda`, i.e. functional ones, are also those who conceived the comprehension itself)*

src/qibocal/auto/transpile.py
```diff
@@ -97,19 +98,65 @@ def execute_transpiled_circuit(
...
+                pulses = getattr(native_gate, native).pulses
+                return PulseSequence(pulses), {}
+
+            backend.compiler[gate] = rule
+
+
 def dummy_transpiler(backend: Backend) -> Passes:
     """
     If the backend is `qibolab`, a transpiler with just an unroller is returned,
     otherwise None.
```
alecandido, line 152: Properly document that this function is not pure, since it has the side effect of modifying the input `backend` (in particular, affecting its `.compiler` attribute).

I know this seems pedantic, but losing track of this changes may have effects later on, when they will become unexpected.

alecandido (commented):
Mostly style comments, but those related to side effects and neglecting the existing `platform.natives` attributes are substantial

src/qibocal/auto/transpile.py
Edoardo-Pedicillo: Yes, but `natives` is already used as variable in this file.

src/qibocal/auto/transpile.py
Edoardo-Pedicillo: In theory yes, I am using Qibolab compiler and NativeGates convention to differentiate between the two.

src/qibocal/auto/transpile.py
Edoardo-Pedicillo: My idea at the beginning was to make the `set_compiler` function in each protocol that executes circuits explicit. I decided to put it here to make my life easier but I will open an issue about it.

src/qibocal/auto/transpile.py
alecandido: It is used locally, so there is no clash.

In particular, where it is used as a local variable, `get_natives` is not used, so there is no danger in shadowing.
In any case, if you want to strongly prevent clashes, you could always append a `_` to the local name (i.e. `natives_`), or anything similar.

However, this one was definitely about style. Feel free to ignore it :)

src/qibocal/auto/transpile.py
alecandido: > My idea at the beginning was to make the `set_compiler` function in each protocol that executes circuits explicit.

No need.

I agree with your choice, since there is nothing protocol-specific, so it should not be repeated over and over.

> to make my life easier

Everyone's life easier. Good choice :)

> I decided to put it here 

Even better, you could have put it here
https://github.com/qiboteam/qibocal/blob/f8f69de13b034cb2f0eccbe3d23e2b690bbf8c54/src/qibocal/auto/execute.py#L263-L264
if it were the single place where the backend was created.

Since unfortunately it is not, instead of this:
https://github.com/qiboteam/qibocal/blob/f8f69de13b034cb2f0eccbe3d23e2b690bbf8c54/src/qibocal/protocols/state_tomography.py#L105-L106
you could actually just make a single function, and invoke as
```py
    backend, transpiler = construct_qibolab_backend(platform)
```
considering that this double functions invocation is repeat over and over.

https://github.com/search?q=repo%3Aqiboteam%2Fqibocal%20construct_backend&type=code

(in any case, this one could be an issue - but it's especially relevant for `0.2`, since we won't maintain `0.1` for long)

src/qibocal/auto/transpile.py
alecandido: Wait a second: this function is used just to feed `set_compiler`, in which you're doing the following
https://github.com/qiboteam/qibocal/blob/e1c00472fe87e9bb23a5db9ea1149aa6fa363644/src/qibocal/auto/transpile.py#L151-L154

To me, that means that you're trying to support rules beyond the default compiler. Including `iSWAP`, if available.

I'm sure I misunderstood something in your reply...

src/qibocal/auto/transpile.py
alecandido: Fine, but is there a reason to differentiate?

I.e. with the current compiler, the `RZ` will always be implemented as virtual phase, and that will be the same for the `Z` as well...

src/qibocal/auto/transpile.py
```diff
@@ -97,19 +104,71 @@ def execute_transpiled_circuit(
...
+
+    return rule
+
+
+def set_compiler(backend, natives_):
+    """
+    Set the compiler to execute the native gates defined by the platform.
+    """
+    compiler = backend.compiler
+    print("EEEEEEE", backend.natives)
```
alecandido, line 147: It seems I forgot this in the previous review: I guess this is just a leftover.

PR #1085, opened 2025-02-13
Updating API documentation

### LINKED ISSUE #1084: API example from documentation
The ["How to use Qibocal as a library" example](https://qibo.science/qibocal/stable/tutorials/advanced.html) from the documentation seems a bit outdated, in particular running
```py
import pathlib

from qibocal.auto.execute import Executor
from qibocal.auto.mode import ExecutionMode
from qibocal.protocols import t1_signal

# allocate platform
platform = create_platform("dummy")

#create executor
executor = Executor.create(
  platform=platform,
  output=pathlib.Path("experiment_data")
)

t1_params = {
    "id": "t1_experiment",
    "targets": [0],  # we are defining here which qubits to analyze
    "operation": "t1_signal",
    "parameters": {
        "delay_before_readout_start": 0,
        "delay_before_readout_end": 20_000,
        "delay_before_readout_step": 50,
    },
}

executor.run_protocol(t1_signal, t1_params, ExecutionMode.ACQUIRE)
executor.run_protocol(t1_signal, t1_params, ExecutionMode.FIT)
```
leads to several errors, such as
```py
TypeError: Executor.create() got an unexpected keyword argument 'output'
```
in the definition of `executor`,
```py
AttributeError: 'dict' object has no attribute 'parameters'
```
as `t1_params` needs to be an `Action` object instead of `dict` and
```py
TypeError: unsupported operand type(s) for /: 'NoneType' and 'str'
```
when the fit is called.

andrea-pasquale:
Closes #1084, it could be expanding a bit more. Let me know if more details are necessary.

### DIFF
--- a/doc/source/tutorials/advanced.rst
+++ b/doc/source/tutorials/advanced.rst
@@ -9,72 +9,49 @@ Qibocal also allows executing protocols without the standard :ref:`interface <in
 In the following tutorial we show how to run a single protocol using Qibocal as a library.
 For this particular example we will focus on the `t1_signal protocol
 <https://github.com/qiboteam/qibocal/blob/main/src/qibocal/protocols/coherence/t1_signal.py>`_ (see also :ref:`t1`).
+The fastest way consists in using the `Executor` class in the following way
 
 .. code-block:: python
 
-    import pathlib
-    from qibolab import create_platform
-
     from qibocal.auto.execute import Executor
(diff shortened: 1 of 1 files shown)


alecandido (approved):

stavros11:
Thanks @andrea-pasquale, this is much more than concise compared to what I was doing.

PR #1089, opened 2025-02-17
Update in serialize.load method. 

jevillegasdatTII:
Currently, when the data key is a tuple (for instance when a QubitPairID is used as key in the stored data) the Jason keys are stored as "[['q1','q2'], 'some_key', 'some_other_key']". This then gets converted into a tuple as (['q1','q2'], 'some_key', ...) . This immediatly generates an error since the first item, a list, is unhashable and therefore cannot be used as a key in the dictionary. Instead, any nested list in the keys need to also be converted to a tuple.

### DIFF
--- a/src/qibocal/auto/serialize.py
+++ b/src/qibocal/auto/serialize.py
@@ -18,15 +18,15 @@ def serialize(raw: dict):
     }
 
 
+def _nested_list_to_tuples(raw):
+    """Convert nested lists to nested tuples."""
+    if isinstance(raw, list):
+        return tuple([_nested_list_to_tuples(item) for item in raw])
+    return raw
+
+
 # TODO: distinguish between tuples and strings
 def load(key):
     """Evaluate key converting string of lists to tuples."""
     raw_load = json.loads(key)
-
-    def to_tuple(raw):
-        if isinstance(raw, list):
-            return tuple([to_tuple(item) for item in raw])
-        else:
-            return raw
-
-    return to_tuple(raw_load)
+    return _nested_list_to_tuples(raw_load)
--- a/src/qibocal/auto/serialize.py
+++ b/src/qibocal/auto/serialize.py
@@ -22,10 +22,11 @@ def serialize(raw: dict):
 def load(key):
     """Evaluate key converting string of lists to tuples."""
     raw_load = json.loads(key)
+
     def to_tuple(raw):
         if isinstance(raw, list):
             return tuple([to_tuple(item) for item in raw])
         else:
             return raw
-        
+
     return to_tuple(raw_load)
--- a/src/qibocal/auto/serialize.py
+++ b/src/qibocal/auto/serialize.py
@@ -22,6 +22,10 @@ def serialize(raw: dict):
 def load(key):
     """Evaluate key converting string of lists to tuples."""
     raw_load = json.loads(key)
-    if isinstance(raw_load, list):
-        return tuple(raw_load)
-    return raw_load
+    def to_tuple(raw):
+        if isinstance(raw, list):
+            return tuple([to_tuple(item) for item in raw])
+        else:
+            return raw
+        
+    return to_tuple(raw_load)


alecandido:
@jevillegasdatTII I hope we'll soon get rid of this manual (de)serialization. In the meanwhile: what are actually `some_key` and `some_other_key`? Why is there a need to go beyond `QubitId` and `QubitPairId` to index the dictionary?

I acknowledge that the recursive converter is a way to turn all lists in tuples. But while the conversion itself is not adding much complexity, accepting nested structures is, and this may arise in other places, propagating the need for handling.

If possible, I'd try to avoid further increasing the complexity of indexing.

(Btw, indexing with qubits and qubit pairs could be avoided as well, that's just my fault: since all qubits in an experiment share the exact same experiment, they could be expanded just in a further external dimension for the data array, since the internal ones are homogeneous. I should have saved the indices in a separate list in the JSON, and match them with the array dimension by position, instead of using them as keys...)

jevillegasdatTII:
Hey @alecandido, In my case I am storing data for a cross-resonance gate calibration as 
```python
class CrossResonanceCNOTData(Data):
         data: dict[(QubitPairId, str, str), 
               npt.NDArray[CrossResonanceType]] = field(default_factory=dict)
```
where the first string is the state of the control qubit (in {0,1}) and the second string is the basis in which I am measuring (in {X,Y,Z}). For reference the protocol iterates over the two states of the control qubit and the 3 basis for the measurement.

I probably can use just the pairID and then store everything else in separate items of the data class (a list with the corresponding setup and basis?), but it felt cleaner doing it this way.

alecandido:
> I probably can use just the pairID and then store everything else in separate items of the data class (a list with the corresponding setup and basis?), but it felt cleaner doing it this way.

As I said, the fact that it feels cleaner is partially my fault, since I structured it as a dictionary in the first place.

However, if you see the index structure, you see that each element how the tuple is a *dimension*, because the dimensions are regular (each index is independent of the others, and you expect all the combinations), and not ragged. And even if the indices were not an exact Cartesian product, the values are expected to have all the same shape anyhow. So, at most, you would need to linearize the indices, which is effectively what you're already doing with a `dict`.

Storing it in a dictionary is fully losing the structure, and adds complexity, requiring cascading fixes (as the tuples' requirement in the first place, and this fix now).

In any case, I'm not directly against this fix temporarily, but I'm suggesting to fully change the paradigm everywhere. Not immediately, but soon (and I'll take care of it).

src/qibocal/auto/serialize.py
```diff
@@ -22,6 +22,11 @@ def serialize(raw: dict):
...
     raw_load = json.loads(key)
-    if isinstance(raw_load, list):
-        return tuple(raw_load)
-    return raw_load
+
+    def to_tuple(raw):
+        if isinstance(raw, list):
+            return tuple([to_tuple(item) for item in raw])
+        else:
+            return raw
```
alecandido, line 30: It would be better to lift this definition top-level. To signal that the function is intended for internal use, just prepend an underscore (i.e. turn the name to `_to_tuple`).

Indeed, Python functions have always the capability to capture the environment. This means that they can become closures, without any explicit marker.

Since, in this case, you do not need that, if it happens it would be purely accidental. So, better to keep it outside, in the spirit of the principle of least power.

src/qibocal/auto/serialize.py
```diff
@@ -22,6 +22,11 @@ def serialize(raw: dict):
 def load(key):
     """Evaluate key converting string of lists to tuples."""
     raw_load = json.loads(key)
-    if isinstance(raw_load, list):
-        return tuple(raw_load)
-    return raw_load
+
+    def to_tuple(raw):
```
alecandido, line 26: Maybe, even better some more explicit name, like `nested_to_tuples`, or `recursive_tuples`.

Names are relevant, since they save times and effort to understand the flow. 
I'm not incredibly skilled myself, but this subtracts nothing to their importance.

src/qibocal/auto/serialize.py
alecandido, line 30:
```suggestion
        if isinstance(raw, list):
            return tuple([to_tuple(item) for item in raw])
        return raw
```
if the former branch is returning, you do not need an `else:` branch (this is flagged by Pylint as well).

jevillegasdatTII: I agree

src/qibocal/auto/serialize.py
jevillegasdatTII: I agree

alecandido (approved):
To me, this is good to go.

I hope it won't be needed for long. But for as long as we keep using dictionaries, this is no worse that what was already in there.

alecandido:
@andrea-pasquale @Edoardo-Pedicillo tests are failing, but it seems pretty random (since they passed on a platform, and failed on others).
https://github.com/qiboteam/qibocal/actions/runs/13370451195/attempts/1?pr=1089

Do you have any idea about why this is happening?

jevillegasdatTII:
I wonder if this comment is related to the same situation with the key deserialization https://github.com/qiboteam/qibocal/blob/7a25e48b442ecddfece20567b32b80fd5f584bf8/src/qibocal/protocols/two_qubit_interaction/virtual_z_phases.py#L67

Edoardo-Pedicillo:
> @andrea-pasquale @Edoardo-Pedicillo tests are failing, but it seems pretty random (since they passed on a platform, and failed on others). [qiboteam/qibocal/actions/runs/13370451195/attempts/1](https://github.com/qiboteam/qibocal/actions/runs/13370451195/attempts/1)
> 
> Do you have any idea about why this is happening?
> 
> EDIT: it keeps failing with a certain frequency, but not always [qiboteam/qibocal/actions/runs/13370451195/job/37353546184](https://github.com/qiboteam/qibocal/actions/runs/13370451195/job/37353546184)

This issue is probably related to the failing of the fit function. I had a chat with @andrea-pasquale and we have some possible solutions (even drop this specific test ...). I will open an issue.

Issue #1095, opened 2025-02-19
Refactor after update to qibolab 0.2


andrea-pasquale:
Now that we have recently completed the update to Qibolab 0.2 in #990. We can proceed with a refactor aimed at simplifying the code and reducing code duplication. Here is a first tentative list of features that we would like to implement:

- [ ] Pass backend to acquisition function. This should solve some of the problems originally addressed in #1076 which proposed as a solution the creation of a local backend inside the acquisition function itself. If we could pass directly the backend to the acquisition function there should be no need to recreate it once again.
- [ ] Connected to #1053 we would like to remove `register_qubit` and find an easier way to process the data retrieved from qibolab. A possible solution suggested by @hay-k consists in mapping each call to `platform.execute()` to a dedicated attribute inside the `Data` class which could be later stored as an array. Any post-processing on the data should be done by external function (not class methods), in order to easily test them. For storing the `Data` class itself we could try to come up with a mechanism to store together all these different arrays or for time being, given that we are not dealing with huge data just store each one in a separated file.
- [ ] Both acquisition and fit function for each protocol should tackle only one qubit/pair at a time. In this way we should be able to remove most of the for loops going on in both the acquisition and the fitting. We could consider to have an `Acquisition` object (target one qubit or pair)
```python
from qibolab import PulseSequence, Sweeper
from qibolab._core.execution_parameters import ExecutionParameters
from dataclasses import dataclass

@dataclass
class Acquisition:
    sequence: PulseSequence
    sweepers: list[Sweepers]
    parameters: ExecutionParameters
``` 
Given two `Acquisition` objects Qibocal could try to perform a parallel execution and throw an error if it is not possible. The user should be able to specify if the execution should be in parallel for all targets.

- [ ] Avoid disconnection when running circuits (this is mainly a Qibolab feature request) to avoid wasting time connecting and reconnecting for protocols like the RB.
- [ ] Restore simulation #913 which might be tricky to do now that we have introduced `CalibrationPlatform` in #1021.
- [ ] Dump data as soon as they are received #919.
- [ ] Add unit test (we could start probably after removing `register_qubit`)
- [ ] Expose fitting_parameters #852 , in particular we don't really know if it makes to store all fitting parameters however we could extract reasonable guess for the fitted parameters from the platform itself and the calibration parameters.

alecandido:
[...]

PR #1097, opened 2025-02-21
Attempt to close #919

### LINKED ISSUE #919: Dump data as soon as they are received
https://github.com/qiboteam/qibocal/pull/909#issuecomment-2209412462

andrea-pasquale:
Initial attempt to close #919.
I know that in theory this is breaking the initial proposal in which `task.run` should not directly perform the dumping, however since both acquisition and fitting are performed there in order to dump the data before performing the post-processing I didn't see another valid solution.

### DIFF
--- a/src/qibocal/auto/history.py
+++ b/src/qibocal/auto/history.py
@@ -74,21 +74,30 @@ class History:
             instance.push(Completed.load(protocol))
         return instance
 
+    def _pending_task_id(self, _id: Id) -> TaskId:
+        """Retrieve the TaskId of a given task."""
+        return TaskId(id=_id, iteration=len(self._tasks[_id]))
+
+    def _executed_task_id(self, _id: Id) -> TaskId:
+        """Retrieve the TaskId of a given task."""
+        return TaskId(id=_id, iteration=len(self._tasks[_id]) - 1)
+
     def push(self, completed: Completed) -> TaskId:
         """Adding completed task to history."""
         id = completed.task.id
         self._tasks[id].append(completed)
-        task_id = TaskId(id=id, iteration=len(self._tasks[id]) - 1)
+        task_id = self._executed_task_id(id)
         self._order.append(task_id)
         return task_id
 
-    @staticmethod
-    def route(task_id: TaskId, folder: Path) -> Path:
+    def task_path(self, task_id: TaskId, folder: Optional[Path]) -> Path:
         """Determine the path related to a completed task given TaskId.
 
         `folder` should be usually the general output folder, used by Qibocal to store
         all the execution results. Cf. :class:`qibocal.auto.output.Output`.
         """
+        if folder is None:
+            return None
         return folder / "data" / f"{task_id}"
 
     def flush(self, output: Optional[Path] = None):
@@ -97,9 +106,7 @@ class History:
         Specifying `output` is possible to select which folder should be considered as
         the general Qibocal output folder. Cf. :class:`qibocal.auto.output.Output`.
         """
-        for task_id, completed in self.items():
-            if output is not None:
-                completed.path = self.route(task_id, output)
+        for _, completed in self.items():
             completed.flush()
 
     # TODO: implement time_travel()
--- a/src/qibocal/auto/output.py
+++ b/src/qibocal/auto/output.py
@@ -193,8 +193,7 @@ class Output:
         self._export_stats()
         (path / META).write_text(json.dumps(self.meta.dump(), indent=4))
 
-        # dump tasks
-        self.history.flush(path)
+        self.history.flush()
 
         # update platform
         if self.platform is not None:
@@ -247,13 +246,14 @@ class Output:
                 and completed.results is not None
             ):
                 raise KeyError(f"{task_id} already contains fitting results.")
-
             # TODO: this is a plain hack, to be fixed together with the task lifecycle
-            # refactor
             self.history._tasks[task_id.id][task_id.iteration] = completed.task.run(
-                platform=self.platform, mode=mode, folder=completed.path
+                platform=self.platform,
+                mode=mode,
+                folder=self.history.task_path(
+                    self.history._executed_task_id(task_id.id), output
+                ),
             )
-            self.history.flush(output)
-
+            self.history.flush()
             if update and completed.task.update:
                 completed.update_platform(platform=self.platform)
--- a/src/qibocal/auto/task.py
+++ b/src/qibocal/auto/task.py
@@ -134,7 +134,6 @@ class Task:
             self.action.targets = targets
 
         completed = Completed(self, folder)
-
         try:
             if platform is not None:
                 if self.parameters.nshots is None:
@@ -153,7 +152,6 @@ class Task:
         except (RuntimeError, AttributeError):
             operation = dummy_operation
             parameters = DummyPars()
-
         if ExecutionMode.ACQUIRE in mode:
             if operation.platform_dependent and operation.targets_dependent:
                 completed.data, completed.data_time = operation.acquisition(
@@ -161,13 +159,14 @@ class Task:
                     platform=platform,
                     targets=self.targets,
                 )
-
             else:
                 completed.data, completed.data_time = operation.acquisition(
                     parameters, platform=platform
                 )
+            completed.dump()
         if ExecutionMode.FIT in mode:
             completed.results, completed.results_time = operation.fit(completed.data)
+            completed.dump()
         return completed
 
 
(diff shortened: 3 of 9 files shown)


alecandido:
> I know that in theory this is breaking the initial proposal in which `task.run` should not directly perform the dumping, however since both acquisition and fitting are performed there in order to dump the data before performing the post-processing I didn't see another valid solution.

To be fair, I'm not even able to locate this proposal...
The only mention of changes about `Task` I've found is #908. Which is just proposing to drop the class, but not the function.

So, I would not worry too much about where it is located, and I will just review the code as soon as you'll be ready.

src/qibocal/protocols/resonator_spectroscopy.py
```diff
@@ -150,14 +150,14 @@ class ResonatorSpectroscopyData(Data):
     will apply a minus to the phase data."""
     data: dict[QubitId, npt.NDArray[ResSpecType]] = field(default_factory=dict)
     """Raw data acquired."""
-    power_level: Optional[PowerLevel] = None
+    power_level: Optional[PowerLevel] = PowerLevel.low
```
alecandido, line 153: Just noticed through scanning: do you still need this to be `Optional` with the new default?

src/qibocal/protocols/state_tomography.py
```diff
@@ -72,10 +72,11 @@ def save(self, path):
 
     @classmethod
     def load(cls, path):
-        circuit = Circuit.from_dict(json.loads((path / CIRCUIT_PATH).read_text()))
-        data = super().load_data(path, DATAFILE)
-        params = super().load_params(path, DATAFILE)
-        return cls(data=data, circuit=circuit, targets=params["targets"])
+        if (path / CIRCUIT_PATH).exists():
```
alecandido, line 75: Is it possible that it does not exist? In case, you should raise an error.

Instead, with the current arrangement you're just returning `None`, because the only `return` statement is scoped within the `if` branch.

src/qibocal/protocols/two_qubit_state_tomography.py
```diff
@@ -60,11 +60,12 @@ def save(self, path):
 
     @classmethod
     def load(cls, path):
-        return cls(
-            data=super().load_data(path, DATAFILE),
-            ideal=super().load_data(path, SIMULATED_DENSITY_MATRIX),
-            simulated=QuantumState.load(path / "simulated.json"),
-        )
+        if (path / "simulated.json").exists():
```
alecandido, line 63: Same as the former

src/qibocal/auto/task.py
```diff
@@ -223,19 +222,22 @@ def results(self, value):
...
-        if self.path is None:
-            raise ValueError("No known path where to dump execution results.")
-
-        self.path.mkdir(parents=True, exist_ok=True)
-        self.task.dump(self.path)
-        if self._data is not None:
-            self._data.save(self.path)
-        if self._results is not None:
-            self._results.save(self.path)
+        if self.path is not None:
```
alecandido, line 225: To make it simpler, just do the opposite
```suggestion
        if self.path is None:
            return None
```

I see you wanted to drop the error,  but not doing anything in a function (as in the case of an empty `else`) is fully equivalent to `return None`.
And you'll save an indentation level (in general, early returns are more readable).

src/qibocal/auto/output.py
```diff
@@ -247,13 +246,14 @@ def process(
...
-                platform=self.platform, mode=mode, folder=completed.path
+                platform=self.platform,
+                mode=mode,
+                folder=self.history.task_path(
+                    self.history._executed_task_id(task_id.id), output
+                ),
             )
-            self.history.flush(output)
-
+            self.history.flush()
```
alecandido, line 257: If everything is supposed to be saved during the execution, you should not need to flush at all.

alecandido: Maybe we can fully get rid of it (including within `Completed` tasks).

It may be still useful, for very long histories, when you just what to load data for a section, while keeping the index of the full history.
But I'm not sure if it's worth investing time on this point now...
(in particular, we're not going to work with very long histories for a lot of time...)

Or, maybe, we could just adopt a different solution, never storing data nested in `History`, and just generate `Completed` object individually.
I.e. you just load the index in `History`, and you load data in a `Completed` object only if you need to access it. If you do not need, no `Completed` instance is generated. If you need to have in multiple places, and it is inconvenient to pass them around, then you'll load them multiple times.
Such that the memory management is just up to the garbage collector, and we never hold an object possibly become huge (as a `History` full of data)

alecandido (commented):
First chunk

src/qibocal/auto/output.py
andrea-pasquale: I decided to keep it at the beginning to avoid breaking the former the implementation. I also agree that probably we don't need it anymore. I dropped in  https://github.com/qiboteam/qibocal/pull/1097/commits/177e2539346053dedb3424cc3b2900171a87dc6f.

src/qibocal/protocols/resonator_spectroscopy.py
```diff
@@ -150,14 +150,14 @@ class ResonatorSpectroscopyData(Data):
...
+    power_level: PowerLevel = PowerLevel.low
     """Power regime of the resonator."""
 
     @classmethod
     def load(cls, path):
         obj = super().load(path)
         # Instantiate PowerLevel object
-        if obj.power_level is not None:  # pylint: disable=E1101
+        if hasattr(obj, "power_level"):  # pylint: disable=E1101
             obj.power_level = PowerLevel(obj.power_level)  # pylint: disable=E1101
```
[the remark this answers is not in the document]
andrea-pasquale: This now happen because I'm trying to make sure that we don't dump twice:
https://github.com/qiboteam/qibocal/blob/177e2539346053dedb3424cc3b2900171a87dc6f/src/qibocal/auto/task.py#L230-L239
The code will now always do a first attempt at loading this is why I added those lines here and similar ones in the state tomography protocols. Perhaps there is a nicer solution. I'm open to suggestions.

alecandido:
While the `.flush()` part of the last comment has been addressed, I would keep discussing the second half of that note.

> Or, maybe, we could just adopt a different solution, never storing data nested in `History`, and just generate `Completed` object individually.
> I.e. you just load the index in `History`, and you load data in a `Completed` object only if you need to access it. If you do not need, no `Completed` instance is generated. If you need to have in multiple places, and it is inconvenient to pass them around, then you'll load them multiple times.
> Such that the memory management is just up to the garbage collector, and we never hold an object possibly become huge (as a `History` full of data)

I'm not sure whether this PR is the best place to do it. On the one side, it is certainly about the data handling. On the other, it is much more specific, addressing an equivalently specific issue (i.e. dumping data immediately, as the title says).
So, maybe we should postpone the discussion to an issue on its own, to avoid bloating this PR and merge it faster.

andrea-pasquale:
> While the `.flush()` part of the last comment has been addressed, I would keep discussing the second half of that note.
> 
> > Or, maybe, we could just adopt a different solution, never storing data nested in `History`, and just generate `Completed` object individually.
> > I.e. you just load the index in `History`, and you load data in a `Completed` object only if you need to access it. If you do not need, no `Completed` instance is generated. If you need to have in multiple places, and it is inconvenient to pass them around, then you'll load them multiple times.
> > Such that the memory management is just up to the garbage collector, and we never hold an object possibly become huge (as a `History` full of data)
> 
> I'm not sure whether this PR is the best place to do it. On the one side, it is certainly about the data handling. On the other, it is much more specific, addressing an equivalently specific issue (i.e. dumping data immediately, as the title says). So, maybe we should postpone the discussion to an issue on its own, to avoid bloating this PR and merge it faster.

I see your point. However I would consider it as low priority compared to the other points that have already been raised in #1095. I will open an issue for now.

src/qibocal/auto/history.py
```diff
@@ -99,33 +99,34 @@ def load(cls, path: Path) -> "History":
...
-    @staticmethod
-    def route(task_id: TaskId, folder: Path) -> Path:
+    def task_path(self, task_id: TaskId, folder: Optional[Path]) -> Path:
         """Determine the path related to a completed task given TaskId.
 
         `folder` should be usually the general output folder, used by Qibocal to store
         all the execution results. Cf. :class:`qibocal.auto.output.Output`.
         """
+        if folder is None:
+            return None
```
alecandido, line 125: Conflicting with the type hint. If this has to be the case, the return type should be `Optional[Path]`

src/qibocal/protocols/resonator_spectroscopies/resonator_spectroscopy.py
```diff
@@ -150,15 +150,15 @@ class ResonatorSpectroscopyData(Data):
...
     """Power regime of the resonator."""
 
     @classmethod
     def load(cls, path):
         obj = super().load(path)
         # Instantiate PowerLevel object
-        if obj.power_level is not None:
-            obj.power_level = PowerLevel(obj.power_level)
+        if hasattr(obj, "power_level"):  # pylint: disable=E1101
+            obj.power_level = PowerLevel(obj.power_level)  # pylint: disable=E1101
```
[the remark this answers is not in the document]
andrea-pasquale: Fixed in https://github.com/qiboteam/qibocal/pull/1097/commits/1bbc87ee8078b68d1d89a7636595cfa7cbc9ab9f

src/qibocal/auto/task.py
```diff
@@ -227,21 +226,21 @@ def results(self, value):
...
+            self.task.dump(self.path)
         if self._data is not None:
-            self._data.save(self.path)
-        if self._results is not None:
+            if self.task.operation.data_type.load(self.path) is None:
+                self._data.save(self.path)
+        if (
+            self.task.operation.results_type.load(self.path) is None
+            and self._results is not None
+        ):
```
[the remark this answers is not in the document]
andrea-pasquale: I think I was overcomplicating some stuff. I've now simplified the implementation.

alecandido (approved):
I didn't think too hard whether it is optimal or not.

But, to be fair, it seems reasonable. At least at a shallow look.
And, at some point, we may have to reconsider quite a large part of what is around it anyhow. So, it should be worth to merge this anyhow.

(If I remember correctly, the rationale for not doing it before was to decouple as much as possible the file handling from the execution - however, this should not happen at the cost of delaying too much saving operations; so, it is certainly better to restore it; so, having fixed it is a good thing on its own)

PR #1099, opened 2025-02-24
Refactor integration weights protocol

andrea-pasquale:
The previous implementation of the protocol to optimize the integration weights was based on LaboneQ and I don't think that we ever used it with QM. I've refactored the code to have a more instrument agnostic version which is assuming that with `AcquisitionType.RAW` the samples are returned without demodulation. 
In the protocol I am now doing the demodulation on software (with the addition of a low pass filter to remove fast-rotating terms $\propto \cos(2\omega_{IF}t)$.

It seems that the readout fidelity improves http://login.qrccluster.com:9000/Clqog-92SlKblXqbMVo-Vw==/, however it might be good if someone else takes a look at the code.

Requires a few changes in the QM driver https://github.com/qiboteam/qibolab/pull/1155


Edoardo-Pedicillo (approved):
It looks good to me,  some documentation  of the fit function is missing.

src/qibocal/protocols/signal_experiments/calibrate_state_discrimination.py
```diff
@@ -34,6 +33,23 @@ class CalibrateStateDiscriminationParameters(Parameters):
...
+    Currently it assumes that instruments with LOs is first one.
+    """
+    controller = next(
+        iter(
+            inst
+            for inst in platform.instruments.values()
+            if isinstance(inst, Controller)
+        )
+    )
+    probe = controller.channels[platform.qubits[qubit].probe]
```
stavros11, line 48: I think it is possible to access the channel directly from the `platform`:
```suggestion
    probe = platform.channels[platform.qubits[qubit].probe]
```
which should also lift the assumption about the instrument being the first one.

tests/platforms/mock/platform.py
```diff
@@ -0,0 +1,45 @@
...
+    """Create dummy hardware configuration based on the dummy instrument."""
+    qubits = {}
+    channels = {}
+    # attach the channels
+    pump_name = "twpa_pump"
+    for q in range(2):
+        drive12 = f"{q}/drive12"
+        qubits[q] = qubit = Qubit.default(q, drive_qudits={(1, 2): drive12})
+        channels |= {
+            qubit.probe: IqChannel(device="pump_name", mixer=None, lo="01/probe_lo"),
```
stavros11, line 22: Not very important but why is the `device="pump_name"`?

stavros11 (approved):

src/qibocal/protocols/signal_experiments/calibrate_state_discrimination.py
andrea-pasquale: Thanks, I was looking exactly for something like this.

tests/platforms/mock/platform.py
andrea-pasquale: Probably I it should have been an f-string or I should have just passed `pump_name`.
Thanks for catching this.

stavros11: I think the `pump_name` is for the TWPA pump local oscillator, which we are already passing to the `AcquisitionChannel`. For the probe channel it should probably be a different device (like qblox, QM, etc.), but for the mock platform it will probably work even without device, since you are not passing any device for the drive and flux either.

andrea-pasquale: Oh I see, at this point we can just remove it if it is not mandatory.

Issue #926, opened 2024-07-09
Alleviate or document environment dependence


alecandido:
After #869 there will be a default executor, already instantiated to be ready to be used.

https://github.com/qiboteam/qibocal/blob/d63bc73efd189cf316c7bd31bde6341af6151a3c/src/qibocal/__init__.py#L10

The best would be to only instantiate it if needed. But this may not be completely possible, in order to retain the `from qibocal.routines import ...` feature.

If present, it will rely on the `QIBO_PLATFORM` environment variable, 
https://github.com/qiboteam/qibocal/blob/d63bc73efd189cf316c7bd31bde6341af6151a3c/src/qibocal/auto/execute.py#L100-L108
thus we should be very careful in documenting that this will interfere with the execution, even when you're not using it.

A more complicate option (but more intuitive for the user) would be to create the executor immediately, but postpone the creation of the platform, e.g. until the first routine call.

andrea-pasquale:
I was thinking about this yesterday... Perhaps it would be better to rely on `dummy` by default to avoid issues.
If `QIBO_PLATFORM` is pointing to an outdated platform the code fail.

alecandido:
Yes, it is definitely something to take care of. I'm not even sure it should be `dummy`, the solution I like the most is the latest.

But most of the time it will already be `dummy`, if that variable is unset (which maybe it should be even on the cluster), so it is not that urgent...

alecandido:
It's a separate issue, but I even wonder whether we need the default executor at all. The context manager (`with Executor.open(...) as e:`) pattern seems actually good enough on its own...

Dropping it would limit even more the code that executes at import-time.

PR #1101, opened 2025-02-25
Using custom `dummy` platform

andrea-pasquale:
In this PR I'm attempting to define a custom "dummy" platform in Qibocal (which is currently called `fake` but if there are better proposal I'm happy to change the name). The motivation started from #1099 where I've noticed that the dummy was missing a local oscillator making it impossible to run the integration weights protocol with the dummy available in Qibolab.

In general I believe that this could be beneficial to further remove some nested dependencies between Qibolab and Qibocal.
For example currently the calibration parameters for `dummy` where stored in Qibocal while the `parameters.json` is currently in Qibolab.
The new `fake` platform only has 2 qubits with a coupler, since I don't believe that for testing purposes we need more than that.

Currently the new platform is added at import time by adding the corresponding directly to `QIBOLAB_PLAFORMS`.

- [x] update documentation

EDIT: after some internal discussions we have decided to rename `fake` to `mock`


andrea-pasquale:
With the latest changes I've defined the `mock`platform which generate random numbers only in `tests`.
For this reason I had to drop the `DEFAULT_EXECUTOR` which was using `dummy` as discussed in #926.
Similarly to what is already happening in #1103 I've added a `platforms` folder in qibocal where I've added a symlink to `mock`.

src/qibocal/auto/execute.py
```diff
@@ -104,7 +104,7 @@ def create(cls, name: str, platform: Union[CalibrationPlatform, str, None] = Non
             platform
             if isinstance(platform, CalibrationPlatform)
             else create_calibration_platform(
-                platform if platform is not None else "dummy"
+                platform if platform is not None else "mock"
```
alecandido, line 107: At this point, we should not need to keep `None` as the default, and just use `"mock"` in place of `None` in the parameter specification.

(of course, it's not strictly related to this PR, and it could have been done even before)

tests/conftest.py
```diff
@@ -2,9 +2,27 @@
...
+
+def set_platform_profile():
+    os.environ[PLATFORMS] = str(Path(__file__).parent / "platforms")
+
+
+@pytest.fixture(scope="module", params=TESTING_PLATFORM_NAMES)
+def platform(request):
+    """Dummy platform to be used when there is no access to QPU."""
+    set_platform_profile()
+    return create_calibration_platform(request.param)
```
alecandido, line 28: Instead of returning, `yield` and then reset the environment. Such that the fixture is fully scoped where it is applied, without leaking state modifications out of its scope.

tests/platforms/mock/platform.py
```diff
@@ -0,0 +1,45 @@
...
+        "dummy": DummyInstrument(address="0.0.0.0", channels=channels),
+        pump_name: DummyLocalOscillator(address="0.0.0.0"),
+    }
+    return Hardware(instruments=instruments, qubits=qubits, couplers=couplers)
+
+
+def create() -> Platform:
+    """Create a dummy platform using the dummy instrument."""
+    hardware = create_mock_hardware()
+    return Platform.load(path=FOLDER, **vars(hardware))
```
alecandido, line 45: This is only needed for `dummy`, to avoid breaking the API (since we introduced it in Qibolab after v0.2.0). However, you can just drop this function, and rename the former `create()`, unless you import this file directly. `create_platform()` will detect it and provide the `Platform.load()` call on its own.

alecandido (commented):
Thanks @andrea-pasquale, everything looks good.

I'm only providing minor suggestions.

platforms/README.md
```diff
@@ -0,0 +1,15 @@
+# Example platforms
+
+In this folder there are a few example platforms that can be used with Qibocal. To use these platforms `platforms` folder needs to be added to `QIBOLAB_PLATFORMS` environment variable. If the qibocal directory is cloned in the `HOME` directory this can be done with
```
Edoardo-Pedicillo, line 3:
```suggestion
In this folder there are a few example platforms that can be used with Qibocal. To use these platforms, `platforms` folder needs to be added to `QIBOLAB_PLATFORMS` environment variable. If the qibocal directory is cloned in the `HOME` directory this can be done with
```
Missing coma.

Edoardo-Pedicillo (approved):

tests/platforms/mock/platform.py
andrea-pasquale: I tried to do exactly as you said but when running test I'm getting an error `RuntimeError: Cannot establish connection to instrument dummy. Error captured: ''dict' object has no attribute 'connect''`.
Given that, if I am not wrong, this feature is only available in qibolab `main` and is not available in a release I suggest to open an issue and skip this comment for now.

alecandido: > Given that, if I am not wrong, this feature is only available in qibolab `main` and is not available in a release I suggest to open an issue and skip this comment for now.

Actually, it should be available in v0.2.3, since https://github.com/qiboteam/qibolab/pull/1123 was introduced in that release.
However, I fixed something after that release in https://github.com/qiboteam/qibolab/pull/1151, and I was trying to remember whether it was needed sometimes or always...

tests/conftest.py
```diff
@@ -2,9 +2,33 @@
...
+    if PLATFORMS in os.environ:
+        del os.environ[PLATFORMS]
+
+
+@pytest.fixture(scope="module", params=TESTING_PLATFORM_NAMES)
+def platform(request):
+    """Dummy platform to be used when there is no access to QPU."""
+    set_platform_profile()
+    yield create_calibration_platform(request.param)
+    unset_platform_profile()
```
alecandido, line 34: To fully revert the status, you should something like the following:
```py
@pytest.fixture(scope="module", params=TESTING_PLATFORM_NAMES)
def platform(request):
    """Dummy platform to be used when there is no access to QPU."""
    old_platforms = os.environ.get(PLATFORMS)
    os.environ[PLATFORMS] = str(Path(__file__).parent / "platforms")
    yield create_calibration_platform(request.param)
    if old_platforms is not None:
        os.environ[PLATFORMS] = old_platforms
    else:
        del os.environ[PLATFORMS]
```
but there is actually an easier way
```suggestion
@pytest.fixture(scope="module", params=TESTING_PLATFORM_NAMES)
def platform(request, monkeypatch):
    """Dummy platform to be used when there is no access to QPU."""
    monkeypatch.setenv(PLATFORMS, str(Path(__file__).parent / "platforms"))
    return create_calibration_platform(request.param)
```
which is essentially doing the same, but already implemented by Pytest.

alecandido (approved):

tests/conftest.py
andrea-pasquale: For sure I like your proposal better than mine :)

PR #1102, opened 2025-02-25
Fix history protocols order

### LINKED ISSUE #1100: History loads protocols in wrong order
When the report is generated using `qq report` the `History` will load which individual `Completed` item:
https://github.com/qiboteam/qibocal/blob/5520bba557c5a116e4eb0bc2186a2f52e45f2755/src/qibocal/auto/history.py#L69-L75

However with this approach the protocols may not match the execution order like in this case http://login.qrccluster.com:9000/k3fLscXxTVqkyP4f-K5YCA==/ where the first single shot classification shown is the one with optimal integration weights and the second one is the one with default integration weights giving a lower readout fidelity. To solve this I see two possible solutions:

1. We need to somehow pass to the `History` this information on how the protocols are sorted
2. A less instrusive way could be to order the protocols based on when they were modified on disk. Something like this
```python
@classmethod
def load(cls, path: Path):
    """To be defined."""
    instance = cls()
    for protocol in sorted((path / "data").glob("*"), key=os.path.getmtime):
        instance.push(Completed.load(protocol))
    return instance
```
which should work in any case apart from some manual file modifications from the user that should not happen.
Currently I would prefer option 2 but if there are other alternatives I would be happy to ear them. Thoughts @Edoardo-Pedicillo @hay-k @alecandido?

andrea-pasquale:
Closes #1100

### DIFF
--- a/src/qibocal/auto/history.py
+++ b/src/qibocal/auto/history.py
@@ -1,5 +1,7 @@
 """Track execution history."""
 
+import json
+import os
 from collections import defaultdict
 from dataclasses import dataclass, field
 from functools import singledispatchmethod
@@ -7,6 +9,9 @@ from pathlib import Path
 from typing import Iterator, Optional
 
 from .task import Completed, Id, TaskId
+
+HISTORY = "history.json"
+"""File where protocols order is dumped."""
 
 
 @dataclass
@@ -28,6 +33,10 @@ class History:
     """
     _order: list[TaskId] = field(default_factory=list)
     """Record of the execution order."""
+
+    @property
+    def _serialized_order(self):
+        return [str(i) for i in self._order]
 
     @singledispatchmethod
     def __contains__(self, elem: Id):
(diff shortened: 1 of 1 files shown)


src/qibocal/auto/history.py
```diff
@@ -101,6 +116,12 @@ def flush(self, output: Optional[Path] = None):
             if output is not None:
                 completed.path = self.route(task_id, output)
             completed.flush()
+        self.dump(output)
+
+    def dump(self, path: Path):
+        """Dumping task order to file."""
+        print(self._order)
```
alecandido, line 123: Leftover

src/qibocal/auto/history.py
```diff
@@ -101,6 +116,12 @@ def flush(self, output: Optional[Path] = None):
             if output is not None:
                 completed.path = self.route(task_id, output)
             completed.flush()
+        self.dump(output)
+
+    def dump(self, path: Path):
```
alecandido, line 121: I would either make this more specific (and internal, prepending a `_`), or just inline the single relevant line in `.flush()`, since otherwise it could be confusing (`.dump()` is usually saving the content, while this would only store the order info)

andrea-pasquale: Adding the line to `.flush()` sounds good to me.

src/qibocal/auto/history.py
```diff
@@ -70,7 +79,13 @@ def items(self) -> Iterator[tuple[TaskId, Completed]]:
     def load(cls, path: Path):
         """To be defined."""
```
Edoardo-Pedicillo, line 80: Could you add a proper docstring?

src/qibocal/auto/history.py
```diff
@@ -29,6 +34,10 @@ class History:
     _order: list[TaskId] = field(default_factory=list)
     """Record of the execution order."""
 
+    @property
+    def _serialized_order(self):
```
Edoardo-Pedicillo, line 38: Missing docstring.

Edoardo-Pedicillo (approved):

PR #1108, opened 2025-03-07
Drive crosstalk protocol

andrea-pasquale:
Implementation of a protocol to estimate drive crosstalk.
I am following section E1 of this paper: https://arxiv.org/pdf/2112.03708
Here are some preliminary measurements: http://login.qrccluster.com:9000/dBwvgJoTRpa_48QuN2TBtg==/

Following the previous reference we have $a_i^{\text{DL} \ Q_i} = 0.0982$ and $a_j^{\text{DL} \ Q_i} = 0.685$ giving us a crosstalk coefficient of

$$ -\log_{10} \frac{a_j^{\text{DL} \ Q_i}}{a_i^{\text{DL} \ Q_i} } \approx - 0.843$$

for qubits B2 and B3 which is ok-ish I guess.

Still left to do before meging:
- [ ] update documentation
- [ ] add drive crosstalk to update
- [ ] add tests


alecandido:
Maybe it would be worth to produce the matrix itself as a visualization, rather than the fit to the cosine
image
What do you think?

alecandido:
> Eventually we could have a script that loops over all possible combination to evaluate the matrix.

I see. It definitely makes sense to keep the atomic building block as a routine, and split the higher level processing as a script.
I had not thought about that. Thanks for pointing out!

PR #1111, opened 2025-03-12
Flux reconstruction

ElStabilini:
I open this draft without aiming to merge anything, it's simply a temporary solution for reconstructing the flux pulse waveform.

I simply created a new routine, identical to `cryoscope` except for the postprocessing, which does not include the filter determination.

@Luca-Ben-Herrmann maybe this can be useful while we work on the final version of the `cryoscope`


ElStabilini:
> Thanks @ElStabilini for pushing this code. I've already implemented yesterday a possible solution which consists in skipping the calculation of the filters if there are already filters for the specified flux line. [#974 (comment)](https://github.com/qiboteam/qibocal/pull/974#issuecomment-2717832453) If you agree with my solution we could simply go ahead with that and close this PR :)

Yes for sure!

PR #1121, opened 2025-03-25
Fix VZ phases protocol

Edoardo-Pedicillo:
This PR addresses issues related to the VZ protocol:  

- The expected fit function is given by $\frac{1}{2} (1+ \cos(\phi+\beta))$.  
- Previously, the formula was evaluated using two RY90 rotations instead of RX90. While no significant differences were expected, we have now strictly implemented the approach originally considered.  
- The formula is evaluated using RZ rotations. As explained in https://github.com/qiboteam/qibolab/pull/1044#issuecomment-2354622956, this involves applying a virtual phase with the opposite angle. In the previous version of the protocol, the virtual phase angle was swept directly, meaning the resulting curve should be the same but with $\phi \rightarrow -\phi$. To maintain clarity, we have chosen to define the beta angles according to RZ rotations. Consequently, the sweep will now range from $-\beta_{\text{stop}}$ to $-\beta_{\text{start}}$, and this will be well documented.


Edoardo-Pedicillo:
This PR is ready for review. Tests are failing because the qibolab branch `sweep_virtualz` has to merge `main`.

Luca-Ben-Herrmann (approved):

src/qibocal/protocols/two_qubit_interaction/virtual_z_phases.py
```diff
@@ -109,114 +113,97 @@ def create_sequence(
...
-    if setup == "X":
-        theta_sequence += control_natives.RX()
+    for _ in range(gate_repetition):
+        # sequence.append(flux_sequence[0])
+        sequence += flux_sequence
+        sequence.append((flux_channel, Delay(duration=dt)))
+
+    # A single RZ is added because qm ignores the first one.
+    # This work for CZ since it commutes with the RZ, but break the iSWAP compatibility.
+    virtual_phases.append(VirtualZ(phase=0))
```
andrea-pasquale, line 181: Here, if I understand correctly only the `virtual_phases` list will contain only one `VirtualZ`, is this expected?

andrea-pasquale (commented):
Thanks @Edoardo-Pedicillo and @Luca-Ben-Herrmann for all the changes.
I have just a few minor comments.
Given that you already put a lot of effort in documenting the protocol it would be nice to update the documentation directly as well at some point. Feel free to do it directly in this PR or open an issue.

src/qibocal/protocols/two_qubit_interaction/virtual_z_phases.py
Edoardo-Pedicillo: Yes, this is exactly what is happening. This is because qm cannot properly execute repetitions of multiple RZ in a row, this is better explained in https://github.com/qiboteam/qibolab/discussions. To solve this, since CZ and RZ commute, we decided to add only one RZ at the end.

pyproject.toml
```diff
@@ -17,7 +17,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = ">=3.9,<3.13"
-qibolab = "^0.2.5"
+qibolab = {git = "https://github.com/qiboteam/qibolab.git"}
```
andrea-pasquale, line 20: Is there something that you need from the main of qibolab? If not we should stick to the last release if possible.

andrea-pasquale (approved):
Thanks for the updates. Just a tiny remark on the qibolab dep.

PR #1126, opened 2025-03-27
Readout optimization with QND

ElStabilini:
I decided to open a new branch pointing at `readout_optimization02` as I also made some changes to the way I save data after acquisition, in addition to the necessary modifications for adding the QND calculation.

The goal for this PR is to come up with a protocol that optimizes the readout frequency and amplitude, considering both the `assignment_fidelity` and the quantum non-demolition-ness (QND). 

Currently, the routine optimizes the frequency and amplitude of readout pulse only with respect to the assignment fidelity and computes the QND without using this quantity for the optimization.

### DIFF
--- a/src/qibocal/protocols/readout_optimization/resonator_optimization.py
+++ b/src/qibocal/protocols/readout_optimization/resonator_optimization.py
@@ -122,7 +122,8 @@ def _acquisition(
     # taking advantage of multiplexing, apply the same set of gates to all qubits in parallel
     amplitudes = {}
     freq_sweepers = {}
-    ro_pulses = {}
+    ro_pulses_m1 = {}
+    ro_pulses_m2 = {}
 
     data = ResonatorOptimizationData(
         resonator_type=platform.resonator_type,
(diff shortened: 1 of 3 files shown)


src/qibocal/protocols/readout_optimization/resonator_optimization.py
```diff
@@ -98,143 +139,263 @@ def _acquisition(
...
+            P_0o_m1_1i = state1_count_1_m1 * state0_count_1_m2 / nshots**2
+            P_0o_1i = P_0o_m0_1i + P_0o_m1_1i
+            P_1o_m0_0i = state0_count_0_m1 * state1_count_0_m2 / nshots**2
+            P_1o_m1_0i = state1_count_0_m1 * state1_count_1_m2 / nshots**2
+            P_1o_0i = P_1o_m0_0i + P_1o_m1_0i
+
+            result = (
+                1 - ((P_0o_1i + P_1o_0i) / 2) if not pi else (P_0o_1i + P_1o_0i) / 2
+            )
+            grids["qnd"][version][j, k] = result
```
andrea-pasquale, line 355: Since this is already available in the `readout_characterization` protocol it could be nice to have this computation as a function that we call here.

andrea-pasquale (commented):
Thanks @ElStabilini for implementing this.
I just have a few comments.

src/qibocal/protocols/readout_optimization/resonator_optimization.py
ElStabilini: I opened the issue here #1206

PR #1136, opened 2025-04-22
Fix `qq upload` from dalma server

stavros11:
`qq upload` does not work when executed from the new QRC workspace server, dalma. After discussing with @scarrazza, this should fix it. I tested and it now works for me, but it would be good if someone else confirms.


andrea-pasquale (approved):
Thanks @stavros11, works on `dalma` for me as well.

scarrazza (approved):

scarrazza:
Thanks @stavros11.

PR #1138, opened 2025-04-23
Update pyproject before release

andrea-pasquale:
Syncing Qibo version with https://github.com/qiboteam/qibolab/pull/1204.
As soon as we release Qibolab I will update the `pyproject` here to point to the new release and then we can release Qibocal as well.


andrea-pasquale:
Thanks @scarrazza, I've updated the Qibolab version. Tests are passing so we should be good to merge and release.

PR #1140, opened 2025-05-02
Re-organizing imports

andrea-pasquale:
The aim of this PR is mainly to fix the current protocols [`__init__.py`](https://github.com/qiboteam/qibocal/blob/main/src/qibocal/protocols/__init__.py) which was getting out of hand.
I am following more or less what it is currently done in Qibolab using `__all__`.
The PR is not urgent and I would advise to merge PRs which contain new protocols before this one is possible to avoid possible conflicts.


Edoardo-Pedicillo (approved):

src/qibocal/protocols/resonator_spectroscopies/resonator_utils.py
alecandido, line 1: I'm pretty sure this should be just moved, and not a new addition.

Actually, I didn't really dig into it, but I expect it does contain 832 lines of code, against the 212 of `src/qibocal/protocols/resonator_utils.py`, just because it got merged with `src/qibocal/utils.py`. Isn't it? 

In which case, reading it would mostly be a waste of time. But asking was cheap enough to be worth :)

alecandido (approved):
Thanks for the reorganization!

src/qibocal/protocols/resonator_spectroscopies/resonator_utils.py
andrea-pasquale: Yes, indeed. I took the opportunity to move a bunch of functions that were used only in the resonator spectroscopies to `resonator_utils.py`.

PR #1145, opened 2025-05-15
Fix AllXY

Luca-Ben-Herrmann:
When no beta_parameter is given, the native Rx pulse should be used.


andrea-pasquale (approved):
Thanks, indeed it makes sense to use the native pulse.

Edoardo-Pedicillo (approved):

PR #1150, opened 2025-05-15
Cryoscope related: Allow use of sampling rate different than 1Gsps

Luca-Ben-Herrmann:


Edoardo-Pedicillo (approved):

andrea-pasquale (approved):
Thanks @Luca-Ben-Herrmann.

@Edoardo-Pedicillo if there are data with a sampling rate different than 1 it would be nice to add another "regression" test similar to what @ElStabilini did for https://github.com/qiboteam/qibocal/blob/main/tests/test_cryoscope.py

Luca-Ben-Herrmann:
@andrea-pasquale Once the fridge is cold I will test the routines with the higher sampling rate (assuming the QM update that would allow us to use the higher sampling rate works).

PR #1162, opened 2025-05-26
Enforce utf-8 encoding on all write operations

alecandido:
Apparently, UTF-8 has been accepted by Python as the default encoding, but it will only happen on py3.15
https://peps.python.org/pep-0686/

Currently, there is no reason why people may expect to be using anything differently on Linux and MacOS. But Windows users may experience some troubles in the meanwhile.

> Developers using macOS or Linux may forget that the default encoding is not always UTF-8.
> For example, using long_description = open("README.md").read() in setup.py is a common mistake. Many Windows users cannot install such packages if there is at least one non-ASCII character (e.g. emoji, author names, copyright symbols, and the like) in their UTF-8-encoded README.md file.
>
> Of the 4000 most downloaded packages from PyPI, 489 use non-ASCII characters in their README, and 82 fail to install from source on non-UTF-8 locales due to not specifying an encoding for a non-ASCII file. [[1]](https://peps.python.org/pep-0597/#id10)

https://peps.python.org/pep-0597/

@sorewachigauyo

### DIFF
--- a/src/qibocal/auto/history.py
+++ b/src/qibocal/auto/history.py
@@ -126,7 +126,9 @@ class History:
             if output is not None:
                 completed.path = self.route(task_id, output)
             completed.flush()
-        (output / HISTORY).write_text(json.dumps(self._serialized_order, indent=4))
+        (output / HISTORY).write_text(
+            json.dumps(self._serialized_order, indent=4), encoding="utf-8"
+        )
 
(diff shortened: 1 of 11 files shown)


andrea-pasquale (approved):

sorewachigauyo (approved):
Works now, thanks

PR #1172, opened 2025-06-04
Remove hardcoded variables from qq upload

### LINKED ISSUE #1171: Generalize qq upload
We should add support to custom host configuration so other labs can adjust their environment and use `qq upload` e.g. by exposing environment variables such as `UPLOAD_HOST`, `TARGET_DIR`, `ROOT_URL` as defined [here](https://github.com/qiboteam/qibocal/blob/main/src/qibocal/cli/upload.py#L17).

andrea-pasquale:
Closes #1171

By adding a new entry in `UploadLab` and in the `from_lab` method we should be able to support more than one lab.
For example it could be possible to upload a report to the Singapore reports by doing something like
```sh
qq upload <report> --lab singapore
```
As soon as the Singapore "database" will be only I will add the corresponding entry.

### DIFF
--- a/src/qibocal/cli/upload.py
+++ b/src/qibocal/cli/upload.py
@@ -20,6 +20,7 @@ class UploadLab(Enum):
     """Lab where the report is uploaded."""
 
     TII = "tii"
+    S14 = "s14"
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


alecandido (approved):
I would consider this a good temporary patch.
There will be time to expand into a more flexible structure.

Thanks @andrea-pasquale!

PR #1181, opened 2025-06-11
Example notebook using `submitit`

stavros11:
Following our discussion, here is the example notebook using qibocal with `submitit` to submit to slurm nodes. I just tried it and despite throwing some warnings/errors during the job submission:
```
sacct: error: slurm_persist_conn_open_without_init: failed to open persistent connection to host:localhost:6819: Connection refused
sacct: error: Sending PersistInit msg: Connection refused
sacct: error: Problem talking to the database: Connection refused
submitit WARNING (2025-06-11 12:41:16,305) - Call #1 - Bypassing sacct error Command '['sacct', '-o', 'JobID,State,NodeList', '--parsable2', '-j', '856825']' returned non-zero exit status 1., status may be inaccurate.
submitit WARNING (2025-06-11 12:41:16,305) - Call #1 - Bypassing sacct error Command '['sacct', '-o', 'JobID,State,NodeList', '--parsable2', '-j', '856825']' returned non-zero exit status 1., status may be inaccurate.
```
in the end  it works and the results come back to the notebook runtime. 

As discussed, it is also possible to do the fitting and show the plotly plots directly in the notebook. Only the report tables are returned as raw html code, however there might be some way (or Python library) to directly render those as actual tables within the notebook.

The only side effect I noticed so far is that the report folder is not generated, but most likely it is possible to generate it with some additional qibocal calls.

This is the only example I tried running, I am not sure how stable `submitit` will be in general, but I'll report in case I find any issues.


alecandido:
Ok, the issue with the other method seems to have been solved. And we may implement sooner or later #1210, to better support fair share on a queue system.

Shall we close this one? @stavros11 @Edoardo-Pedicillo @andrea-pasquale

andrea-pasquale:
We can close it from my side.

PR #1187, opened 2025-06-13
Solving issues with `R` native in some experiments 

### LINKED ISSUE #1185: Qibocal ignores double RX90 substitution for pi-pulse rotation in some routines
Currently in https://github.com/qiboteam/qibocal/blob/main/src/qibocal/protocols/allxy/allxy.py#L168-L181, the `ALLXY` sequence creates the 90 degree rotation by dividing the amplitude of the pi-pulse by half. 

We have a discussion about this last year in https://github.com/qiboteam/qibolab/issues/1066 and it may give more useful results to use the RX90 pulse if it is available.

andrea-pasquale:
Closes #1185


src/qibocal/protocols/allxy/allxy.py
Edoardo-Pedicillo, line 1: Not for this PR, but we should migrate allxy to circuits, so this if conditionals will be directly solved in the compiler.

Edoardo-Pedicillo (approved):

sorewachigauyo (approved):
Tested the allxy and seems to be working
[rx90-fix-tests.tar.gz](https://github.com/user-attachments/files/21177009/rx90-fix-tests.tar.gz)

PR #1189, opened 2025-06-17
Minor fixes for iSWAP chevron protocol

andrea-pasquale:
The fitting procedure for the chevron protocol with an `iSWAP` was incorrect because we were just using the same procedure used for the `CZ`, while for the `iSWAP` the duration of the gate should be half the oscillation period.
Screenshot 2025-06-17 at 15 34 13

I've also removed an additional X pulse for the chevron sequence that was added to improve the plot for the `CZ` case which is not necessary for the `iSWAP`.


alecandido (approved):

Edoardo-Pedicillo (approved):
Looks good to me, thanks.

PR #1194, opened 2025-06-23
Add mechanism to skip update for qubits

### LINKED ISSUE #1193: Skip specific qubits when updating platform
As @sorewachigauyo it could be useful to avoid updating the parameters of specific qubits when running `qq update`, since often the fitting procedure might fail only for certain qubits. My idea would be to add a `skip-qubits` flag:

```sh
qq update <report> --skip-qubits 0 1
```

andrea-pasquale:
Closes #1193

The current proposal is implementing https://github.com/qiboteam/qibocal/issues/1193#issue-3167523019
I still need to add tests and documentation (and eventually improve the implementation) but so far it seems to work as expected.

EDIT: 
Here is the syntax to skip the update for certain qubits:

```sh
qq update <output_folder> --skip-qubits 0 1
```
Currently only single qubits are supported (not pairs).

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -6,6 +6,7 @@ import pytest
 from click.testing import CliRunner
 
 from qibocal import create_calibration_platform
+from qibocal.auto.serialize import load
 from qibocal.cli._base import command
 
 test_runcards_dir = pathlib.Path(__file__).parent / "runcards"
@@ -75,8 +76,8 @@ def test_skip_qubits_option(skip_qubits, tmp_path, monkeypatch, platform):
             catch_exceptions=False,
         )
         new_platform = create_calibration_platform("mock")
-        for i in new_platform.qubits:
-            if i in list(skip_qubits):
+        for i in platform.qubits:
+            if i in [load(q) for q in skip_qubits]:
                 assert old_platform.config(
                     old_platform.qubits[i].acquisition
                 ) == new_platform.config(new_platform.qubits[i].acquisition)


andrea-pasquale:
> The updating of the single qubit gate is skipped, but not the frequency updating, e.g. from Ramsey
> 
> Here, qubit 14 had a bad Ramsey fit, which gets updated even with `qq update --skip-qubits 14`
> 
> [skip-qubit-test.tar.gz](https://github.com/user-attachments/files/21570070/skip-qubit-test.tar.gz)

Thanks @sorewachigauyo https://github.com/qiboteam/qibocal/pull/1194/commits/1406c5fd699ff845ac353cbc034ab4e41cf1c741 should fix the issue. 
I believe that the zip is empty, according to tests it should work correctly now.

alecandido:
> I believe that the zip is empty, according to tests it should work correctly now.

Maybe I'm misunderstanding the comment, but the provided `.zip` file is not empty...

sorewachigauyo:
Sorry for the lack of response

I tried using this with multiple qubits e.g. `qq update flipping --skip-qubits 1 9 13` which returns the following error
```
Usage: qq update [OPTIONS] folder
Try 'qq update -h' for help.

Error: Got unexpected extra arguments (9 13)
```

alecandido:
This is because that's not the way you should use *multiple* options in Click
https://click.palletsprojects.com/en/stable/options/#multiple-options

So, the correct (and verbose) way is:
```bash
qq update flipping --skip-qubits 1 --skip-qubits 9 --skip-qubits 13
```

A short alternative could be provided for the argument name, in order to make it a bit more pleasant.

sorewachigauyo (approved):

PR #1199, opened 2025-07-02
Set acquisition to zero before executing time of flight

Edoardo-Pedicillo:


andrea-pasquale (approved):
Thanks @Edoardo-Pedicillo for this fix.
If you want you could also add an `_update` function so that we update the tof correctly.
We should check if all qibolab instruments are ok with putting 0 as tof, if not we probably  need to modify this special case directly in qibolab.

alecandido:
Is QM preventing time-of-flight to be null?

In principle, the time-of-flight is just a delay before starting the acquisition.
Ok, there may be some non-zero delay anyhow, because you may use the same processor for the `play` and `acquire` instruction, but I would have expected that this is not accounted in the time-of-flight (which I would have put on top)

Edoardo-Pedicillo:
> Thanks @Edoardo-Pedicillo for this fix. If you want you could also add an `_update` function so that we update the tof correctly. 

This feature is implemented in https://github.com/qiboteam/qibocal/pull/1199/commits/a1c16ec2313da26535e060b4189b98794d760c5a .

>We should check if all qibolab instruments are ok with putting 0 as tof, if not we probably need to modify this special case directly in qibolab.

I have tested the protocol with QM (http://login.qrccluster.com:9000/U1KNHznnQ3epLy4FE3okaQ==/) and it doesn't accept 0 as time of flight (it should be a multiple of 4 greater or equal 24), to solve this issue I have defined an upper level constant `MINIMUM_TOF`, that is the default value, and set it to 24.

Thanks to @andrea-pasquale, we have imporved also the tof fit that seems more robust know.

Edoardo-Pedicillo:
> (it should be a multiple of 4 greater or equal 24)

The fit is not taking this limitation into account, so it could happen the value provided by the protocol is not allowed. This should be taken into consideration by qibolab.

andrea-pasquale (approved):

alecandido (approved):

PR #1208, opened 2025-07-22
update qibolab dep

Edoardo-Pedicillo:
Update qibolab dep to the new version so that it is published by pypi.


andrea-pasquale:
> Thanks @Edoardo-Pedicillo, did you test this on hardware?

There should not be any difference since the current qibocal's main branch was pointing to qibolab's main branch.

Issue #1205, opened 2025-07-21
Move analysis in _fit() function for resonator optimization


ElStabilini:
As title says: do the post-processing directly in the fitting instead of acquisition.

Issue #1206, opened 2025-07-21
Compute QND using the implementation in `readout_characterization`


ElStabilini:
Since the QND calculation is already available in the `readout_characterization` protocol, it could be useful to use this function in the `resonator_optimization` protocol instead of directly computing QND.

Thank you @andrea-pasquale for the suggestion.

ElStabilini:
Related to the QND calculation in the `resonator_optimization` protocol, it may also be interesting to investigate the reason for the observed pattern of both QND and QND-pi as a function of the amplitude and frequency of the readout resonator.

What we saw from report is similar to the following image

<img width="1529" height="357" alt="Image" src="https://github.com/user-attachments/assets/bc7d4b66-0355-4f7c-b9c5-1d9d88be3648" />
[report](http://login.qrccluster.com:9000/brbo8k-_RUmeDOraTnI8ZQ==/)

while we expected a plot  similar to the fidelity one.

PR #1211, opened 2025-07-25
feat: add compute_qnd function

ElStabilini:
First attempt to solve #1206


ElStabilini:
This PR should also close #1205, which was probably a problem only for the optimization without QND measurement.

andrea-pasquale (approved):

PR #1217, opened 2025-07-29
Fix update in resonator spectroscopy at low power

andrea-pasquale:
When running at low power we were not updating the `dressed_resonator_frequency` which is what we are scanning around.


Luca-Ben-Herrmann (approved):
Thanks for this, I am confirming that the fix works: http://login.qrccluster.com:9000/dxm2DLDxROSJ1O2qfINVIw==

PR #1219, opened 2025-07-29
Fix readout pulses for `drag_simple` with unrolling 

andrea-pasquale:
As title says.


Luca-Ben-Herrmann (approved):
I confirm that this fixes the bug.

PR #1222, opened 2025-07-30
Sweep offset of external DC source

stavros11:
Adds the possibility to sweep the current of an external DC source (such as the SPI rack: https://github.com/qiboteam/qibolab/pull/1260) for the resonator and qubit flux routines.

A more elegant solution would be to implement software sweepers directly on the SPI driver, so that we don't need to modify the routines (which may be many more). However, that solution requires some rework of the `Platform` for multi-instrument support, which we decided to postpone for now and provide modify the routines directly as a simpler solutions.

### DIFF
--- a/src/qibocal/protocols/flux_dependence/__init__.py
+++ b/src/qibocal/protocols/flux_dependence/__init__.py
@@ -2,9 +2,9 @@ from .cryoscope import cryoscope
 from .flux_amplitude_frequency import flux_amplitude_frequency
 from .flux_gate import flux_gate
 from .qubit_crosstalk import qubit_crosstalk
-from .qubit_flux_dependence import qubit_flux
+from .qubit_flux_dependence import qubit_flux, qubit_flux_dc
 from .qubit_vz import qubit_vz
-from .resonator_flux_dependence import resonator_flux
+from .resonator_flux_dependence import resonator_flux, resonator_flux_dc
 
 __all__ = [
     "qubit_flux",
--- a/src/qibocal/protocols/flux_dependence/__init__.py
+++ b/src/qibocal/protocols/flux_dependence/__init__.py
@@ -2,9 +2,11 @@ from .cryoscope import cryoscope
 from .flux_amplitude_frequency import flux_amplitude_frequency
 from .flux_gate import flux_gate
 from .qubit_crosstalk import qubit_crosstalk
-from .qubit_flux_dependence import qubit_flux, qubit_flux_dc
+from .qubit_flux_dependence import qubit_flux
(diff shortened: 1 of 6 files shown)


alecandido:
> A more elegant solution would be to implement software sweepers directly on the SPI driver, so that we don't need to modify the routines (which may be many more). However, that solution requires some rework of the `Platform` for multi-instrument support, which we decided to postpone for now and provide modify the routines directly as a simpler solutions.

Actually, this is certainly feasible (at some point).
But while thinking again about it, I was wondering which would be the optimal/desired way to support it.

Essentially, sweepers are just targeting channels. So, all you're doing is specifying a `ChannelId`.
To reuse the exact same routines, we should just allow:
1. multiple instruments to provide the same channels, 
2. simultaneous sweeps on multiple instruments (only updating parameters in the relevant location, but we need loops everywhere), and then 
3. disable sweeping in the controller through configurations.

I.e. the routine will contain a `Sweeper(parameter=Parameter.offset, values=..., channels=["0/flux"])`, but then we still need to pass this information
https://github.com/qiboteam/qibocal/blob/c3e78e0509cc900ebb2682808883d87af2763f0d/src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py#L149-L151
somewhere, because we really want to support two different behaviors: sweeping the flux pulse, or the bias.
That somewhere will be in instruments configurations (e.g. in the `DcConfig`, toggling "bias vs pulse").

Is it acceptable like this? Or having multiple instruments on the same channel sounds less than great?

stavros11:
> Is it acceptable like this? Or having multiple instruments on the same channel sounds less than great?

One issue I can think of with this approach on that particular example is that we will have two controllable offsets that go to that channel, one from the external current source (SPI) and one from the controller (QM). If we use a `DcConfig` to set it, we only have one channel entry which will be mapped to one of the two (ideally the SPI), so we won't be able to control the QM offset if we need to. 

Of course this issue can be solved just by defining another `Config` (for example `BiasTeeConfig`) which supports two offsets. Also, this may not be even necessary in practice for this example, since the whole point of the bias tee (if I understand it correctly) is to block the DC from one port (the controller) and only let it from the current source.

Another issue with using the same channel is that we will be passing redundant information to the instruments through the configs. For example the `DcConfig` containing the SPI offset will also be passed to QM, even though it is not relevant. This may create issues for example for https://github.com/qiboteam/qibolab/pull/1265, since even though all QM related parameters will be unaffected, the `configs` passed to `play` will still be changing. I guess this can also be addressed by doing the cache check in a different way, but I am not sure if there are any other side-effects (given that passing redundant info is not ideal).

PR #1223, opened 2025-07-31
Add function for baseline estimation (ALS)

alecandido:
Based on [P. Eilers and H. Boelens, 2005](http://zanran_storage.s3.amazonaws.com/www.science.uva.nl/ContentPages/443199618.pdf).

> [!NOTE]
> Since it is not available any longer at the original URL, you can get the paper from the Internet Archive.
> [archived version](https://web.archive.org/web/20160914041040/http://zanran_storage.s3.amazonaws.com/www.science.uva.nl/ContentPages/443199618.pdf)

The function is not yet used in any protocol, and integrating it into the resonator spectroscopy workflow will require some more effort, since it needs manual tuning of the algorithm's parameters.

Cf.

| image | image |
|--------|--------|

### DIFF
--- a/src/qibocal/protocols/utils.py
+++ b/src/qibocal/protocols/utils.py
@@ -8,7 +8,7 @@ import plotly.graph_objects as go
 from numpy.typing import NDArray
 from plotly.subplots import make_subplots
 from qibolab._core.components import Config
-from scipy import constants
+from scipy import constants, sparse
 from scipy.optimize import curve_fit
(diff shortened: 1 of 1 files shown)


alecandido:
Actually, I should add a test (a closure one).

But not even the other fits are tested. So, I could postpone also this one, and just add to the heap.

andrea-pasquale (approved):
Thanks @alecandido

PR #1229, opened 2025-08-12
Improving flux fitting by using different mask

andrea-pasquale:
In this PR I'm trying to improve the performance of the fit by doing the following:

1. Adding more parameters (in this way we avoid to rely on outdated parameters from the JSON files) + using a slight different formula
2. Instead of using a mask based on the confidence interval I'm using a different approach where after averaging the signal for each column (which should compensate any unwanted noise) I'm just calculating the peaks using `find_peaks` with a suitable `prominence`.

Here is how the fitting are looking now: http://login.qrccluster.com:9000/f4DP80QOStCzwaBP2XQxAQ==/
Compared to before where the fitting was failing: http://login.qrccluster.com:9000/x9YShaUsRU6rRJRAnmRd1w==/

I've also decided to remove the information related to the phase since it is not relevant for the fit. If we can confirm this I can refactor also the acquisition to simplify it removing the phase.

### DIFF
--- a/src/qibocal/protocols/utils.py
+++ b/src/qibocal/protocols/utils.py
@@ -768,35 +768,28 @@ def extract_feature(
     Given a dataset of the form (x, y, z) where a spike or a valley is expected,
     this function discriminate the points (x, y) with a signal, from the pure noise
     and return the first ones.
-
-    A first mask is construct by looking at `ci_first_mask` confidence interval for each y bin.
-    A second mask is applied by looking at `ci_second_mask` confidence interval to remove outliers.
-    `feat` could be `min` or `max`, in the first case the function will look for valleys, otherwise
-    for peaks.
-
     """
-
+    x_ = np.unique(x)
+    y_ = np.unique(y)
     # background removed over y axis
-    z_ = z.reshape(len(np.unique(y)), len(np.unique(x)))
+    z_ = z.reshape(len(y_), len(x_))
     z_ = z_ / np.mean(z, axis=0)
     normalized_z = z_.reshape(z.shape)
 
     # filter data using find_peaks
-    filtered_bias = []
-    filtered_freq = []
-    unique_bias = np.unique(y)
-    unique_freq = np.unique(x)
-    for i in unique_bias:
+    filtered_y = []
+    filtered_x = []
+    for i in y:
         signal_fixed_y = normalized_z[y == i]
         peak, _ = find_peaks(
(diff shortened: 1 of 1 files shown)


andrea-pasquale:
I took the opportunity to refactor and modify slightly the plots. Here are two examples
Screenshot 2025-08-13 at 11 49 05
Screenshot 2025-08-13 at 11 48 51

src/qibocal/protocols/flux_dependence/utils.py
```diff
@@ -288,24 +266,35 @@ def transmon_readout_frequency(
...
+            - transmon_frequency(
+                xi=xi,
+                xj=xj,
+                w_max=w_max,
+                d=d,
+                normalization=normalization,
+                offset=offset,
+                crosstalk_element=crosstalk_element,
+                charging_energy=charging_energy,
+            )
```
alecandido, line 282: Since the transmon frequency is computing twice with the exact same parameters, I would store it in a variable, for better clarity (and marginal efficiency).

src/qibocal/protocols/flux_dependence/qubit_crosstalk.py
```diff
@@ -120,10 +120,6 @@ def _acquisition(
     freq_sweepers = []
     offset_sweepers = []
 
-    delta_frequency_range = np.arange(
-        -params.freq_width / 2, params.freq_width / 2, params.freq_step
-    )
-
     for qubit in targets:
         natives = platform.natives.single_qubit[qubit]
         charging_energy[qubit] = platform.calibration.single_qubits[
```
lballerio: same as lines 143-144-145, remove for redundancy?

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
lballerio: Decide whether we want not to save and process phase data

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
```diff
@@ -70,7 +69,6 @@ class QubitFluxResults(Results):
         ("freq", np.float64),
         ("bias", np.float64),
         ("signal", np.float64),
-        ("phase", np.float64),
     ]
 )
 """Custom dtype for resonator flux dependence."""
```
lballerio: typo in the comment: correct:

"""Custom dtype for qubit flux dependence."""

Issue #1008, opened 2024-10-11
Resonator phases and cable delay


alecandido:
Both in the plain resonator spectroscopy, and in the pumped up resonator flux dependence (the version of the spectroscopy with an additional scan over the flux pulse) the phases in the transmission coefficient are affected by the cable delay, as described in #876.

In #917, the cable delay has been subtracted for the specific case of the full-fledged circle fit of the resonator spectroscopy.
https://github.com/qiboteam/qibocal/blob/6789ac85b096310087415369f04cad9b88eac717/src/qibocal/protocols/resonator_utils.py#L14-L42
It happened there, because that fit is the first to finally use the information on the phase for the fit, while the other fits in both protocols are just using the information in the magnitude.

Despite what is used in the fit, the phases are shown in the report, and they are largely made irrelevant by the presence of the cable delay (well, before they were even more irrelevant, since cable delay-drifted *and* wrapped).

So, we should ensure a consistent treatment of the phase information, even beyond the unwrapping.

Moreover, even in the `s21` fit of the resonator spectroscopy the cable delay is separately evaluated, despite the "time of flight" protocol having already measured it, potentially.
So, we should make it possible to propagate the information separately measured by the time of flight (if available) to always preprocess consistently the transmission coefficient phases.

alecandido:
> Maaaybeee not???, we would need to pass also the frequency at which we are measuring, so I am not sure if it makes sense here. The cable delay doesnt mean much for the phase correction when the frequency is not changing (so only for spectroscopies, and even only single tone (redout) spectroscopies)

Actually, why not?

We can decide where to put this method, but imagine that we have a `Readout` object including the `.cable_delay` attribute. Then, we could use that attribute to compute the phase, with something like:

```py
class Readout:
	def phase(self, bare: np.ndarray, frequencies: Optional[np.ndarray] = None):
         res = ...
         if frequencies is None:
             return res
         return res - (frequencies - frequencies[0]) * self.cable_delay
```

If we lift the function to the [`QubitCalibration`](https://github.com/qiboteam/qibocal/blob/7b4fd4a7b03c7dc3e98efedb979a6c68e9c435fd/src/qibocal/calibration/calibration.py#L125) we'd get also access to the bare and dressed resonator frequencies, and we could turn the parameter into
```py
frequencies: Union[np.ndarray, Literal["bare", "dressed"], None]
```
[...]

PR #1233, opened 2025-08-18
New mask for resonator punchout

andrea-pasquale:
Propagate new mask for punchout protocol http://login.qrccluster.com:9000/q4Qk4JyRSY-fSj_Uuc767w==/
I'm also refactoring some parts of the code (removing `register_qubit`)

### DIFF
--- a/src/qibocal/protocols/resonator_spectroscopies/resonator_punchout_attenuation.py
+++ b/src/qibocal/protocols/resonator_spectroscopies/resonator_punchout_attenuation.py
@@ -12,8 +12,8 @@ from qibocal.auto.operation import Data, Parameters, QubitId, Results, Routine
 from qibocal.calibration import CalibrationPlatform
 from qibocal.result import magnitude, phase
 
-from ..utils import HZ_TO_GHZ, reshaping_raw_signal, table_dict, table_html
-from .resonator_utils import fit_punchout, punchout_extract_feature, punchout_mask
+from ..utils import HZ_TO_GHZ, table_dict, table_html
+from .resonator_utils import fit_punchout, punchout_extract_feature
 
 __all__ = ["resonator_punchout_attenuation", "ResonatorPunchoutAttenuationData"]
 
@@ -210,9 +210,14 @@ def _fit(data: ResonatorPunchoutAttenuationData) -> ResonatorPunchoutAttenuation
 
     for qubit in data.qubits:
         filtered_x, filtered_y = data.filtered_data(qubit)
-        bare_freq, readout_freq, ro_val, fit_flag = fit_punchout(
-            filtered_x, -filtered_y
-        )
+
+        if filtered_x is None or filtered_y is None:
+            bare_freq = readout_freq = ro_val = None
+            fit_flag = False
+        else:
+            bare_freq, readout_freq, ro_val, fit_flag = fit_punchout(
+                filtered_x, -filtered_y
+            )
 
         if fit_flag:
             readout_freqs[qubit] = readout_freq
@@ -243,15 +248,16 @@ def _plot(
         horizontal_spacing=0.1,
         vertical_spacing=0.2,
         subplot_titles=(
-            "Normalised Signal [a.u.]",
+            "Raw Signal [a.u.]",
             "Phase [rad]",
         ),
     )
 
     frequencies, attenuations, qubit_signal = data.grid(target)
-    _, _, qubit_signal = reshaping_raw_signal(frequencies, attenuations, qubit_signal)
-    qubit_signal = punchout_mask(qubit_signal).ravel()
     qubit_phase = data.phase(target).ravel()
+    delay = np.mean(np.gradient(qubit_phase)) / (np.mean(np.gradient(frequencies)))
+    qubit_phase_compensation = delay * (frequencies - np.mean(frequencies))
+    qubit_phase -= qubit_phase_compensation
     frequencies *= HZ_TO_GHZ
 
     fig.add_trace(
--- a/src/qibocal/protocols/resonator_spectroscopies/resonator_punchout.py
+++ b/src/qibocal/protocols/resonator_spectroscopies/resonator_punchout.py
@@ -3,12 +3,13 @@ from typing import Optional
 
 import numpy as np
 import plotly.graph_objects as go
+from plotly.subplots import make_subplots
 from qibolab import AcquisitionType, AveragingMode, Parameter, PulseSequence, Sweeper
 
 from qibocal import update
 from qibocal.auto.operation import Data, Parameters, QubitId, Results, Routine
 from qibocal.calibration import CalibrationPlatform
-from qibocal.result import magnitude
+from qibocal.result import magnitude, phase
 
 from ..utils import HZ_TO_GHZ, scaling_slice, table_dict, table_html
 from .resonator_utils import fit_punchout, punchout_extract_feature
@@ -63,6 +64,9 @@ class ResonatorPunchoutData(Data):
 
     def signal(self, qubit: QubitId) -> np.ndarray:
         return magnitude(self.data[qubit])
+
+    def phase(self, qubit: QubitId) -> np.ndarray:
+        return phase(self.data[qubit])
 
     def grid(self, qubit: QubitId) -> tuple[np.ndarray]:
         x, y = np.meshgrid(self.frequencies[qubit], self.amplitudes)
@@ -156,7 +160,13 @@ def _fit(data: ResonatorPunchoutData) -> ResonatorPunchoutResults:
     for qubit in data.qubits:
         filtered_x, filtered_y = data.filtered_data(qubit)
 
-        bare_freq, readout_freq, ro_val, fit_flag = fit_punchout(filtered_x, filtered_y)
+        if filtered_x is None or filtered_y is None:
+            bare_freq = readout_freq = ro_val = None
+            fit_flag = False
+        else:
+            bare_freq, readout_freq, ro_val, fit_flag = fit_punchout(
+                filtered_x, filtered_y
+            )
 
         if fit_flag:
             readout_freqs[qubit] = readout_freq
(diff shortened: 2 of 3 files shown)


src/qibocal/protocols/utils.py
```diff
@@ -257,55 +257,6 @@ def cumulative(input_data, points):
...
-                bare_freq = np.max(mask_freq)
-            ro_val = np.max(mask_amps[mask_freq == best_freq])
-
-        low_freqs[qubit] = best_freq
-        high_freqs[qubit] = bare_freq
-        ro_values[qubit] = ro_val
-
-    return [low_freqs, high_freqs, ro_values]
-
-
```
lballerio: `fit_punchout` function oved in `resonator_punchout.py` script

lballerio (dismissed):

src/qibocal/protocols/flux_dependence/qubit_crosstalk.py
```diff
@@ -276,10 +276,9 @@ def fit_function(x, crosstalk_element, offset):
             crosstalk_matrix[target_qubit][flux_qubit] = (
                 popt[0] * data.matrix_element[target_qubit]
             )
-        except RuntimeError as e:  # pragma: no cover
-            log.error(
-                f"Off-diagonal flux fit failed for qubit {flux_qubit} due to {e}."
-            )
+
+        except (RuntimeError, ValueError) as e:  # pragma: no cover
+            log.error(f"Error in qubit_crosstalk protocol fit: {e} ")
```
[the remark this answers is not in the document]
lballerio: yeah this commit I think was done before the change of branch to merge the PR into (from main to stabilize_extract_feature)

lballerio (commented):
the backbone of the PR is the same as my last review but with some small changes for handling better the case where no signal (i.e. peaks) is found.

alecandido (commented):
Mostly fine.

In any case, we should wait for #1319 before merging this one (better merging PRs from the base, to avoid inflating diffs).

alecandido (commented):
The comments from the previous review have not been addressed yet :)

lballerio:
@alecandido sorry, my bad!

alecandido (commented):
I did not test it myself, but it seems fine. However, please solve the conflicts with the base branch :)

lballerio:
Hi @jevillegasdatTII, please test it and review it, tell me what to improve

jevillegasdatTII:
The fit (or plot of the fit) needs a check when more than one qubit is used.
http://login.qrccluster.com:9000/HPNRT5wDR6aXyH1d8rsLfA==

src/qibocal/protocols/resonator_spectroscopies/resonator_punchout_attenuation.py
```diff
@@ -159,24 +177,6 @@ def _acquisition(
             lo_channel = ro_los[qubit]
             updates.append({lo_channel: {"power": attenuation}})
```
jevillegasdatTII: This will need to change for compatibility with the new qibolab definition of `lo.power` to 

```
updates.append({lo_channel: {"power": -attenuation}})
```

lballerio: Honestly now I don't know what you modified in qibolab for attenuation. Also it depend on which qibolab implementation main should use...(To be discussed maybe with @alecandido)

src/qibocal/protocols/resonator_spectroscopies/resonator_punchout_attenuation.py
```diff
@@ -288,11 +281,11 @@ def _plot(
     fig.update_yaxes(autorange="reversed", row=1, col=1)
     fig.update_yaxes(autorange="reversed", row=1, col=2)
 
-    if fit is not None:
+    if fit is not None and fit.successful_fit[target]:
         fig.add_trace(
             go.Scatter(
                 x=[fit.readout_frequency[target] * HZ_TO_GHZ],
-                y=[fit.readout_attenuation[target]],
+                y=[-fit.readout_attenuation[target]],
```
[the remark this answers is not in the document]
lballerio: probably you are right, I think it was just a fix for working with QM

alecandido:
> * It's possible that we want to set the amplitude of the ro pulse to 1 for this calibration instead of using the one in the parameters.json, what do you think @alecandido ?

Since this was posted 4 minutes before the other comment, I'm pretty sure that the fix proposed for https://github.com/qiboteam/qibocal/issues/1339#issuecomment-3805597048 should also work for this one ;)

> * Additionally, there are no longer any plots for the phase, which if we just introduce the correction for the phase delay before plotting also give very useful information (this delay can be taken as the parameters' acquisition delay, or calculated as 1 / the slope of the phase plot background).

Indeed: it is useful to report there, but we want to avoid extracting this information once more.
We already produce the cable delay in two different places:
- the time-of-flight experiments, based on scope mode acquisition and edge detection
- the resonator spectroscopy itself, for which it makes more sense to make a refined fit, rather than fitting in the same way all lines of the punchout
	- it is also not that simple to choose a specific one automatically: though it should not be too relevant for the cable delay, the fit could be much better if we can rely on the resonance being visible (such that we can fit it), or at least centered (such that we can cut out)

For this, we have #1008, and the proposal is to store this value in `Calibration` (and thus `calibration.json`), such that we can use to subtract the cable delay automatically from all the phases computed (instead of repeating for each protocol, we can update the phase-computation function to accept an optional cable delay
https://github.com/qiboteam/qibocal/blob/7b4fd4a7b03c7dc3e98efedb979a6c68e9c435fd/src/qibocal/result.py#L71-L79

lballerio:
> There is still an issue with the fit when used for an attenuation punch-out, otherwise everything is ready (I just pushed a small fix of some bugs). The fit error might be given that the range of attenuation is positive but he feature extraction method was designed to use values as amplitude, the call to the fit method would think high power is low and vice versa:

Regarding this issue I think this is simply due to the fact that you are not measuring the high power branch of the signal. In fact the code for the fit is:

```python
low_freq_idx = (filtered_x - np.min(filtered_x)) <= tol_step
high_freq_idx = (np.max(filtered_x) - filtered_x) <= tol_step

low_freq = np.median(filtered_x[low_freq_idx])
high_freq = np.median(filtered_x[high_freq_idx])

low_freq_amp = np.max(filtered_y[low_freq_idx])
high_freq_amp = np.max(filtered_y[high_freq_idx])

if low_freq_amp > high_freq_amp:
        readout_freq = high_freq
        bare_freq = low_freq
        ro_val = high_freq_amp
else:
        readout_freq = low_freq
        bare_freq = high_freq
        ro_val = low_freq_amp
```

where tol_step is essentially a small tolerance. Anyway, the way it should work these lines is the following:

1. it takes the filtered signal (which should be equal to simply the punchout signal and removing all the surrounding noise) and takes all the points that are closer than tolerance to either the maximum or the minimum frequencies of the signal.
2. then, once found the max (min) frequencies, we estimate the related max (min) frequency amplitude and simply compare them in the last if statement: if min freq amplitude > max freq amplitude it means that the high attenuation  branch of the punchout signal is at high frequency values and viceversa if min freq amplitude < max freq amplitude.
3. finally we save all the necessary information accordingly.

This code might be a bit deficitary if you are only measuring just one branch, hence half of the full signal, since low and high frequencies might be very similar and very close.

Essentially here we are assuming that we are doing a full experiment and we are not missing out anything. I am pretty sure that if you run a full experiment also catching the other branch you will not see this problem.

Then, it might be suboptimal this workflow, plus now we are still lacking of a saturation controller for both punchouts, we can also might thing about a simple check for understanding if we measured just one branch of the full signal or not, but I don't know how prioritary is this fix.
Also keep in mind that @alecandido  and I decided to put on hold saturation and other refinements in order to merge this new signal detection which seems working and this PR is simply for introducing a sort of fit and signal detection for punchout, which I think was kind of still lacking, and that's it, in theory we still suggest users to take fit with a pinch of salt and keep updating the fitted value as before.

The issue you are facing now I think is mostly due to the fact that you did not saturate the signal, in this case for low attenuation regime, and in theory a flag should be risen. But as discussed before, this saturation control is out of this PR scope and it will be addressed in the future.

src/qibocal/protocols/resonator_spectroscopies/resonator_punchout_attenuation.py
```diff
@@ -190,50 +187,49 @@ def _acquisition(
...
 
-    readout_freq, bare_freq, readout_param = fit_punchout(data, fit_type)
+    readout_freqs = {}
+    bare_freqs = {}
+    ro_values = {}
+    successful_fit = {}
+
+    for qubit in data.qubits:
+        filtered_x, filtered_y = data.filtered_data(qubit)
+        bare_freq, readout_freq, ro_val, fit_flag = fit_punchout(filtered_x, filtered_y)
```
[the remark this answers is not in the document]
jevillegasd: It's more like to prevent errors when filtered_x(y) found no peaks.
```suggestion
        if filtered_x is not None and filtered_y is not None:
            bare_freq, readout_freq, ro_val, fit_flag = fit_punchout(filtered_x, -filtered_y)
        else:
            bare_freq, readout_freq, ro_val, fit_flag = None, None, None, False
            
```

src/qibocal/protocols/resonator_spectroscopies/resonator_punchout_attenuation.py
```diff
@@ -190,50 +187,49 @@ def _acquisition(
...
+
+        if fit_flag:
+            readout_freqs[qubit] = readout_freq
+            bare_freqs[qubit] = bare_freq
+            ro_values[qubit] = ro_val
```
jevillegasd: And we need to change the power back to att
```suggestion
            ro_values[qubit] = -ro_val
```

src/qibocal/protocols/resonator_spectroscopies/resonator_punchout_attenuation.py
```diff
@@ -250,24 +246,19 @@ def _plot(
...
-
-    # Normalize signal for each attenuation level
-    for i in range(n_attenuations):
-        qubit_data.signal[i * n_freq : (i + 1) * n_freq] = norm(
-            qubit_data.signal[i * n_freq : (i + 1) * n_freq]
-        )
+    frequencies, attenuations, qubit_signal = data.grid(target)
+    _, _, qubit_signal = reshaping_raw_signal(frequencies, attenuations, qubit_signal)
+    qubit_signal = punchout_mask(qubit_signal).ravel()
+    qubit_phase = data.phase(target).ravel()
```
jevillegasd: This could be a way to remove the phase background without knowing the electrical delay 
```suggestion
   qubit_phase = data.phase(target).ravel()

    # remove electrical delay from phase data
    delay = np.mean(np.gradient(qubit_phase)) / (np.mean(np.gradient(frequencies)))
    qubit_phase_compensation = delay * (frequencies - np.mean(frequencies))
    qubit_phase -= qubit_phase_compensation
```

PR #1247, opened 2025-08-19
Fixing QND

andrea-pasquale:
As @ElStabilini pointed out in #1211 the computation of the QND was quite weird...
In this PR I'm following more accurately https://arxiv.org/pdf/2106.06173 to get the correct value.
I still need to propagate the changes in the `readout_optimization` protocol.

### DIFF
--- a/src/qibocal/protocols/utils.py
+++ b/src/qibocal/protocols/utils.py
@@ -29,7 +29,7 @@ LEGEND_FONT_SIZE = 20
 TITLE_SIZE = 25
 EXTREME_CHI = 1e4
 KB = constants.k
-HBAR = constants.hbar
+H = constants.h
 """Chi2 output when errors list contains zero elements"""
 COLORBAND = "rgba(0,100,80,0.2)"
 COLORBAND_LINE = "rgba(255,255,255,0)"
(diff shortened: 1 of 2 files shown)


andrea-pasquale:
Screenshot 2025-08-19 at 18 37 18
Now I am getting values around 93% which seems to be more consistent with what I have seen in the literature.

ElStabilini:
Thank you @andrea-pasquale, I had a quick look at the code and seems more similar to what I expected.

I am not sure, but this may also be related to the problem mentioned [here](https://github.com/qiboteam/qibocal/issues/1206#issuecomment-3096860966).

src/qibocal/protocols/utils.py
```diff
@@ -188,50 +188,51 @@ def effective_qubit_temperature(
...
+    p_o_i0 = inverse_lambda_m @ np.array([p__m0_i0, p__m1_i0])
 
-    lambda_m = [
-        [state0_count_0_m1 / nshots, state0_count_1_m1 / nshots],
-        [state1_count_0_m1 / nshots, state1_count_1_m1 / nshots],
-    ]
+    if not pi:
+        qnd = 1 - (p_o_i1[0] + p_o_i0[1]) / 2
+    else:
+        qnd = 1 - (p_o_i1[1] + p_o_i0[0]) / 2
```
alecandido, line 216: I'd suggest you to go immediately for arrays, and keeping some indices implicit.

This part could be expressed as:
```py
p_m1 = np.mean([m1_state_0, m1_state_1], axis=1)
p_m2 = np.mean([m2_state_0, m2_state_1], axis=1)

lambda_m = np.stack([1 - p_m1, p_m1])
lambda_m2 = np.stack([1 - p_m2, p_m2])

p_o = np.linalg.inv(lambda_m) @ lambda_m2

qnd = 1 - np.sum(np.diag(p_o[::-1] if not pi else p_o)) / 2
```

The numerical result is *almost* identical (the only difference comes from the further replacement of `np.linalg.pinv` with `np.linalg.inv` - not sure why you preferred the pseudo-inverse, ma the numerical exact inverse is fine for me).

andrea-pasquale: Thanks for the suggestion. Why are you using `lambda_m2` as well? In the original calculation I'm not using it.

andrea-pasquale:
Here are the latest results http://login.qrccluster.com:9000/MK68uzq1SXKmsiB9FCrCQA==/

PR #1256, opened 2025-09-10
Long cryoscope protocol

### LINKED ISSUE #1235: Implement long-cryoscope
See Sect. III in https://arxiv.org/pdf/2503.04610v1

andrea-pasquale:
Closes #1235 
Currently only the acquisition is implemented, I will work on the fitting next.
Here is what the output is looking like http://login.qrccluster.com:9000/RhPMAh9oRum8jwSEOaTDLA==/


andrea-pasquale:
It seems that the filters are working. I was able to run a two qubit RB with just a rectangular envelope.
Screenshot 2025-09-19 at 17 14 00
The filters evaluation is tricky I will soon update the code with it.

alecandido (commented):
Ok, I believe I was still missing the main protocol. But apparently I left some comments pending, and forgot about them.

I'm just publishing those, and I will complete the review at a later stage. If needed

src/qibocal/protocols/flux_dependence/__init__.py
lballerio: importing long_cryoscope

src/qibocal/protocols/flux_dependence/utils.py
```diff
@@ -30,6 +30,19 @@ def bias_range(self) -> np.ndarray:
...
+    x: list[float], estimate: list[float], threshold: float = 0.001
+) -> list[float]:
+    """Normalize input array to 1.
+
+    Estimate is a guess for the mean. If the normalization
+    is off by a factor larger than 5% estimate is ignored."""
+    mean_ = np.mean(x[len(x) // 2 :])
+    if np.array(np.abs(mean_ / estimate)).max() > threshold:
+        return (np.array(x) / mean_).tolist()
+    return np.abs(np.array(x) / estimate).tolist()
```
lballerio: I don't get why guessing the mean and why computing the mean to just half of the points.

lballerio: I don't get why guessing the mean and why computing the mean to just half of the points.

Also I think the condition in the if statement is not correct, maybe better do this way:

```suggestion
    Estimate is a guess for the mean. If the normalization
    is off by a factor larger than 5% estimate is ignored."""
    mean_ = np.mean(x[len(x) // 2 :])
    if np.array(np.abs((mean_  - estimate) / mean_)).max() > threshold:
        return (np.array(x) / mean_).tolist()
    return np.abs(np.array(x) / estimate).tolist()
```
?

Also in docstring they say the threshold is 5%, but then the default value for `threshold` is `0.001`.

Also both `x`and `estimate` are defined as 1D list of floating numbers, but then `np.mean(x)` should always return an array (hence in the following line `.max()` is useless); also, since `estimate` should be an estimation for the signal mean, why is it a list? shouldn't it be a scalar?


maybe we can simply rewrite the whole function as:

```python
def normalize(
    x: list[float]
) -> list[float]:
    """Normalize input array to 1."""
    mean_ = np.mean(x[len(x)//2 :])
   return (np.array(x) / mean_).tolist()
```

src/qibocal/update.py
```diff
@@ -238,17 +244,22 @@ def kernel(kernel: np.ndarray, platform: Platform, qubit: QubitId):
...
-    """Update flux pulse feedforward parameter in platform for specific qubit."""
-    platform.update(
-        {f"configs.{platform.qubits[qubit].flux}.filter.feedforward": feedforward}
-    )
+    old_iir = [
+        filter_ for filter_ in old_filters if isinstance(filter_, ExponentialFilter)
+    ]
+    if fir is not None:
+        old_filters = old_iir + [FiniteImpulseResponseFilter(coefficients=fir)]
+    platform.update({f"configs.{platform.qubits[qubit].flux}.filters": old_filters})
```
lballerio: why are we keeping track of old iir filters?

src/qibocal/protocols/flux_dependence/cryoscope.py
```diff
@@ -268,15 +263,14 @@ def _acquisition(
...
+    p0 = [tau_guess, 1, 1]
 
     def expmodel(t, tau, exp_amplitude, g):
-        return step_response / (g * (1 + exp_amplitude * np.exp(-t / tau)))
+        return y / (g * (1 + exp_amplitude * np.exp(-t / tau)))
 
-    popt, _ = curve_fit(expmodel, t, target, p0=init_guess)
+    popt, _ = curve_fit(expmodel, x, target, p0=p0)
 
     return popt
```
lballerio: for this configuration I got the following error:

``
  File "/nfs/users/lorenzo.ballerio/test/.venv/lib/python3.10/site-packages/scipy/optimize/_minpack_py.py", line 1019, in curve_fit
raise RuntimeError("Optimal parameters not found: " + errmsg) 
RuntimeError: Optimal parameters not found: Number of calls to function has reached maxfev = 800. 
``

Maybe there is a problem in the initial guess of the fitting parameters.

lballerio: the order of the initial guess in `p0`is changed but in `expmodel` the order of the parameters is the same.

lballerio (commented):
when running:

```python

        flux_ampl_exp = e.flux_amplitude_frequency(amplitude_max=0.1,
                                            amplitude_min=-0.25,
                                            amplitude_step=0.005,
                                            duration=60, 
                                            targets=["B1"])
            
        cryoscope_exp = e.cryoscope(duration_max=80,
                                    duration_min=1,
                                    duration_step=1,
                                    flux_pulse_amplitude=0.25 if q == 'B2' else -0.20,
                                    nshots=default_nshots,
                                    relaxation_time=default_relax_time,
                                    fir=40,
                                    targets=["B1"])
        
        long_dist_exp = e.long_cryoscope(duration_max=5000,
                                    duration_min=100,
                                    duration_step=500,
                                    flux_pulse_amplitude=0.25 if q == 'B2' else -0.20,
                                    freq_step=5e6,
                                    freq_width=3.0e8,
                                    nshots=default_nshots,
                                    targets=["B1"]
                                    )
```

on `qw21q-b` using QM with the filters already present in the `parameters.json`, the following error rises up:

```
qm.exceptions.OpenQmException: Can not open QM, see the following errors:  │
CONFIG ERROR in key "controllers.con4.analogOutputs[1].filter.feedforward" │
[Analog output filters] : Analog output port number 1 in controller con4 ha│
s too many feedforward taps (Max number of taps with 3 feedback taps is: 23│
)      
```

probably an error in updating the filters' parameters.

src/qibocal/protocols/flux_dependence/long_cryoscope.py
```diff
@@ -0,0 +1,379 @@
...
+        return delay, freq
+
+    def step_reponse(self, qubit: QubitId) -> list[float]:
+        """Compute expected frequency by averaging over last half."""
+        _, freq = self.filtered_data(qubit)
+        freq_ghz = (freq - self.frequency[qubit]) * HZ_TO_GHZ
+        p = np.poly1d(self.flux_coefficients[qubit])
+        amplitude = [
+            max((p - f).roots) if True else min((p - f).roots) for f in freq_ghz
+        ]
```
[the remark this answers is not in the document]
alecandido: I guess it was a temporary value, which should have been exchanged for the proper one, once found.

Try to understand why it may be either a minimum or a maximum, and then we make a choice

src/qibocal/update.py
alecandido: Because they can be added in multiple steps.

In particular, the cryoscope fits the short-time behavior (~100 ns) with both an FIR and a few IIR (1-3, I do not remember the exact number). Then, the long-time distortion protocol tries to extract other exponential corrections, on a different timescale (~us), and provide them as ***additional*** IIR corrections.

Since you have different exponentials, you want to keep all of them. Not replace one for the other.

pyproject.toml
```diff
@@ -16,8 +16,8 @@ classifiers = [
 
 
 [tool.poetry.dependencies]
+qibolab = { git = "https://github.com/qiboteam/qibolab" }
```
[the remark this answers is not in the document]
alecandido: Good point. Please, update the dependency, and then run `poetry lock`, to also update the lock file (i.e. `poetry.lock`)

alecandido (changes requested):

PR #1275, opened 2025-10-29
ramsey updates qubit frequency as well

jevillegasdatTII:
As discussed in #1274 , updating both the qubit and the drive frequencies makes sense for this test.

### DIFF
--- a/src/qibocal/protocols/ramsey/ramsey_signal.py
+++ b/src/qibocal/protocols/ramsey/ramsey_signal.py
@@ -308,7 +308,7 @@ def _update(
         update.drive_frequency(results.frequency[target][0], platform, target)
         platform.calibration.single_qubits[
             target
-        ].qubit.frequency_01 = results.frequency[target]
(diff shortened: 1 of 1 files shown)


src/qibocal/protocols/ramsey/ramsey_signal.py
```diff
@@ -306,6 +306,9 @@ def _update(
 ):
     if results.detuning is not None:
         update.drive_frequency(results.frequency[target][0], platform, target)
+        platform.calibration.single_qubits[
+            target
+        ].qubit.frequency_01 = results.frequency[target]
```
andrea-pasquale:
```suggestion
        ].qubit.frequency_01 = results.frequency[target][0]
```
Since as you see above the frequency is a tuple containing the value and the error.

jevillegasdatTII: true. Corrected.

PR #1288, opened 2025-11-20
Ramsey initial fit parameters guess changed

lballerio:
The aim of this change is to improve the fit of a sinusoidal signal specific for ramsey experiment, which in the previous version was failing from times to times.
This issue is believed to result from an inefficient initial estimate of the fit parameters, which likely causes the approximation to occasionally converge to local minima.

Screenshot from 2025-11-20 14-22-25

Fit with the previous initial guess.

Screenshot from 2025-11-20 14-29-16

Fit with the new initial guess.

### DIFF
--- a/src/qibocal/protocols/ramsey/utils.py
+++ b/src/qibocal/protocols/ramsey/utils.py
@@ -13,11 +13,31 @@ PERR_EXCEPTION = [1] * 5
 """Fit errors to handle exceptions; their choice has no physical meaning
 and is meant to avoid breaking the code."""
 THRESHOLD = 0.5
-"""Threshold parameters for find_peaks to guess
-   frequency for sinusoidal fit."""
+"""Threshold parameters for find_peaks to guess frequency for sinusoidal fit."""
 DAMPED_CONSTANT = 1.5
-"""For damped oscillations, the factor is not easily determined, since the percentile depends on how much of the decay you see.
-By dividing the amplitude guess by this constant, the estimation will lend closer to the actual amplitude.
+"""Scaling factor to recover amplitude from quantiles.
+
+Measuring intermediate quantiles is less noise sensitive then meauring extremal points
+(minimum and maximum), but it is not a direct measurement of the amplitude itself.
+For pure sinusoidal oscillations, the scaling from the value associated to a given
+quantile and the amplitude is asymptotically fixed, for a large number of oscillations.
+Assuming that samples are dense enough that they could be represented by the continuous
+distribution, essentially projecting a uniform measure over an interval through a single
+sinusoidal oscillation.
+
+.. todo::
+
+    Move the above paragraph to the Rabi equivalent, and replace here with a reference
+
+However, for damped oscillations, the factor is not easily determined, since the
+value associated to a certian quantile depends on the observation window extent, and the
+ratio between the decay rate and the oscillation.
+
+Assuming a mild decay, and we can approximate this factor with the same one for the
+pure oscillation. This can be assumed to be slightly decreased because of the dampening,
+but there is no general control about how much.
+By reducing the amplitude by this rough 30%, the estimation will lend closer to the
+actual amplitude. We rely anyhow on the fit to determine the precise value.
 """
 
 


src/qibocal/protocols/ramsey/utils.py
```diff
@@ -90,13 +90,17 @@ def fitting(x: list, y: list, errors: list = None) -> list:
 
     period = fallback_period(guess_period(x, y))
     omega = 2 * np.pi / period
+    idx = np.argmin(x)
+    mean_sig = (max(y) + min(y)) / 2
+    amplitude_guess = (max(y) - min(y)) / 2
```
alecandido: Even this one is potentially prone to noise, as any feature which depends on single points.
The smaller the group of points involved, the higher the noise sensitivity.

The two options I have in mind to get an estimate here are:
- the variance, which uses all the points (but also sensitive to outliers, as the average)
- some quantiles, e.g. the 0.2 and 0.8 quantile, and then you rescale with the sine function

src/qibocal/protocols/ramsey/utils.py
```diff
@@ -90,13 +90,17 @@ def fitting(x: list, y: list, errors: list = None) -> list:
 
     period = fallback_period(guess_period(x, y))
     omega = 2 * np.pi / period
+    idx = np.argmin(x)
```
alecandido: Isn't this equivalent to
```suggestion
    idx = 0
```
?

src/qibocal/protocols/ramsey/utils.py
```diff
@@ -90,32 +95,40 @@ def fitting(x: list, y: list, errors: list = None) -> list:
 
     period = fallback_period(guess_period(x, y))
     omega = 2 * np.pi / period
+    median_sig = np.median(y)
+    q80 = np.quantile(y, 0.8)
+    q20 = np.quantile(y, 0.2)
+    amplitude_guess = abs(q80 - q20)
+    sin_phase = (y[0] - median_sig) / amplitude_guess
```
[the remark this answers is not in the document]
lballerio: so you suggest to put `phase_guess=0`?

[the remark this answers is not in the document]
alecandido: Ah, yes, you're right

src/qibocal/protocols/ramsey/utils.py
```diff
@@ -90,32 +95,39 @@ def fitting(x: list, y: list, errors: list = None) -> list:
 
     period = fallback_period(guess_period(x, y))
     omega = 2 * np.pi / period
+    median_sig = np.median(y)
+    q80 = np.quantile(y, 0.8)
+    q20 = np.quantile(y, 0.2)
+    amplitude_guess = abs(q80 - q20)
```
alecandido: This is supposed to be divided by `1.62`

```py
[ins] In [1]: import numpy as np
         ...: x = np.linspace(0, 10*2*np.pi, 10000)
         ...: y = np.sin(x)
         ...: amp_guess_var = 1 / np.std(y)
         ...: q80 = np.quantile(y, 0.8)
         ...: q20 = np.quantile(y, 0.2)
         ...: amp_guess_quant = (q80 - q20)

[ins] In [2]: amp_guess_quant
Out[2]: np.float64(1.6178492793570904)
```

https://github.com/qiboteam/qibocal/pull/1288#discussion_r2634886024

alecandido (commented):
This I forgot myself in previous reviews

src/qibocal/protocols/ramsey/utils.py
```diff
@@ -90,32 +95,39 @@ def fitting(x: list, y: list, errors: list = None) -> list:
 
     period = fallback_period(guess_period(x, y))
     omega = 2 * np.pi / period
+    median_sig = np.median(y)
+    q80 = np.quantile(y, 0.8)
+    q20 = np.quantile(y, 0.2)
+    amplitude_guess = abs(q80 - q20) / 1.62  # normalization factor
```
alecandido: As discussed, let's even put `1.5` as well. And since this appears to be a random constant, just place it top-level, and document with the considerations in the comment.

Issue #1300, opened 2025-11-22
Move to ranges for parameters


alecandido:
Sweepers are now standard in Qibolab: they accept ranges defined as `(start, stop, step)`.
They could also allow for an array of values, but that's rarely an actual option for most drivers.

In any case, we could even allow for both, making something like
```py
SweptPoints = Union[range, ArrayLike]
```

But the main idea is that we do not need to rely any longer on `par_start`, `par_stop`, `par_step`, `par_width` - which were mostly needed to make it more familiar for the runcards.
In scripts, generating ranges is just simpler: you type less, and you can pass actual range objects.

> [!NOTE]
> In specific cases, we could make the parameter's range optional, and still allow for something like width and step, in order to support some default central value (e.g. the known sweetspot).
> I.e. using for those parameters `Union[SweptPoints, tuple[int, int]]`.
> However, I would always allow for passing range.

All in all, it means that a flux dependency call, instead of looking like this:
```py
e.qubit_flux(
    nshots=200,
    relaxation_time=10_000,
    bias_step=0.002,
    bias_width=0.06,
    freq_step=1_000_000,
    freq_width=20_000_000,
    drive_amplitude=0.015,
)
```
it should be
```py
e.qubit_flux(
    nshots=200,
    relaxation_time=10_000,
    bias=(0.06, 0.002),
    frequency=(20_000_000, 1_000_000),
    drive_amplitude=0.015,
)
# or, also supported
e.qubit_flux(
    nshots=200,
    relaxation_time=10_000,
    bias=(0.06, 0.002),  # <- this could also be a range
    frequency=range(6_130_000_000, 6_150_000_000, 1_000_000),
    drive_amplitude=0.015,
)
```

> [!NOTE]
> This should also help making the interface more standard and familiar.

> [!TIP]
> We could make `range()` mandatory, but also allow for `(float, float, int)` tuples, with the `linspace()` semantics.
> Making it a single type, they could be interchangeable. They will all resolved internally in a coherent way.
>
> **Implementation:** since sweepers will always be implemented with increments, the corresponding `step` for the `linspace()` semantics will be inferred, just dropping a 0.1% fraction, to make sure a value significantly close to `stop` will always be included in the swept values.

PR #1306, opened 2025-11-27
Punchout with attenuation

jevillegasdatTII:
### RO Resonator Power shift using attenuation (log of Power)
The plot reverses the vertical axis so that in increase with power (this is reversed compared to same routine in Qibocal 0.1)

image

### DIFF
--- a/src/qibocal/update.py
+++ b/src/qibocal/update.py
@@ -102,7 +102,7 @@ def lo_attenuation(
 ):
     """Update LO attenuation value in platform for specific qubit."""
     channel = getattr(platform.qubits[qubit], channel_type)
-    lo_channel = getattr(platform.channels[channel], "lo")
+    lo_channel = platform.channels[channel].lo
     platform.update({f"configs.{lo_channel}.power": attenuation})
 
 


src/qibocal/protocols/resonator_spectroscopies/resonator_punchout_attenuation.py
```diff
@@ -0,0 +1,375 @@
...
+        )
+
+    @property
+    def delta_frequency_range(self) -> npt.NDArray[np.float64]:
+        span = self.freq_width
+        return np.arange(
+            -span / 2,
+            (span / 2 - (span / 2) % self.freq_step) + self.freq_step,
+            self.freq_step,
+        )
```
alecandido: I understand that all the juggling for the maximum values is to make sure that a single point beyond that will be included. And no other further one.

The most basic observation is that, instead of repeating your trick twice, at least a helper function could be introduced, suitably documenting the purpose of the manipulation (i.e. "(start, stop, step), ensuring that one point beyond max will be included").

However, I'm not sure in general it is a good idea to include a point beyond the maximum value specified. Since it could be unexpected to the user (though here they are carefully placed in the harmless direction, i.e. maximum attenuation = minimum power).

Instead, I would propose to rely on standard semantics of a `(start, stop, step)` range.
If that is not what you want, consider using the `linspace()` semantics, and replace `step` by `npoints`, which will make sure to include the extremes (possibly internally translating to a step, since increments are easier for hardware, but etching an $\varepsilon$ (e.g. a 0.1%) from the inferred step value, to make sure the maximum will be included).

At some point, this won't be required per-protocol, but it will be provided as a standard service and interface, ready-made for all protocols' parameters, and consistent for the end user #1300

jevillegasdatTII: I typically use a `np.range(min, max + step, step) `  and this works fine if the range is a multiple of the step, but it can get messy when it isn't, and that's why I include those modulos. But I agree this can all be the responsibility of the user to keep the section here concise (and under control of what the user intended). I will update it to just 
```python
return np.arange(
            self.min_attenuation,
            self.max_attenuation,
            self.step_attenuation,
        )
```
so that is more homogenous with other protocols.

PR #1316, opened 2025-12-09
`nshots` fix for Ramsey signal

rodolfocarobene:
Quite self-explanatory, the nshots parameter was missing in this routine and it was not updated from the runcard.
I quickly looked all the other routines and it seems to me that this was the only one where it was missing.


alecandido:
Agreed, this is definitely an error. Thanks for fixing it!

PR #1318, opened 2025-12-16
Add sweeper support for 1Q RB

sorewachigauyo:
Instead of using circuits in the standard RB, we manually transpile the U3 Cliffords and generate a pulse sequence. As the decomposition gives a fixed RX & RZ sequence, we can sweep the RZ angles corresponding to each random sequence for a given circuit depth.

This PR also provides an interface to do RB on multiple qubits in parallel, but they are currently sharing the same sequence to make execution simpler and decrease the number of sweepers/sweep parameters.

Currently there is an issue with the VZ sweepers on my hardware so I'm waiting for that to be resolved before testing on QPU, but I've done some loose checks to test that both the decomposed circuit output and the manual transpilation give the same output.

TODO
~~Write tests~~
- [x] Test on QPU
- [x] Test parallel execution

### DIFF
--- a/src/qibocal/protocols/randomized_benchmarking/standard_rb_sweeper.py
+++ b/src/qibocal/protocols/randomized_benchmarking/standard_rb_sweeper.py
@@ -1,15 +1,14 @@
 from collections import defaultdict
 
 import numpy as np
-from qibo import Circuit
 from qibo.gates import U3, Unitary
 from qibo.transpiler.unitary_decompositions import u3_decomposition
 from qibolab import (
     AcquisitionType,
     AveragingMode,
     Parameter,
-    Pulse,
     PulseSequence,
+    Readout,
     Sweeper,
     VirtualZ,
 )
@@ -27,10 +26,14 @@ from qibocal.protocols.randomized_benchmarking.utils import (
     RB_Generator,
     RBData,
     RBType,
+    add_inverse_layer,
+    layer_circuit,
     setup,
 )
 
 __all__ = ["standard_rb_sweeper"]
+
+NUM_VZ_PER_CLIFFORD = 3
 
 
 def _acquisition(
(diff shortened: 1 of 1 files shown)


sorewachigauyo:
Tests on QPU

The parallel RB seems to work but eats up a lot of compile time in return on my instrument.

[rb-tests.zip](https://github.com/user-attachments/files/24187669/rb-tests.zip)

alecandido:
> For the keysight, the number of sweep points (`niter x depth x 3`) limit for hardware is 24576, so 10 x 800 x 3 was already quite close. After that we fallback to software sweeping.

Apparently, QM is limiting the amount of points per value sweeper
image
Though, not declaring exactly how many. And not declaring how many loops are available in parallel.

With Qblox instead (or Qblox-like architecture with value sweepers available) we would need a different compilation from the current one, specific for this experiment: if we relied on one sweeper per `VZ`, it is as I said before. But, if we unravel all the values in a single 1D array (instead of `(niter, 3 * depth)`), we could just increment a single iterator for all the `VZ` values. But this would require a dedicated infrastructure to increment it every time it is used (and value sweepers are not available anyhow).

In any case, if for you is working, and on a practical scale, that's just great.
I already took a look into the implementation, and I haven't seen anything particularly strange. So, as soon as you'll be ready, I believe we could even merge this fast.

src/qibocal/protocols/randomized_benchmarking/standard_rb.py
```diff
@@ -11,7 +11,7 @@
 from .fitting import exp1B_func
 from .utils import RBData, StandardRBResult, fit, number_to_str, rb_acquisition
 
-__all__ = ["standard_rb", "StandardRBParameters", "RBData", "_plot"]
+__all__ = ["standard_rb", "StandardRBParameters", "RBData", "_plot", "_fit", "_update"]
```
alecandido: I would say that even including `_plot` was an error (I did not realize before).

The purpose of `__all__` is to publicly export outside `qibocal` itself. Instead, the role of the leading `_`is to keep it private to the present module.

In principle, there is no reason to name the fit function `_fit` (unfortunately, in practice at some point Qibocal started reading the name of this function, which should have never happened - it should have relied on the name of the `Routine` object attribute).
Thus, I initially named the first ones `_acquisition`, `_fit`, `_plot` because they were only relevant for their own modules, and they had no role outside.

In summary:
- long(er) term, if you use them outside the module, give some meaningful name; if nothing else, avoid the leading underscore: `fit` is fine if it's the only one in the module, such that it will be scoped by the module name as `standard_rb.fit`
- right now, just strip all the elements with the prepended `_` from `__all__` - they are certainly not part of the user interface

sorewachigauyo: Fixed in https://github.com/qiboteam/qibocal/pull/1318/commits/a53888490df445b289036bdfe84454a3a85a9b22

src/qibocal/protocols/randomized_benchmarking/standard_rb_sweeper.py
```diff
@@ -0,0 +1,165 @@
...
+                    + [(qubit_drive, vz_phi)]
+                )
+
+        for qubit_id in targets:
+            qubit = platform.qubits[qubit_id]
+            sequence.align([qubit.drive, qubit.acquisition])
+
+            ro_channel, ro_pulse = platform.natives.single_qubit[qubit_id].MZ()[0]
+            ro_pulses[qubit_id] = ro_pulse
+            sequence += [(ro_channel, ro_pulse)]
```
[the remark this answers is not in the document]
sorewachigauyo: Resolved in https://github.com/qiboteam/qibocal/pull/1318/commits/138156cbc97151dbb69693cc3165bdc1df4bf95c

I did `sequence += qseq` instead because I didn't want and need to invoke the channel synchronization

sorewachigauyo:
https://github.com/qiboteam/qibocal/pull/1318/commits/138156cbc97151dbb69693cc3165bdc1df4bf95c tests

[index.html](https://github.com/user-attachments/files/24228938/index.html)

PR #1319, opened 2025-12-16
Stabilize extract feature

lballerio:
this PR means to improve extract_feature function in protocols/utils.py.
Before that the signal was selected from the background noise by first rescaling the whole signal by the signal average and then by exploiting find_peaks function from scipy.signal.

This approach indeed is quite problematic when the SNR of the signal il not big enough, and it leads to an error since find_peaks doesn't find any peak at all.
This error can be recreated for example by running qubit_flux experiment for small number of shots and low relaxation time for qubits with low SNR (in the case of qw21q-b chip B1).
find_peaks_fail

We addressed this problem from scratch, by first applying to the input signal a custom mask made by a ZCA transformation and a Gaussian filter, then by rescaling from 0 to 1 the masked signal; the normalized signal is then used in find_peaks. then the peaks are used in a clustering algorithm for classify real signal to background noise. This is the result for same qubit B1 for qubit_flux experiment with same parameters:
clustering_good

report of the previous approach:
http://login.qrccluster.com:9000/fSGyUI49QDGSzGp9bFXsrg==

report of the improved one:
http://login.qrccluster.com:9000/DPlqPIoiQk6OJ4YcfYeDCg==

zipped repo of the whole experiment
[improved_mask_clustering.zip](https://github.com/user-attachments/files/24186029/improved_mask_clustering.zip)

### DIFF
--- a/src/qibocal/protocols/utils.py
+++ b/src/qibocal/protocols/utils.py
@@ -49,7 +49,7 @@ Since it is given in a 3D-space, with a compressed vertical dimension, and the h
 this distance correspond to diagonally adjacent pixels, with some additional leeway for the extra dimension.
 """
 DISTANCE_Z = 0.5
-"""See :py::const:DISTANCE_XY."""
+"""See :const:`DISTANCE_XY`."""
 
 
 class PowerLevel(str, Enum):


src/qibocal/protocols/utils.py
```diff
@@ -757,39 +762,131 @@ def table_html(data: dict) -> str:
...
 
 
-def extract_feature(
-    x: np.ndarray,
-    y: np.ndarray,
-    z: np.ndarray,
-    find_min: bool,
-):
-    """Extract feature using confidence intervals.
+def zca_whiten(X):
```
alecandido: Actually, a `whiten` implementation already exists in SciPy.
It's even a bit worse: there are even two implementations in there
- [`scipy.cluster.vq.whiten`](https://scipy.github.io/devdocs/reference/generated/scipy.cluster.vq.whiten.html#scipy.cluster.vq.whiten)
- [`scipy.stats.Covariance.whiten`](https://scipy.github.io/devdocs/reference/generated/scipy.stats.Covariance.whiten.html)

However, the function is small enough that we can even keep it as it is, as it also gives better control over the implementation (apparently, whitening is not unique, and the ZCA whitening is just an option).

alecandido (commented):
One batch of comments about the whitening

src/qibocal/protocols/utils.py
```diff
@@ -757,39 +762,248 @@ def table_html(data: dict) -> str:
...
+    #   W_zca = E * D * E.T
+    W = np.dot(np.dot(E, D), E.T)
+
+    X_white = np.dot(X, W)
+
+    return X_white
+
+
+def custom_filter_mask(matrix_z: np.ndarray):
+    """Applying a mask compsosed by first a ZCA transformation and then a gaussian filter with variance 1."""
```
alecandido: A ZCA transformation is not a convolutional mask. And the Gaussian is not really custom...
```suggestion
def filter_data(matrix_z: np.ndarray):
    """Filter data with a ZCA transformation and then a unit-variance Gaussian."""
```

src/qibocal/protocols/utils.py
```diff
@@ -757,39 +762,248 @@ def table_html(data: dict) -> str:
...
+    return zca_gauss_z
+
+
+def scaling_global(sig: np.ndarray) -> np.ndarray:
+    """Min–max scaling over the whole np.ndarray (global)."""
+    sig_min = np.min(sig)
+    return (sig - sig_min) / (np.max(sig) - sig_min)
+
+
+def scaling_slice(sig: np.ndarray, axis: int) -> np.ndarray:
```
alecandido:
```suggestion
def scaling_slice(sig: np.ndarray, axis: Optional[int]) -> np.ndarray:
```

Let's not restrict the range of options the NumPy accepts ;)

src/qibocal/protocols/utils.py
```diff
@@ -757,39 +762,248 @@ def table_html(data: dict) -> str:
...
+    for i, values in enumerate([xs, ys]):
+        sizes[i] = np.max(values) - np.min(values)
+    return np.sqrt((sizes**2).sum())
+
+
+def build_clustering_data(peaks_dict: dict, z: np.ndarray):
+    """Preprocessing of the data to cluster."""
+    peaks_x = peaks_dict["x"]["idx"]
+    peaks_y = peaks_dict["y"]["idx"]
+    peaks_sf = z[peaks_y, peaks_x]
```
alecandido: Do you remember the reason for the `_sf`?

Otherwise, maybe just name it `_z` :)

src/qibocal/protocols/utils.py
```diff
@@ -757,39 +762,248 @@ def table_html(data: dict) -> str:
...
+
+    diag = horizontal_diagonal(peaks_x, peaks_y)
+    diag = np.sqrt(2)
+    return np.stack((peaks_x, peaks_y, scaling_global(peaks_sf) * diag)).T
```
alecandido: Since the function is all related to peaks, maybe we can stop prepending `peaks_` to all the variables

alecandido (commented):
First batch

src/qibocal/protocols/utils.py
```diff
@@ -757,39 +762,248 @@ def table_html(data: dict) -> str:
...
+            peaks["x"]["val"].append(x[x_idx])
+            peaks["y"]["idx"].append(y_idx)
+            peaks["y"]["val"].append(y_val)
+
+    peaks_dict = {
+        feat: {kind: np.array(vals) for kind, vals in smth.items()}
+        for feat, smth in peaks.items()
+    }
+
+    return peaks_dict
```
alecandido:
```suggestion
    return {
        feat: {kind: np.array(vals) for kind, vals in smth.items()}
        for feat, smth in peaks.items()
    }
```

src/qibocal/protocols/utils.py
```diff
@@ -757,39 +762,248 @@ def table_html(data: dict) -> str:
...
-                filtered_y.append(i)
-                filtered_x.append(x_[j])
+    peaks_dict = peaks_finder(x_, y_, z_masked_norm)
 
-    return np.array(filtered_x), np.array(filtered_y)
+    # normalizing peaks for clustering
+    peaks = build_clustering_data(peaks_dict, z_masked)
+
+    # clustering
+    distance = np.sqrt(2) + 0.5
```
alecandido: This is a constant. Constants in Python are just named in `SCREAMING_CASE` and placed top-level.

So, just put it top-level (i.e. outside any function), and please document this value.
```py
DISTANCE = np.sqrt(2) + 0.5
"""Minimum distance for separate clusters.

Clusters below this distance will be merged.
Since it is given in a 3D-space, with a compressed vertical dimension, and the horizontal plane measured in pixels, this distance correspond to diagonally adjacent pixels, with some additional leeway for the extra dimension.
"""
```

src/qibocal/protocols/utils.py
```diff
@@ -757,39 +762,248 @@ def table_html(data: dict) -> str:
...
+def scaling_global(sig: np.ndarray) -> np.ndarray:
+    """Min–max scaling over the whole np.ndarray (global)."""
+    sig_min = np.min(sig)
+    return (sig - sig_min) / (np.max(sig) - sig_min)
+
+
+def scaling_slice(sig: np.ndarray, axis: int) -> np.ndarray:
+    """Min–max scaling over a specific axis of the np.ndarray."""
+    sig_min = np.min(sig, axis=axis)
+    return ((sig.T - sig_min) / (np.max(sig, axis=axis) - sig_min)).T
```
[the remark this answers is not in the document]
lballerio: @alecandido did you test `scaling_global` with this new implementation? because `np.expand_dims` requires the axis to e an `int`, so I would rewrite the whole function as:

```suggestion
    expand = lambda a: np.expand_dims(a, axis) if axis is not None else a
    sig_min = expand(np.min(sig, axis=axis))
    return (sig - sig_min) / (expand(np.max(sig, axis=axis)) - sig_min)
```

src/qibocal/protocols/utils.py
lballerio: no I do not ahahahaah

src/qibocal/protocols/utils.py
```diff
@@ -757,39 +762,243 @@ def table_html(data: dict) -> str:
...
+    for i, values in enumerate([xs, ys]):
+        sizes[i] = np.max(values) - np.min(values)
+    return np.sqrt((sizes**2).sum())
+
+
+def build_clustering_data(peaks_dict: dict, z: np.ndarray):
+    """Preprocessing of the data to cluster."""
+    peaks_x = peaks_dict["x"]["idx"]
+    peaks_y = peaks_dict["y"]["idx"]
+    peaks_sf = z[peaks_y, peaks_x]
```
lballerio:
```suggestion
    x_ = peaks_dict["x"]["idx"]
    y_ = peaks_dict["y"]["idx"]
    z_ = z[peaks_y, peaks_x]
```
(so we do not keep the same name for `z`)

src/qibocal/protocols/utils.py
```diff
@@ -757,39 +762,243 @@ def table_html(data: dict) -> str:
...
+
+    diag = horizontal_diagonal(peaks_x, peaks_y)
+    diag = np.sqrt(2)
+    return np.stack((peaks_x, peaks_y, scaling_global(peaks_sf) * diag)).T
```
lballerio:
```suggestion
    return np.stack((x_, y_, scaling_global(z_) * rescaling_fact)).T
```

src/qibocal/protocols/utils.py
```diff
@@ -35,10 +36,6 @@
 H = constants.h
 COLORBAND = "rgba(0,100,80,0.2)"
 COLORBAND_LINE = "rgba(255,255,255,0)"
-CONFIDENCE_INTERVAL_FIRST_MASK = 99
-"""Confidence interval used to mask flux data."""
-CONFIDENCE_INTERVAL_SECOND_MASK = 70
-"""Confidence interval used to clean outliers."""
 DELAY_FIT_PERCENTAGE = 10
 """Percentage of the first and last points used to fit the cable delay."""
 STRING_TYPE = "<U100"
```
lballerio:
```suggestion
STRING_TYPE = "<U100"

DISTANCE = np.sqrt(2) + 0.5
"""Minimum distance for separate clusters.

Clusters below this distance will be merged.
Since it is given in a 3D-space, with a compressed vertical dimension, and the horizontal plane measured in pixels, this distance correspond to diagonally adjacent pixels, with some additional leeway for the extra dimension.
"""
```

src/qibocal/protocols/utils.py
```diff
@@ -757,39 +762,243 @@ def table_html(data: dict) -> str:
...
-                filtered_y.append(i)
-                filtered_x.append(x_[j])
+    peaks_dict = peaks_finder(x_, y_, z_masked_norm)
 
-    return np.array(filtered_x), np.array(filtered_y)
+    # normalizing peaks for clustering
+    peaks = build_clustering_data(peaks_dict, z_masked)
+
+    # clustering
+    distance = np.sqrt(2) + 0.5
```
lballerio:
```suggestion
    # clustering
```

src/qibocal/protocols/utils.py
```diff
@@ -281,16 +278,24 @@ def fit_punchout(data: Data, fit_type: str):
...
         signal = qubit_data.signal
         mask_freq, mask_amps = extract_feature(freqs, amps, signal, data.find_min)
-        if fit_type == "amp":
-            best_freq = np.max(mask_freq)
-            bare_freq = np.min(mask_freq)
+
+        if mask_freq.size == 0:  # mask_freq and mask_amps have always the same shape
+            best_freq = 0
+            bare_freq = 0
+            ro_val = 0
```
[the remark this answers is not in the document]
lballerio: won't this raise an error in `_fit`?

src/qibocal/protocols/utils.py
```diff
@@ -757,39 +762,253 @@ def table_html(data: dict) -> str:
...
+    It slices the dataset along a preferred direction (`y` dimension, corresponding to the flux bias) and for each slice it determines the biggest peaks
+    by using `scipy.signal.find_peaks` routine.
+    It returns a dictionary `peaks_dict` containing all the features for the computed peaks.
     """
+
+    # filter data using find_peaks
+    peaks = {"x": {"idx": [], "val": []}, "y": {"idx": [], "val": []}}
+    for y_idx, y_val in enumerate(y):
+        signal_fixed_y = z[y_idx]
+        peak, info = find_peaks(signal_fixed_y, prominence=0.2)
```
alecandido: At this point, the name is more confusing than helpful
```suggestion
        peak, info = find_peaks(z[y_idx], prominence=0.2)
```

If it was used more than once, the assignment was almost mandatory. And even when it is used once, it may be worth.
But in this case, the name is not really clarifying that much the value. And it is possibly more confusing.

src/qibocal/protocols/utils.py
```diff
@@ -757,39 +762,251 @@ def table_html(data: dict) -> str:
...
+    hdb = HDBSCAN(copy=True, min_cluster_size=2)
+    hdb.fit(peaks)
+    labels = hdb.labels_
 
-    return np.array(filtered_x), np.array(filtered_y)
+    DISTANCE = 1.5 * scaling_factor  # very heuristic
+    # Minimum distance for separate clusters.
+    # Clusters below this distance will be merged.
+    # Since it is given in a 3D-space, with a compressed vertical dimension, and the horizontal plane measured in pixels,
+    # this distance correspond to diagonally adjacent pixels, with some additional leeway for the extra dimension.
```
alecandido: Still wrong: put it outside ***all*** functions.

To be more explicit: no indentation at all.

alecandido:
https://peps.python.org/pep-0008/#constants

alecandido (dismissed):

alecandido:
@lballerio as you say, there is no theoretically-perfect recipe. We are going a bit by trial and error.

However, I would revert what you did, based on a couple of more general arguments.

1. The `DISTANCE = 5` is surprisingly high, since that's only used by the cluster-merging algorithm, and it is measured in pixels: we want to reconnect clusters which are actually close, and usually this means that the boundaries are almost *touching* each other, so they are either 1-pixel distant horizontally (same frequency, bias different by a single step), or diagonally adjacent, thus $\sqrt{2}$ distance. This is not taking into account the vertical separation in `z`, but the idea of having $1.5$ was exactly to keep it a bit larger than $\sqrt{2}$ for that reason. If you increase even beyond $2$, you start gathering together clusters which are not strictly adjacent horizontally. This could be useful, in some cases, but it is also likely to bring in noise - it should be done very carefully, when we are confident we could separate noise to that extent.
2. I strongly disagree about giving a fixed dimension to the `z` axis, since this will keep the ratio with the `xy`-plane floating. The idea to scale with the horizontal diagonal was exactly because that one is more or less scaling with the number of pixel. Instead of using `sqrt(x**2 + y**2)` we could have used `sqrt(xy)`. The purpose is just to compute something which is scaling with the amount of pixels, but also scales linearly with the individual dimensions (if you double both `x` and `y`, we'd like to double also `z`, not multiply it by four). If instead is left free-floating, then it could be irrelevant for many pixels, but the fewer pixels you take, the more `z` becomes relevant. Which I believe it is completely uncontrolled and unintended. For this reason, I would keep the `horizontal_diagonal()` scaling, and rather suppress by a larger number (either divide by 20, or, at most, even multiply by 0, and completely disregard that dimension for some time - leaving an issue to remind we could investigate, to extract even more information, that could potentially better distinguish signal and noise)

I see the "very heuristic" approach. And I can not provide much more than heuristic. But if the heuristic is motivated, it has a chance to generalize. If it's just uneducated trial and error, it is likely that you will keep changing those numbers for a very long time (i.e. forever...)

alecandido:
What you did in 6e135fc also does not look right: having applied the scale factor to the `DISTANCE` it propagates on all dimensions, not just `z`. Thus, 1.5 becomes a relative value once again, and it loses its meaning in pixel units.

If you need a larger `DISTANCE`, you should motivate it. Not just try to inflate the value somewhere else.
The problem is still generalization: "just working" on the example you have at a hand is almost meaningless. While, if the "motivated" algorithm (still heuristically) fails on the example, that's meaningful, and it should inform the reasoning about a different algorithm. Which should still be motivated as well, on a general ground.

We need a reliable algorithm, especially to run automatically in pipelines. And this means that it should necessarily work on never-seen-before data.

lballerio:
I am still thinking to merge the clusters in a 3D space, where hence the concept of pixel looses a bit its meaning, and this is also why I decided to rescale the distance according to how much we stretched the z-direction. So the alternative I might see is to merge only in the x-y plane where then pixel is meaningful. Note that this is valid only for the `merging` part, not the `clustering` part (where we'll keep the z-axis). Anyway by doing this we'll draw back in edbe370 since `DISTANCE` parameter will only be an arbitrary constant value given which however symbolizes the min distance in pixel between two clouds in the xy plane.
Anyway putting `DISTANCE=1.5` might still be too small since you're cutting off clusters from the same signal but let's say distant 3 pixels (i.e. `ResonatorPunchout`, even though we can also treat high and low ampl signals as two separate signals).
what do you think @alecandido?

src/qibocal/protocols/utils.py
```diff
@@ -757,39 +770,266 @@ def table_html(data: dict) -> str:
...
+
+    indices_list = np.arange(len(labels)).astype(int)
+    indexed_labels = np.stack((labels, indices_list)).T
+    data = np.vstack((data.T, indices_list))
+
+    clusters = [data[:, labels == lab] for lab in unique_labels if lab >= 0]
+    noise_points = data[:, labels < 0]
+
+    for i in range(noise_points.shape[1]):
+        clusters.append(np.array(noise_points[:, i])[:, np.newaxis])
```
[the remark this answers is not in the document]
lballerio: no no you are right, it is indeed redundant

src/qibocal/protocols/utils.py
```diff
@@ -35,14 +36,20 @@
...
-"""Confidence interval used to mask flux data."""
-CONFIDENCE_INTERVAL_SECOND_MASK = 70
-"""Confidence interval used to clean outliers."""
 DELAY_FIT_PERCENTAGE = 10
 """Percentage of the first and last points used to fit the cable delay."""
 STRING_TYPE = "<U100"
 
+MAX_PIXEL_DISTANCE = 2
+"""Maximul pixel distance of signals to be merged in a single cluster"""
+DISTANCE_XY = 1.5 * MAX_PIXEL_DISTANCE  # very heuristic
```
alecandido: Ok, I'm not sure exactly what you mean here. But if you use just a single constant, I would actually suggest naming only that one (in this case `DISTANCE_XY`) and documenting that value, even if originated from multiple parts.

Of course, if it becomes much more complex than a product of two numbers, it could be worth splitting it.

My guess is that you want to use the `1.5` as the distance between diagonally-adjacent points (plus some mild extra), and then also merge points which are at distance 2 in the diagonal.

If that's what you mean, fine, it could be inferred. But it is not really written there...

alecandido: Ok, this is just a documentation issue. Just try to add more details, explaining it to someone who did not work at the original implementation

alecandido (approved):
Minor review for the docstrings

PR #1321, opened 2025-12-16
Long qubit spectroscopy

jevillegasdatTII:
This is a possible implementation of the loooong spectroscopy.
The fit and report functions are the same as the opriginal qubit spectrscopy, and the main difference is that in the acquisition there is a for loop that runs the spectroscopy updating the LO each time. I tested this with one qubit and it looks like its working well.

newplot (4)

I am currently hardcoding the maximum IF frequency before splitting the job, for qblox this may be available in the codes validator of the qcm (qrm) driver. But for other electronics I don't know. Maybe qibolab would need to expose something that can be accessed to be able to have a generic implementation (+- 300 MHz should work for QM and QBlox)

I was having issues with the validation of the `nco_freq`:  When setting the initial values (lo and nco frequencies) the validator compares against whatever is on the json specification after updates (which for a very wide spectroscopy may set an nco too far from the updated lo  frequency), even though the nco frequency is later swept in a reasonable range. So a dirty fix was to also send an update of the drive frequency to some value in range, I used
```python
update_dict[drive_channels[qubit]] = {
                "frequency": platform.config(drive_channels[qubit]).frequency + lo_offset
            }
```

### DIFF
--- a/src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
+++ b/src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
@@ -284,13 +284,4 @@ def _update(
 
 qubit_spectroscopy = Routine(_acquisition, _fit, _plot, _update)
 """Qubit Spectroscopy routine.
-
-A qubit spectroscopy sweeps the drive frequency around the qubit
-frequency at which the qubit's transition from |0> to |1> occurs.
-Typically a long pulse is used to ensure a narrow frequency spectrum of the
-drive, and helping the qubit to get excited to a superposition state.
-
-When the frequency sweep width exceeds the IF bandwidth range (+/- 300 MHz),
-the routine automatically splits the sweep into multiple batches, adjusting
-the LO frequency accordingly for each batch.
 """
--- a/doc/source/protocols/qubit_spectroscopy/qubit_spectroscopy.rst
+++ b/doc/source/protocols/qubit_spectroscopy/qubit_spectroscopy.rst
@@ -22,6 +22,10 @@ to the transition frequency  :math:`w_{01}` some of the population will
 move to the excited state. If the drive pulse is long enough it will be
 generated a maximally mixed state with :math:`\rho \propto I` :cite:p:`Baur2012RealizingQG, gao2021practical`.
 
+When the frequency bandwidth measured exceeds the IF bandwidth range (+/- 300 MHz),
+the routine automatically splits the sweep into multiple batches, adjusting
+the LO frequency accordingly for each batch.
+"""
 
 Parameters
 ^^^^^^^^^^


jevillegasdatTII:
> @jevillegasdatTII maybe we should add a quick mixer calibration every time the LO is moved (as something to be done, possibly an issue).
> 
> It could be even optional, controlled by a flag exposed in the routine's parameters.
> 
> If we add room for something like this, then this should definitely become a Qibocal feature, and we should leave it outside Qibolab. We could even make it a general function, such that we could reuse for all routines which are performing frequency sweeps.

Yes! That would be ideal. I haven't checked on the status of the update of mixer info, is it working already for qblox??

alecandido:
> I haven't checked on the status of the update of mixer info, is it working already for qblox??

I have a PR in which I'm setting them. But I did not have the chance to test it yet.

Let me rebase it on top of the filters, and we can resume from there!

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
```diff
@@ -69,91 +69,223 @@ class QubitSpectroscopyData(ResonatorSpectroscopyData):
...
+    freq_width: int, freq_step: int, max_if_bandwidth: int = 300_000_000
+):
+    """
+    Calculate frequency batches for wideband spectroscopy.
+
+    """
+    # If freq_width fits within the IF bandwidth, no batching needed
+    if freq_width <= 2 * max_if_bandwidth:
+        delta_frequency_range = np.arange(-freq_width / 2, freq_width / 2, freq_step)
+        return [{"delta_freq_range": delta_frequency_range, "lo_offset": 0}]
```
alecandido: This should not be needed: your algorithm below should give the same result if the condition is matched.

The only difference would be the `lo_offset`, which could be solved just by using:
```py
if len(batches) == 1:
   batches[0]["lo_offset"] = 0.0
```
which should make it completely equivalent.

However, I believe that the implication is still a bit wrong: even if the span falls in between the bandwidth, it's not enough to keep the LO still. You need to match the more restrictive condition to be in an interval centered around the LO.

In that case, the condition depends on the actual frequencies, and not only on the scan width.
I would treat it outside this function.

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
```diff
@@ -69,91 +69,223 @@ class QubitSpectroscopyData(ResonatorSpectroscopyData):
...
+        delta_frequency_range = batch["delta_freq_range"]
+        lo_offset = batch["lo_offset"]
+
+        # ----->>> Remover after tests
+        if len(batches) > 1:
+            print(
+                f"Executing batch {batch_idx + 1}/{len(batches)}: "
+                f"LO offset = {lo_offset / 1e6:.1f} MHz, "
+                f"sweep range = [{delta_frequency_range[0] / 1e6:.1f}, {delta_frequency_range[-1] / 1e6:.1f}] MHz"
+            )
```
alecandido: Ok, I guess here your comment is already explicit enough :P

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
```diff
@@ -69,91 +69,223 @@ class QubitSpectroscopyData(ResonatorSpectroscopyData):
...
-        f0 = platform.config(platform.qubits[qubit].drive).frequency
-        signal = magnitude(result)
-        _phase = phase(result)
-        if len(signal.shape) > 1:
-            error_signal = np.std(signal, axis=0, ddof=1) / np.sqrt(signal.shape[0])
-            signal = np.mean(signal, axis=0)
-            error_phase = np.std(_phase, axis=0, ddof=1) / np.sqrt(_phase.shape[0])
-            _phase = np.mean(_phase, axis=0)
+        # Handle errors
+        if all_error_signals[qubit][0] is not None:
```
alecandido: Uhm, I assume that your intention here is to check whether one is not `None`, assuming the rest will be the same.

I'm not sure how much the assumption is reliable, or required. I don't see it likely to run a qubit spectroscopy in classification... (though it is certainly possible...)

I'm inclined to either remove the option at all, are make the correct check:
```suggestion
        if all(x is not None for x in all_error_signals[qubit]):
```
which is just more conservative.

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
```diff
@@ -69,91 +69,223 @@ class QubitSpectroscopyData(ResonatorSpectroscopyData):
...
+        lo_offset = batch_center
+
+        # Frequency range relative to the batch center (LO position)
+        delta_freq_range = np.arange(
+            batch_start - batch_center, batch_end - batch_center, freq_step
+        )
+
+        batches.append({"delta_freq_range": delta_freq_range, "lo_offset": lo_offset})
+
+    return batches
```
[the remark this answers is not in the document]
jevillegasdatTII: Yes! Very clean :) 
Actually, we don't even need to manually add the last value to steps since np.arange already returns the start value for each batch. I am leaving this as:

```python
def _calculate_batches(
    freq_width: int, max_if_bandwidth: int = 300_000_000
):
    batch_starts = np.arange(-freq_width / 2, freq_width / 2, 2 * max_if_bandwidth)
    batch_ends = np.append(batch_starts[1:], freq_width / 2)
    batch_limits = np.stack((batch_starts, batch_ends))
    lo_offsets = batch_limits.sum(axis=0) / 2 if len(batch_starts) > 1 else [0]
    return np.vstack((batch_limits, lo_offsets)).T
```

PR #1331, opened 2026-01-09
nex initial guess for rabi fitting

lballerio:
this PR aims to make the same fixes as the ones used in [PR #1288](https://github.com/qiboteam/qibocal/pull/1288) but for Rabi experiments.
The main part of this PR is simply redefining the initial guess of the fitting parameters in a more robust way, in order to decrease the number of failing fits.

Screenshot from 2026-01-09 13-10-45

rabi_good_plot

### DIFF
--- a/src/qibocal/protocols/rabi/length_signal.py
+++ b/src/qibocal/protocols/rabi/length_signal.py
@@ -147,7 +147,7 @@ def _fit(data: RabiLengthSignalData) -> RabiLengthSignalResults:
         x = (rabi_parameter - x_min) / (x_max - x_min)
         y = (voltages - y_min) / (y_max - y_min) - 1 / 2
 
-        pguess = utils.rabi_initial_guess(x, y, "length")
+        pguess = utils.rabi_initial_guess(x, y, "length", signal=True)
 
         try:
             popt, _, pi_pulse_parameter = utils.fit_length_function(
--- a/src/qibocal/protocols/rabi/utils.py
+++ b/src/qibocal/protocols/rabi/utils.py
@@ -17,7 +17,7 @@ from ..utils import (
     table_html,
 )
 
-QUANTILE_CONSTANT = 1.6
+QUANTILE_CONSTANT = 1.5
 """Scaling factor to recover signal amplitude from quantiles.
 
 Measuring intermediate quantiles is less noise sensitive then meauring extremal points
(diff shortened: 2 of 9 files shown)


src/qibocal/protocols/ramsey/utils.py
```diff
@@ -15,29 +15,7 @@
...
-However, for damped oscillations, the factor is not easily determined, since the
-value associated to a certian quantile depends on the observation window extent, and the
-ratio between the decay rate and the oscillation.
-
-Assuming a mild decay, and we can approximate this factor with the same one for the
-pure oscillation. This can be assumed to be slightly decreased because of the dampening,
-but there is no general control about how much.
-By reducing the amplitude by this rough 30%, the estimation will lend closer to the
-actual amplitude. We rely anyhow on the fit to determine the precise value.
+"""See :py:const:`rabi.utils.DAMPED_CONSTANT` for details.
```
alecandido:
```suggestion
"""See :const:`rabi.utils.DAMPED_CONSTANT` for details.
```

src/qibocal/protocols/rabi/utils.py
```diff
@@ -7,7 +7,29 @@
 from qibocal.auto.operation import Parameters, QubitId
 from qibocal.update import replace
 
-from ..utils import COLORBAND, COLORBAND_LINE, table_dict, table_html
+from ..utils import COLORBAND, COLORBAND_LINE, angle_wrap, table_dict, table_html
+
+DAMPED_CONSTANT = 1.5
```
alecandido: The damped version we only need for the Rabi duration. And, in principle, it is not as needed as for Ramsey, since in Ramsey we explicitly want to see the decay. While in Rabi we mostly focus on shorter durations.

In any case, the Rabi amplitude is not a damped oscillation, since there is no amplitude-dependent decay. In that case, the original value proposed should be a better fit.

src/qibocal/protocols/ramsey/utils.py
```diff
@@ -15,21 +15,11 @@
 THRESHOLD = 0.5
 """Threshold parameters for find_peaks to guess frequency for sinusoidal fit."""
 DAMPED_CONSTANT = 1.5
-"""Scaling factor to recover amplitude from quantiles.
+"""See :const:`rabi.utils.DAMPED_CONSTANT` for details.
```
alecandido: Isn't it called `QUANTILE_CONSTANT` now?

alecandido (commented):
Minor comments, it is mostly good. Thanks :)

src/qibocal/protocols/rabi/length_signal.py
```diff
@@ -148,7 +148,13 @@ def _fit(data: RabiLengthSignalData) -> RabiLengthSignalResults:
         y = (voltages - y_min) / (y_max - y_min) - 1 / 2
 
         period = fallback_period(guess_period(x, y))
-        pguess = [0, np.sign(y[0]) * 0.5, period, 0, 0]
+        median_sig = np.median(y)
+        q80 = np.quantile(y, 0.8)
+        q20 = np.quantile(y, 0.2)
+        amplitude_guess = abs(q80 - q20) / utils.QUANTILE_CONSTANT
+
+        pguess = [median_sig, amplitude_guess, period, np.pi / 2, 0]
```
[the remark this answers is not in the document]
lballerio: if I remember correctly the guessed phase is because of the fitting function, that for ramsey I think is `np.sin` and rabi is `np.cos`, so are either `np.pi` or `np.pi/2` in order to get the right excited state probability for `t=0` or `amp=0`.

PR #1278, opened 2025-10-30
Simplify single shot protocols

andrea-pasquale:
Closes #1241
In this PR I'm planning to simplify the code related to the single experiments.
Some modifications include:

- [x] Use LDA to compute angle threshold
- [x] Drop outdated parameters related to ML classifiers
- [x] Simplify plot using directly plotly
- [x] Add confusion matrix
- [x] Propagate changes to qutrit classification
- [x] Plot qutrit fit 
- [x] Update readout optimization protocols

PR #1334, opened 2026-01-14
SSRO: add average states markers in plots and change the scaling so that the clusters are more visible.

jevillegasdatTII:
Add average states markers in ssro plots and change the scaling so that the clusters are more visible.


lballerio (approved):
looks fine for me, don't know whether to merge after merging #1278 or not...indeed this PR is very simple, so for me we can merge...let me ask @alecandido if he agrees or if he want to wait.

PR #1342, opened 2026-01-15
small correction for merging function - prevent edge cases

lballerio:
I found out a small bug in #1319 that in corner cases might lead to no peak detection. This small fix handle these cases (never happened so far on the test made), but appeared in another experiment which exploits this function.


alecandido:
Maybe just add a comment right above, briefly explaining why it is required. It will be useful when we will read the code again in some months :)

PR #1356, opened 2026-01-28
Add pre-commit to dev dependencies and fix installation docs

elisendaortiz:
The developer installation instructions reference `pre-commit install` but pre-commit is not included in dev dependencies, causing the command to fail.

**Changes:**
- Added pre-commit to dev dependencies
- Updated installation instructions in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,8 @@ For development make sure to install the package using [`poetry`](https://python
 ```sh
 git clone https://github.com/qiboteam/qibocal.git
 cd qibocal
-poetry install
-pre-commit install
+poetry install --with dev
+poetry run pre-commit install
 ```
 
 ## Minimal working example
--- a/poetry.lock
+++ b/poetry.lock
@@ -127,6 +127,18 @@ files = [
 ]
 
 [[package]]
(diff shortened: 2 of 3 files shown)


alecandido:
Hi @elisendaortiz, thanks for your PR!

In general, contributions to our projects are more than welcome, and we are open to any kind of suggestion.

However, specifically for the proposed changes, there is a motivation why `pre-commit` is not listed among the development dependencies.
It is because the recommended way to use Poetry environments is through `poetry run`, which means that your environment will often (or always) be inactive when you run `git commit`. This would cause an issue, since the Git hook will be registered by in `.git/hooks` folder, and activated (made executable with the conventional name), but the CLI provided by the Python package will not be available, since only present in the environment.
This would prevent committing from outside the Python environment, which seems quite an arbitrary and inconvenient limitation - this is why we'd recommend installing `pre-commit` CLI (coming with the Python package) outside the Poetry environment.

You could now chose to close the PR if you do not want to commit any further refactor, or repurpose to explain the rationale detailed above.

Thanks again, and we are looking forward to further contributions ;)

PR #1377, opened 2026-02-10
resonator punchout: correctly detect case where there is no fit

RoyStegeman:
Fit failure was not detected correctly and therefore qibocal would exit with an error, which meant that the data would not be plotted either


lballerio (approved):
looks fine to me, thanks @RoyStegeman!

PR #1385, opened 2026-02-18
update minum delay to be compatible with Quantum Machines

jevillegasd:
A tof routine while using OPX1000 returns
```
Invalidation ERROR time of flight@elements.0/acquisition: Time of flight must be a multiple of 4 and greater or equal than 28 (got 24)
```
This can be changed in the `MINIMUM_TOF` constant (this PR), or perhaps be part of the instrument bounds (part of the driver or in the configs json)


alecandido (approved):
Sorry for the very late reply. I was postponing because of the failing tests, but they are actually failing at random... The change is minimal, and certainly fine.

scarrazza (approved):

Issue #1367, opened 2026-02-05
Punchout data representation


alecandido:
We could get a more comprehensive representation of punchout data, just adding more plots. At the same time, this would lift the constraint on the current ones, such that we can always represent raw data - which are a more faithful representation of the experimental results.

### Semi-Raw Data Plotting

In practice, pure raw data are hardly desired: if the data are hidden just behind a simple projection, then better to plot the suitably projected data.

For the punchout, we could apply this general idea for two instances:

- **phase information:** display the unwrapped phase with cable-delay subtracted
- **(amplitude sweep-only) gradient:** present signal after known gradient subtraction

> [!NOTE]
> #### Why is there a gradient?
> The reason essentially boils down to the usage of pulses.
>
> #### How to subtract it?
> Just divide by the amplitude.
> > If there is any point at zero amplitude, it can be safely discarded.

### Filtered Data Visualization

Add plots that showcase filtered data as supplementary visualizations.

This could be useful to understand the fit behavior: since the peaks-finding algorithm just works on

PR #1386, opened 2026-02-18
Normalize signal data to the amplitude of the pulse sent

jevillegasd:
I know I said that I prefer to have the raw data in the plots. But when we sweep the amplitude of the pulse the instruments does not normalize the signal it reads against that different amplitude, and in consequence, especially for large ranges, the plot of amplitude doesn't show much in the punch out (with @alecandido  we talked about this the other day). 

So we can go from
newplot(60)

to:
newplot(59)

I also added a dummy scaler for the signal visualization (color) to cover 99% of the data range that helps prevent outliers changing the color range.


lballerio (approved):
looks fine for me

lballerio:
can you send me the data?
Idk why the estimated frequency is so far off the signal

alecandido:
This is actually related to #1367.

I also agree with this PR, and it is the best trade-off with a single plot. But I would keep the issue there, to more extensively address the reporting. All in all, adding more plots is not coming to any significant expense, and it could be useful.

PR #1393, opened 2026-02-26
Refactor circuit execution: remove `QibolabBackend` and allow for hardware averaging

### LINKED ISSUE #1358: Optimize RB execution
Even for the basic Qibocal RB, based on the execution of multiple sequences by Qibolab, we have huge delays just because of the overhead caused by using the `QibolabBackend`.

### Single shots

Since the `QibolabBackend` is tuned for circuits execution by end-users, it is bound to return all shots individually (well, we could even change this - but it is not that relevant...), while here we clearly want to average shots on board: the variance will be anyhow computed on the iterations.

Forcing single shots download put a strong pressure on the acquisition memory, which becomes the limitation for unrolling.

#### Implementation

This is quite simple: it is sufficient to instantiate the default compiler and re-assign the downloaded measurements. Consider the following as just a mnemomic sketch
```py
from qibolab import Compiler

compiler = Compiler.default()
sequences = []
measmap = {}
for circ in circuits:
    seq, mmap = compiler.compile(circ, platform)
    sequences.append(seq)
    measmap |= mmap

res = platform.execute(sequences, acquisition_type=AveragingMode.CYCLIC)

return {meas_id: res[acq_id] for meas_id, acq_id in measmap.items()}
```

### LINKED ISSUE #1399: Results order may differ from acquisitions order
Right now `execute_circuits` relies on the assumption that the readout dict is ordered in the same way as the list of sequences. While qibolab works like this for the time being, we should not rely on it. 

For example, improving the batching algorithm in qibolab would break this behavior of the execute function.

RoyStegeman:
This PR changes circuit execution in qibocal by replacing the qibolabbackend with functions native to qibocal. This simplifies the workflow and provides more flexibility to qibocal. The original motivation was a performance improvement of the RB protocol which is achieved in this PR by executing circuits in a CYCLIC manner, and avoiding disconnecting and connecting on every circuit execution as was done before. **While originally motivated by the RB, it propagates similar changes to all protocols using circuits to unify the impementation.** 

The user-input for `unrolling` in the RB is removed. This is not a physics choice relevant for the user and qibocal should just execute protocols as efficiently as possible. 

Sidenote: I did not propagate the change to the recently merged RB with sweepers protocol since this protocol is not to be used by general users and instead  only intended for use by a contributor https://github.com/qiboteam/qibocal/pull/1318

The main difference relevant for developers is that `execute_circuits` now returns a list Counters objects, one per circuit, encoding the number of counts for each state. E.g. for a two-qubit system the Counter object may look like `Counter({'00': 1, '10': 2, '01': 3, '11': 4})`. The rationale for this choice is that so far none of the protocols need the full list of samples so there is no point in providing it. Then while RB only needs the rate of ground state observations, protocols like tomographies need to know the frequency of all states, thus this format provides the minimal information required by the full set of protocols. 

Finally, while working on this PR I made a number of changes improving type-hints and docstrings.

**Summary:** for an RB with `depths=2**np.arange(11)`, `niter=100`, and `nshots=1000` on a specific platform, this PR together with https://github.com/qiboteam/qibolab/pull/1385 leads to a reduction from 1.5h+ to 10 minutes.

### DIFF
--- a/src/qibocal/auto/execute.py
+++ b/src/qibocal/auto/execute.py
@@ -26,6 +26,20 @@ PLATFORM_DIR = "platform"
 """Folder where platform will be dumped."""
 
 
+def _validate_targets(targets):
+    if not isinstance(targets, list):
+        raise TypeError(
+            f"targets must be a list, got {type(targets).__name__} with value: {targets}"
+        )
+
+    invalid_targets = [t for t in targets if not isinstance(t, (tuple, int, str))]
+    if invalid_targets:
+        raise TypeError(
+            "targets must contain only qubit IDs (int/str) or qubit pairs (tuple), "
+            f"got invalid entries: {invalid_targets}"
+        )
+
+
 def _register(name: str, obj):
     """Register object as module.
 
@@ -277,14 +291,7 @@ class Executor:
         if update is not None:
             self.update = update
         if targets is not None:
-            if not isinstance(targets, list) or not all(
-                isinstance(t, (tuple, int, str)) for t in targets
-            ):
-                raise TypeError(
-                    "targets must be a list of qubit IDs (of type int/str) or qubit "
-                    f"pairs (of type tuple), got {type(targets).__name__} with value: "
-                    f"{targets}"
-                )
+            _validate_targets(targets)
             self.targets = targets
 
         # generate output folder
--- a/src/qibocal/auto/execute.py
+++ b/src/qibocal/auto/execute.py
@@ -277,6 +277,14 @@ class Executor:
         if update is not None:
             self.update = update
         if targets is not None:
+            if not isinstance(targets, list) or not all(
+                isinstance(t, (tuple, int, str)) for t in targets
+            ):
+                raise TypeError(
+                    "targets must be a list of qubit IDs (of type int/str) or qubit "
+                    f"pairs (of type tuple), got {type(targets).__name__} with value: "
+                    f"{targets}"
+                )
             self.targets = targets
 
         # generate output folder
--- a/src/qibocal/auto/operation.py
+++ b/src/qibocal/auto/operation.py
@@ -205,14 +205,17 @@ class Data(AbstractData):
     @property
     def qubits(self) -> list[QubitId]:
         """Access qubits from data structure."""
+        # TODO: what is this for? Seems like a hacky way to implement something. Also
+        # means we can't do multi qubit in a sensbile way since, but isntead rely on the
+        # `pairs` property below, since tuples are now used for whatever this is.
         if set(map(type, self.data)) == {tuple}:
-            return list({tuple(sorted(q[:2])) for q in self.data})
+            return list({q[0] for q in self.data})
         return [q for q in self.data]
 
     @property
     def pairs(self):
-        """Access qubit pairs ordered alphanumerically from data structure."""
-        return list({tuple(sorted(q[:2])) for q in self.data})
+        """Access qubit pairs from data structure."""
+        return list({tuple(q[:2]) for q in self.data})
 
     def register_qubit(self, dtype, data_keys, data_dict):
         """Store output for single qubit.
--- a/src/qibocal/auto/operation.py
+++ b/src/qibocal/auto/operation.py
@@ -9,7 +9,7 @@ from typing import Callable, Generic, NewType, Optional, TypeVar, Union
 
 import numpy as np
 import numpy.typing as npt
-from qibolab import AcquisitionType, AveragingMode, Platform, Qubit
+from qibolab import Platform, Qubit
 
 from qibocal.config import log
 
@@ -73,8 +73,11 @@ class Parameters:
     """Wait time for the qubit to decohere back to the `gnd` state."""
     hardware_average: bool = False
     """By default hardware average will be performed."""
-    classify: bool = False
-    """By default qubit state classification will not be performed."""
+
+    # # TODO: this is an inherent propery of the expirment so I don't see why is thould be
+    # # exposed as a parameter. Also thin it's not used anyway, so remove it.
+    # classify: bool = False
+    # """By default qubit state classification will not be performed."""
 
     @classmethod
     def load(cls, input_parameters):
@@ -99,23 +102,24 @@ class Parameters:
             setattr(instantiated_class, parameter, value)
         return instantiated_class
 
-    @property
-    def execution_parameters(self):
-        """Default execution parameters."""
-        averaging_mode = (
-            AveragingMode.CYCLIC if self.hardware_average else AveragingMode.SINGLESHOT
-        )
-        acquisition_type = (
-            AcquisitionType.DISCRIMINATION
-            if self.classify
-            else AcquisitionType.INTEGRATION
-        )
-        return dict(
-            nshots=self.nshots,
-            relaxation_time=self.relaxation_time,
-            acquisition_type=acquisition_type,
-            averaging_mode=averaging_mode,
-        )
+    # # TODO: I think this is not used anywhere, so remove it.
+    # @property
+    # def execution_parameters(self):
+    #     """Default execution parameters."""
+    #     averaging_mode = (
+    #         AveragingMode.CYCLIC if self.hardware_average else AveragingMode.SINGLESHOT
+    #     )
+    #     acquisition_type = (
+    #         AcquisitionType.DISCRIMINATION
+    #         if self.classify
+    #         else AcquisitionType.INTEGRATION
+    #     )
+    #     return dict(
+    #         nshots=self.nshots,
+    #         relaxation_time=self.relaxation_time,
+    #         acquisition_type=acquisition_type,
+    #         averaging_mode=averaging_mode,
+    #     )
 
 
 class AbstractData:
--- a/src/qibocal/auto/operation.py
+++ b/src/qibocal/auto/operation.py
@@ -201,9 +201,10 @@ class Data(AbstractData):
     @property
     def qubits(self) -> list[QubitId]:
         """Access qubits from data structure."""
-        # TODO: what is this for? Seems like a hacky way to implement something. Also
-        # means we can't do multi qubit in a sensbile way since, but isntead rely on the
-        # `pairs` property below, since tuples are now used for whatever this is.
+        # TODO: I don't understand the rationale of this function. Perhaps it would be
+        # nice if we could do Data.qubits and get a list of either qubits or pairs
+        # instead of having a separate function for each case even though they serve the
+        # same purpose just for different routines.
         if set(map(type, self.data)) == {tuple}:
             return list({q[0] for q in self.data})
         return [q for q in self.data]
--- a/src/qibocal/auto/operation.py
+++ b/src/qibocal/auto/operation.py
@@ -201,10 +201,9 @@ class Data(AbstractData):
     @property
     def qubits(self) -> list[QubitId]:
         """Access qubits from data structure."""
-        # TODO: I don't understand the rationale of this function. Perhaps it would be
-        # nice if we could do Data.qubits and get a list of either qubits or pairs
-        # instead of having a separate function for each case even though they serve the
-        # same purpose just for different routines.
+        # TODO: In the two-qubit case, a set of the first elements of the tuples is
+        # returned. This behaviour is not reflected in the name of the property so may
+        # lead to confusion and should therefore be changed.
         if set(map(type, self.data)) == {tuple}:
             return list({q[0] for q in self.data})
         return [q for q in self.data]
--- a/src/qibocal/auto/operation.py
+++ b/src/qibocal/auto/operation.py
@@ -206,7 +206,7 @@ class Data(AbstractData):
     def qubits(self) -> list[QubitId]:
         """Access qubits from data structure."""
         if set(map(type, self.data)) == {tuple}:
-            return list({q[0] for q in self.data})
+            return list({tuple(sorted(q[:2])) for q in self.data})
         return [q for q in self.data]
 
     @property
--- a/src/qibocal/auto/operation.py
+++ b/src/qibocal/auto/operation.py
@@ -9,7 +9,7 @@ from typing import Callable, Generic, NewType, Optional, TypeVar, Union
 
 import numpy as np
 import numpy.typing as npt
-from qibolab import Platform, Qubit
+from qibolab import AcquisitionType, AveragingMode, Platform, Qubit
 
 from qibocal.config import log
 
@@ -73,11 +73,8 @@ class Parameters:
     """Wait time for the qubit to decohere back to the `gnd` state."""
     hardware_average: bool = False
     """By default hardware average will be performed."""
-
-    # # TODO: this is an inherent propery of the expirment so I don't see why is thould be
-    # # exposed as a parameter. Also thin it's not used anyway, so remove it.
-    # classify: bool = False
-    # """By default qubit state classification will not be performed."""
+    classify: bool = False
+    """By default qubit state classification will not be performed."""
 
     @classmethod
     def load(cls, input_parameters):
@@ -102,24 +99,23 @@ class Parameters:
             setattr(instantiated_class, parameter, value)
         return instantiated_class
 
-    # # TODO: I think this is not used anywhere, so remove it.
-    # @property
-    # def execution_parameters(self):
-    #     """Default execution parameters."""
-    #     averaging_mode = (
-    #         AveragingMode.CYCLIC if self.hardware_average else AveragingMode.SINGLESHOT
-    #     )
-    #     acquisition_type = (
-    #         AcquisitionType.DISCRIMINATION
-    #         if self.classify
-    #         else AcquisitionType.INTEGRATION
-    #     )
-    #     return dict(
-    #         nshots=self.nshots,
-    #         relaxation_time=self.relaxation_time,
-    #         acquisition_type=acquisition_type,
-    #         averaging_mode=averaging_mode,
-    #     )
+    @property
+    def execution_parameters(self):
+        """Default execution parameters."""
+        averaging_mode = (
+            AveragingMode.CYCLIC if self.hardware_average else AveragingMode.SINGLESHOT
+        )
+        acquisition_type = (
+            AcquisitionType.DISCRIMINATION
+            if self.classify
+            else AcquisitionType.INTEGRATION
+        )
+        return dict(
+            nshots=self.nshots,
+            relaxation_time=self.relaxation_time,
+            acquisition_type=acquisition_type,
+            averaging_mode=averaging_mode,
+        )
 
 
 class AbstractData:
--- a/src/qibocal/auto/transpile.py
+++ b/src/qibocal/auto/transpile.py
@@ -147,7 +147,7 @@ def execute_transpiled_circuits(
     The input `transpiler` is optional, but it should be provided if the backend
     is `qibolab`.
     For the qubit map look :func:`dummy_transpiler`.
-    This function returns the list of transpiled circuits and the execution results.
+    This function returns a list of the execution results.
     """
     transpiled_circuits = transpile_circuits(
         circuits,
@@ -155,7 +155,7 @@ def execute_transpiled_circuits(
         platform,
         transpiler,
     )
-    return transpiled_circuits, execute_circuits(
+    return execute_circuits(
         platform,
         compiler,
         transpiled_circuits,
--- a/src/qibocal/auto/transpile.py
+++ b/src/qibocal/auto/transpile.py
@@ -1,18 +1,14 @@
 from typing import Optional
 
 from qibo import Circuit, gates
-from qibo.backends import Backend
 from qibo.transpiler.pipeline import Passes
 from qibo.transpiler.unroller import NativeGates, Unroller
-from qibolab import PulseSequence
+from qibolab import AveragingMode, PulseSequence
 from qibolab._core.compilers import Compiler
 from qibolab._core.native import NativeContainer
-from qibo.result import MeasurementOutcomes
 
 from qibocal.auto.operation import QubitId
 from qibocal.config import raise_error
-from qibolab import AveragingMode
-import numpy as np
 
 REPLACEMENTS = {
     "RX": "GPI2",
@@ -83,11 +79,14 @@ def execute_circuit(circuit, platform, compiler, initial_state=None, nshots=1000
     sequence, _measurement_map = compiler.compile(circuit, platform)
 
     # TODO?: pass options dict
-    readout = platform.execute([sequence], nshots=nshots, averaging_mode=AveragingMode.SINGLESHOT)
+    readout = platform.execute(
+        [sequence], nshots=nshots, averaging_mode=AveragingMode.SINGLESHOT
+    )
 
     result = list(readout.values())[0]
 
     return result
+
 
 def execute_circuits(platform, compiler, circuits, initial_states=None, nshots=1000):
     """Executes multiple quantum circuits with a single communication with
@@ -118,23 +117,15 @@ def execute_circuits(platform, compiler, circuits, initial_states=None, nshots=1
         )
 
     # TODO: Maybe these loops can be parallelized
-    sequences, measurement_maps = zip(
+    sequences, _measurement_maps = zip(
         *(compiler.compile(circuit, platform) for circuit in circuits)
     )
 
     readout = platform.execute(sequences, nshots=nshots)
 
-    results = []
-    for circuit, measurement_map in zip(circuits, measurement_maps):
-        results.append(
-            MeasurementOutcomes(circuit.measurements, self, nshots=nshots)
-        )
-        for gate, sequence in measurement_map.items():
-            samples = [readout[acq.id] for _, acq in sequence.acquisitions]
-            gate.result.backend = self
-            gate.result.register_samples(np.array(samples).T)
-    return results
+    result = list(readout.values())
 
+    return result
 
 
 def execute_transpiled_circuits(
@@ -162,8 +153,12 @@ def execute_transpiled_circuits(
         platform,
         transpiler,
     )
-    return transpiled_circuits, execute_circuits(platform, compiler,
-        transpiled_circuits, initial_states=initial_states, nshots=nshots
+    return transpiled_circuits, execute_circuits(
+        platform,
+        compiler,
+        transpiled_circuits,
+        initial_states=initial_states,
+        nshots=nshots,
     )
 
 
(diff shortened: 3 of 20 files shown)


scarrazza (changes requested):
@RoyStegeman thanks a lot for this. My main concern here are tests, in particular I believe we are missing some qubit map error handling (invalid types, out-of-bounds), averaging mode edge cases, circuit size extremes (deep circuits, empty circuits), memory constraints and error recovery and state mapping and data aggregation logic. I have added some suggestions above.

What is the status of testing this on QPU?

In terms of general comments, we might take this PR to discuss about the following points:
- While the current refactoring is a great step forward, I thinkt that creating a high-level "Orchestrator" or "Experiment Runner" that wraps set_compiler, dummy_transpiler, transpile_circuits, and execute_circuits into a single call could further reduce boilerplate in protocols.
- As you noted in the code's TODOs, the compilation of circuits in execute_circuits could be parallelized to improve performance for experiments with many circuits. We should reference to some issue or open a new one.
- The logic for averaging_mode in execute_circuits could be further generalized to handle multi-state readout (beyond just "0" and "1") more robustly if needed in the future (and open a new issue).
- You are not touching the sphinx docs, it might be worth checking it to reflect the removal of the backend abstraction and the shift towards the new execution and data handling patterns.

RoyStegeman (commented):
Thanks for the review @scarrazza! I largely agreed with your comments and implemented them or left an explanation. 

I tested the protocols on QPU. Due to limitations of the available calibrations some I could only test if they ran while for others the results have been validated more rigorously. 

Regarding the general comments: 
- my TODO was indeed because I would probably wrap `transpile_circuits` and `execute_circuits` in a single function. Since we're not defining a state there no need for a class, unless that class also contains the platform so it would be the existing `Executor`. 
- The parallelization of that loop is a qibolab issue but either way not a bottleneck for qibocal. I removed the comment here. 
- multi-state readout is hardware dependent and even that is not so simple. E.g. the qblox firmware supports binary discrimination per sequencer so doing qutrits requires two sequencers and still a subsequent discrimination step in software. Averaging at the level of sequencers will not work so avaraging cannot be done in hardware. This doesn't mean that averaging in hardware will be entirely impossible since it may be possible to use a register in one of the modules to count states conditional on the output of multiple sequencers, but this is far beyond the scope of what we can do here.
- The docs are almost exclusively a list of protocols and don't describe the internal mechanisms, so these changes don't affect what is written in the docs.

src/qibocal/auto/transpile.py
```diff
@@ -19,9 +27,9 @@
 def transpile_circuits(
     circuits: list[Circuit],
     qubit_maps: list[list[QubitId]],
-    backend: Backend,
-    transpiler: Optional[Passes],
-):
+    platform: Platform,
+    transpiler: Passes,
+) -> list[Circuit]:
```
scarrazza: Could you plase update the transpile_circuits docstring to include Args: and Returns: sections, matching the new standard set for other functions in the file?

scarrazza (approved):

src/qibocal/auto/execute.py
```diff
@@ -277,6 +277,14 @@ def init(
...
             self.update = update
         if targets is not None:
+            if not isinstance(targets, list) or not all(
+                isinstance(t, (tuple, int, str)) for t in targets
+            ):
+                raise TypeError(
+                    "targets must be a list of qubit IDs (of type int/str) or qubit "
+                    f"pairs (of type tuple), got {type(targets).__name__} with value: "
+                    f"{targets}"
+                )
```
alecandido: Could split this as a `_validate_targets(targets)`?

It is quite self-documenting, but it makes it more modular.
And, for improved clarity, I would implement it with early returns
```suggestion
            err = TypeError(
                "targets must be a list of qubit IDs (of type int/str) or qubit "
                f"pairs (of type tuple), got {type(targets).__name__} with value: "
                f"{targets}"
            )
            if not isinstance(targets, list):
                raise err
            if not all(isinstance(t, (tuple, int, str)) for t in targets):
                raise err
```
which would be even better with specialized (and more explicit) errors for the two distinct cases
- either it is not a list, which is one kind of problem
- or it does not contain only valid elements, which is a different one

Eventually, I hope we can pass this sort of validation to Pydantic, and just express it through type hints.

src/qibocal/auto/operation.py
```diff
@@ -201,14 +201,18 @@ class Data(AbstractData):
     @property
     def qubits(self) -> list[QubitId]:
         """Access qubits from data structure."""
+        # TODO: I don't understand the rationale of this function. Perhaps it would be
+        # nice if we could do Data.qubits and get a list of either qubits or pairs
+        # instead of having a separate function for each case even though they serve the
+        # same purpose just for different routines.
```
alecandido: To give an answer (which I would rather ignore in this PR), I would attribute this to legacy behavior.

The initial design for Qibocal was not accounting for pairs, but only single qubit protocols. Then, pairs have been added.
When it has been realized that handling them separately was just confusing and redundant, the *targets* concept has been developed, to have a more agnostic list of target object. Delegating the validation of the target structure to the individual protocols.

Despite having developed a consistent alternative, the package has never been entirely reviewed to ensure consistency.

src/qibocal/auto/operation.py
```diff
@@ -201,14 +201,18 @@ class Data(AbstractData):
...
         if set(map(type, self.data)) == {tuple}:
             return list({q[0] for q in self.data})
         return [q for q in self.data]
 
     @property
     def pairs(self):
-        """Access qubit pairs ordered alphanumerically from data structure."""
-        return list({tuple(sorted(q[:2])) for q in self.data})
+        """Access qubit pairs from data structure."""
+        return list({tuple(q[:2]) for q in self.data})
```
alecandido, line 214: Again an explanation, not (much of) a request.

The reason why they were sorted was to ensure that you could calibrate the same CZ irrespective of the pair sorting. I.e. `CZ(0, 2) == CZ(2, 0)`, with all the related protocols.

However, not all gates (and even less protocols) are symmetric. Not because they are *manifestly symmetric* (same operations on the qubits), nor because there is an intrinsic sorting (high vs low frequency qubit).
E.g.
- there are two distinct `CNOT` gates possible for each pair
- the flux crosstalk does not have to be symmetric - also because the objects involved are distinct, i.e. flux lines and SQUIDs (which, despite the natural bijective association, are actually connected in an all-to-all topology, if you do not put any threshold on the amount of field)

In conclusion, it is definitely better not to sort pairs.
But, we have to take into account this may have been assumed by some protocols, and possibly propagate this change.
While not advocating for an immediate extensive review, we should append this to the list to keep in mind while reviewing protocols.

src/qibocal/auto/transpile.py
```diff
@@ -1,10 +1,18 @@
-from typing import Optional
+# This file contains functions to transpile and execute quantum circuits.
+#
+# TODO: Since these functions are always used in the same way, we should probably
+# provide a single function that takes care of setting the compiler, transpiler, doing
+# the transpilation and execution in a single call. This would mean that set_compiler
+# and dummy_transpiler are called for every circuit execution instead of just once per
+# protocol, so I'm not convinced that's what should be done.
```
alecandido: Well, since you are always account for a batch of circuits (both `transpile_circuits()` and `execute_circuits()` have `circuits: list[Circuit]` as a parameter), I do not see why you would need to call multiple times per protocol.

If they are always used together, they could well be piped into one single call.

src/qibocal/auto/transpile.py
```diff
@@ -35,83 +43,187 @@ def transpile_circuits(
...
+
+    Args:
+        circuits: List of quantum circuits to transpile and pad.
+        qubit_maps: List of qubit maps, one per circuit. Each qubit map maps physical
+            qubit IDs to logical qubit indices.
+        platform: The quantum platform to transpile circuits for.
+        transpiler: The transpiler to apply to the circuits.
+
+    Returns:
+        List of transpiled and padded Circuit instances, one per input circuit.
```
alecandido: In general, there is no strong need to document each and every argument of a function.

The style I recommend for docstrings is to just write them in plain text. And avoid any structured section.
If there is any argument which needs a deeper explanation, it can be mentioned explicitly in the text.

I know NumPy has a nice rendering for individual arguments, but they are also making that effort only for specific public functions. On top of which, each public function is receiving its own entire page in the documentation.

This instead is an internal function. And, if rendered, it could be part of a quite infinite page (also because of poor choices about hierarchical structure), which makes it rather useless, cf. [Qibo API reference](https://qibo.science/qibo/stable/api-reference/qibo.html) (it can be done way better, but Qibo is not the only package to suffer this).

Instead, I'm suggesting to adopt the same style used in Python standard library docs, e.g. [`shutil.copytree`](https://docs.python.org/3/library/shutil.html#shutil.copytree) for a single function, and the whole module page as an example of rendering.

A line like
```py
        transpiler: The transpiler to apply to the circuits.
```
is just quite redundant, and adding no information: we know it is a transpiler from the parameter name, and we know it will be applied to circuits from the function name.

Let's try to dilute a bit less our code :)

src/qibocal/auto/transpile.py
```diff
@@ -148,42 +260,65 @@ def rule(gate: gates.Gate, natives: NativeContainer) -> PulseSequence:
     return rule
 
 
-def set_compiler(backend, natives_):
-    """
-    Set the compiler to execute the native gates defined by the platform.
+def set_compiler(platform: Platform) -> Compiler:
```
alecandido: Well, maybe this function should not be called `set_compiler()` any longer, since it is now just returning it.

src/qibocal/auto/transpile.py
```diff
@@ -35,83 +43,187 @@ def transpile_circuits(
...
-        for i, qubit_map in enumerate(qubit_maps):
-            qubit_map = map(lambda x: qubits.index(x), qubit_map)
-            qubit_maps[i] = list(qubit_map)
+    # QubitId can be integers or strings. pad_circuit only works with integer qubit IDs,
+    # so if the qubit maps contain string IDs, we convert them to integer indices based
+    # on the platform's qubit order.
+    _qubit_maps: list[list[int]] = [
+        [q if isinstance(q, int) else qubits.index(q) for q in qubit_map]
+        for qubit_map in qubit_maps
+    ]
```
alecandido: You could split this as its own (internal) function, adding the comment as documentation.

Even the type hint is already nice as return type ;)

src/qibocal/auto/transpile.py
```diff
@@ -35,83 +43,187 @@ def transpile_circuits(
...
+    """
+    if len(gate.qubits) != 1:
+        raise ValueError(
+            f"Measurement gate must measure a single qubit. "
+            f"Got gate with {len(gate.qubits)} qubits."
+        )
+    if len(sequence.acquisitions) != 1:
+        raise ValueError(
+            f"Measurement sequence must have exactly one acquisition. "
+            f"Got {len(sequence.acquisitions)} acquisitions."
```
alecandido, line 105:
```suggestion
    if len(gate.qubits) != 1:
        raise ValueError(
            "Measurement gate must measure a single qubit. "
            f"Got gate with {len(gate.qubits)} qubits."
        )
    if len(sequence.acquisitions) != 1:
        raise ValueError(
            "Measurement sequence must have exactly one acquisition. "
            f"Got {len(sequence.acquisitions)} acquisitions."
```

Weird that it did not get fixed by `pre-commit`.

src/qibocal/auto/transpile.py
```diff
@@ -35,83 +43,187 @@ def transpile_circuits(
...
+        )
+    if gate.qubits[0] not in qubit_map:
+        raise KeyError(f"Qubit {gate.qubits[0]} not found in qubit map: {qubit_map}.")
+    if sequence.acquisitions[0][1].id not in readout:
+        raise KeyError(
+            f"Acquisition ID {sequence.acquisitions[0][1].id} not found in readout results."
+        )
```
alecandido: Half of the function is about `gate` and `qubit_map` (also used together), and the other half about `sequence` and `readout` (also used together). But there is no crosstalk between the two groups.

It would make sense to split in two functions with two arguments each.

In a sense, the principle is just *modularity*, declined as "the fewer variables in a scope the better".

src/qibocal/auto/transpile.py
```diff
@@ -35,83 +43,187 @@ def transpile_circuits(
...
+        "Number of circuits and qubit maps must match."
     )
-    return transpiled_circuits, backend.execute_circuits(
-        transpiled_circuits, initial_states=initial_states, nshots=nshots
+
+    sequences, measurement_maps = zip(
+        *(compiler.compile(circuit, platform) for circuit in circuits)
     )
 
+    # TODO?: pass options dict
```
alecandido: As an optional argument, it could be fine. Since this is going to be used by every protocol, it may be useful to delegate to them the choice of options - especially considering the platform updates.
Though, the platform updates themselves could also be used to modify the object which is passed as the parameter. Which would mostly leave the acquisition type as spare option.

In case of doubts, we could still do it. It's quite cheap anyhow.

src/qibocal/auto/transpile.py
```diff
@@ -35,83 +43,187 @@ def transpile_circuits(
...
+                    "states in a multi-qubit setup."
+                )
+            if len(measurement_map) != 1:
+                raise ValueError(
+                    "Averaging mode CYCLIC requires exactly one measurement acquisition "
+                    "per circuit."
+                )
+            phys_to_logic_mapping = {q: i for i, q in enumerate(qubit_map)}
+            [(gate, sequence)] = measurement_map.items()
+            _validate_measurement(gate, sequence, phys_to_logic_mapping, readout)
```
alecandido:
```suggestion
            [(gate, sequence)] = measurement_map.items()
            _validate_measurement(gate, sequence, qubit_map, readout)
```
Inlining the enumeration is not adding significant information. Rather, I'd design the check to just take into account the keys, and find the index if needed.

Actually, if I'm not wrong, it is already using only the dictionary keys. Any reason for the enumeration?

src/qibocal/auto/transpile.py
```diff
@@ -35,83 +43,187 @@ def transpile_circuits(
...
+                    "Use SINGLESHOT instead. The reason is that the excited state probability "
+                    "individual qubits (which is what CYCLIC extracts) does not provide full "
+                    "information about the probability distribution of the full set of basis "
+                    "states in a multi-qubit setup."
+                )
+            if len(measurement_map) != 1:
+                raise ValueError(
+                    "Averaging mode CYCLIC requires exactly one measurement acquisition "
+                    "per circuit."
+                )
```
alecandido: I would not mention the `SINGLESHOT` and `CYCLIC` modes in exceptions raised by Qibocal.
Since Qibolab is only a transitive dependency, from the point of view of Qibocal's user it is an implementation detail. While I'm quite in favor of leaving exceptions bubble from underlying levels, such that they are possibly handled only when a strategy is actually available[*], if the error is raised in a certain place, it should refer its inputs, and not some other derived value (which would result quite obscure even to its direct caller).

[*]: which could even be just "avoid crashing, close everything you have still open, give a generic error message, and log the full traceback on temp file". But it should be implemented only by the user facing layer, such that it is also only implemented once, and not by all components in the stack

src/qibocal/auto/transpile.py
```diff
@@ -35,83 +43,187 @@ def transpile_circuits(
...
-    is `qibolab`.
-    For the qubit map look :func:`dummy_transpiler`.
-    This function returns the transpiled circuit and the execution results.
-    """
+    countslist = []
+    if averaging_mode.average:
+        # NOTE: averaging mode only makes sense for a two state readout. If there are
+        # more states it would have to be conditional since the excited state
+        # probability of individual qubits does not provide full information about the
+        # probability distribution of the full set of basis states.
```
alecandido: When you discuss the "multiple states" I immediately think to qutrits. But here you mean multiple qubits (which is explained later). Maybe it could be rephrased to avoid confusion (the note itself is fine to have)

src/qibocal/auto/transpile.py
```diff
@@ -35,83 +43,187 @@ def transpile_circuits(
...
+            for gate, sequence in measurement_map.items():
+                _validate_measurement(gate, sequence, phys_to_logic_mapping, readout)
+                logical_qubit = phys_to_logic_mapping[gate.qubits[0]]
+                result[logical_qubit] = readout[sequence.acquisitions[0][1].id]
+            # The inverse sorting is to have little-endian bitstring notation, which
+            # means that the qubit with the smallest qubitId is the most significant bit
+            # in the output string (on the right).
+            invsorted_result = sorted(result)[::-1]
+            arr = np.stack([result[q] for q in invsorted_result]).astype(int)
+            countslist.append(Counter("".join(map(str, col)) for col in arr.T))
```
alecandido: Both `if` and `else` branches could be standalone functions.

alecandido: I wonder whether we should just use tuples/arrays instead of strings. 

Just hyperbolically, if we had 11 states per local space, you should start using the hex representation to write. And that also has a limit.

Why not using directly a sequence of integers, instead of characters?

alecandido (commented):
I finally completed the review of the `auto` part[*].
I will proceed with the protocols, but I expect them to require less effort, since they should mostly reflect the reviewed changes, and propagate them.

[*]: btw, maybe we should rename it `exec`, `infra`, `orchestration`, `coordination`. I don't have any great name for it, which is the reason why now it has a terrible one (when I created it, 3y ago, I was working under the request of "something to automate"), since it is mostly providing the layout for the protocols' definition, and their execution

src/qibocal/protocols/randomized_benchmarking/standard_rb.py
```diff
@@ -118,7 +112,14 @@ def _plot(
     fig = go.Figure()
     fitting_report = ""
     x = data.depths
-    raw_data = data.extract_probabilities(qubit)
+    single_qubit = isinstance(target, int)
+    raw_data = np.array(
+        [
+            val["survival_probs"]
+            for key, val in data.data.items()
+            if (key[0] if single_qubit == 1 else key[:2]) == target
```
alecandido: What is this?

And: is `single_qubit` semantically a boolean flag?

src/qibocal/protocols/randomized_benchmarking/standard_rb_sweeper.py
```diff
@@ -23,14 +23,22 @@
...
 
+try:
+    from qibocal.protocols.randomized_benchmarking.utils import (
+        RB_Generator,
+        setup,
+    )
+except ImportError:
+    "ImportError: a significant rewrite of the treatment of circuits means that this "
+    "protocol is no longer working. See https://github.com/qiboteam/qibocal/pull/1393.\n\n"
+    "The ImportError is caught to avoid breaking the entire package."
```
alecandido, line 40: Maybe we should mention Paul in here, to pre-announce he will to do something

src/qibocal/protocols/utils.py
```diff
@@ -234,6 +236,26 @@ def compute_qnd(
     return qnd, lambda_m.tolist(), lambda_m2.tolist()
 
 
+def marginalize_qubit_counts(counts: Counter, indices: Sequence[int] | int):
```
alecandido: `Counter` admits subscription, as `dict` (e.g. I expect `Counter[str, int]` to be the suitable type hint here).

src/qibocal/auto/operation.py
RoyStegeman: Yes but the main thing that confuses me is that in case it's a tuple of two qubits, what is returned is the set (turned into a list) of only the first qubit in the pairs

src/qibocal/protocols/utils.py
```diff
@@ -234,6 +236,26 @@ def compute_qnd(
...
+        indices: Qubit positions to marginalize over.
+
+    Returns:
+        Counter of the marginal distribution.
+    """
+    out = Counter()
+    indices_list = [indices] if isinstance(indices, int) else indices
+    for state, count in counts.items():
+        # -1-i because logical qubit 0 is on the right. See execute_circuits.
+        reduced = "".join(state[-1 - i] for i in indices_list)
```
alecandido: While this could even be more efficient, I would say it is quite negligible in this context, and I'd rather use `reversed()` on `state` once, to make it a bit more intuitive.

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -30,6 +31,32 @@
...
+    circuit: Circuit
+    index: CircuitIndex
+
+
+@dataclass
+class IndexedResult:
+    """An execution result paired with its (qubit, depth, iteration) CircuitIndex."""
+
+    result: Counter
+    index: CircuitIndex
```
alecandido: For new objects, I would recommend using immediately Pydantic models.

We mostly need Pydantic to support serialization and deserialization, which are not always strictly required. However, there still two considerations that I would consider relevant to always use it:
1. simplicity: since Pydantic models are effectively a replacement for dataclasses, once we pay the price of having it as a dependency, better to stick to a single solution
2. "you never know", in the sense that things that are not required to be serialized today, they may turn useful tomorrow - possibly even indirectly, as parts of other objects (and, to serialize with Pydantic, it has to be either built-in, or Pydantic itself)

src/qibocal/auto/transpile.py
RoyStegeman: I mean either, unless the current qblox driver supports hardware averaging over qutrit states.

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -84,7 +111,7 @@
 
 RBType = np.dtype(
     [
-        ("samples", np.int32),
+        ("survival_probs", np.float64),
     ]
 )
 """Custom dtype for RB."""
```
alecandido, line 124: If our `dtype` is not composite (just a single float), do we actually need to still make it composite?

src/qibocal/auto/operation.py
alecandido: Ok, this is not what I read in this `TODO` comment. Where is this happening? Or did you already fix it?

src/qibocal/auto/transpile.py
alecandido: Even in that case, it would make no sense. The only reason why averaging shots makes sense is that, since you know the full amount you performed, the average is a counting itself.

But then, the only thing it is checked in the following is the number of qubits. So, the comment is more generic than the check (not really a big deal)

alecandido (commented):
Sorry, I'm not done yet with all the comments, but it does not allow me to reply from the "Conversation" tab, until I release the other pending comments...

(I should have done it from the "Files changed" tab, at least I could have just posted the individual replies...)

sorewachigauyo:
Tested the `standard_rb`, `standard_rb_2q` and `standard_rb_2q_inter` on QPU

[sq-rb.tar.gz](https://github.com/user-attachments/files/26627751/sq-rb.tar.gz)
[2q-rb.tar.gz](https://github.com/user-attachments/files/26627739/2q-rb.tar.gz)
[2q-irb.tar.gz](https://github.com/user-attachments/files/26627747/2q-irb.tar.gz)

src/qibocal/auto/operation.py
RoyStegeman: It is called from `optimize.py`, `standard_rb_2q_inter` and `snz_optimize_t_idle_vs_t_tot`, and something that is simply looped over. `optimize` is not so clear to me what it is, while since `snz` is symmetric, there indeed forcing a sorting makes sense, although for the 2q RB I don't think it makes sense in the general case because you may want to use it for asymmetric gates. 

Either way, this should only break behavior up to the point that without sorting users have to ensure to pass the qubitids in the pair in the order that they want to store in the platform.

src/qibocal/auto/operation.py
RoyStegeman: It was indeed not so clear from the TODO. I updated the comment

RoyStegeman:
That's a very useful check. Thanks @sorewachigauyo!

src/qibocal/auto/transpile.py
RoyStegeman: I opened an issue that is a bit more general, but also includes this point: https://github.com/qiboteam/qibocal/issues/1437

src/qibocal/auto/transpile.py
RoyStegeman: Good point, I removed explicit references to CYCLIC and SINGLESHOT

src/qibocal/auto/transpile.py
RoyStegeman: The main motivation was to make clear the purpose of `qubit_maps` by giving the name `phys_to_logic_mapping`, but indeed it's also possible and clear to take the index and call that `logical_qubit`.

src/qibocal/auto/transpile.py
RoyStegeman: > Just hyperbolically, if we had 11 states per local space, you should start using the hex representation to write. And that also has a limit.

That's far out of scope for qibocal, so I don't believe this is something we have to consider. 

> Why not using directly a sequence of integers, instead of characters?

I don't have a strong opinion, but I think strings are a bit more readable and nicer to print for a low number of qubits which is what we mainly work with. Other than the limit of 10 states (which is mainly a theoretical limitation), I don't think there is a much a benefit to doing a sequence?

src/qibocal/auto/transpile.py
RoyStegeman: Good point, I renamed it `build_native_gate_compiler`

src/qibocal/protocols/utils.py
RoyStegeman: The `Counter` bitstring is big-endian while `indices` (renamed `qubit_id`) are e.g. 0 for qubit 0, so the rightmost qubit in the bitstring. I think it's better for understanding if the code reflects this more directly, but I also added a more detailed comment.

src/qibocal/protocols/randomized_benchmarking/utils.py
RoyStegeman: Without reworking also `register_qubit`, yes we need to keep it composite.

src/qibocal/auto/transpile.py
alecandido: > Other than the limit of 10 states (which is mainly a theoretical limitation), I don't think there is a much a benefit to doing a sequence?

Reduce the amount of code needed for the conversion?

src/qibocal/protocols/randomized_benchmarking/standard_rb.py
RoyStegeman: The key was either `(qubit, depth)` in the single qubit case, or `(qubit0, qubit1, depth)` in the 2q case. I slightly rewrote it such that the key is now `(qubit, depth)` or `((qubit0, qubit1), depth)`, so always the first element of the key can be taken and no condition is needed.

src/qibocal/protocols/randomized_benchmarking/standard_rb_sweeper.py
RoyStegeman: @sorewachigauyo

alecandido (approved):

PR #1402, opened 2026-03-17
Plugins

### LINKED ISSUE #508: Plugin API

alecandido:
Closes #508

Prototype:
```py
from typing import Any, Callable, Optional

from pydantic import BaseModel, ConfigDict


class Model(BaseModel):
    model_config = ConfigDict(frozen=True, arbitrary_types_allowed=True)


# TODO: dummy replacements
Completed = int
Routine = Callable[[], int]


class CalibrationProtocol:
    """Protocol to calibrate a chip."""

    def __call__(
        self,
        *args: Any,
        parameters: Optional[dict] = None,
        update: bool = True,
        **kwargs: Any,
    ) -> Completed:
        """Invoke calibration experiment.

        The signature depend on the specific protocol. For the core set, the documentation
        can be found at <url>. Otherwise compare the documentation for the Qibocal extension
        providing the protocol invoked.
        """
        _ = self, parameters, update, args, kwargs
        return 42


placeholder = CalibrationProtocol()


class BaseSet:
    a: CalibrationProtocol = placeholder
    b: CalibrationProtocol = placeholder
    c: CalibrationProtocol = placeholder


def a(**parameters: Any) -> int:
    print(parameters)
    return len(parameters)


def b(**parameters: Any) -> int:
    return len(parameters) * 10


def c(**parameters: Any) -> int:
    return len(parameters) * -7


PROTOCOLS: dict[str, Routine] = {"a": a, "b": b, "c": c}


class Executor(Model, BaseSet):
    def model_post_init(self, context: Any) -> None:
        _ = context
        for name, protocol in PROTOCOLS.items():
            object.__setattr__(self, name, self.wrapped_protocol(protocol))

    def wrapped_protocol(self, protocol: Routine):
        def wrapper(
            *args: Any,
            parameters: Optional[dict] = None,
            update: bool = True,
            **kwargs: Any,
        ) -> Completed:
            if parameters is None:
                parameters = {"args": args} | kwargs
            return self.run_protocol(protocol, parameters=parameters, update=update)

        return wrapper

    def run_protocol(
        self, protocol: Routine, parameters: dict, update: bool
    ) -> Completed:
        _ = update
        return protocol(**parameters)


if __name__ == "__main__":
    x = Executor()
    print(x.a(a=3, b=7, c=12341))
```

- [x] check and update docs, especially for
  - the deprecated `Executor.name`
  - the shift of arguments (mostly internal, but affects `Executor.init()`)
- [x] add a page about how to extend Qibocal with a plugin
- [x] check consistency of `Executor`'s usage
  - it is passing tests, but it may confuse the former `name` for a `path`, or things like that


scarrazza (approved):

src/qibocal/auto/output.py
```diff
@@ -42,7 +42,6 @@ def tot(self) -> float:
 class Metadata:
     """Execution metadata."""
 
-    title: str
```
RoyStegeman, line 45: `title` is still being called in two places.

https://github.com/qiboteam/qibocal/blob/4c9d352d7fe75a096e73b83e4478244e1b475d9b/src/qibocal/cli/compare.py#L27

https://github.com/qiboteam/qibocal/blob/4c9d352d7fe75a096e73b83e4478244e1b475d9b/src/qibocal/cli/compare.py#L57

src/qibocal/auto/execute.py
```diff
@@ -102,31 +99,23 @@ class Executor:
...
-    def __post_init__(self):
+    def model_post_init(self, context: Any) -> None:
         """Register as a module, if a name is specified."""
+        # explicitly unused
+        _ = context
+
         if self.name is not None:
             _register(self.name, self)
+        if len(self.targets) == 0:
+            object.__setattr__(self, "targets", list(self.platform.qubits))
```
RoyStegeman: Should this be the default behaviour? What if the protocol to be executed is a two-qubit gate?

src/qibocal/auto/execute.py
```diff
@@ -72,22 +63,28 @@ def _register(name: str, obj):
         setattr(parent_module, child_name, obj)
 
     sys.modules[qualified] = obj
-    obj.name = obj.__name__ = qualified
+    obj.__name__ = qualified
```
RoyStegeman: This ensured that `Executor.name` and `Executor.__name__` were the same. Why is this no longer needed?

src/qibocal/auto/execute.py
alecandido: Not affected by this PR. It was already the default before.
https://github.com/qiboteam/qibocal/blob/883e5c5d22763ccbd73dbf8b028dfd267fb0d427/src/qibocal/auto/execute.py#L127

In any case, the problem can not be solved here, unless we decide to remove the default completely. I.e. we make it mandatory to specify the targets for each protocol, which is an option (though I would do in a different PR anyhow).

A way to solve the problem would be to specify a set of qubits as default, and then pass to the 2q protocols the pairs in the platforms' connectivity whose qubits are both part of the default targets.

However, this seems exceedingly convoluted, so I would do one of two things:
- remove the protocols' default in full (i.e. the `Executor.targets` attribute, not only its default value), which is possibly the simplest and clearer option
- or we start labeling 2q protocols consistently (if we are not already doing), and raise a more explicit error when the protocol chosen is a 2q, and no target is specified - which is what is being done now, with a more messy error, and also means just removing the default, but only for a subset of protocols

src/qibocal/auto/execute.py
alecandido: `Executor.name` is deprecated, since it was there to allow the registration as a Python module - which we do not want to support any longer.

This function still allows for the same trick for arbitrary objects, not relying on the object to have a `.name` attribute. I'm still thinking whether to drop it, or to expose publicly - to make it still possible, but with an explicit action by the user.

In any case, most likely I will just scrap everything as soon as I'll start implementing #1456

But I will update the docstrings.

src/qibocal/auto/output.py
alecandido: Done in 75d37ee

The title usage has been simply suppressed. However, the whole `qq compare` is deprecated itself, since it is not been used, and heavily relies on HTML manipulation, which is easily broken. (And no one is using it)
I would just drop it, but not in this PR.

If you need to compare two reports side-by-side, just split the window in two tabs, or use two windows.

src/qibocal/auto/execute.py
RoyStegeman: I think removing the default should be fine. In many (probably even most) cases the targets will be a subset of all qubits so it has to be passed manually anyway, and I don't think this makes the interface unfriendly to users.

RoyStegeman (approved):
I tested by running an experiment.

src/qibocal/auto/execute.py
RoyStegeman: I tested and I believe it can be removed safely. Also the default value for `targets` in the initialization of `Executor` should be removed such that there is a more sensible error sooner.

src/qibocal/auto/execute.py
alecandido: Done in e64004a

src/qibocal/auto/execute.py
alecandido: Done in d816f43

PR #1131, opened 2025-04-04
Cross resonance protocol for 0.2 version

andrea-pasquale:
In this PR I'm planning to convert the protocols for cross-resonance to the 0.2 version.
Right now this PR includes only one routine where I sweep the duration of the cross-resonance pulse `cross_resonance_length`. 
I was able to verify the implementation of the protocol using https://github.com/qiboteam/qibolab/pull/1186 with the platform `qutrits` which I have uploaded in this PR (which is based on some parameters from IBM chips).
Here is a report with the protocol: http://login.qrccluster.com:9000/V_RjRIyXT66pwqcRvCdukQ==/

PR #1414, opened 2026-03-31
Cross Resonance calibration

### LINKED ISSUE #821: Cross Resonance Routine
To implement a basic routine for the calibration of cross-resonance (CR) gates.

This involves sending a pulse on the control qubit with the frequency of the target qubit. We vary the pulse length to observe a Rabi oscillation. Based on the state of the control qubit, there will be a splitting of the Rabi frequencies, and we can select a pulse length $t$ such that the phase difference between the two oscillations is $\pi$. This then implements a CNOT gate.

image

A basic CR calibration is as follows:
```python

sweep = np.arange(0, 1000, 20)
res1 = np.zeros(len(sweep))
res2 = np.zeros(len(sweep))

crtl_pi_pulse = platform.create_RX_pulse(qubit=CRTL_QUBIT, start=5)
tgt_ro_pulse = platform.create_qubit_readout_pulse(qubit=TGT_QUBIT, start=crtl_pi_pulse.finish)
cr_pulse = platform.create_RX_pulse(qubit=TGT_QUBIT, start=crtl_pi_pulse.finish)
cr_pulse.channel = crtl_pi_pulse.channel
cr_pulse.amplitude = 1

# We first perform a CR sweep when the control qubit is at the |0> state.
# This is equivalent to a typical Rabi length experiment
ps = PulseSequence(*[cr_pulse, tgt_ro_pulse])
for idx, t in enumerate(sweep):
    cr_pulse.duration = t
    tgt_ro_pulse.start = cr_pulse.finish
    res1[idx] = platform.execute_pulse_sequence(ps, opts)[tgt_ro_pulse.serial].magnitude

# Next, we add a pi-pulse on the control qubit at the start to get results for the |1> state.
ps.add(crtl_pi_pulse)
for idx, t in enumerate(sweep):
    cr_pulse.duration = t
    tgt_ro_pulse.start = cr_pulse.finish
    res2[idx] = platform.execute_pulse_sequence(ps, opts)[tgt_ro_pulse.serial].magnitude
```

The frequency of each Rabi oscillation is proportional to the amplitude of `cr_pulse`, so to obtain an optimal parameter for the amplitude, a 2D chevron of CR pulse amplitude and length can be performed.
``` python
sweeper_amplitude = Sweeper(
    Parameter.amplitude,
    amplitude_range,
    pulses=[cr_pulse],
    type=SweeperType.FACTOR,
)
sweeper_duration = Sweeper(
    Parameter.duration,
    duration_range,
    pulses=[cr_pulse],
    type=SweeperType.ABSOLUTE,
)
results = platform.sweep(
    ps,
    opts,
    sweeper_duration,
    sweeper_amplitude,
)
# and repeat for the other state of the control qubit
```

By also simultaneously driving the target qubit on resonance, an approximate interaction forms
$A_{CR} (\omega_{ZX} ZX + \omega_{IX} IX) + A_t \omega_{R2} IX $
where $A_{CR}$ is the amplitude of `cr_pulse`, $\omega_{ZX}, \omega_{IX}$ is determined by the target qubit Rabi frequency $\omega_{R2}$, effective qubit-qubit coupling, detunings and anharmonicites and $A_t$ is the amplitude of the second drive. 

A basic experiment with the second drive is as follows:

```python
crtl_pi_pulse = platform.create_RX_pulse(qubit=CRTL_QUBIT, start=5)
tgt_ro_pulse = platform.create_qubit_readout_pulse(qubit=TGT_QUBIT, start=crtl_pi_pulse.finish)
cr_pulse = platform.create_RX_pulse(qubit=TGT_QUBIT, start=crtl_pi_pulse.finish)
#cr_pulse.relative_phase = np.pi
IX_pulse = platform.create_RX_pulse(qubit=TGT_QUBIT, start=crtl_pi_pulse.finish)
IX_pulse.amplitude = 0.4
cr_pulse.channel = crtl_pi_pulse.channel
cr_pulse.amplitude = 1

ps = PulseSequence(*[cr_pulse, tgt_ro_pulse, IX_pulse])
sweeper_duration = Sweeper(
    Parameter.duration,
    duration_range,
    pulses=[cr_pulse, IX_pulse],
    type=SweeperType.ABSOLUTE,
)
results_crtl_gnd = platform.sweep(
    ps,
    opts,
    sweeper_duration
)
ps.add(crtl_pi_pulse)
results_crtl_exc = platform.sweep(
    ps,
    opts,
    sweeper_duration
)
```

lballerio:
This PR contains the whole calibration of the Cross Resonance gate as described by the [CR IBM paper](https://arxiv.org/abs/1603.04821).

The main method exploited is the so called `Hamltonian Tomography`, defined as the process of inferring a quantum system’s Hamiltonian from observed dynamics. The algorithm is quite simple: after preparing known states, we let them evolve, then measure outcomes, and fit a model to estimate the terms in the theoretical Hamiltonian model.
This PR gives the freedom of the user to calibrate the CR in different levels of accuracy:

- Simple CR calibration: by simply sweeping over time (or control qubit amplitude) we determine the pulse parameters; the sequence is composed by simply one pulse applied on the control qubit with the frequency of the target one.

- Echo CR sequence calibration: simply in this case the pulse sequence is more convoluted, since it consists on two CR pulses out-of-phase interleaved by two $\pi$-pulses on the control qubit; this pulse sequence improves the performances of the whole gate since it refocuses specific sources of noise.

- CR with Cancellation calibration: we add to the echo sequence two out-of phase pulses to the target qubit which are synched to the CR pulses; this calibration refocuses the remaining sources of noise, but additional parameters (i.e. cancellation and CR pulses phases and cancellation amplitude) have to be calibrated.

Here is a scheme of the different calibrations:

- the simple CR calibration is simply the first green pulse
- the echo calibration is the sum of the green and the blue pulses
- the complete calibration is the whole picture

Screenshot From 2026-03-31 17-11-40

this is a followup of PR #1131.
In addition, this PR requires a new emulator version from [Qibolab #1416](https://github.com/qiboteam/qibolab/pull/1416), [Qibolab #1412](https://github.com/qiboteam/qibolab/pull/1412) and [Qibolab #1401](https://github.com/qiboteam/qibolab/pull/1401).

Some suggestions for eventual improvements:
- https://arxiv.org/abs/2303.01427
- https://iopscience.iop.org/article/10.1088/2058-9565/ad895b


src/qibocal/protocols/utils.py
```diff
@@ -1043,3 +1025,109 @@ def baseline_als(data: NDArray, lamda: float, p: float, niter: int = 10) -> NDAr
...
+
+
+def guess_period(x, y):
+    """Return fft period estimation given a sinusoidal plot."""
+    fft = np.fft.rfft(y)
+    fft_freqs = np.fft.rfftfreq(len(y), d=(x[1] - x[0]))
+    mags = np.abs(fft)
+    mags[0] = 0
+    return 1 / fft_freqs[np.argmax(mags)]
+
```
RoyStegeman: I think this should simply invert `guess_frequency_numpyfied` to avoid the duplication of code.

src/qibocal/protocols/utils.py
```diff
@@ -1043,3 +1025,109 @@ def baseline_als(data: NDArray, lamda: float, p: float, niter: int = 10) -> NDAr
...
+
+def guess_frequency_numpyfied(x: np.ndarray, y: np.ndarray, axis: int = -1):
+    """Numpyfied version of :func:`guess_period` but here we work on frequencies."""
```
RoyStegeman: Maybe just call it `guess_frequency` and extend a bit the docstrings to explain in more detail the working of the axis. Using 'Numpyfied' to mean it's also working for an additional dimension is not so obvious to me.

src/qibocal/update.py
```diff
@@ -188,6 +188,12 @@ def iSWAP_amplitude(amp: float, platform: Platform, pair: QubitPairId):
     )
 
 
+def cnot_sequence(cr_sequence: PulseSequence, platform: Platform, pair: QubitPairId):
+    platform.update(
+        {f"native_gates.two_qubit.{f'{pair[0]}-{pair[1]}'}.CNOT": cr_sequence}
+    )
```
alecandido: For consistency, you may use `TypeAdapter(QubitPairId).dump_json(pair)`, with `TypeAdapter` imported from Pydantic.

In this way, you do not have to duplicate the convention here for how to extract a pair, and you are automatically consistent.
https://pydantic.dev/docs/validation/latest/concepts/types/#adding-validation-and-serialization

src/qibocal/result.py
```diff
@@ -78,6 +78,7 @@ def phase(iq: npt.NDArray):
     return np.unwrap(np.arctan2(iq_[0], iq_[1]))
 
 
-def probability(values: npt.NDArray, state: int = 0):
+def probability(values: npt.NDArray, state: int = 0) -> npt.NDArray:
     """Return the statistical frequency of the specified state."""
-    return np.sum(values == state, axis=0) / len(values)
+    means = np.sum(values == state, axis=0) / len(values)
```
alecandido: I believe you could just use `np.mean()` for this
```suggestion
    means = np.mean(values == state, axis=0)
```

But test it.

src/qibocal/result.py
```diff
@@ -78,6 +78,7 @@ def phase(iq: npt.NDArray):
...
+    return np.clip(means, 0, 1)
```
alecandido: Also this one I'm not sure how it is useful: if you average an array of Booleans (effectively, an array of 0s and 1s), the value falls in the `[0, 1]` interval by design.

doc/source/protocols/cross_resonance/amplitude.png
alecandido, line 1: Do we want to replace this with any more updated picture?

The `control_amplitude` protocol is now an Hamiltonian tomography as well. Not sure about the `off_res_crosstalk` variants

doc/source/protocols/cross_resonance/length.png
alecandido, line 1: Same as above

src/qibocal/protocols/two_qubit_interaction/cross_resonance/__init__.py
```diff
@@ -0,0 +1,15 @@
+from .cancellation_amplitude import hamiltonian_tomography_canc_amplitude
+from .cancellation_phase import hamiltonian_tomography_canc_phase
+from .control_amplitude import hamiltonian_tomography_cr_amplitude
+from .length import hamiltonian_tomography_cr_length
```
alecandido: The names of the two modules are a bit inconsistent

src/qibocal/protocols/two_qubit_interaction/cross_resonance/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+from .off_res_crosstalk_amp import cr_crosstalk_amplitude
+from .off_res_crosstalk_length import cr_crosstalk_length
```
alecandido: What are these protocols?

doc/source/protocols/cross_resonance/cross_resonance.rst
alecandido, line 1: Given this is more a "protocol suite" than a single protocol, we should put some care in documenting it properly, to avoid confusion about the roles of the various protocols.

In this sense, this page is quite fundamental.

doc/source/protocols/cross_resonance/cross_resonance.rst
```diff
@@ -0,0 +1,188 @@
...
+The probability of the target qubit is fitted in both cases to a dumped cosine functions.
+It is possible to extract the effective coupling as
+
+.. math::
+
+    \text{J}_\text{eff}/ 2 \pi = \frac{f^{\pi}_\text{Rabi} - f_\text{Rabi}}{2}
+
+
+where :math:`f^{\pi}_\text{Rabi}` and :math:`f_\text{Rabi}` are the frequencies of the
+fitted Rabi oscillations on the target qubit.
```
alecandido, line 92: This protocol does not exist any longer (to the best of my knowledge) since it has been replaced by `hamiltonian_tomography_cr_length` (unless it is `cr_crosstalk_length`, I'm still confused about that).

doc/source/protocols/cross_resonance/cross_resonance.rst
```diff
@@ -0,0 +1,188 @@
...
+
+
+The expected output is the following:
+
+.. image:: amplitude.png
+
+Post-processing
+^^^^^^^^^^^^^^^
+
+The probability of the target qubit is fitted in both cases to a cosine function.
```
alecandido, line 133: Same as above

src/qibocal/protocols/two_qubit_interaction/cross_resonance/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+
+__all__ = [
+    "hamiltonian_tomography_cr_amplitude",
+    "hamiltonian_tomography_cr_length",
+    "hamiltonian_tomography_canc_phase",
+    "hamiltonian_tomography_canc_amplitude",
```
alecandido: If we decide to only provide the Hamiltonian tomography variant of these protocols, prepending them by `hamiltonian_tomography_` is just redundant.

alecandido: Moreover, we may consider to disable the X and Y components in the basic amplitude and length, for a faster calibration (not sure if useful at any time). However, it may still be more convenient to have them as flags limiting the full protocol (seen as a performance tradeoff), rather than fully separate protocols.

Even if agreed, this would be just a performance improvement, and it should be addressed in a separate PR (in case, let's just open an issue).

doc/source/protocols/cross_resonance/cross_resonance.rst
```diff
@@ -0,0 +1,188 @@
...
+      nshots: 2000
+      pulse_amplitude: 0.1
+      pulse_duration_end: 400
+      pulse_duration_start: 10
+      pulse_duration_step: 20
+
+
+The expected output is the following:
+
+.. image:: tomography_length.png
```
alecandido, line 180: This is definitely inconsistent. If nothing else, at least the `operation:` key

alecandido (commented):
Very first batch.

Still missing all the protocols and Quinn-Fernandes

src/qibocal/protocols/rabi/utils.py
```diff
@@ -365,21 +365,23 @@ def fit_length_function(
...
+                (y_max - y_min) * popt[1] * np.exp(x_min * popt[4] / (x_max - x_min)),
+                popt[2] * (x_max - x_min),
+                popt[3] - 2 * np.pi * x_min / popt[2] / (x_max - x_min),
+                popt[4] / (x_max - x_min),
+            ]
+        )
 
     pi_pulse_parameter = popt[2] / 2 * period_correction_factor(phase=popt[3])
-    return popt, perr, pi_pulse_parameter
+    return popt.tolist(), perr.tolist(), pi_pulse_parameter
```
alecandido: `popt` seems to be a `list` if coming from the `if` branch. In which case, it would raise an error for `.tolist()`. The LSP should tell you about this

src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_amplitude.py
```diff
@@ -0,0 +1,338 @@
...
+
+__all__ = ["hamiltonian_tomography_cr_amplitude"]
+
+HamiltonianTomographyCRAmplType = np.dtype(
+    [
+        ("prob_target", np.float64),
+        ("error_target", np.float64),
+        ("prob_control", np.float64),
+        ("error_control", np.float64),
+        ("x", np.float64),
```
alecandido: Maybe call it `amplitude`, to make it more explicit

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
```diff
@@ -0,0 +1,338 @@
...
+    """HamiltonianTomographyCRAmplitude runcard inputs."""
+
+    pulse_duration: float
+    """Duration of CR pulse [ns]."""
+    control_ampl_start: float
+    """Amplitude of cancellation pulse."""
+    control_ampl_end: float
+    """Final amplitude of CR pulse."""
+    control_ampl_step: float
+    """Step CR pulse amplitude."""
```
alecandido: For this, since we did not publish yet (and no one wrote the docs), we could directly go for the range syntax, without any compatibility mechanism (since there is no legacy code to support)

src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_amplitude.py
```diff
@@ -0,0 +1,338 @@
...
+    control_ampl_start: float
+    """Amplitude of cancellation pulse."""
+    control_ampl_end: float
+    """Final amplitude of CR pulse."""
+    control_ampl_step: float
+    """Step CR pulse amplitude."""
+    control_phase: float = 0.0
+    """Phase of the CR pulse on the control qubit"""
+    target_phase: float = 0.0
+    """Phase of the Cancellation pulse on the target qubit"""
```
alecandido, line 57: In a sense, better more parameters than less. But... what do we need these for?

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
```diff
@@ -0,0 +1,338 @@
...
+    control_phase: float = 0.0
+    """Phase of the CR pulse on the control qubit"""
+    target_phase: float = 0.0
+    """Phase of the Cancellation pulse on the target qubit"""
+    target_amplitude: float = 0.0
+    """Amplitude of the Cancellation pulse on the target qubit"""
```
alecandido: Since this is the base protocol, also support the explicit exclusion of the cancellation pulse, adding a `None` value, which should be the default (which is practically the same of a pulse with zero amplitude - but it is not the same for the compilation).

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
```diff
@@ -0,0 +1,338 @@
...
+    """Data acquisition for Hamiltonian tomography CR protocol.
+
+    We measure the expectation values X,Y and Z on the target qubit after
+    applying the CR sequence specified by the input parameters. We repeat the
+    measurement twice for each target qubit, once with the control qubit in state 0
+    and once with the control qubit in state 1.
+
+    We store the probability of the control qubit and the expectation value of the target qubit.
+
+    """
```
alecandido: As for the Ramsey, better moving this docstring to the `Routine` object down below

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
```diff
@@ -0,0 +1,338 @@
...
+        control_phase=params.control_phase,
+        target_amplitude=params.target_amplitude,
+        target_phase=params.target_phase,
+    )
+
+    for pair in targets:
+        control, target = pair
+
+        for basis in Basis:
+            for setup in SetControl:
```
alecandido: You should unroll it. Which just means collecting all the sequences, in a `sequences` list, and pass it to `platform.execute()` in a single batch.

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
```diff
@@ -0,0 +1,338 @@
...
+    data = HamiltonianTomographyCRAmplData(
+        echo=params.echo,
+        cr_duration=params.pulse_duration,
+        control_phase=params.control_phase,
+        target_amplitude=params.target_amplitude,
+        target_phase=params.target_phase,
+    )
+
+    for pair in targets:
+        control, target = pair
```
alecandido: Different pairs should be executed in parallel. If overlapping, just raise an error (as for Ramsey ZZ)

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
```diff
@@ -0,0 +1,338 @@
...
+
+        for basis in Basis:
+            for setup in SetControl:
+                sequence, cr_pulses, _, _, _ = cross_resonance_experiment(
+                    platform=platform,
+                    control=control,
+                    target=target,
+                    duration=params.pulse_duration,
+                    control_amplitude=params.control_ampl_end,
```
alecandido: Since you're sweeping it, no need to give any actual value to the control amplitude. Pass it a `0.0`

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
```diff
@@ -0,0 +1,338 @@
...
+                updates = []
+                updates.append(
+                    {
+                        platform.qubits[control].drive_extra[target]: {
+                            "frequency": platform.config(
+                                platform.qubits[target].drive
+                            ).frequency
+                        }
+                    }
+                )
```
alecandido:
```suggestion
                updates = [
                    {
                        platform.qubits[control].drive_extra[target]: {
                            "frequency": platform.config(
                                platform.qubits[target].drive
                            ).frequency
                        }
                    }
                ]
```

src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_amplitude.py
```diff
@@ -0,0 +1,338 @@
...
+                    dict(
+                        x=amp_sweeper.values,
+                        prob_target=1 - 2 * prob_target,
+                        error_target=(
+                            2 * np.sqrt(prob_target * (1 - prob_target) / params.nshots)
+                        ).tolist(),
+                        prob_control=prob_control,
+                        error_control=np.sqrt(
+                            prob_control * (1 - prob_control) / params.nshots
+                        ).tolist(),
```
alecandido, line 198: By now, these formulae are really everywhere. We should move them to a single place in Qibocal.

Not recommended in this PR, but we should open an issue for this.

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
```diff
@@ -0,0 +1,338 @@
...
+) -> HamiltonianTomographyCRAmplResults:
+    """Post-processing function for HamiltonianTomographyCRAmpl.
+
+    We fit the expectation values using the Eq. S10 from the paper https://arxiv.org/pdf/2303.01427.
+    Afterwards, we extract the Hamiltonian terms from the fitted parameters.
+
+    """
+    fitted_parameters, cr_gate_ampls = tomography_cr_fit(
+        data=data,
+    )
```
alecandido: Stupid remark: Python formatters are a bit dumb: if you put a comma in a list of parameters, they will split them o multiple lines. Regardless of whether they fit or not on a single line.

This clearly happened with this function:
```suggestion
    fitted_parameters, cr_gate_ampls = tomography_cr_fit(data=data)
```

Possibly it had more arguments before, and when you started deleting them as lines, they remained on separated lines.

(occasionally it is also useful to force them on multiple lines - but I mostly use it as an editing tool, since it is easier to navigate them once they are already on multiple lines)

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
```diff
@@ -0,0 +1,338 @@
...
+        fitting_report = ""
+    return figs, fitting_report
+
+
+def _update(
+    results: HamiltonianTomographyCRAmplResults,
+    platform: CalibrationPlatform,
+    target: QubitPairId,
+):
+    target = target[::-1] if target not in results.cr_amplitudes else target
```
alecandido: Actually, why?

We need the tuples to be sorted in the correct way

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
```diff
@@ -0,0 +1,338 @@
...
+        platform.qubits[target[0]].drive,
+        VirtualZ(phase=np.pi / 2),
+    )
+
+    new_cr_seq = (
+        PulseSequence([control_single_qubit_operation, target_single_qubit_operation])
+        | cr_seq
+    )
+
+    getattr(update, f"{results.native.lower()}_sequence")(new_cr_seq, platform, target)
```
alecandido: This is an extra-convoluted way to write
```suggestion
    update.cnot_sequence(new_cr_seq, platform, target)
```
since it can not be anything else than a CNOT.

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
```diff
@@ -0,0 +1,338 @@
...
+
+    control_single_qubit_operation = (
+        platform.qubits[target[0]].drive,
+        VirtualZ(phase=np.pi / 2),
+    )
+
+    new_cr_seq = (
+        PulseSequence([control_single_qubit_operation, target_single_qubit_operation])
+        | cr_seq
+    )
```
alecandido: It's the first protocol I review, but I'm pretty sure this is everywhere, to make the gate out of the pulse.
Better to make a function for this.

src/qibocal/protocols/two_qubit_interaction/cross_resonance/fitting.py
alecandido, line 1: The whole module is named `fitting`. So, no need to repeat `fit_`/`_fit` i the name of each function.

Moreover, they are the functional form. Which are just functions, that you may use both for fitting and evaluation. And they are performing no fit by themselves.
So, I would also update the docstrings.

src/qibocal/protocols/two_qubit_interaction/cross_resonance/fitting.py
```diff
@@ -0,0 +1,73 @@
...
+    t: np.ndarray, wx: float, wy: float, wz: float, w: float, gamma: float
+) -> np.ndarray:
+    """Fitting Y expectation value for CR tomography.
+
+    See https://arxiv.org/pdf/2303.01427 Eq. S10.
+    """
+    return (
+        (-w * wx * np.sin(w * t) - wy * wz * np.cos(t * w) + wy * wz)
+        / w**2
+        * np.exp(-abs(gamma) * t)
```
alecandido: Why only in this function you put the absolute value for `gamma`? I'm pretty sure we could assume `gamma` to be positive (also because of the usual `bounds`). And, if it is not, it's a problem also in the other functions.

src/qibocal/protocols/two_qubit_interaction/cross_resonance/fitting.py
```diff
@@ -0,0 +1,73 @@
...
+    gamma: float,
+) -> np.ndarray:
+    """Simulateneous fit for X, Y and Z expectation values.
+
+    We also constrain w to be the norm of the w vector with component
+    wx, wy, wz."""
+
+    w = np.sqrt(wx**2 + wy**2 + wz**2)
+    t1, t2, t3 = np.split(t, 3)
+    gamma = np.abs(gamma)
```
alecandido: Btw, the `np.exp(-gamma*t)` is a multiplicative factor common to all the three components.

Instead of adding it to each and every component, just multiply the concatenation.

src/qibocal/protocols/two_qubit_interaction/cross_resonance/fitting.py
```diff
@@ -0,0 +1,73 @@
+"""Fitting function for CR tomography."""
+
+import numpy as np
+
+
+def fit_Z_exp(
```
alecandido: Use lower-case names (snake-case, precisely) for function names. In case of doubts, PEP 8
https://peps.python.org/pep-0008/#function-and-variable-names

src/qibocal/protocols/two_qubit_interaction/cross_resonance/fitting.py
alecandido: Same for `/ w**2`

src/qibocal/protocols/two_qubit_interaction/cross_resonance/fitting.py
```diff
@@ -0,0 +1,73 @@
...
+    w = np.sqrt(wx**2 + wy**2 + wz**2)
+    t1, t2, t3 = np.split(t, 3)
+    gamma = np.abs(gamma)
+    return np.concatenate(
+        [
+            fit_X_exp(t1, wx, wy, wz, w, gamma),
+            fit_Y_exp(t2, wx, wy, wz, w, gamma),
+            fit_Z_exp(t3, wx, wy, wz, w, gamma),
+        ]
+    )
```
alecandido: Moreover, the structure of the three terms is always the same:

$$
a \cos(\omega t) + b \sin(\omega t) + c
$$

Instead of repeating it over and over, define the `np.array([np.cos(w*t), np.sin(w*t), 1])`, and multiply for the array of coefficients, which are the only things you need to keep separate.

alecandido (commented):
Still missing 5 protocols and 3 `utils` files.

Not even sure why we have 5 protocols, since I only expected 4 (amplitude, length, cancellation phase, cancellation amplitude). But I already wrote this somewhere else, and I will eventually figure out what are the `off_res*` ones

src/qibocal/protocols/two_qubit_interaction/cross_resonance/off_res_crosstalk_amp.py
alecandido, line 1: Maybe we could postpone this to a different PR, and merge into Rabi. We could specify what we read with the usual `targets=` argument, and add a `drive=` argument, which is a matching list, and, if not specified, it is assumed to be the same of `targets=`.

src/qibocal/protocols/two_qubit_interaction/cross_resonance/off_res_crosstalk_length.py
alecandido, line 1: Same as above

src/qibocal/update.py
lballerio: do you mean:

```python
{f"native_gates.two_qubit.{f'{TypeAdapter(QubitPairId).dump_json(pair)}'}.CNOT": cr_sequence}
```
?

src/qibocal/result.py
lballerio: yeah was something made mostly for the emulator, in case of some numerical error accumulated.
But as we already checked this clipping is already done in the qibolab side, so it is safe to delete.

src/qibocal/result.py
lballerio: yeah I kept it as implemented before but it is better to use `np.mean` instead.

doc/source/protocols/cross_resonance/amplitude.png
lballerio: we already discussed but I think is better to repeat and leave words written down on server :laughing:...
`off_res_crosstalk` experiments are simply crosstalk rabi experiments, but we are considering whether addressing these two protocols in another PR and fuse them within the `rabi` scripts since they are the same experiment.

lballerio: Also, do we want to plot also the control qubit state?

doc/source/protocols/cross_resonance/cross_resonance.rst
lballerio: actually there was already a first implementation of the hamiltonian tomography, but yeah we are no longer relying on that. I think documentation will be the last thing I'll fix before merging.

doc/source/protocols/cross_resonance/cross_resonance.rst
lballerio: why is it inconsistent?

doc/source/protocols/cross_resonance/cross_resonance.rst
lballerio: I agree, I'll keep doc updating as last fix.

src/qibocal/protocols/rabi/utils.py
lballerio: you're right, silly error

lballerio: Btw do you know why `perr` is manipulated only for `signal==False`?

src/qibocal/protocols/two_qubit_interaction/cross_resonance/__init__.py
lballerio: change done in 998bae646d46faa608995c3d1aec8bc38ef6116c

src/qibocal/protocols/two_qubit_interaction/cross_resonance/__init__.py
lballerio: these are the Cross Rabi experiments, now I'll remove from this PR and add in a new one

src/qibocal/protocols/two_qubit_interaction/cross_resonance/__init__.py
lballerio: If we do only Z we cannot then tune the cancellation pulse, then we are essentially doing the Cross Rabi experiment.
We could create a wrapper protocol calling the general `rabi` protocols.

src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_amplitude.py
lballerio: here I am calling `x` because then I can use the same fitting module for both `amplitude` and `duration` experiments

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
lballerio: perfect, even better

src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_amplitude.py
lballerio: I think it depend how do you interpret these package of experiments...
In fact, if we consider them as Hamiltonian Tomography, where then we estimate the contributions of each Hamiltonian term, then we should be as most flexible in the parameters as possible. Otherwise, if we only consider them for calibrating from scratch the CR pulse, then I agree with you and can be deleted, also with `target_amplitude` I would say.

Or maybe can be used in maintaining the calibration?

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
lballerio: ok perfect

doc/source/protocols/cross_resonance/amplitude.png
alecandido: > we already discussed but I think is better to repeat and leave words written down on server 😆...
> `off_res_crosstalk` experiments are simply crosstalk rabi experiments, but we are considering whether addressing these two protocols in another PR and fuse them within the `rabi` scripts since they are the same experiment.

I had the same idea, and put an explicit comment below 😅 
https://github.com/qiboteam/qibocal/pull/1414#discussion_r3348962878

> Also, do we want to plot also the control qubit state?

Not a bad idea, but I would open an issue and deal in a separate PR. This one is already massive enough, that I would focus as much as possible on merging

src/qibocal/protocols/two_qubit_interaction/cross_resonance/__init__.py
alecandido: Indeed, you're right. It would be just cross-Rabi, but a conditional one, i.e. with the extra feature of toggling the qubit related to the drive channel, producing two series of data.

So, it is somewhere in between cross-Rabi and an actual CR

doc/source/protocols/cross_resonance/cross_resonance.rst
alecandido: Maybe *inconsistent* was misleading, as a word. Let's say *outdated*

src/qibocal/protocols/rabi/utils.py
alecandido: To be fair, that's a good point. It doesn't seem meaningful, since we want to extract the variances.

However, even that is wrong: we care about covariances, since we are not just using `popt`, but we are actually turning it into another `popt` (which is shadowing, and changing the meaning of the variable - we truly need a different name for the variable assigned with the `if: ... else: ...` branch). So, we need error propagation, and the distributions of `popt` are not at all independent.

Right now, `perr` is just outright wrong

src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_amplitude.py
alecandido: I see. Then, maybe, it is just better to move this in one of the `utils` modules, rather than keeping it in one, and using it in both protocols. Rather than defining it in one of them

src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_amplitude.py
alecandido: Ok, let's discuss about this

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
lballerio: adressed in b764d4f0c734b0a68413ee0794cd83a64caa1543

doc/source/protocols/cross_resonance/amplitude.png
lballerio: replacement addressed in 998bae646d46faa608995c3d1aec8bc38ef6116c

lballerio: > Not a bad idea, but I would open an issue and deal in a separate PR. This one is already massive enough, that I would focus as much as possible on merging.

Ok perfect

doc/source/protocols/cross_resonance/length.png
lballerio: replacement addressed in 998bae646d46faa608995c3d1aec8bc38ef6116c

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
lballerio: addressed in 998bae646d46faa608995c3d1aec8bc38ef6116c

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
lballerio: I agree, even though it has to be tested on real clusters to see if there is some memory issue

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
lballerio: yeah I know, just a placeholder, but I can absolutely set to `0.0`.

lballerio: addressed in 998bae646d46faa608995c3d1aec8bc38ef6116c

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
lballerio: addressed in 998bae646d46faa608995c3d1aec8bc38ef6116c

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
lballerio: was so dumb that I've already fixed even before reading this comment 😆!
Fixed in 998bae646d46faa608995c3d1aec8bc38ef6116c

src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_amplitude.py
lballerio: Put in #1517

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
lballerio: addressed in 998bae646d46faa608995c3d1aec8bc38ef6116c

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
lballerio: I agree, I also think is quite useless. I think I inherited from some very old version...

lballerio: I think was intended to be more an additional check, but honestly is useless... if things are done consistently i cannot find a reason why to keep it.

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
lballerio: as before, I also noted this very bad writing before reading this comment 😆 

Addressed in  998bae646d46faa608995c3d1aec8bc38ef6116c

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
lballerio: written as issue about unroll for CR calibration #1518

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
lballerio: addressed in 998bae646d46faa608995c3d1aec8bc38ef6116c

src/qibocal/protocols/two_qubit_interaction/cross_resonance/fitting.py
lballerio: addressed in 998bae646d46faa608995c3d1aec8bc38ef6116c

src/qibocal/protocols/two_qubit_interaction/cross_resonance/fitting.py
lballerio: addressed in 1404d858ebccc7ee3476cb0c692632817c2e8e0b

src/qibocal/protocols/two_qubit_interaction/cross_resonance/fitting.py
lballerio: yeah, you're right and I remember we already discussed about that.
I think I found slightly different performances if together with the bounds we also impose `np.abs(gamma)`, but honestly I don't remember exactly.

lballerio: but then when the function is called in `ham_tomography_utils.py`, in function `scipy_curve_fit_optimizer` we actually impose positivity with the `bounds`.

Anyway, addressed in 998bae646d46faa608995c3d1aec8bc38ef6116c

src/qibocal/protocols/two_qubit_interaction/cross_resonance/fitting.py
lballerio: addressed in 998bae646d46faa608995c3d1aec8bc38ef6116c

src/qibocal/protocols/two_qubit_interaction/cross_resonance/fitting.py
lballerio: addressed in 998bae646d46faa608995c3d1aec8bc38ef6116c

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
lballerio: ok thanks

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
lballerio: addressed in 998bae646d46faa608995c3d1aec8bc38ef6116c

src/qibocal/protocols/two_qubit_interaction/cross_resonance/off_res_crosstalk_amp.py
lballerio: > Maybe we could postpone this to a different PR, and merge into Rabi. We could specify what we read with the usual `targets=` argument, and add a `drive=` argument, which is a matching list, and, if not specified, it is assumed to be the same of `targets=`.

I agree

src/qibocal/protocols/utils.py
lballerio: addressed in 56f44603094e7ae4273971d2ce9f26bfed1bf77c

src/qibocal/protocols/utils.py
lballerio: addressed in 56f44603094e7ae4273971d2ce9f26bfed1bf77c

src/qibocal/result.py
lballerio: addressed in 56f44603094e7ae4273971d2ce9f26bfed1bf77c

src/qibocal/result.py
lballerio: addressed in 56f44603094e7ae4273971d2ce9f26bfed1bf77c

src/qibocal/protocols/two_qubit_interaction/cross_resonance/control_amplitude.py
lballerio: addressed in 998bae646d46faa608995c3d1aec8bc38ef6116c

src/qibocal/protocols/rabi/utils.py
lballerio: ok, so do you have any plan for that?

src/qibocal/protocols/two_qubit_interaction/cross_resonance/__init__.py
lballerio: addressed in 998bae646d46faa608995c3d1aec8bc38ef6116c

src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_amplitude.py
lballerio: yeah I agree...addressed in d18d0f3ee

alecandido (approved):

scarrazza (approved):

PR #1416, opened 2026-04-01
Add support for parallel two qubit benchmarking

sorewachigauyo:
Instead of launching the benchmarking sequentially per qubit pair, we use the same circuit across all pairs and run the benchmarking in parallel to provide some speedup. 

Tests on QPU
[rb2q-parallel-test.tar.gz](https://github.com/user-attachments/files/26400447/rb2q-parallel-test.tar.gz)

### DIFF
--- a/src/qibocal/protocols/randomized_benchmarking/standard_rb_2q.py
+++ b/src/qibocal/protocols/randomized_benchmarking/standard_rb_2q.py
@@ -8,7 +8,13 @@ from qibocal.protocols.randomized_benchmarking.standard_rb import (
 )
 
 from ...calibration.calibration import TwoQubitCalibration
-from .utils import RB2QData, StandardRBResult, fit, twoq_rb_acquisition
+from .utils import (
+    RB2QData,
+    StandardRBResult,
+    fit,
+    twoq_rb_acquisition,
+    twoq_rb_acquisition_parallel,
+)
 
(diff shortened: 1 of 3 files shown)


RoyStegeman:
Hi @sorewachigauyo, there is currently a PR waiting for review that changes a lot about circuit execution and RB in particular https://github.com/qiboteam/qibocal/pull/1393. The PR mainly changes the use of the backend and the use of `MeasurementOutcomes`. I would suggest rebasing on top of that. 

Passing a list of `targets`, that can be qubit pairs, is expected to execute the protocol for all qubitids in that list simultaneously for any protocol. At the moment this is not the case for RB, but it means that what you are implementing should be the standard/only behavior if a list with multiple pairs is passed as input, and thus there is no need for a `parallel` flag. 

As a general comment regarding implementation: we should avoid multiple calls to `excute_circuit` and instead pass the full list of compiled sequences to the `platform.execute` at once (see e.g. the snippet below). The platform should then take care of the unrolling and batching, and execute the circuits in the most efficient way. 

https://github.com/qiboteam/qibocal/blob/2dbbf0a37b107c6f1f315eadf6fefe201e776c6a/src/qibocal/auto/transpile.py#L170

sorewachigauyo:
Moved to https://github.com/qiboteam/qibocal/pull/1490

PR #1417, opened 2026-04-01
small cleanup of README

RoyStegeman:
I had a few minutes to kill while waiting for the qpu to be available... 

- The link to the docs badge was broken and putting the same link twice seems a bit overkill
- A changelog is fine, but maybe more suitable once we have a v1 release, and changes.md is by now outdated


scarrazza (approved):
Thanks @RoyStegeman !

lballerio (approved):
sounds fine for me.
Btw the problem with the first fix was only the duplicate link or am I missing something else?

RoyStegeman:
> Btw the problem with the first fix was only the duplicate link or am I missing something else?

No, the link for the badge was based on old workflows so had to be updated

Issue #1419, opened 2026-04-01
protocols (with qblox) are slow despite unrolling


RoyStegeman:
### Flipping
I ran flipping with the following settings
```
            delta_amplitude=0.05
            nflips_max=30
            nflips_step=1
            unrolling=True
```

And executing the experiment took a factor 15 more than the pulse-execution itself.  
```
INFO:qibocal.config:Executing mode AUTOCALIBRATION on flipping.
INFO:qibolab._core.platform.platform:Minimal execution time: 9.084 s
[Qibocal 0.2.5|INFO|2026-04-01 18:18:11]: Finished acquisition in 141.44 seconds
```

### drag_tuning

```
            beta_start=-4,
            beta_end=4,
            beta_step=0.1,
            nflips=5,
            unrolling=True,
```
```
INFO:qibocal.config:Executing mode AUTOCALIBRATION on drag_tuning.
INFO:qibolab._core.platform.platform:Minimal execution time: 24.160 s
[Qibocal 0.2.5|INFO|2026-04-01 18:35:12]: Finished acquisition in 376.39 seconds
```

jevillegasd:
Same if I run a calibration with 100 shots

```
  - id: motzoi_calibration
    operation: drag_simple
    parameters:
      beta_start: -2
      beta_end: 1
      beta_step: 0.2
      nshots: 100
      unrolling: true
```

or with 1000 shots, the time of the experiment is roughly the same, and in general terribly slow (so unrolling is still not really used?)
(100 shots)
```
INFO:qibocal.config:Executing mode AUTOCALIBRATION on motzoi_calibration.
[38;20m[Qibocal 0.2.5|INFO|2026-04-14 19:45:51]: Finished acquisition in 98.18 seconds[0m
INFO:qibocal.config:Finished acquisition in 98.18 seconds
[38;20m[Qibocal 0.2.5|INFO|2026-04-14 19:45:51]: Finished fit in less than 1 second.[0m
INFO:qibocal.config:Finished fit in less than 1 second.
```
(1k shots)
```
[38;20m[Qibocal 0.2.5|INFO|2026-04-14 19:41:07]: Executing mode AUTOCALIBRATION on motzoi_calibration.[0m
INFO:qibocal.config:Executing mode AUTOCALIBRATION on motzoi_calibration.
[38;20m[Qibocal 0.2.5|INFO|2026-04-14 19:42:49]: Finished acquisition in 101.05 seconds[0m
[...]

PR #1434, opened 2026-04-07
improvements for drag_tuning and drag_simple protocols

### LINKED ISSUE #1420: issues with drag protocol
The fit fails regularly:
- http://login.qrccluster.com:9000/6DbcQ8f3Sqa56S-V6PEU9w==

the "Optimal Beta Param" may show two sigfigs, such 3.0, but only one of them is a genuine sigfig. So it will show 3.0 also for 3.3333. 

The optimal beta parameter is determined from the fit, but only from hose points in the fitted function evaluated at the values of beta where data was taken. 

The error on the beta param is hardcoded to be 0

The protocol is also slow due to the singleshot nature (see #1419), but the chi2 is being provided so that has to change.

RoyStegeman:
addresses #1420

This fixes some issues with the standard drag protocol (not drag_simple). Except for the default unrolling and CYCLIC, which is also done for drag_simple. 

- Issue: The "Optimal Beta Param" may show two sigfigs, such 3.0, but only one of them is a genuine sigfig - for example it will show 3.0 also for 3.3333.
  Fix: This was actually related to the issue below regarding the fact that the error was simply passsed as 0, because the script that prepares the table matches the sigfigs of the central beta param to be the same as the error. Either way, we resolve this by showing 5 sigfigs for the beta param.

- Issue: The error on the beta param is hardcoded to be 0
   Fix: Since it is 0 anyway, there is no point in showing it since the error on chi2 is not so interesting.

- Issue: The optimal beta parameter is determined from the fit, but only from those points in the fitted function evaluated at the values of beta where data was taken. 
   Fix: Compute the max from the sinusoidal functional form. If multiple maxima are shown in the beta window, take the value with the smallest absolute value.  

- Issue: The protocol is also slow due to the singleshot nature (see #1419), but the chi2 is being provided so that has to change. 
    Fix: We now always do unrolling and it is no longer a user choice. We also do cyclic instead of singleshot. Uncertainties are computed assuming binomial distribution so that didn't depend on having shots. 

- `unrolling` has been removed at the level of `DragTuningParameters`, so the performance improvements (cyclic and unrolling) have also been propagated to drag_simple. For that protocol there were not the other issues. 

**Before**
image

**After**
image

> [!NOTE]  
> There are two protocols to determine the beta parameter of the drag pulse, but presumably one suffices so we may also think about scrapping one if there is no benefit to keeping both.


src/qibocal/protocols/utils.py
```diff
@@ -319,6 +319,13 @@ def round_report(
     rounded_values = []
     rounded_errors = []
     for value, error in measure:
+        if isinstance(error, str):
+            rounded_values.append(
+                np.format_float_scientific(value, trim="-").replace("e+00", "")
+            )
+            rounded_errors.append(error)
+            continue
```
alecandido: I'm not sure why the error is supposed to be a string, instead of a float. I would try to standardize to just one type. And, since it is a number, it would make sense to make it a float (for `"not determined"` just use `float("nan")`, which is standard also for invalid values in floating point algebra, or in libraries like Pandas for missing values).

alecandido: In general, all this function is a bit of a mess. I understand the purpose is to make an easier comparison between the value and the error, but `(1e6, 1e2)` is arguably easier than `(1, 0.0001) * 1e6`. And the imbalanced exponential syntax is just coming with [Python default float formatters](https://docs.python.org/3/library/string.html#formatspec:~:text=The%20available%20presentation%20types%20for%20float%20and%20Decimal%20values%20are) (not even NumPy).

While not necessarily perfect[*], a good enough solution for formatting value and errors could be using the `#.3g` and `#.2g` respectively (or `#.4g` and `#.2g`).

If nothing else, it would greatly simplify the implementation.

[*]: I would like to tune a bit more how to represent numbers smaller than 1, and to control independently sig figs and when it starts using the exponent, e.g. start using the exponent only `>=1000`, but still use just 2 sig figs

src/qibocal/protocols/flipping.py
```diff
@@ -185,7 +184,8 @@ def _acquisition(
                 result = results[ro_pulse.id]
             else:
                 result = results[i][ro_pulse.id]
-            prob = probability(result, state=1)
+            # prob = probability(result, state=1)
```
alecandido: This was used only because of `SINGLESHOT`. You can just drop

src/qibocal/protocols/drag/drag.py
```diff
@@ -132,44 +129,32 @@ def _acquisition(
...
-                    (qubit),
-                    dict(
-                        prob=np.array([prob]),
-                        error=np.array([np.sqrt(prob * (1 - prob) / params.nshots)]),
-                        beta=np.array([beta]),
+    # execute the pulse sequences
+    results = platform.execute(sequences, **options)
+    for beta, ro_pulses in zip(beta_param_range, all_ro_pulses):
+        for qubit in targets:
+            ground_state_prob = 1 - results[ro_pulses[qubit].id]
```
alecandido: Why recording ground? Usually we report the excited state probability

alecandido (commented):
Just having a quick look

src/qibocal/protocols/drag/drag.py
RoyStegeman: This was just a refactor, even before the probability was that of the ground state

src/qibocal/protocols/utils.py
RoyStegeman: The previous implementation used a placeholder value of 0 for the beta uncertainty because it was not determined. This placeholder was needed for symmetry with the chi2, because the table included the chi2 uncertainty. Since the chi2 uncertainty (and arguably even its central value) is hardly relevant, the best solution in the context of PR is to simply remove both uncertainties from the table. 

I opened an issue specifically regarding this function: https://github.com/qiboteam/qibocal/issues/1482

src/qibocal/protocols/drag/drag.py
RoyStegeman: Actually I swapped it in a later commit. The y-axis is now excited state probability

RoyStegeman:
I thought about deduplicating the plotting functions, but since they both loop over `SEQUENCES` this either means making helper functions at the level of the individual `fig.add_trace calls`, or rethinking the `Data` class such that we can for example add a dimension to loop over. 

Probably the latter is nicer, but it will not be a quick thing as we had in mind to start off the deduplication. It's further complicated by the fact that types have not been handled consistently between inheritances of `AbstractData`, so that will have to be handled with some extra care.

src/qibocal/protocols/drag/drag.py
```diff
@@ -186,88 +180,100 @@ def _fit(data: DragTuningData) -> DragTuningResults:
...
-                    qubit_data.error,
+                    qubit_data["error"],
                 ),
-                np.sqrt(2 / len(prob)),
+                np.sqrt(2 / len(qubit_data["prob"])),
             )
         except Exception as e:
             log.warning(f"drag_tuning_fit failed for qubit {qubit} due to {e}.")
-    return DragTuningResults(betas_optimal, fitted_parameters, chi2=chi2)
+    return DragTuningResults(betas_optimal, fitted_parameters, chi2=chi2[0])
```
sorewachigauyo: This change breaks the execution

Hotfixing it makes it work

[drag.tar.gz](https://github.com/user-attachments/files/27584015/drag.tar.gz)

RoyStegeman: Good point, thanks for pointing that out. We don't actually use the chi2 anymore for the plotting downstream so that's why I missed it (and for me the 0 key existed).

src/qibocal/protocols/drag/drag.py
```diff
@@ -86,7 +89,9 @@ def _acquisition(
     """
 
     data = DragTuningData()
-    beta_param_range = np.arange(params.beta_start, params.beta_end, params.beta_step)
+    beta_param_range = np.arange(
+        params.beta_start, params.beta_end, params.beta_step
+    ).tolist()
```
alecandido: You could start introducing a `.beta` as a range, on the blueprint of what @lballerio did in #1474, i.e.
https://github.com/qiboteam/qibocal/blob/479a66a7e5b5276db104ab55864787fe8f1aaf2c/src/qibocal/protocols/ramsey/acquisition.py#L27-L31
(preserving the compatibility with the current interface), and with #1300 in mind (but not yet implementing it in full).

This seems a small addition to this PR, but it will contribute moving forward our transition program

src/qibocal/protocols/utils.py
```diff
@@ -56,7 +56,7 @@
 
 
 class FeatExtractionError(Exception):
-    def __init__(self, *args):
+    def __init__(self, *args) -> None:
```
alecandido: You can also annotate `*args`, with `*args: Any`.
https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

However, if you actually have to use `Any` there is of course no advantage in annotating, since it is already the implicit type.

src/qibocal/protocols/utils.py
```diff
@@ -56,7 +56,7 @@
...
         super().__init__(*args)
```
alecandido: If nothing is done in it, no need to specify an `.__init__()` method at all. Just give the class an empty body
```suggestion
    pass
```

src/qibocal/protocols/utils.py
```diff
@@ -297,7 +299,7 @@ def cumulative(input_data, points):
     return np.searchsorted(np.sort(points), np.sort(input_data))
 
 
-def eval_magnitude(value):
+def eval_magnitude(value) -> int:
```
alecandido: Also annotate the input: either as a `float`, or `np.ndarray` (possibly `npt.NDArray[float]`)

src/qibocal/protocols/utils.py
```diff
@@ -709,7 +713,7 @@ def table_dict(
     names: list[str],
     values: list,
     display_error=False,
-) -> dict:
+):
```
alecandido: Why did you remove?

`dict` is not very accurate. But better `dict` than nothing, isn't it?

alecandido: Btw, it could even be a `TypedDict`, since the structure is actually quite rigid and respected.

tests/test_fit_functions.py
```diff
@@ -160,11 +167,54 @@ def test_rabi_fit():
...
 
 
+def _approx(*args, **kwargs):
+    """
+    Helper function to make pytest.approx less stringent by default. This is useful for
+    fitting tests where many floating-point and RNG errors accumulate.
+    """
+    if "rel" not in kwargs:
+        kwargs["rel"] = 1e-5
+    return pytest.approx(*args, **kwargs)
```
alecandido: Is this needed strictly required for all parameters? If it's a regression test, it should not be that much required.
It seems to me there should be no random number involved in those tests (least squares fitting is iterative, but deterministic)

tests/test_fit_functions.py
alecandido, line 1: I know it is not extra-long, but I would start giving these tests a structure, and split the new additions (DRAG-related) into their own module.

alecandido (commented):
Still need to review most of `drag.py` and `drag_simple.py`. But here there are some first comments

tests/test_fit_functions.py
RoyStegeman: Maybe not for all, but for some it is, and 1e-5 seems restrictive enough to me.

src/qibocal/protocols/drag/drag.py
```diff
@@ -33,22 +38,53 @@
 
 
 # TODO: add errors in fitting
+# TODO: for ranges of floats we should probably support only linspace, meaning nsteps
+# instead of stepsize. The reason being that the endpoint inclusion becomes ambiguous
+# with floats
```
alecandido: True, but here I would just use the range syntax regardless, and just quote NumPy's statement to the user.
And, then, leave it free to use the step size, understanding the consequences of it.

Sometimes (which is actually often), you may not care that much about the overall number of points, and not even whether the end point is included, but it is just convenient writing "200 kHz", rather than computing it.
If the tradeoffs are understood, it should be reasonable.

---

> When using a non-integer step, such as 0.1, it is often better to use [`numpy.linspace`](https://numpy.org/doc/stable/reference/generated/numpy.linspace.html#numpy.linspace).

Image

https://numpy.org/doc/stable/reference/generated/numpy.arange.html

src/qibocal/protocols/drag/drag.py
```diff
@@ -33,22 +38,53 @@
...
+            )
+        if has_beta:
+            return self.beta
+        if has_all_beta_fields:
+            return (self.beta_start, self.beta_end, self.beta_step)
+        raise ValueError(
+            "Must define either `beta` tuple or all of `beta_start`, `beta_end`, "
+            "`beta_step`."
+        )
+
```
alecandido: To me this is even overly refined: even checking whether the range is set or not, and then deciding which of the two would have had priority over the other (which should be the old one, just for the sake of stability) it would have been already quite fine.

I do not care that much, we could even keep it like this, for the time being. This will be scrapped anyhow when the transition will be completed.

src/qibocal/protocols/drag/drag.py
```diff
@@ -96,15 +133,16 @@ def _acquisition(
...
+            drag = qd_pulse.model_copy(
+                update={
+                    "envelope": Drag(
+                        rel_sigma=qd_pulse.envelope.rel_sigma,
+                        beta=beta_param,
+                    )
+                }
             )
-            drag_negative = replace(drag, relative_phase=np.pi)
+            drag_negative = drag.model_copy(update={"relative_phase": np.pi})
```
alecandido, line 141: This `replace()` function is pretty useless, it was just a device to transition from `dataclasses` to Pydantic.

However, we have this other function in Qibolab
https://github.com/qiboteam/qibolab/blob/af313eccccdf49e429d1cbae291a39355200274c/src/qibolab/_core/parameters.py#L214-L223
which is instead used to address exactly the replacement of nested objects, in a bit more straightforward way.

In particular, if `Drag` were even slightly more complex, you would have needed a double call to `.model_copy()`
```py
drag = qd_pulse.model_copy(
    update={
        "envelope": qd_pulse.envelope.model_copy(update={"beta": beta_param})
    }
)
```

To avoid a large number of nested replacements, the strategy in the cited function is to dump the object to a dictionary, modify it in place, and then reload it into an object. But, being encapsulated, it should be a good enough interface for these kinds of replacements.
In this case, it would look like:
```py
drag = replace(qd_pulse, {"envelope.beta": beta_param})
```

Anyhow, it would be a suggestion worth to be implemented in a separate PR.

src/qibocal/protocols/drag/drag.py
```diff
@@ -186,88 +213,100 @@ def _fit(data: DragTuningData) -> DragTuningResults:
...
-        prob_min = np.min(prob)
-        prob_max = np.max(prob)
-        normalized_prob = (prob - prob_min) / (prob_max - prob_min)
-
         # normalize beta
-        beta_params = qubit_data.beta
+        beta_params = qubit_data["beta"]
         beta_min = np.min(beta_params)
         beta_max = np.max(beta_params)
         normalized_beta = (beta_params - beta_min) / (beta_max - beta_min)
```
alecandido: Since you removed the other, why not even this one?

The translation and scaling are just translating as period and phase, which are fit parameters. Not sure why not using directly in the original space, instead of operating the transformation.

The scaling makes sense for peak finding, since it gives us a reference for the prominence. But, if that is not used, and we are even providing a guess for the period[*], we could well avoid the transformation.

[*]: check the Rabi/Ramsey fit, and ask @lballerio, to improve the estimates for the initial value guesses and the domain bounds - this is just the nth sinusoidal fit, and we have seen poor results with similar definitions in the other fits

src/qibocal/protocols/drag/drag.py
```diff
@@ -186,88 +213,100 @@ def _fit(data: DragTuningData) -> DragTuningResults:
...
+
+            period_fit = translated_popt[2]
+            phase_fit = translated_popt[3]
+
+            # calculate the smallest and largest k for which the minimum lies in the
+            # beta interval. Minima of drag_fit(x, offset, amplitude, period, phase)
+            # occur for x = period * (k + 1/2 - phase / (2*pi)), for integer k.
+            phase_2pi = phase_fit / (2 * np.pi)
+            k_min = math.ceil(beta_min / period_fit + phase_2pi - 0.5)
+            k_max = math.floor(beta_max / period_fit + phase_2pi - 0.5)
```
alecandido: Not sure if it makes any sense, but for the sake of uniformity I would replace functions from `math` with the analog ones from `numpy`.

src/qibocal/protocols/drag/drag.py
```diff
@@ -186,88 +213,100 @@ def _fit(data: DragTuningData) -> DragTuningResults:
...
+                betas_optimal[qubit] = min(candidate_betas, key=abs)
+            else:
+                # If no analytical minimum lies in the beta interval, minimum is
+                # fixed at one of the interval boundaries.
+                left_value = drag_fit(beta_min, *translated_popt)
+                right_value = drag_fit(beta_max, *translated_popt)
+                betas_optimal[qubit] = (
+                    beta_min if left_value <= right_value else beta_max
+                )
+
```
alecandido: Here I would propose a different approach: instead of making all the conditions in terms of `if: ...  else:  ...`, you could just add a penalty term, which is synced to the periodicity of the function (since you just determined it - and it is the same one you use in the `if`s).

Here is a brief script motivating the following penalty term:
```py
amplitude*np.abs(np.floor(x / period + phase / (2 * np.pi) % 1))
```
to be added to the results of the `drag_fit`.

<details><summary>Plotting script</summary>

```py
import numpy as np

def drag_fit(x, offset, amplitude, period, phase):
    return offset + amplitude * np.cos(2 * np.pi * x / period + phase)

x = np.linspace(-20, 15, 1000)
offset = 0.1
amplitude = 1.3
period = 2 * np.pi * 0.8
phase = np.pi * -0.5
y = drag_fit(x, offset, amplitude, period, phase)
yp = y + amplitude*np.abs(np.floor(x / period + phase / (2 * np.pi) % 1))
idx = np.argmin(yp)

fig, axs = plt.subplots(
    2, 1, figsize=(10, 6), sharex=True, gridspec_kw={"hspace": 0}
)
axs[0].plot(x, y)
axs[1].plot(x, yp)
axs[0].scatter(x[idx], y[idx], c="r", marker="x")
axs[1].scatter(x[idx], yp[idx], c="r", marker="x")
for ax in axs:
    ax.vlines(0, 0, 1, transform=ax.get_xaxis_transform(), color="gray", linestyle="--")
```

</details>

Image

Since it only depends on fitting parameters (and it is a quick NumPy function, as the fitting one), you can just evaluate it on an arbitrary array of points, and get its `np.argmin()`. Of course, it makes sense that the range is somehow related to the one given by the user. So, an intuitive choice may be just `np.arange(beta_min, beta_max, 1000)`.

alecandido: The above penalty term will also choose the minimum closest to zero, if at least one minimum is in the chosen window.

In order not to alter the minima positions, it has to be locally flat around them. An easy choice is to make the penalty discontinuous at the positions of maxima.

If your window was positioned around a maximum, that was already a problematic choice on its own. But the only thing happening here is that one of the two branches will be preferred.

The only truly problematic situation is the following:

Image

(same script of the above, just `x` in the range `[-8, -2]`)

In which one minimum is found, but a non-minimum value is preferred on a different branch.
However, this could be tuned by lowering arbitrarily the weight carried by the penalty term.
Indeed, I scaled it by `amplitude`, but if you scale by `amplitude/10` or `/100` you can control the maximum amount of error made. And the minimization will work for any amount > 0.

alecandido (commented):
Some proposals for `drag.py`. I'm now moving to `drag_simple.py`

src/qibocal/protocols/drag/drag_simple.py
```diff
@@ -43,6 +53,9 @@ class DragTuningSimpleData(DragTuningData):
     """DragTuningSimple acquisition outputs."""
 
     def __getitem__(self, key: tuple[QubitId, str]):
+        # TODO: the input type breaks Liskov so probably the base class AbstractData
+        # should be reconsidered.
```
alecandido, line 57: What is `Liskov`?

src/qibocal/protocols/drag/drag_simple.py
```diff
@@ -67,7 +83,9 @@ def _acquisition(
     """
 
     data = DragTuningSimpleData()
-    beta_range = np.arange(params.beta_start, params.beta_end, params.beta_step)
+    beta_range = np.arange(
+        params.beta_start, params.beta_end, params.beta_step
+    ).tolist()
```
alecandido: You should be able to use the same `np.arange(*params.beta_range)` as in the other protocol

alecandido (commented):
Ok, no fit touched. This was easier :P

src/qibocal/protocols/drag/drag.py
RoyStegeman: For beta perhaps the number of points can be more relevant, but having said that we will indeed review the interface anyway so I can also remove the comment? 

Where would you quote numpy's statment? In the docstring of `beta`?

alecandido: If you wish, right now we could even do that. But, eventually, I would just put in the documentation for ranges inputs. Which will be a guide we should add just in Sphinx, and link to the type docstring.

We should do this while addressing #1300. Let me add a note there

src/qibocal/protocols/drag/drag.py
alecandido: I turned this into #1494, if you wish we could resolve here

src/qibocal/protocols/drag/drag.py
alecandido: Note appended to the OP of #1300. If you wish, you could remove the comment and resolve

src/qibocal/protocols/drag/drag_simple.py
RoyStegeman: It's a principle in OOP that I learned about recently 😄 . It's one of the SOLID principles, that tbf are quite obvious, but known enough that it has its own [wiki page](https://en.wikipedia.org/wiki/SOLID). 

The Liskov substitution principle just says that subclasses should not break the behaviour of the parent class, which this one does. `DragTuningSimpleData` is a child class of `AbstractData`, but `__getitem__` here has a different input type from `__getitem__` in the parent class.

src/qibocal/protocols/drag/drag.py
RoyStegeman: Yes arguably it's a bit much, but I was aware it would be revised later on so I'd say let's leave it like this for now.

src/qibocal/protocols/drag/drag.py
RoyStegeman: Numpy returns a numpy float while the math variant returns an integer, and since it should be in integer that makes more sense to me.

src/qibocal/protocols/drag/drag.py
RoyStegeman: This suggestion has the benefit of getting rid of the `if-else` statements, but at the expense of 1) taking the preferred beta value from an `argmin` instead of computing from the fitted parameters, and 2) the issue you point out and having to introduce a hyperparameter which still doesn't fully remove the issue it. 

1\) I could accept although even on its own I'm not sure its worth it, but 2) seems a cure worse than the disease. 

While we're discussing this, I'm not fully sure if there is a reason to choose one minimum over the other. I selected the smallest absolute value since it has the smallest impact on the waveforms and thus is presumably more favorable, at least because it means fewer flipping-drag iterations.

src/qibocal/protocols/drag/drag.py
RoyStegeman: I undid also the normalization of beta. Since I did not observe the fit failing, @lballerio suggested only removing the bounds on the phase.

alecandido: Exactly. And you can estimate the amplitude and offset initial values using median and quantiles.

Btw, we should better deduplicate the tools for sinusoidal fits... this could be an issue.

RoyStegeman: > And you can estimate the amplitude and offset initial values using median and quantiles.

Since I had not seen the fits failing, @lballerio suggested limiting the changes to this. Indeed if we want to use it for all sinusoidal fits, it has to be taken from a common place and not reimplemented many times.

src/qibocal/protocols/drag/drag.py
alecandido: I believe the problem of the two hyperparameters (the number of points and penalty scaling) it is not really a problem: they are just resulting in a precision loss. But the numerical precision is not unlimited anyhow, and they can easily be brought in a regime where they are not the limiting factor (the accuracy of the data, and consequently the fit, will be much more impactful).
There will also be a tiny performance price, since you could determine the same without sampling the function. But it is NumPy, it will take little to nothing - I would not worry too much about this kind of performances (until they are reasonable).

The clear benefit of the penalty term is that it is much easier to interpret (because it is much easier to read/write), and makes mathematically explicit the choice that we are making.

src/qibocal/protocols/drag/drag.py
alecandido: `np.floor(...).astype(int)` should be good enough. The reason why NumPy is not doing the casting by default is that you are possibly loosing numerical accuracy. In this case, it is because you can represent very large numbers in floating notation, because of the exponent. And it is to exceed what can be represented by a fixed size integer.

I'm not exactly sure what `math.floor()` is doing, but Python integers are bigints, which makes them quite terrible for array operations (because they are array themselves, and you need dynamic memory - as in all Python objects), but they do not cause precision loss. But, as said, I'm not sure `math.floor()` is making use of that.

I know it is a bit pedantic, but I would be inclined to accept NumPy's reasoning. Also because of a general bias towards NumPy for math operations (I trust they are more careful on that than Python itself, whose main focus is not numerics). And because I like the operations to always be vectorization-ready (even when it is useless... just a full disclosure).

In any case, in practice they are pretty much equivalent here. So, if you have a strong preference for that, we could even keep `math`. I do not expect any actual problem.

src/qibocal/protocols/drag/drag_simple.py
alecandido: Ah, makes sense. I would have complained that it breaks the type hint. But this Liskov seems a more sophisticated quote xD

In any case, I perfectly agree. I would not address in this PR, but this `TODO` is to be kept

src/qibocal/protocols/drag/drag.py
RoyStegeman: The inputs to `math.floor` are still floats so they will be limited by the float representation and not benefit from the python long ints. Having said that, in this case we're not close to the 53(iirc?) bit precision limit provided by floats. 

I'm happy to standardize numpy for operations and explicitly call `astype(int)` though.

alecandido: What I mean is that you could easily represent as a float `1e100` (ok, as a double, it would overflow in single precision), but then there is no way you can do that as n integer: 4-bytes unsigned integers arrive to `~4e9` (i.e. `2**32`), and 8-bytes (same as double) arrive up to `~1.6e19` (i.e. `2**64`).
This is because in one case you have an explicit exponent, and you make holes in your representation.

In practice, unless you are implementing a prime numbers sieve, I guess this only matters when you start using small numbers' layouts, e.g. on a single byte, to aggressively save space. Which is truly not our case.

src/qibocal/protocols/drag/drag.py
alecandido: Indeed: maybe just open an issue about this, and that's it for this PR.

RoyStegeman: https://github.com/qiboteam/qibocal/issues/1495

src/qibocal/protocols/drag/drag.py
RoyStegeman: I think this is resolved?

alecandido: I think so as well

src/qibocal/protocols/drag/drag.py
RoyStegeman: I believe we can (in fact, should) have only one hyperparameter since the sampling density also determines the resolution with which we can distinguish two minima of the sinusoid. I implemented it in https://github.com/qiboteam/qibocal/pull/1434/commits/d2b7cf94ec75a00024111044c8e77d42445024f9

src/qibocal/protocols/drag/drag.py
```diff
@@ -229,33 +229,30 @@ def _fit(data: DragTuningData) -> DragTuningResults:
...
+            _offset, amplitude, period, phase = popt
+
+            # Evaluate drag_fit on a dense grid and select the minimum closest to
+            # beta=0. A penalty term breaks equality between equally deep minima of the
+            # sinusoidal fit and favours those points closest to beta=0.
+            #
+            # The penalty factor has to exceed the variation in drag_fit between
+            # adjacent grid points near a minimum, which from a Taylor expansion is
+            # (2*pi/period* beta_step)**2. The O(beta_step**4) term is negative so the
+            # quadratic bound is safe.
```
alecandido, line 241: Good point!

alecandido (commented):
Tests not passing (most likely just a precision issue)

src/qibocal/protocols/drag/drag.py
```diff
@@ -186,88 +209,91 @@ def _fit(data: DragTuningData) -> DragTuningResults:
...
+                beta_params.min(), beta_params.max(), sampling_points
+            )
+            beta_step = (beta_params.max() - beta_params.min()) / sampling_points
+            # beta_step is divided by two because the worst-case scenario is where a
+            # minimum is in the middle between two sampled points
+            penalty_factor = (
+                amplitude * 0.5 * (2 * np.pi / period * (beta_step / 2)) ** 2
+            )
+            penalty = penalty_factor * np.abs(
+                np.floor(beta_grid / period + phase / (2 * np.pi))
```
alecandido, line 253: Since it is not that trivial, it would be optimal to store your calculation somewhere, e.g. a notebook in `doc/dev/`.

In any case, we also have quite some incentive to move forward, so this should be definitely addressed in a different PR, since the explanation provided in the comments is enough to reconstruct the argument.

alecandido (approved):

RoyStegeman:
@qiboteam/maintainer this is ready for review.

scarrazza (approved):

PR #1443, opened 2026-04-14
update qibolab dependency to 0.2.13 for tests and 0.2.8 to run

RoyStegeman:
while the current qibocal should work for qibolab 0.2.8 or higher, the tests use a platform without bounds and for that reason require qibolab 0.2.13 or higher. 

In recent versions of numpy only 0-dimensional arrays can be converted to Python scalars, but it was attempted for 1-dimensional arrays containing only a single element.


scarrazza (approved):

pyproject.toml
```diff
@@ -33,6 +33,7 @@ cma = "^3.3.0"
 optional = true
 
 [tool.poetry.group.tests.dependencies]
+qibolab = "^0.2.13"
```
alecandido: Maybe we could leave here a comment, to motivate why a more recent version is needed. Just to remind ourselves.

alecandido (approved):
Tiny comment, but the PR is fine. Thanks!

PR #1444, opened 2026-04-15
Purcell fit

### LINKED ISSUE #1352: Purcell Filter spectroscopy fit
In this [paper from Wallraff's](https://arxiv.org/pdf/2307.07765) group they use a simple model for readout with Purcell filters given by

$$
    |S_{\rm{out, in}}|(\omega) = (A + k(\omega-\omega_0)) \left|\cos(\phi) - e^{i\phi}\frac{\kappa_p( -2i\Delta_r^{g/e})}{4J^2 + (\kappa_p - 2i\Delta_p)(- 2i\Delta_r^{g/e})}\right|,
$$

It would be nice to implement this as well in the resonator spectroscopy and allow a user to tell the fit function whcih model to use (or detect automatically if needed). This could be in addition to the current check of whether the resonator is a 3D cavity or a 2D one. 

#### Other relevant references

- Schoelkopf's group (Yale), 2010 https://arxiv.org/abs/1003.0142
    - it should be the original work introducing Purcell filters
- Martinis's group (Santa Barbara), 2014 https://arxiv.org/abs/1401.0257
- Nakamura's group (RIKEN), 2022 https://arxiv.org/pdf/2202.06202

Giuhcs:
In this (draft) PR I am adding the option to fit a Purcell model in a resonator spectroscopy, as requested in #1352.

For a readout resonator coupled to a Purcell filter, the expected output $S_{out,in}$ can be modelled following an input-output prescription, resulting in a spectrum profile described by (see [Wallraf](https://arxiv.org/pdf/2307.07765))

$$S_{out,in}(\omega)=\left(A-k(\omega-\omega_0)\right)\left[\cos\phi-e^{i\phi}\frac{-2i\Delta_r^{e,g}\kappa_p}{4J^2-2i\Delta_r^{e,g}\left(\kappa_p-2i\Delta_p\right)}\right].$$

At this point, the featured fit is performed (after a removal of the baseline by means of Asymmetric Least Square) in the amplitude part of the signal, following the above reference. From that, one should be able to extract estimations for $\kappa_p$, $J$, $\phi$, as well as the frequencies $\omega_p$ and $\omega_r^{e,g}$ for the Purcell filter and readout resonator, respectively.

### DIFF
--- a/src/qibocal/protocols/resonator_spectroscopies/resonator_spectroscopy.py
+++ b/src/qibocal/protocols/resonator_spectroscopies/resonator_spectroscopy.py
@@ -21,7 +21,7 @@ from ..utils import (
     lorentzian_fit,
     readout_frequency,
 )
-from .resonator_utils import s21, s21_fit, s21_spectroscopy_plot, spectroscopy_plot, purcell_fit, purcell_s_out_in
+from .resonator_utils import s21, s21_fit, s21_spectroscopy_plot, spectroscopy_plot, purcell_fit, purcell_s_out_in, purcell_spectroscopy_plot
 
 __all__ = ["resonator_spectroscopy", "ResonatorSpectroscopyData", "ResSpecType"]
 
@@ -73,7 +73,7 @@ FITS = {
         s21_spectroscopy_plot,
     ),
     "purcell": ResonatorSpectroscopyFit(
-        purcell_s_out_in,
+        lambda *params: np.abs(purcell_s_out_in(*params)),
         purcell_fit,
         chi2_reduced,
         lambda z: (z.signal, z.phase),
--- a/src/qibocal/protocols/resonator_spectroscopies/resonator_spectroscopy.py
+++ b/src/qibocal/protocols/resonator_spectroscopies/resonator_spectroscopy.py
@@ -21,7 +21,7 @@ from ..utils import (
     lorentzian_fit,
     readout_frequency,
 )
-from .resonator_utils import s21, s21_fit, s21_spectroscopy_plot, spectroscopy_plot
+from .resonator_utils import s21, s21_fit, s21_spectroscopy_plot, spectroscopy_plot, purcell_fit, purcell_s_out_in
 
 __all__ = ["resonator_spectroscopy", "ResonatorSpectroscopyData", "ResSpecType"]
 
@@ -72,6 +72,14 @@ FITS = {
         lambda z: (z.error_signal, z.error_phase),
         s21_spectroscopy_plot,
     ),
+    "purcell": ResonatorSpectroscopyFit(
+        purcell_s_out_in,
+        purcell_fit,
+        chi2_reduced,
+        lambda z: (z.signal, z.phase),
+        lambda z: (z.error_signal, z.error_phase),
+        purcell_spectroscopy_plot,
+    ),
 }
 """Dictionary of available fitting routines for ResonatorSpectroscopy."""
 
@@ -88,7 +96,7 @@ class ResonatorSpectroscopyParameters(Parameters):
     """Power regime (low or high). If low the readout frequency will be updated.
     If high both the readout frequency and the bare resonator frequency will be updated."""
     fit_function: str = "lorentzian"
-    """Routine function (lorentzian or s21) to fit data with a model."""
+    """Routine function (lorentzian, s21 or purcell) to fit data with a model."""
     phase_sign: bool = True
     """Several instruments have their convention about the sign of the phase. If True, the routine
     will apply a minus to the phase data."""
@@ -152,7 +160,6 @@ class ResonatorSpectroscopyData(Data):
     """Raw data acquired."""
     power_level: PowerLevel = PowerLevel.low
     """Power regime of the resonator."""
-
 
 def _acquisition(
     params: ResonatorSpectroscopyParameters,
@@ -232,7 +239,6 @@ def _acquisition(
             ),
         )
     return data
-
 
 def _fit(
     data: ResonatorSpectroscopyData,
--- a/src/qibocal/protocols/resonator_spectroscopies/resonator_spectroscopy.py
+++ b/src/qibocal/protocols/resonator_spectroscopies/resonator_spectroscopy.py
@@ -21,7 +21,15 @@ from ..utils import (
     lorentzian_fit,
     readout_frequency,
 )
-from .resonator_utils import s21, s21_fit, s21_spectroscopy_plot, spectroscopy_plot, purcell_fit, purcell_s_out_in, purcell_spectroscopy_plot
+from .resonator_utils import (
+    purcell_fit,
+    purcell_s_out_in,
+    purcell_spectroscopy_plot,
+    s21,
+    s21_fit,
+    s21_spectroscopy_plot,
+    spectroscopy_plot,
+)
 
 __all__ = ["resonator_spectroscopy", "ResonatorSpectroscopyData", "ResSpecType"]
 
@@ -161,6 +169,7 @@ class ResonatorSpectroscopyData(Data):
     power_level: PowerLevel = PowerLevel.low
     """Power regime of the resonator."""
 
+
 def _acquisition(
     params: ResonatorSpectroscopyParameters,
     platform: CalibrationPlatform,
@@ -239,6 +248,7 @@ def _acquisition(
             ),
         )
     return data
+
 
 def _fit(
     data: ResonatorSpectroscopyData,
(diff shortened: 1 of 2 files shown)


sorewachigauyo:
For some test data from our QPU, it looks like the fitting is failing to find both troughs
```  File "/mnt/home/nqch_paul/qibolab-qcs-2.6.4/qibocal/src/qibocal/protocols/resonator_spectroscopies/resonator_spectroscopy.py", line 276, in _fit
    fit_result = fit.fit(
                 ^^^^^^^^
  File "/mnt/home/nqch_paul/qibolab-qcs-2.6.4/qibocal/src/qibocal/protocols/resonator_spectroscopies/resonator_utils.py", line 1047, in purcell_fit
    w_l_guess, w_h_guess = frequencies[peaks]
    ^^^^^^^^^^^^^^^^^^^^
ValueError: not enough values to unpack (expected 2, got 1)
```
image

[purcell-fit-test.tar.gz](https://github.com/user-attachments/files/26792733/purcell-fit-test.tar.gz)

RoyStegeman:
@sorewachigauyo thanks for trying! We do not have any data from hardware yet ourselves to be able to test it, so we can use this data to improve the fitting algorithm

src/qibocal/protocols/resonator_spectroscopies/resonator_utils.py
```diff
@@ -958,3 +962,310 @@ def fit_punchout(filtered_x, filtered_y):
...
+        * (k_p * (-2j * (w - w_r)))
+        / (4 * J**2 + (k_p - 2j * (w - w_p)) * (-2j * (w - w_r)))
+    )
+
+
+def purcell_fit(
+    data: NDArray,
+    resonator_type=None,
+    fit=None,
+) -> tuple[NDArray, NDArray]:
```
sorewachigauyo, line 1006: The fit dumping is failing because the default JSON encoder doesn't serialize numpy arrays

src/qibocal/protocols/resonator_spectroscopies/resonator_utils.py
```diff
@@ -958,3 +962,310 @@ def fit_punchout(filtered_x, filtered_y):
...
+    fig_raw = make_subplots(
+        rows=1,
+        cols=2,
+        horizontal_spacing=0.1,
+        vertical_spacing=0.1,
+        specs=[
+            [{"rowspan": 2}, {}],
+            [None, {}],
+        ],
+    )
```
sorewachigauyo, line 1113: This is failing because `specs` is invalid for the 1 row x 2 column configuration

sorewachigauyo:
I hacked some fixes together to try to make it run

image

Here is some data

[purcell_test2.tar.gz](https://github.com/user-attachments/files/26975582/purcell_test2.tar.gz)

Giuhcs:
Hello @sorewachigauyo ! Thanks for the feedback! (and for the new data (: ) I'll keep working on it ;)

src/qibocal/protocols/resonator_spectroscopies/resonator_spectroscopy.py
```diff
@@ -72,6 +80,14 @@ class ResonatorSpectroscopyFit:
         lambda z: (z.error_signal, z.error_phase),
         s21_spectroscopy_plot,
     ),
+    "purcell": ResonatorSpectroscopyFit(
+        lambda *params: np.abs(purcell_s_out_in(*params)),
+        purcell_fit,
+        chi2_reduced,
+        lambda z: (z.signal, z.phase),
+        lambda z: (z.error_signal, z.error_phase),
+        purcell_spectroscopy_plot,
```
RoyStegeman, line 89: `values` is the tuple `(z.signal, z.phase)`, but `chi2` is `chi2_reduced`, which expects as input only the signal, so this is inconsistent. 

https://github.com/qiboteam/qibocal/blob/258cdd27e822a3afd59725c00dfe714a3489641c/src/qibocal/protocols/resonator_spectroscopies/resonator_spectroscopy.py#L290-L298

src/qibocal/protocols/resonator_spectroscopies/resonator_utils.py
```diff
@@ -958,3 +962,310 @@ def fit_punchout(filtered_x, filtered_y):
...
+    peaks, properties = find_peaks(
+        -(np.abs(z) - z_filt) / abs(min(np.abs(z) - z_filt)), height=0.0, prominence=0.2
+    )  # height filters peaks above 0, with relative height to its neighbours equal to "prominence"
+    rel_height = 0.5  # hyperparameter, using standard value for peak width determination at half height of the peak
+    widths = peak_widths(-(np.abs(z) - z_filt), peaks, rel_height=rel_height)
+
+    # selecting the two widest peaks and ordering then
+    widest_peaks_indices = sorted(
+        range(len(widths[0])), key=lambda i: widths[0][i], reverse=True
+    )[:2]
```
RoyStegeman, line 1052: Probably `np.argsort` is more readable.

src/qibocal/protocols/resonator_spectroscopies/resonator_utils.py
```diff
@@ -958,3 +962,310 @@ def fit_punchout(filtered_x, filtered_y):
...
+    widest_peaks_frequencies = [frequencies[peaks[i]] for i in widest_peaks_indices]
+    widest_peaks_widths = [widths[0][i] for i in widest_peaks_indices]
+    if widest_peaks_frequencies[0] > widest_peaks_frequencies[1]:
+        w_l_guess, w_h_guess = widest_peaks_frequencies[1], widest_peaks_frequencies[0]
+        k_l_guess, k_h_guess = widest_peaks_widths[1], widest_peaks_widths[0]
```
RoyStegeman, line 1057: `widest_peaks_widths` is the width in samples, but here and after you're putting it on the same footing as the w in Hz

src/qibocal/protocols/resonator_spectroscopies/resonator_utils.py
```diff
@@ -958,3 +962,310 @@ def fit_punchout(filtered_x, filtered_y):
...
+    rel_height = 0.5  # hyperparameter, using standard value for peak width determination at half height of the peak
+    widths = peak_widths(-(np.abs(z) - z_filt), peaks, rel_height=rel_height)
+
+    # selecting the two widest peaks and ordering then
+    widest_peaks_indices = sorted(
+        range(len(widths[0])), key=lambda i: widths[0][i], reverse=True
+    )[:2]
+    widest_peaks_frequencies = [frequencies[peaks[i]] for i in widest_peaks_indices]
+    widest_peaks_widths = [widths[0][i] for i in widest_peaks_indices]
+    if widest_peaks_frequencies[0] > widest_peaks_frequencies[1]:
```
RoyStegeman, line 1055: What if only a single peak is found? It would raise an error and fail the script instead of proceeding to generate the plot without the fit

src/qibocal/protocols/resonator_spectroscopies/resonator_utils.py
```diff
@@ -958,3 +962,310 @@ def fit_punchout(filtered_x, filtered_y):
...
+    # plotting fit if it succeeded
+    if fit is not None:
+        params = fit.fitted_parameters[qubit]
+        purcell_fitted = purcell_s_out_in(frequencies, *params)
+        # recovering baseline from the absolute signals
+        lamda = (
+            1e6  # hyperparameter, recommended from ALS paper to be between 1e6 and 1e9
+        )
+        p = 0.99  # hyperparameter
+        z_als = baseline_als(data=np.abs(signal), lamda=lamda, p=p)
```
RoyStegeman, line 1203: Instead of refitting, perhaps the baseline should be stored. Either way, here `p` is different from before so you are reconstructing a baseline different from the one that was removed.

src/qibocal/protocols/resonator_spectroscopies/resonator_utils.py
```diff
@@ -958,3 +962,310 @@ def fit_punchout(filtered_x, filtered_y):
...
+        eq1 = 0.5 * (w_r + w_p) + 0.5 * np.real(expr) - w_h_guess
+        eq2 = 0.5 * (w_r + w_p) - 0.5 * np.real(expr) - w_l_guess
+        eq3 = 0.5 * k_p - np.imag(expr) - k_h_guess
+        eq4 = 0.5 * k_p + np.imag(expr) - k_l_guess
+
+        return [eq1, eq2, eq3, eq4]
+
+    # solving and getting the intitial guesses for w_r, w_p, k_p and J
+    solution = fsolve(
+        equations, [600, 650, 1, 2]
```
RoyStegeman, line 1078: How were these values determined? Please write at least a comment explaining the choise

RoyStegeman (commented):
I'm just leaving a few comments after having a look. I also notice that the file is now almost 1300 lines, unless there is some obstacle to doing this I think it makes sense to separate them into their own files.

RoyStegeman:
@Giuhcs I'm closing this PR, assuming you don't have plans to finish it. If I'm wrong please just open it again.

PR #1449, opened 2026-04-18
JAZZ protocol

RoyStegeman:
This PR makes the following changes:

1. Implement JAZZ protocol in `zz_interaction` folder, together with ZZ-related classes (`ZZInteractionData`, `ZZInteractionParameters`, `ZZInteractionResults`) plus module `utils.py`, where functions for processing and plotting are defined.
2. Moving Ramsey-ZZ experiment from `ramsey` folder to `zz_interaction` one and refactor of this file according to the aforementioned classes and methods.
3. Adding measurement errors for all Ramsey experiments in `ramsey`.
4. Adding documentation for ZZInteraction experiments (both JAZZ and RamseyZZ)
5. Fixing Ramsey non-ZZ documentation
6. Fixing bug in Cross Resonance documentation (simply fixing headers)
7. Fixing typehints in `execute.py` and `task.py` files.

In the following link RamseyZZ and JAZZ protocols are shown for (0,1) qubit pair from QPU169 chip:
http://login.qrccluster.com:9000/arJHeNW8SFuqMwODfko9Pw==


sorewachigauyo:
Here is some test data from our QPU

We moved the coupler slightly to exaggerate the ZZ effects, but this also changed the dressed frequencies of the qubit pair so the XY pulses for the exaggeration are slightly off.

[jazz_test.tar.gz](https://github.com/user-attachments/files/27197251/jazz_test.tar.gz)
[ramsey_zz_exagerrated.tar.gz](https://github.com/user-attachments/files/27197255/ramsey_zz_exagerrated.tar.gz)
[jazz_test_exaggerated.tar.gz](https://github.com/user-attachments/files/27197257/jazz_test_exaggerated.tar.gz)

src/qibocal/protocols/two_qubit_interaction/jazz.py
```diff
@@ -0,0 +1,196 @@
...
+)
+
+
+@dataclass
+class JAZZParameters(Parameters):
+    """JAZZ runcard inputs."""
+
+    start_wait: float
+    end_wait: float
+    nsteps: int
```
lballerio: as far as I understood we want to transition from this syntax to taking a range tuple, so even though is ugly we could accept either one or the other.

src/qibocal/protocols/two_qubit_interaction/jazz.py
```diff
@@ -0,0 +1,196 @@
...
+    # X(pi)
+    for channel, pulse in q0_natives.RX():
+        sequence += [(channel, pulse)]
+        sequence += [(ro_channel, Delay(duration=pulse.duration))]
+        sequence += [(q1_drive_channel, Delay(duration=pulse.duration))]
+    # X(pi) on q1
+    for channel, pulse in q1_natives.RX():
+        sequence += [(channel, pulse)]
+        sequence += [(ro_channel, Delay(duration=pulse.duration))]
+        sequence += [(q0_drive_channel, Delay(duration=pulse.duration))]
```
lballerio: do we really have at the moment a list with more than 1 element for RX?

src/qibocal/protocols/two_qubit_interaction/jazz.py
```diff
@@ -0,0 +1,196 @@
...
+    # delay
+    sequence += [(q0_drive_channel, delay)]
+    sequence += [(ro_channel, delay)]
+    # Y(pi/2)
+    for channel, pulse in q0_natives.R(theta=np.pi / 2, phi=np.pi / 2):
+        sequence += [(channel, pulse)]
+        sequence += [(ro_channel, Delay(duration=pulse.duration))]
+    # MZ
+    sequence += [(ro_channel, ro_pulse)]
```
lballerio: maybe we can also measure, save and plot even the q1 state, might be informative

src/qibocal/protocols/two_qubit_interaction/jazz.py
```diff
@@ -0,0 +1,196 @@
...
+
+def _acquisition(
+    params: JAZZParameters,
+    platform: CalibrationPlatform,
+    targets: list[QubitPairId],
+) -> JAZZData:
+    """Data acquisition for JAZZ"""
+
+    if len(targets) != 1:
+        raise ValueError("simultaneous pair is not supported for JAZZ.")
```
lballerio: In Ramsey ZZ we decided to open the protocol to multiple `QubiPairId` and simply check if some qubits are repeated.
Better to choose something consistent for both, whatever the decision is.

lballerio (commented):
I agree that this protocol belongs to the two qubit interaction folder, maybe might be also reasonable to move ZZ experiment to the same folder and maybe renaming a bit (not necessary to do here, just a stream of thoughts).

RoyStegeman:
Since in the end this PR does much more than the title suggests, could you please update the title and also make the description more detailed? E.g. the file that is now `ramsey_zz.py` corresponds to the same experiment as the old `zz_coupling.py`. However, it was not simply moved, but more changes were made to the content as well. 

For next time: probably this should have been at least three PRs, one for introducing jazz, another for all the refactoring related to Ramsey zz, and a third for minor changes in otherwise unrelated files (typehints and the general error message). If changes that serve a common purpose are grouped together that makes reviewing a lot simpler.

lballerio:
Hi @RoyStegeman I agree, I also thought to divide in multiple smaller PRs but I thought it was too late.
However in general, despite the absolute number of modified files is not small for sure, the changes are not massive in files outside `zz_interaction` folder.
If you want I can split the content of this PR into smaller PRs.

RoyStegeman:
The changes may not be large, but as an example: it is very quick to copy a function from one file to another, but it can take much more time to reconstruct afterwards that this is what has happened if a reviewer has to look across diffs of, in this case, 22 files with as only explanation the (not so descriptive) commit messages. 

With this in mind, it is helpful if a reviewer can either go commit-by-commit, and each commit is a self-contained step with a clear commit message that makes sense in the overall plan (which in practice may involve merging commits and writing helpful messages just before asking for a review). Or alternatively, it is indeed helpful if this kind of separation is done by splitting into different PRs. 

If you are still able do do either of those things, that would help a lot.

src/qibocal/auto/execute.py
RoyStegeman, line 39: https://github.com/qiboteam/qibocal/pull/1624/changes#r3637902931

src/qibocal/auto/task.py
```diff
@@ -258,4 +258,4 @@ def update_platform(self, platform: Platform):
             try:
                 self.task.operation.update(self.results, platform, qubit)
             except KeyError:
-                log.warning(f"Skipping update of qubit {qubit} due to error in fit.")
+                log.warning(f"Skipping update of qubit {qubit}.")
```
RoyStegeman: It would be useful to know why it's being skipped.

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
+Residual ZZ Interaction
+=======================
+
+In superconducting transmon quantum processors, qubits are never perfectly isolated.
+Even when no two-qubit gate is being executed, neighboring qubits experience weak
+always-on interactions mediated by the coupling network. One of the most important
```
RoyStegeman:
```suggestion
always-on interactions. One of the most important
```
What is "the coupling network"? Is it jargon? If it means nothing specific but is just to say that qubits are not perfectly isolated and thus coupled in some way, then that was covered by saying there are always-on interactions.

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+undesired interactions is the **residual ZZ coupling**, which produces a state-dependent
+frequency shift.
+
+For two coupled qubits, restricted to the computational subspace
```
RoyStegeman:
```suggestion
For two coupled qubits restricted to the computational subspace
```

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+:math:`\{|00\rangle, |01\rangle, |10\rangle, |11\rangle\}`, the
+effective static Hamiltonian can generally be written as
```
RoyStegeman:
```suggestion
effective static Hamiltonian can be written as
```
What do you mean by generally?

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+
+.. math::
+
+   H/\hbar = -\frac{\omega_0}{2} ZI
+           - \frac{\omega_1}{2} IZ
+           - \frac{\zeta}{4} ZZ,
+
+where :math:`\zeta` is the residual ZZ interaction strength.
+
+where :math:`\omega_1, \omega_2` are the (dressed) qubit frequencies
```
RoyStegeman:
```suggestion
where :math:`\zeta` is the residual ZZ interaction strength, :math:`\omega_1, \omega_2` are the (dressed) qubit frequencies
```

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+and :math:`\zeta` is the residual ZZ interaction rate.
+In a simple three-level-per-qubit model, the ZZ rate can be
```
RoyStegeman:
```suggestion
In a simple qutrit model, the ZZ rate can be
```

RoyStegeman:
```suggestion
In a three-level-per-qubit model, the ZZ rate can be
```
What is simple about the model?

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+frequency shift.
+
+For two coupled qubits, restricted to the computational subspace
+:math:`\{|00\rangle, |01\rangle, |10\rangle, |11\rangle\}`, the
+effective static Hamiltonian can generally be written as
+
+.. math::
+
+   H/\hbar = -\frac{\omega_0}{2} ZI
+           - \frac{\omega_1}{2} IZ
```
RoyStegeman:
```suggestion
   H/\hbar = -\frac{\omega_1}{2} ZI
           - \frac{\omega_2}{2} IZ
```
Everywhere else you're using qubits 1 and 2.

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+In a simple three-level-per-qubit model, the ZZ rate can be
+approximated in the **dispersive regime** as
+
+.. math::
+
+   \zeta_{ZZ} \approx \frac{2 g^2 \alpha_1 \alpha_2}
+   {(\Delta_{12} - \alpha_1)(\Delta_{12} + \alpha_2)}
+
+where :math:`\Delta_{12} = \omega_1 - \omega_2` is the qubit-qubit
+detuning and :math:`\alpha_{1,2}` are the qubit anharmonicities and :math:`g` is the
```
RoyStegeman:
```suggestion
where :math:`\Delta_{12} = \omega_1 - \omega_2` is the detuning between qubits 1 and 2, :math:`\alpha_{1,2}` are the qubit anharmonicities, and :math:`g` is the
```

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+**coupling strength**, which depends on the physical capacitance of the circuits.
```
RoyStegeman:
```suggestion
**coupling strength**. The latter is determined by the physical capacitance of the circuits.
```

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+
+The ZZ term shifts the transition frequency of one qubit depending on the state of
+its neighbor:
+
+.. math::
+
+   \omega_\mathrm{target}^{(|1\rangle)}
+   - \omega_\mathrm{target}^{(|0\rangle)}
+   = \zeta.
```
RoyStegeman: You'll need to introduce this notation. Or leave it out - arguably it doesn't add much that wasn't clear from the sentence above.

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+
+Consequences include:
+
+* coherent phase accumulation during idle periods;
+* reduced gate fidelity;
+* increased crosstalk;
```
RoyStegeman, line 41:
```suggestion
```
I guess this one is by definition?

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+The ZZ term shifts the transition frequency of one qubit depending on the state of
+its neighbor:
+
+.. math::
+
+   \omega_\mathrm{target}^{(|1\rangle)}
+   - \omega_\mathrm{target}^{(|0\rangle)}
+   = \zeta.
+
+Consequences include:
```
RoyStegeman: Perhaps this should be phrased a bit nicer and something along the lines of "ZZ coupling is a relevant effect that we need to account for durning calibration because neglecting to do so may result in .... "

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+
+Two common characterization experiments are JAZZ and Ramsey ZZ.
+Both experiments described below exploit the same basic idea: put one
+qubit (the "spectator" or "control") in :math:`|0\rangle` or
+:math:`|1\rangle`, and use a Ramsey-type measurement on the other
+qubit (the "target") to detect the resulting frequency shift.
+
+
+Ramsey ZZ (Conditional Ramsey)
+------------------------------
```
RoyStegeman: Probably "Ramsey ZZ (Conditional Ramsey)" and "JAZZ (Joint Amplification of ZZ)" should be subsection of "Estimating Residual ZZ".

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+
+Given that through a Ramsey experiment we can measure carefully the frequency of the qubit,
```
RoyStegeman:
```suggestion
Given that through a Ramsey experiment we can carefully measure the frequency of the qubit,
```

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+
+
+From :ref:`ramsey` we know that both signals oscillates according to
+
+.. math::
+
+   P_{|1\rangle}(t) = \frac{1}{2}\Big[1 + e^{-t/T_2^{*}}
+   \cos\big(\Omega{c} \, t + \phi_0\big)\Big]
+
+where :math:`\delta_c` is the target frequenncy conditioned on
```
RoyStegeman:
```suggestion
where :math:`\delta_c` is the target frequency conditioned on
```

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+
+* coherent phase accumulation during idle periods;
+* reduced gate fidelity;
+* increased crosstalk;
+* limitations on simultaneous gate execution.
+
+Estimating Residual ZZ
+----------------------
+
+Two common characterization experiments are JAZZ and Ramsey ZZ.
```
RoyStegeman: citations for JAZZ and Ramsey ZZ

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+Both experiments described below exploit the same basic idea: put one
```
RoyStegeman:
```suggestion
Both experiments are described below and exploit the same basic idea: put one
```

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+qubit (the "spectator" or "control") in :math:`|0\rangle` or
+:math:`|1\rangle`, and use a Ramsey-type measurement on the other
+qubit (the "target") to detect the resulting frequency shift.
```
RoyStegeman:
```suggestion
qubit (the "target") to detect the state-dependent frequency shift.
```

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+
+
+Ramsey ZZ (Conditional Ramsey)
+------------------------------
+
+Given that through a Ramsey experiment we can measure carefully the frequency of the qubit,
+by repeating it while putting one of the neighbor qubits in
+state :math:`\ket{1}` we can have an estimate on the ZZ coupling :math:`\zeta`.
+For a more detailed discussion about Ramsey experiment see :ref:`ramsey`.
```
RoyStegeman:
```suggestion
For a more detailed discussion of the Ramsey experiment see :ref:`ramsey`.
```

doc/source/protocols/zz_interaction/zz_interaction.rst
```diff
@@ -0,0 +1,157 @@
...
+:math:`|1\rangle`, and use a Ramsey-type measurement on the other
+qubit (the "target") to detect the resulting frequency shift.
+
+
+Ramsey ZZ (Conditional Ramsey)
+------------------------------
+
+Given that through a Ramsey experiment we can measure carefully the frequency of the qubit,
+by repeating it while putting one of the neighbor qubits in
+state :math:`\ket{1}` we can have an estimate on the ZZ coupling :math:`\zeta`.
```
RoyStegeman:
```suggestion
we can have estimate the ZZ coupling :math:`\zeta` by repeating the experiment while preparing one of the neighboring qubits in state :math:`\ket{1}`.
```

doc/source/protocols/zz_interaction/zz_interaction.rst
RoyStegeman, line 1: I think it would clarify a lot if we could show the pulse-sequence for the experiments.

doc/source/protocols/index.rst
```diff
@@ -45,4 +45,5 @@ In this section we introduce the basics of all protocols supported by ``qibocal`
     chsh
     twpa_calibration/twpa
     cross_resonance/cross_resonance
+    zz_interaction/zz_interaction
```
RoyStegeman: Since the residual ZZ interactions can be measured directly after calibrating the single qubit gates, I think it makes sense to put this section a bit higher to reflect its position in the calibration steps.

src/qibocal/update.py
```diff
@@ -255,7 +256,10 @@ def readout_coupling(g: float, platform: Platform, qubit: QubitId):
 
 
 def pair_coupling(g: list[float], platform: Platform, pair: QubitPairId):
-    platform.calibration.two_qubits[pair].coupling = g
+    if pair in platform.calibration.two_qubits:
+        platform.calibration.two_qubits[pair].coupling = g
+    else:
+        platform.calibration.two_qubits |= {pair: TwoQubitCalibration(coupling=g)}
```
RoyStegeman:
```suggestion
    platform.calibration.two_qubits.setdefault(pair, TwoQubitCalibration()).coupling = g
```

src/qibocal/protocols/ramsey/processing.py
```diff
@@ -22,7 +22,7 @@
 
 from .acquisition import RamseyResults
 
-MAXIMUM_FIT_POINTS = 1_000
+MAXIMUM_FIT_POINTS = 500
 """maximum number of points to use when plotting fit results."""
```
RoyStegeman: This is a strange variable name/description, since it is the exact number of plotting points, not the max. 

Also, this is just about aestethics of 1D plots, which I don't think is a choice that should be made unique to Ramsey. Either it should be a parameter we can set globally and at some point we may want a global style sheet, but for the time being I think the number of points for which the plotted function is evaluated can just be part of the fitting function. I.e. do `linspace(start,stop,500)` instead of `linspace(start,stop,MAXIMUM_FIT_POINTS)`

src/qibocal/protocols/ramsey/processing.py
```diff
@@ -198,15 +203,27 @@ def signal_plot(
...
     yaxis_title: str,
+    error_bar: NDArray | None = None,
 ) -> tuple[list[go.Figure], str]:
     """Create a signal scatter plot and optional fit report."""
 
+    error_y = (
+        dict(
+            type="data",
+            array=error_bar,
+            visible=True,
```
RoyStegeman:
```suggestion
```

This is already the default and I don't think it clarifies anything by explicitly passing it.

src/qibocal/protocols/zz_interaction/jazz.py
```diff
@@ -0,0 +1,307 @@
...
+    ZZInteractionResults,
+    ZZIntType,
+    coupling_strength,
+    create_report_table,
+    signal_plot,
+    zz_update,
+)
+
+DAMPED_CONSTANT = 1.5
+"""See :const:`rabi.utils.QUANTILE_CONSTANT` for details."""
```
RoyStegeman: Why look at `QUANTILE_CONSTANT`? Are they the same thing?

src/qibocal/protocols/zz_interaction/jazz.py
```diff
@@ -0,0 +1,307 @@
...
+    ZZInteractionParameters,
+    ZZInteractionResults,
+    ZZIntType,
+    coupling_strength,
+    create_report_table,
+    signal_plot,
+    zz_update,
+)
+
+DAMPED_CONSTANT = 1.5
```
RoyStegeman: `DAMPED_CONSTANT` is also defined for the regular Ramsey protocol. Are they physically the same thing or is their agreement more coincidental and could they also be independently set?

src/qibocal/protocols/zz_interaction/jazz.py
```diff
@@ -0,0 +1,307 @@
...
+            # here we compute coupling as a frequency
+            coupling[pair] = coupling_strength(
+                omega1=data.qubit_freqs[target],
+                omega2=data.qubit_freqs[spectator],
+                anharmonicity1=data.anharmonicity[target],
+                anharmonicity2=data.anharmonicity[spectator],
+                zz=zz_zeta,
+            )
+
+        except Exception as e:
```
RoyStegeman:
```suggestion
        except RuntimeError as e:
```

src/qibocal/protocols/zz_interaction/utils.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+@dataclass
+class ZZInteractionResults(Results):
+    """Container for ZZ-interaction experiments results."""
+
+    zz: dict[QubitPairId, list[float]] = field(default_factory=dict)
+    """Estimated ZZ coupling of a qubit pair."""
+    coupling: dict[QubitPairId, list[float]] = field(default_factory=dict)
+    """Estimated coupling strenght of a qubit pair."""
```
RoyStegeman:
```suggestion
    """Estimated coupling strength of a qubit pair."""
```

src/qibocal/protocols/zz_interaction/utils.py
```diff
@@ -0,0 +1,198 @@
...
+    signal_style = dict(color="red" if label == "I" else "blue")
+
+    target_scatters.append(
+        go.Scatter(
+            x=delays,
+            y=signal.targ_prob,
+            error_y=dict(
+                type="data",
+                array=signal.targ_error,
+                visible=True,
```
RoyStegeman:
```suggestion
```
And below as well

src/qibocal/protocols/zz_interaction/utils.py
```diff
@@ -0,0 +1,198 @@
...
+    targ, spect = target
+    fitting_report = table_html(
+        table_dict(
+            [targ] * 2,
+            [
+                f"ZZ  with {targ} [Hz]",
+                f"Coupling with {spect} [Hz]",
+            ],
+            [
+                fit.zz[target] if target in fit.zz else (0, 0),
```
RoyStegeman:
```suggestion
                fit.zz.get(target, (0, 0)),
```

src/qibocal/protocols/zz_interaction/utils.py
```diff
@@ -0,0 +1,198 @@
...
+                fit.coupling[target] if target in fit.coupling else (0, 0),
```
RoyStegeman:
```suggestion
                                fit.zz[target] if target in fit.zz else (0, 0),
```

RoyStegeman (commented):
- > 4. Adding documentation for ZZInteraction experiments (both JAZZ and RamseyZZ)

	I think the documentation is awkwardly written and should be revised. At some point I stopped leaving comments with suggestions for change, so no comment does not mean nothing can be improved in that part of the docs. You can use my examples as a guide if you agree that they make it smoother to read. The descriptions of the experiment in docstrings or documentation are quite short and without a pulse diagram or citation they are too limited. 

- > 7. Fixing typehints in `execute.py`, `task.py` and `calibration.py` files.

	`calibration.py` was not touched. Did you forget to push changes?

- Please also add tests. 

- Could you share an example report for each protocol?

- The tools we use through pre-commit wrap lines after column 88. While not enforced for docstrings or inline comments, in general (just covering some exception that may exist, such as long hyperlinks) it makes sense to respect the same wrapping.

src/qibocal/protocols/zz_interaction/jazz.py
lballerio: Yes they are the same thing, we can also think of moving this variable in an outer folder, but I would think carefully because I am afraid it will disappear among all the millions of functions/variables that already exists

lballerio:
> calibration.py was not touched. Did you forget to push changes?

no you're right, my mistake, was in another PR i think

src/qibocal/auto/execute.py
lballerio: addressed in 8228f8e

src/qibocal/auto/task.py
lballerio: addressed in 8228f8e

src/qibocal/update.py
lballerio: addressed in 8228f8e

src/qibocal/protocols/ramsey/processing.py
lballerio: addressed in 8228f8e

src/qibocal/protocols/ramsey/processing.py
lballerio: addressed in 8228f8e

src/qibocal/protocols/zz_interaction/jazz.py
lballerio: addressed in 8228f8e

src/qibocal/protocols/zz_interaction/jazz.py
lballerio: addressed in 8228f8e

src/qibocal/protocols/zz_interaction/utils.py
lballerio: addressed in 8228f8e

src/qibocal/protocols/zz_interaction/utils.py
lballerio: addressed in 8228f8e

src/qibocal/protocols/zz_interaction/utils.py
lballerio: addressed in 8228f8e

src/qibocal/protocols/zz_interaction/utils.py
lballerio: addressed in 8228f8e

src/qibocal/protocols/zz_interaction/jazz.py
RoyStegeman: If they are hyperparameters, for me it's fine to repeat it, you may for whatever reason decide to make it a slightly different value in one protocol than in another. But if they are the same thing and there is no debate, then putting it in a shared folder is arguably the best thing. Either way, that is not super important, but I do think it makes sense to indicate that they are the same by giving them the same name and explaining in a short comment why it is repeated. E.g. calling all of them the same, but append the protocol name QUANTILE_CONSTANT_JAZZ

RoyStegeman: [This other comment](https://github.com/qiboteam/qibocal/pull/1449#discussion_r3643898131) is related, so I'll resolve it and we can discuss both in one place. By making JAZZ and Ramsey share the same instance of DAMPED_CONSTANT while elsewhere you are using QUANTILE_CONSTANT, but then still referencing QUANTILE_CONSTANT as an explanation for DAMPED_CONSTANT, it becomes very unclear how they are related. Based on your answer here I understand it now, but without that I would be wondering: are they the same because they are the same physical object, or is it a coincidence?  Since JAZZ and Ramsey share the same object, is it the same physical object for those two but a different one for Rabi? ect...

RoyStegeman:
I cannot approve since I'm the one who opened it, so you'll have to do that yourself. I left one comment unresolved, but I'll leave it up to you to address it and (unless you decide to change more) don't have to see this PR again before it's merged. 

I see you addressed all my points except the documentation. But as we discussed, since we have specific plans to prioritize documentation in the very short term, I will not insist here. 

The only comment I did not resolve is regarding the confusion about `QUANTILE_CONSTANT` vs `DAMPING_CONSTANT`. My preferred solution would be to either put it in a single location as also mentioned, but I understand why you don't like it, so as an alternative I think doing QUANTILE_CONSTANT_[protocol name] for each protocol independently with a small comment for why they are repeated while having the long explanation in only one of them, is also acceptable. But having two names for the same thing, and then share in some cases but no in other is just confusing.

lballerio:
OK I see your point and I think is a reasonable solution.
Then at some point we should, as we said, move it outside in a more general function, but it will take time to make some order.

lballerio (approved):

scarrazza (approved):

PR #1459, opened 2026-04-28
RB: make inner loop over circuit-depth for better packing of batches

RoyStegeman:
Since doing the optimal packing for many sequences into batches is computationally inefficient, qibolab relies on approximate methods. For now, the best result I obtained is by doing a small optimization in qibolab, on top of an inner loop over depths. This is mostly empirical, and at some point we may change qibolab to perform it's own ordering first, but since this works best after some empirical tests, I'd like to make this change in qibocal. 

Still to be tested on qpu.

### DIFF
--- a/src/qibocal/protocols/randomized_benchmarking/utils.py
+++ b/src/qibocal/protocols/randomized_benchmarking/utils.py
@@ -382,9 +382,9 @@ def _generate_indexed_circuits(
 
     inv_file = getattr(params, "file_inv", None)
 
-    for depth in params.depths:
-        for target in targets:
-            for iteration in range(params.niter):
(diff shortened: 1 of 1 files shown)


alecandido (approved):

scarrazza:
@RoyStegeman did you test it on qpu?

scarrazza (approved):

RoyStegeman:
No, the calibration is still ongoing so we don't have a platform to test yet. I did this change locally before while working on the main RB PR and it worked fine. The order of the loops should not matter since the UUID is used to map the circuit to the result, but still I'd like to run it as a sanity check before merging.

PR #1463, opened 2026-04-30
remove unsupported `hardware_average` for resonator spectroscopy from docs

### LINKED ISSUE #1469: remove unsupported `hardware_average` for resonator spectroscopy
The option to do hardware average for resonator spectroscopy is not supported but the flag exist and it is still mentioned in the docs.

RoyStegeman:
The option to do hardware average for resonator spectroscopy is not supported but the flag exist and it is still mentioned in the docs. This PR removes any references to it. 

In most cases doing hardware average will be exactly what we want, possibly to the point that we don't even want to offer downloading the full list of shots as an option. Another issue will have to be opened, either about replacing SINGLESHOT with CYCLIC, or at least one that adds the option for CYCLIC and that should be the default. Personally I vote for dropping SINGLESHOT, but what do you think?


alecandido (approved):

alecandido:
> Personally I vote for dropping SINGLESHOT, but what do you think?

I would also agree to remove the `SINGLESHOT` option for this protocol. It's pretty much useless and expensive, it would only be there to deceive the user...

scarrazza (approved):

scarrazza:
> > Personally I vote for dropping SINGLESHOT, but what do you think?
> 
> I would also agree to remove the `SINGLESHOT` option for this protocol. It's pretty much useless and expensive, it would only be there to deceive the user...

me too

PR #1466, opened 2026-04-30
allow to manually select center of resonator freq and bias spectroscopy

RoyStegeman:
While for re-calibration it makes sense to take the default value from the calibration.json. It should also be possible to define the sweeping window without modifying the calibration.json. 

I'm not sure if this is the way I want to solve it, so therefore this PR Is a draft.


alecandido:
Would you consider starting implementing this, #1300, just for this protocol?

If you do not agree on the proposed interface, we can discuss about it.

RoyStegeman:
Closed in favor of #1595

Issue #1464, opened 2026-04-30
Error in estimating qubit coupling from Ramsey ZZ experiment


lballerio:
In the Ramsey ZZ experiment we first determine the ZZ interaction term of the qubit pair by measuring the difference in the signal frequencies, then we can exploit this equation:

$$
\zeta \approx 2g^2\Big(\frac{1}{\Delta - \eta_2} - \frac{1}{\Delta + \eta_1}\Big)
$$

where $\eta_i$ is the anharmonicity of the i-th qubit and $\Delta = \omega_1 - \omega_2$.

> [!NOTE]
> Since ZZ coupling is symmetric, we see that the aforementioned equation is symmetric to $1\longleftrightarrow 2$ exchange.

However in Qibocal $|\Delta|$ is used, which then breaks this symmetry.

PR #1467, opened 2026-04-30
Fixing Ramsey ZZ coupling strength estimation

lballerio:
Linked to isse #1464.

We see that in the simulated-qpu167 if we perform RamseyZZ on a single pair for both control-target combinations, the previous version computes an asymmetric ocupling strength g and also a signal-insensitive zz:

http://login.qrccluster.com:9000/9d2iORfMSDqkDzJUmP5x9g==

with the fixed routine we find the correct strength in both cases:
http://login.qrccluster.com:9000/lnFIyU95QBWfXeRPve1XYw==

also tested on real hardware, namely qpu168, we find decent agreement between the two cases:
http://login.qrccluster.com:9000/DhL2MZlZR2ydWs18lCttXg==
repeated with different detuning:
http://login.qrccluster.com:9000/dmS4kqB0RViiDqe3rljzew==


src/qibocal/protocols/ramsey/ramsey_zz.py
```diff
@@ -23,21 +21,59 @@
 
 __all__ = ["ramsey_zz"]
 
+EPS = 1  # Hz
+
+
+class AnharmError(Exception):
```
alecandido:
```suggestion
class AnharmonicityError(Exception):
```

Since this could get raised, and thus flow outside this file, better to give it an explicit name.

In Qibocal we are not yet paying attention to this, but uncaught exceptions are part of the public interface of the package.

src/qibocal/protocols/ramsey/zz_coupling.py
```diff
@@ -23,21 +21,59 @@
 
 __all__ = ["ramsey_zz"]
 
+EPS = 1  # Hz
```
alecandido, line 40: Add docstring

src/qibocal/protocols/ramsey/ramsey_zz.py
```diff
@@ -78,23 +114,31 @@ def _acquisition(
...
+            qubit: platform.calibration.single_qubits[qubit].qubit.anharmonicity
+            for qubit in targets
+        }
+        spectator_anharmonicity = platform.calibration.single_qubits[
+            params.spectator_qubit
+        ].qubit.anharmonicity
+    except KeyError:
+        raise AnharmError(
+            "One or more anharmonicities are not calibrated yet, calibrate e-f transition for all qubits."
+        )
```
alecandido: I would raise this during the fitting, rather than in here. You can still acquire data even without the anharmonicities calibrated.

The only reason why you see them here is that in the fitting function there is no access to the `Platform` object, so the values are set into the data object to pass down to it.
Not a great design, but I would not change it here.

My advice is to just set the `..._anharmocit*` variables to `None`, if the `KeyError` is caught. Then, you could raise during the fit, or just skip it. And possibly, still plot the raw data

lballerio: I agree about the `Platform` comment, is not the best but ok, neither the worst, and I guess it is actually part of the reason why I put that `try-except` here. Also I thought there was no point in even starting the experiment without having all the necessary parameters already calibrated, which I think is also not super bad.

alecandido: The only parameters required are those for the acquisition.

Indeed, if we are missing parameters which are then required for the extraction of the relevant physical quantities is not the best. But at the same time is not a good reason to just fail

lballerio: ok.
Not super convinced about that but I'll stay with you idea.
Maybe the only 'benefit' is that still you have the data but fit fails, then you calibrate e-f transitions (then also anharmonicities) and after that you go back to fit.

alecandido: > Not super convinced about that but I'll stay with you idea.

No need, counterarguments are good: we need to converge out of many different needs. You may have something in mind which I'm overlooking

> Maybe the only 'benefit' is that still you have the data but fit fails, then you calibrate e-f transitions (then also anharmonicities) and after that you go back to fit.

You should be able to measure a ZZ even without the anharmonicities, since they are just a difference in the measured frequencies (no further post-processing).
The only missing bit would be the coupling strength, which is the reason why anharmonicities are requested.

Isn't it?

src/qibocal/protocols/ramsey/acquisition.py
```diff
@@ -97,57 +97,84 @@ def waits(self) -> npt.NDArray:
...
-                f"Cannot run Ramsey experiment on qubit {target_qubit} if it is already in Ramsey sequence."
-            )
-            natives = platform.natives.single_qubit[target_qubit]
-            sequence += natives.RX()
+        full_sequence += qubit_seq
+        full_ro_sequence += qubit_ro_seq
+        full_delays += qubit_delays
 
-    return sequence, delays + flux_pulses
+    return full_sequence, full_ro_sequence, full_delays
```
alecandido, line 176: I may be missing something, but you are now returning the readout sequence separately, but never using it separately.

Why not returning just `full_sequence + full_ro_sequence, full_delays`?

src/qibocal/protocols/ramsey/acquisition.py
```diff
@@ -97,57 +97,84 @@ def waits(self) -> npt.NDArray:
...
-) -> tuple[PulseSequence, list[PulseLike]]:
+) -> tuple[PulseSequence, PulseSequence, list[Pulse | Delay]]:
     """Pulse sequence used in Ramsey (detuned) experiments.
 
     The pulse sequence is the following:
 
     RX90 -- wait -- RX90 -- MZ
     """
-    delays = 2 * len(targets) * [Delay(duration=wait)]
+    delays = [Delay(duration=wait), Delay(duration=wait)]
```
alecandido: If you want them to be identical, but with different IDs, use list comprehension
```suggestion
    delays = [Delay(duration=wait) for _ in range(2)]
```

The difference is that the code for the element is essentially a function, and gets called over and over. Instead, if you do `[x()] * 2`, you first construct a one element list, and then duplicate its elements (essentially, the pointer)

src/qibocal/protocols/ramsey/acquisition.py
```diff
@@ -97,57 +97,84 @@ def waits(self) -> npt.NDArray:
...
-        if flux_pulse_amplitude is not None
-        else []
-    )
+    # TODO: move flux lines outside this function
+    if flux_pulse_amplitude is not None:
+        flux_pulse = [
+            Pulse(duration=0, amplitude=flux_pulse_amplitude, envelope=Rectangular())
+        ]
+    else:
+        flux_pulse = []
```
alecandido:
```suggestion
    flux_pulse = [
        Pulse(duration=0, amplitude=flux_pulse_amplitude, envelope=Rectangular())
    ] if flux_pulse_amplitude is not None else []
```
as usual :)

src/qibocal/protocols/ramsey/zz_coupling.py
```diff
@@ -27,103 +37,286 @@
...
+"""we add 1Hz when computing Delta frequency between the two qubit frequencies in order to avoid numerical error."""
+
+
+class AnharmonicityError(Exception):
+    def __init__(self, *args):
+        super().__init__(*args)
+
+
+def compute_coupling_strength(
```
alecandido:
```suggestion
def coupling_strength(
```
as usual

src/qibocal/protocols/ramsey/zz_coupling.py
```diff
@@ -27,103 +37,286 @@
...
+    delta_q = omega1 - omega2
+    xi = 2 g**2 (1 / (delta_q - alpha_2) - 1 / (delta_q + alpha_1))
+    where delta_q is the difference in frequency and alpha_i is the anharmonicity
+    """
+
+    if anharmonicity1 == 0 or anharmonicity2 == 0:
+        raise AnharmonicityError
+
+    # adding an eps to avoid numerical issues
+    delta_qubit_freq = omega1 - omega2 + EPS
```
alecandido: I would guess `detuning` to be the most frequent name

src/qibocal/protocols/ramsey/zz_coupling.py
```diff
@@ -27,103 +37,286 @@
...
+    denominator = 1 / (delta_qubit_freq - anharmonicity2) - 1 / (
+        delta_qubit_freq + anharmonicity1
+    )
+
+    # here we compute coupling as a frequency and do error propagation
+    return [
+        float(np.sqrt(np.abs(zz[0] / 2 / denominator))),
+        float(zz[1] / 2 / np.sqrt(2 * abs(denominator * zz[0]))),
+    ]
```
alecandido, line 78: In principle, to properly propagate the error, we should also account for the error on the frequencies.

In any case, it's not deadly relevant. And we would need a suitable scheme to pass around this information.

src/qibocal/protocols/ramsey/zz_coupling.py
```diff
@@ -27,103 +37,286 @@
...
+    spect_ro_delay = Delay(duration=ramsey_duration)
+    spect_ro_seq.append((spect_ro_ch, spect_ro_delay))
+    spect_ro_seq.append((spect_ro_ch, spect_ro_pulse))
+
+    # sweeping over spectator delay before measuring it
+    spectator_sweeper = Sweeper(
+        parameter=Parameter.duration,
+        # here we are adding to the spectator sweeper the duration of the PI/2 pulses
+        # durations for all target qubita
+        range=(ramsey_duration, fin_t + ramsey_duration - init_t, step_t),
```
alecandido: I guess the reason why it is not
```suggestion
        range=(init_t + ramsey_duration, fin_t + ramsey_duration, step_t),
```
is that `init_t` is already accounted for in `ramsey_duration`.

However, maybe it is better to subtract it there, and do something like:
```py
bare_duration = ramsey_duration - init_t
```
and then add it back. At least, you could comment it, and there would be a clear place where to fix it, in case we change convention on placeholder values.

src/qibocal/protocols/ramsey/zz_coupling.py
```diff
@@ -27,103 +37,286 @@
...
-            targets=targets,
-            target_qubit=params.target_qubit if setup == "X" else None,
+    if len(qubits_set) != len(qubits_list):
+        log.warning(
+            "In the pair list there are repeated qubits: "
+            "Parallel execution is not possible."
         )
+        parallel_execution = False
+    else:
+        parallel_execution = True
```
alecandido: I would say, just raise an error. As pointed out by @RoyStegeman some time ago, we do not need another API for sequential execution.

On top of it, we should avoid warnings as much as possible. The objective is to eliminate logging from most of Qibocal (especially protocols). An error will solve also this problem

src/qibocal/protocols/ramsey/zz_coupling.py
```diff
@@ -142,113 +335,129 @@ def _fit(data: RamseyZZData) -> RamseyZZResults:
...
+                omega1=data.qubit_freqs[target],
+                omega2=data.qubit_freqs[spectator],
+                anharmonicity1=data.anharmonicity[target],
+                anharmonicity2=data.anharmonicity[spectator],
+                zz=zz[pair],
+            )
+        except AnharmonicityError:
+            log.warning(
+                f"e-f transitions are not calibrated for qubit pair {pair}, cannot compute coupling strength"
+            )
```
alecandido: You could also avoid the warning: the coupling strength will be missing in the result, and you could explain in the docs why.

src/qibocal/protocols/ramsey/acquisition.py
lballerio: yeah I was thinking to do it but then it seemed an overshoot for a 2 items list

src/qibocal/protocols/ramsey/acquisition.py
lballerio: this is the same output layout as `single_qubit_ramsey_sequence`, which also returns `sequence` and `ro_sequence`. I think, by doing this, we are more flexible since we can append other pulses before the RO without being convoluted.

src/qibocal/protocols/ramsey/acquisition.py
lballerio: yes...actually we can also think to delete it and add only in the ramsey-flux experiment.

src/qibocal/protocols/ramsey/zz_coupling.py
lballerio: ok I see the point, does this just come from the LO and NCO specs?

src/qibocal/protocols/ramsey/zz_coupling.py
lballerio: ok

src/qibocal/protocols/ramsey/acquisition.py
alecandido: Then, I would suggest to have both functions. And, whenever you do not need the extra flexibility, you just use the wrapper one

src/qibocal/protocols/ramsey/acquisition.py
alecandido: Fine even for 2 items.

It would not matter even if it was longer, since it is more explicit about the two item being exactly equal, but not the same.
But it is even shorter ;)

src/qibocal/protocols/ramsey/zz_coupling.py
alecandido: Not really: the frequencies used in the `denominator` are those determined from the various spectroscopies. So, we have fitting errors also for those.

Hopefully, this error is small enough to be negligible.
On the frequencies themselves it certainly is. Usually, we determine multi-GHz frequencies with an error at the level of the MHz or less.
However, on detunings the error is certainly more sizable, since the differences are more of the order of tens or hundreds of MHz.

alecandido (commented):
Multiple comments of the past review are not yet addressed. And I replied to those where the conversation moved forward

src/qibocal/protocols/ramsey/acquisition.py
lballerio: ok

src/qibocal/protocols/ramsey/zz_coupling.py
lballerio: I see your point, but then we also need to propagate the frequency error in the `calibration.json` because at the time being we do not.

lballerio:
I think this PR is ready to merge, I tested on both qpu165, qpu167 and platinum and gives satisfying results. 
The only thing it is worthy to discuss about is whether (I think we should) we bring back error estimation in all RAMSEY experiments (which were removed in #1474) and chi2 and how to estimate.

src/qibocal/protocols/ramsey/processing.py
```diff
@@ -140,7 +140,7 @@ def process_fit(
     """Processing Ramsey fitting results."""
 
     delta_fitting = popt[2] / (2 * np.pi)
-    if detuning is not None:
+    if detuning is not None and detuning != 0:
```
alecandido: Most likely you want to use `isclose()` (e.g. the one from NumPy), or something like that. Exact floating point comparison is rarely desirable

alecandido (approved):
Just a minor comment. Let's merge

src/qibocal/protocols/ramsey/processing.py
lballerio: addressed in 8d48ad16e8e23a08efc7b3159dbf2b648210437a

lballerio:
rebased to main in 8e166d4015be1296b932f54772bfe067b4485f56

scarrazza (approved):

PR #1473, opened 2026-05-05
purge `hardware_average` and `classify` from `Parameters` Class

### LINKED ISSUE #1418: Remove `Parameters.hardware_average` and replace it with protocol-specific uncertainty flags
The current flag exposes a hardware-level acquisition detail that users should not need to think about. Instead, qibocal should choose the most efficient acquisition strategy automatically. 

For many protocols, especially 2D acquisitions, downloading shot-level data only to estimate uncertainties is unnecessary overhead. In those cases, a user-facing `hardware_average` option is misleading and should not exist.

For protocols where uncertainties are useful, mainly some 1D acquisitions, we should expose a flag with user-facing meaning, such as `download_shots_for_uncertainties`. This makes clear what the user achieves through their choice. The fact that it is slower should be mentioned in the docs, but choosing the right name for the flag (such as maybe the suggested one) will also imply this tradeoff.

RoyStegeman:
closes https://github.com/qiboteam/qibocal/issues/1418

These flags could always be provided as input but were ignored in all protocols except qubit spectroscopy. There is no point in exposing them to the user since the acquisition type and averaging mode should be fixed depending on the protocol. 

SINGLESHOT -> CYCLIC for qubit spectroscopy will be done in a separate PR

### DIFF
--- a/src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
+++ b/src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
@@ -3,7 +3,14 @@ from dataclasses import dataclass, field
 from typing import Optional
 
 import numpy as np
-from qibolab import Delay, Parameter, PulseSequence, Sweeper
+from qibolab import (
+    AcquisitionType,
+    AveragingMode,
+    Delay,
+    Parameter,
+    PulseSequence,
+    Sweeper,
+)
 from qibolab._core.components import IqChannel
 
 from qibocal.auto.operation import Parameters, QubitId, Results, Routine
@@ -44,8 +51,6 @@ class QubitSpectroscopyParameters(Parameters):
     """Drive pulse duration [ns]. Same for all qubits."""
     drive_amplitude: Optional[float] = None
     """Drive pulse amplitude (optional). Same for all qubits."""
-    hardware_average: bool = True
-    """By default hardware average will be performed."""
 
 
 @dataclass
@@ -174,7 +179,10 @@ def _acquisition(
             [sequence],
             [sweepers],
             updates=batch_updates,
-            **params.execution_parameters,
+            averaging_mode=AveragingMode.SINGLESHOT,
+            acquisition_type=AcquisitionType.INTEGRATION,
+            nshots=params.nshots,
+            relaxation_time=params.relaxation_time,
         )
 
         # Collect results from this batch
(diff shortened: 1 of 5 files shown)


src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
```diff
@@ -174,7 +179,10 @@ def _acquisition(
             [sequence],
             [sweepers],
             updates=batch_updates,
-            **params.execution_parameters,
+            averaging_mode=AveragingMode.SINGLESHOT,
```
alecandido: The default value for `hardware_average` was already `True` here. Why do you want to shift it to `SINGLESHOT` in this PR?

lballerio (approved):
not any big comment on that, so `execution_parameters` is only used in qubit spectroscopies?

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy_ef.py
```diff
@@ -130,7 +137,10 @@ def _acquisition(
             }
             for q in targets
         ],
-        **params.execution_parameters,
+        averaging_mode=AveragingMode.SINGLESHOT,
```
alecandido: This was not `hardware_average = True`, but having to choose one, I would follow the more popular qubit spectroscopy, since this is seldom used anyhow (and you would expect to behave in the same way).

alecandido (changes requested):

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
RoyStegeman: Because the resonator protols still have the functionality to handle SINGLESHOT, which has to be removed. This just serves as a reminder that work is needed (which will be done first thing after this is merged).

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy_ef.py
RoyStegeman: The reasoning is the same as the other case

alecandido (approved):

scarrazza (approved):
Thanks @RoyStegeman, please remove `execution_parameters` and `classify` from the docs: https://github.com/qiboteam/qibocal/blob/7f6fa4984c59af202ce131bd9773543f291d1b6d/doc/source/conf.py#L41

PR #1474, opened 2026-05-05
Refactoring Ramsey experiments into CYCLIC

lballerio:
In this PR we refactor Ramsey experiment to align to the new drivers capabilities.
In particular we want to switch every single protocols to use CYCLIC averaging mode (shots average in situ) rather than SINGLESHOT mode (downloading every single shot) to improve performances, especially for Qblox architecture.
When we do CYCLIC though we loose the information about the measurement uncertainty, which anyhow is not a big deal since before was not used much. 

> [!NOTE]
 In this PR we had to modify the dependances also in T2 experiments, `cryoscope` and `flux-gate` for not crashing the whole repo, but still they have to be refactored and it will be addressed in a future PRs.


src/qibocal/protocols/ramsey/acquisition.py
alecandido, line 1: Since `ramsey` is already the folder, you could even rename the file just `acquisition.py`. Same for the other modules

src/qibocal/protocols/ramsey/acquisition.py
```diff
@@ -0,0 +1,186 @@
...
+@dataclass
+class RamseyParameters(Parameters):
+    """Ramsey runcard inputs."""
+
+    delay_between_pulses_start: float
+    """Initial delay between RX(pi/2) pulses in ns."""
+    delay_between_pulses_end: float
+    """Final delay between RX(pi/2) pulses in ns."""
+    delay_between_pulses_step: float
+    """Step delay between RX(pi/2) pulses in ns."""
```
alecandido:
```suggestion
    delay: tuple[float, float, float] | None = None
    """..."""
    delay_between_pulses_start: float | None = None
    """Initial delay between RX(pi/2) pulses in ns."""
    delay_between_pulses_end: float | None = None
    """Final delay between RX(pi/2) pulses in ns."""
    delay_between_pulses_step: float | None = None
    """Step delay between RX(pi/2) pulses in ns."""
```
#1300

src/qibocal/protocols/ramsey/acquisition.py
```diff
@@ -0,0 +1,186 @@
...
+
+    @property
+    def waits(self):
+        """
+        Return a list with the waiting times without repetitions.
+        """
+        qubit = next(iter(self.data))
+        return np.unique(self.data[qubit].wait)
+
+    def compute_qubit_signal(self, qubit: QubitId) -> npt.NDArray[np.float64]:
```
alecandido:
```suggestion
    def qubit_signal(self, qubit: QubitId) -> npt.NDArray[np.float64]:
```
Just redundant

src/qibocal/protocols/ramsey/acquisition.py
```diff
@@ -0,0 +1,186 @@
...
+        """
+        Return the signal magnitude for a given qubit.
+        """
+        return magnitude(collect(self.data[qubit].i, self.data[qubit].q))
+
+
+def ramsey_sequence(
+    platform: Platform,
+    targets: list[QubitId],
```
alecandido, line 102: Since they do not have a relation among them, you could write it for a single target, and just call the function multiple times.

src/qibocal/protocols/ramsey/acquisition.py
```diff
@@ -0,0 +1,186 @@
...
+
+    RX90 -- wait -- RX90 -- MZ
+    """
+    delays = 2 * len(targets) * [Delay(duration=wait)]
+    if flux_pulse_amplitude is not None:
+        flux_pulses = len(targets) * [
+            Pulse(duration=0, amplitude=flux_pulse_amplitude, envelope=Rectangular())
+        ]
+    else:
+        flux_pulses = []
```
alecandido:
```suggestion
    flux_pulses = len(targets) * [
            Pulse(duration=0, amplitude=flux_pulse_amplitude, envelope=Rectangular())
        ] if flux_pulse_amplitude is not None else []
```

The usual `if` expression, instead of the statement.

src/qibocal/protocols/ramsey/acquisition.py
```diff
@@ -0,0 +1,186 @@
...
+    params: RamseyParameters,
+    return_probs: bool,
+) -> dict[PulseId, Results]:
+    """Execute Ramsey experiment on the platform."""
+
+    updates = []
+    updates += [
+        {platform.qubits[q].probe: {"frequency": readout_frequency(q, platform)}}
+        for q in targets
+    ]
```
alecandido:
```suggestion
    updates = [
        {platform.qubits[q].probe: {"frequency": readout_frequency(q, platform)}}
        for q in targets
    ]
```

No need to initialize empty, if you have mandatory values.

alecandido: Did you understand why are we moving the readout frequency?

src/qibocal/protocols/ramsey/classification.py
```diff
@@ -0,0 +1,196 @@
...
+    will be constructed using `np.arange`.
+    It is possible to detune the drive frequency using the parameter `detuning` in
+    RamseyParameters which will increment the drive frequency accordingly.
+    Currently when `detuning==0` it will be performed a sweep over the waiting values
+    if `detuning` is not zero, all sequences with different waiting value will be
+    executed sequentially. By providing the option `unrolling=True` in RamseyParameters
+    the sequences will be unrolled when the frequency is detuned.
+    The following protocol will display on the y-axis the probability of finding the ground
+    state, therefore it is advise to execute it only after having performed the single
+    shot classification. Error bars are provided as binomial distribution error.
```
alecandido: This docstring is more suitable for the `Routine` object (`ramsey`, defined at the bottom) then for the acquisition itself.

src/qibocal/protocols/ramsey/classification.py
```diff
@@ -0,0 +1,196 @@
...
+                ],
+                [
+                    fit.delta_phys[target],
+                    fit.delta_fitting[target],
+                    fit.frequency[target],
+                    fit.t2[target],
+                ],
+                display_error=True,
+            )
+        )
```
alecandido: This could be split in its own (internal) function

alecandido (commented):
First batch. Just missing `processing.py` and `signal.py`

src/qibocal/protocols/ramsey/acquisition.py
lballerio: I also think it has to be moved in the `signal.py` module otherwise if it is a general method inherited to all Ramsey experiment it will give an error when called in a classification protocol

src/qibocal/protocols/ramsey/acquisition.py
lballerio: we might even think to remove it from the general sequence.
But this removal better to address in #1467 with also the single qubit sequence function

src/qibocal/protocols/ramsey/acquisition.py
lballerio: addressed in #1467.

src/qibocal/protocols/ramsey/acquisition.py
lballerio: this is actually a good question.
I saw this update in several protocols, honestly I've always taken it as granted even if was a bit obscure the sense of it to me.
Looking at the code of `readout_frequency` ad the input we give in ramsey it is likely they are updating the readout frequency to be either the probe frequency or the resonator dressed frequency, and this second case we know might be suboptimal in general.
So maybe better to remove and directly use the calibrated probe frequency.

src/qibocal/protocols/ramsey/classification.py
lballerio: you mean a single function for all ramsey protocols?

alecandido (approved):

src/qibocal/protocols/ramsey/classification.py
```diff
@@ -0,0 +1,150 @@
...
+
+    data: dict[QubitId, npt.NDArray[RamseyProbType]] = field(default_factory=dict)
+    """Raw data acquired."""
+
+
+def _acquisition(
+    params: RamseyParameters,
+    platform: CalibrationPlatform,
+    targets: list[QubitId],
+) -> RamseyData:
```
scarrazza:
```suggestion
) -> RamseyProbData:
```

src/qibocal/protocols/ramsey/zz_coupling.py
```diff
@@ -0,0 +1,301 @@
...
+
+    """
+    waits = data.waits
+    popts: dict[QubitId, list[float]] = {}
+    freq_measure: dict[QubitId, list[float]] = {}
+    t2_measure: dict[QubitId, list[float]] = {}
+    delta_phys_measure: dict[QubitId, list[float]] = {}
+    delta_fitting_measure: dict[QubitId, list[float]] = {}
+    zz: dict[QubitId, list[float]] = {}
+    coupling: dict[QubitId, list[float]] = {}
```
scarrazza:
```suggestion
    zz: dict[QubitId, float] = {}
    coupling: dict[QubitId, float] = {}
```

tests/calibration_scripts/rx_calibration.py
```diff
@@ -32,10 +32,6 @@
         detuning=1_000_000,
         update=False,
     )
-    if ramsey_output.results.chi2[target][0] > 2:
```
scarrazza, line 35: If you remove this we loose the control over the fit quality. Maybe we could consider an alternative (e.g., residual norm check).

src/qibocal/protocols/ramsey/acquisition.py
```diff
@@ -0,0 +1,189 @@
...
+    Rectangular,
+    Sweeper,
+)
+
+from qibocal.auto.operation import Data, Parameters, QubitId, Results
+
+
+class InputError(Exception):
+    def __init__(self, *args):
+        super().__init__(*args)
```
scarrazza:
```suggestion
    pass
```

tests/calibration_scripts/rx_calibration.py
lballerio: Yeah the fact is that if we don't compute anymore error on the data `chi2` looses a bit its meaning

src/qibocal/protocols/ramsey/processing.py
```diff
@@ -178,3 +161,79 @@ def process_fit(
...
 
     return new_frequency, t2, delta_phys_measure, delta_fitting_measure, popt
+
+
+def fit_plot(
+    target: QubitId, fit: RamseyResults, waits: NDArray, fig: go.Figure
+) -> str:
+    """Generate the fit trace and summary table for Ramsey data."""
+
+    fit_waits = np.linspace(min(waits), max(waits), 20 * len(waits))
```
scarrazza: Could you please remind me how many points we typically use for the sweep? If too many, could cap this to a max value, e.g. `min(20 * len(waits), 2000)`.

lballerio: Actually for ramsey maybe hundreds is a bit eccessive

scarrazza: Ok, then keep this as it is, plotly can handle that well.

scarrazza (approved):

tests/calibration_scripts/rx_calibration.py
lballerio: another thing @alecandido was that this check was in a code that was not used but with random generated data

PR #1476, opened 2026-05-07
flipping: CYLIC + refactoring

### LINKED ISSUE #1478: Make flipping use CYCLIC

RoyStegeman:
remove the unroling flag (always do unrolling now) and use CYCLIC. Also some refactoring to make it more developer- and user-friendly.

### DIFF
--- a/src/qibocal/protocols/flipping.py
+++ b/src/qibocal/protocols/flipping.py
@@ -70,6 +70,22 @@ class FlippingParameters(Parameters):
     rx90: bool = False
     """Calibration of native pi pulse, if true calibrates pi/2 pulse"""
 
+    def __post_init__(self):
+        if not isinstance(self.nflips_max, int):
+            raise TypeError(
+                f"nflips_max must be int, got {type(self.nflips_max).__name__}"
+            )
+        if not isinstance(self.nflips_step, int):
+            raise TypeError(
+                f"nflips_step must be int, got {type(self.nflips_step).__name__}"
+            )
+        if not isinstance(self.rx90, bool):
+            raise TypeError(f"rx90 must be boolean, got {type(self.rx90).__name__}")
+        if self.nflips_max <= 0:
+            raise ValueError("nflips_max must be greater than 0.")
+        if self.nflips_step <= 0:
+            raise ValueError("nflips_step must be greater than 0.")
+
 
 @dataclass
 class FlippingResults(Results):
@@ -128,10 +144,8 @@ def _acquisition(
         resonator_type=platform.resonator_type,
         delta_amplitude=params.delta_amplitude,
         pulse_amplitudes={
-            qubit: (
-                platform.natives.single_qubit[qubit].RX90()
-                if params.rx90
-                else platform.natives.single_qubit[qubit].RX()
+            qubit: getattr(
+                platform.natives.single_qubit[qubit], "RX90" if params.rx90 else "RX"
             )[0][1].amplitude
             for qubit in targets
         },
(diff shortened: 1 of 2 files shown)


src/qibocal/protocols/flipping.py
```diff
@@ -170,30 +161,26 @@ def _acquisition(
...
-            ]
-            if params.unrolling:
-                result = results[ro_pulse.id]
-            else:
-                result = results[i][ro_pulse.id]
-            prob = probability(result, state=1)
+            acq_channel = platform.qubits[qubit].acquisition
+            ro_pulses = list(sequence.channel(acq_channel))
+            if not ro_pulses:
+                continue  # Skip if no acquisition pulse found
```
alecandido: Here you should rather `assert`, not silently skip. Isn't it?

RoyStegeman: You're correct

src/qibocal/protocols/flipping.py
```diff
@@ -167,33 +164,33 @@ def _acquisition(
...
+        sequences,
+        acquisition_type=AcquisitionType.DISCRIMINATION,
+        averaging_mode=AveragingMode.CYCLIC,
+        nshots=params.nshots,
+        relaxation_time=params.relaxation_time,
+    )
 
-    for i in range(len(sequences)):
+    for idx, sequence in enumerate(sequences):
+        flips = flips_sweep[idx]
```
alecandido:
```suggestion
    for flips, sequence in zip(flips_sweep, sequences):
```

src/qibocal/protocols/flipping.py
```diff
@@ -66,14 +65,27 @@ class FlippingParameters(Parameters):
...
+        if not isinstance(self.nflips_step, int):
+            raise TypeError(
+                f"nflips_step must be int, got {type(self.nflips_step).__name__}"
+            )
+        if not isinstance(self.rx90, bool):
+            raise TypeError(f"rx90 must be boolean, got {type(self.rx90).__name__}")
+        if self.nflips_max <= 0:
+            raise ValueError("nflips_max must be greater than 0.")
+        if self.nflips_step <= 0:
+            raise ValueError("nflips_step must be greater than 0.")
```
alecandido, line 87: > [!NOTE]
> We are still using `dataclasses`, so this is a suggestion for later.
> Possibly the next PR, when turning to Pydantic :P

I would check whether with Pydantic there is a better way to raise user-friendly errors.

As a start, I already know that the type errors are coming for free (from validation), and the minima/maxima can also be enforced in the type.
So, in this instance, it would be just a matter of formatting, if we care about that.

src/qibocal/protocols/flipping.py
```diff
@@ -167,33 +164,33 @@ def _acquisition(
...
-            ]
-            if params.unrolling:
-                result = results[ro_pulse.id]
-            else:
-                result = results[i][ro_pulse.id]
-            prob = probability(result, state=1)
+            acq_channel = platform.qubits[qubit].acquisition
+            assert acq_channel is not None
+            acq_channel_pulses = list(sequence.channel(acq_channel))
+            ro_pulse = acq_channel_pulses[-1]  # Last pulse is Readout
```
alecandido: In principle, you could assert even that, if you feel it to be relevant information. Of course, it is not necessary for the way the `ro_pulse` variable is subsequently used.

alecandido (approved):
Minor suggestions, but it looks good.

Thanks!

src/qibocal/protocols/flipping.py
RoyStegeman: I thought about this, but since these errors are user-facing and not just for developers I think (at least the default) pydantic errors are not so nice. I didn't check if there is a way to make them more user friendly.

alecandido: The good thing of Pydantic errors is that they contain all the required information: the traceback, and the details about what is missing where.
For as long as we care about the traceback, then I would keep those as the default.

If instead we want to make a more friendly UI, than I would catch these errors at some point, avoid even the traceback, and just tell the user what's wrong with his input, based on Pydantic's exception (i.e. reformat, print, and exit).
However, we have to make sure to isolate this at an external enough level, and give the option to circumvent it: if not in an interactive environment, tracebacks are well appreciated. And so are verbose Pydantic messages.

scarrazza (approved):

PR #1481, opened 2026-05-07
Investigate Ramsey phase and amplitude origin

alecandido:
<p align="center">
image
</p>


alecandido:
This notebook is not really "production ready". But there is not even any pressure for it to be. Consequently, the PR may survive for long, with no development.

However, I find the notebook rather useful. Both for the explanation itself, but also to quickly experiment with Ramsey-like protocols - or even with more general Bloch-sphere evolution. For as long as they are piecewise constant rotations.

At this point, there may be some value in just merging it.

scarrazza (approved):

RoyStegeman (approved):
Since it's in the `dev/` directory it doesn't have to be production level, and it being useful for development is enough reason to keep it.

PR #1489, opened 2026-05-11
Refactor `standard_rb_sweeper`

sorewachigauyo:
In line with PR https://github.com/qiboteam/qibocal/pull/1393, to refactor `standard_rb_sweeper` in using the new RB API and cyclic acquisition (since the results have to be averaged anyway).

Test results on QPU
[rb-test.tar.gz](https://github.com/user-attachments/files/27582200/rb-test.tar.gz)

### DIFF
--- a/src/qibocal/protocols/randomized_benchmarking/standard_rb_sweeper.py
+++ b/src/qibocal/protocols/randomized_benchmarking/standard_rb_sweeper.py
@@ -1,6 +1,5 @@
-from collections import defaultdict
-
 import numpy as np
+from qibo.backends import NumpyBackend
 from qibo.gates import U3, Unitary
 from qibo.transpiler.unitary_decompositions import u3_decomposition
 from qibolab import (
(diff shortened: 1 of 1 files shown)


alecandido (commented):
To me, it seems just fine. But since @RoyStegeman took care of the RB rewriting, I asked him to take a look at the modifications, since mostly related to the rewriting itself.

RoyStegeman (approved):
Thanks, this seems fine to me. Nice that it only takes 4 minutes for these statistics

src/qibocal/protocols/randomized_benchmarking/standard_rb_sweeper.py
```diff
@@ -66,8 +57,12 @@ def _acquisition(
         RBData: The depths, samples and ground state probability of each experiment in the scan.
     """
 
-    data, backend = setup(params, platform, single_qubit=True)
-    rb_gen = RB_Generator(params.seed)
+    rb_gen = RBGenerator(params.seed)
+    npulses_per_clifford = 2
+    data = setup_data(
+        params, npulses_per_clifford=npulses_per_clifford, single_qubit=True
+    )
```
scarrazza, line 64: The value `npulses_per_clifford = 2` is hardcoded within the _acquisition function. If this value needs to change for different hardware or RB variations, it requires a code change rather than a parameter update.
```suggestion
    data = setup_data(
        params, npulses_per_clifford=getattr(params, 'npulses_per_clifford', 2), single_qubit=True
    )
```

scarrazza (approved):

PR #1490, opened 2026-05-11
Add support for parallel benchmarking

sorewachigauyo:
Following PR https://github.com/qiboteam/qibocal/pull/1393, this is a rework of PR https://github.com/qiboteam/qibocal/pull/1416 with the new RB and circuit execution APIs.

For each depth/iteration combination, we generate a benchmarking circuit for each qubit pair and fuse them together into a larger circuit. These circuits are then accumulated onto an array which is passed for batched execution.

It may be nice to add a flag to re-use the same circuit for all pairs to save time on sequence generation.

Tests on QPU
(without parallel benchmarking)
[rb-2q-test.tar.gz](https://github.com/user-attachments/files/27583241/rb-2q-test.tar.gz)

(with parallel benchmarking)
[rb-2q-test.tar.gz](https://github.com/user-attachments/files/27592429/rb-2q-test.tar.gz)

### DIFF
--- a/src/qibocal/protocols/randomized_benchmarking/utils.py
+++ b/src/qibocal/protocols/randomized_benchmarking/utils.py
@@ -34,7 +34,7 @@ from .fitting import fit_exp1B_func
 
 
 class CircuitIndex(BaseModel):
-    """Tracks the (qubits, depth, iteration) CircuitIndex of a circuit."""
+    """Tracks the (depth, iteration) CircuitIndex of a circuit."""
 
     model_config = ConfigDict(frozen=True, extra="forbid")
 
@@ -413,7 +413,7 @@ def _execute_indexed_circuits(
     indexed_circuits: list[IndexedCircuit],
     params: Parameters,
     platform: CalibrationPlatform,
-    targets: list[QubitId],
+    qubit_map: list[QubitId],
     averaging_mode: AveragingMode = AveragingMode.SINGLESHOT,
 ) -> list[IndexedResult]:
     """Execute indexed circuits and return results paired with their indices.
@@ -422,13 +422,13 @@ def _execute_indexed_circuits(
         indexed_circuits: List of IndexedCircuit objects to execute.
         params: Experiment parameters.
         platform: CalibrationPlatform to execute on.
-        targets: List of target qubit IDs.
+        qubit_map: List of physical qubit IDs.
 
     Returns:
         List of IndexedResult objects with execution results paired with their indices.
     """
 
-    qubit_maps = [targets] * len(indexed_circuits)
+    qubit_maps = [qubit_map] * len(indexed_circuits)
     circuits = []
     for indexed_circuit in indexed_circuits:
         circuits.append(indexed_circuit.circuit)
@@ -492,7 +492,7 @@ def rb_acquisition(
         indexed_circuits=indexed_circuits,
         params=params,
         platform=platform,
-        targets=targets,
+        qubit_map=targets,
         averaging_mode=AveragingMode.CYCLIC
         if len(targets) == 1
         else AveragingMode.SINGLESHOT,
@@ -560,7 +560,7 @@ def twoq_rb_acquisition(
         indexed_circuits=indexed_circuits,
         params=params,
         platform=platform,
-        targets=list(chain.from_iterable(targets)),
+        qubit_map=list(chain.from_iterable(targets)),
     )
 
     # Create a dict of the form {(qubit[0], qubit[1], depth): list[result]}.
--- a/src/qibocal/protocols/randomized_benchmarking/utils.py
+++ b/src/qibocal/protocols/randomized_benchmarking/utils.py
@@ -38,7 +38,6 @@ class CircuitIndex(BaseModel):
 
     model_config = ConfigDict(frozen=True, extra="forbid")
 
-    targets: list[QubitId]
     depth: int
     iteration: int
 
@@ -360,7 +359,7 @@ def setup_data(
 def _generate_indexed_circuits(
     params: Parameters,
     rb_gen: RBGenerator,
-    targets: list[QubitId | QubitPairId],
+    targets: list[QubitId] | list[QubitPairId],
     inverse_layer: bool = True,
     interleave: str | None = None,
 ) -> list[IndexedCircuit]:
@@ -402,7 +401,6 @@ def _generate_indexed_circuits(
 
             add_measurement_layer(full_circuit)
             index = CircuitIndex(
-                targets=list(chain.from_iterable(targets)) if two_qubit else targets,
                 depth=depth,
                 iteration=iteration,
             )
@@ -415,28 +413,24 @@ def _execute_indexed_circuits(
     indexed_circuits: list[IndexedCircuit],
     params: Parameters,
     platform: CalibrationPlatform,
+    targets: list[QubitId],
     averaging_mode: AveragingMode = AveragingMode.SINGLESHOT,
 ) -> list[IndexedResult]:
     """Execute indexed circuits and return results paired with their indices.
 
     Args:
         indexed_circuits: List of IndexedCircuit objects to execute.
-        targets: List of target qubit IDs.
         params: Experiment parameters.
         platform: CalibrationPlatform to execute on.
+        targets: List of target qubit IDs.
 
     Returns:
         List of IndexedResult objects with execution results paired with their indices.
     """
 
-    qubit_maps = []
+    qubit_maps = [targets] * len(indexed_circuits)
     circuits = []
     for indexed_circuit in indexed_circuits:
-        target = indexed_circuit.index.targets
-        if isinstance(target, (list, tuple)):  # Multi-qubit, default
-            qubit_maps.append(list(target))
-        else:  # Single-qubit
-            qubit_maps.append([target])
         circuits.append(indexed_circuit.circuit)
 
     transpiler = build_native_gate_transpiler(platform)
@@ -498,6 +492,7 @@ def rb_acquisition(
         indexed_circuits=indexed_circuits,
         params=params,
         platform=platform,
+        targets=targets,
         averaging_mode=AveragingMode.CYCLIC
         if len(targets) == 1
         else AveragingMode.SINGLESHOT,
@@ -565,6 +560,7 @@ def twoq_rb_acquisition(
         indexed_circuits=indexed_circuits,
         params=params,
         platform=platform,
+        targets=list(chain.from_iterable(targets)),
     )
 
     # Create a dict of the form {(qubit[0], qubit[1], depth): list[result]}.
--- a/src/qibocal/protocols/randomized_benchmarking/utils.py
+++ b/src/qibocal/protocols/randomized_benchmarking/utils.py
@@ -1,6 +1,7 @@
 import pathlib
 from collections import Counter, defaultdict
 from dataclasses import dataclass, field
+from itertools import chain
 from numbers import Number
 from typing import Union
 
@@ -27,23 +28,23 @@ from qibocal.protocols.randomized_benchmarking.dict_utils import (
     load_cliffords,
     separator,
 )
-from qibocal.protocols.utils import significant_digit
+from qibocal.protocols.utils import marginalize_qubit_counts, significant_digit
 
 from .fitting import fit_exp1B_func
 
 
 class CircuitIndex(BaseModel):
-    """Tracks the (qubit, depth, iteration) CircuitIndex of a circuit."""
+    """Tracks the (qubits, depth, iteration) CircuitIndex of a circuit."""
 
     model_config = ConfigDict(frozen=True, extra="forbid")
 
-    qubit: QubitId | QubitPairId
+    targets: list[QubitId]
     depth: int
     iteration: int
 
 
 class IndexedCircuit(BaseModel):
-    """A circuit paired with its (qubit, depth, iteration) CircuitIndex."""
+    """A circuit paired with its (qubits, depth, iteration) CircuitIndex."""
 
     # arbitrary_types_allowed is needed to allow the Circuit type to be a field.
     model_config = ConfigDict(frozen=True, extra="forbid", arbitrary_types_allowed=True)
@@ -359,7 +360,7 @@ def setup_data(
 def _generate_indexed_circuits(
     params: Parameters,
     rb_gen: RBGenerator,
-    targets,  # list[QubitId] or list[QubitPairId]
+    targets: list[QubitId | QubitPairId],
     inverse_layer: bool = True,
     interleave: str | None = None,
 ) -> list[IndexedCircuit]:
@@ -380,16 +381,32 @@ def _generate_indexed_circuits(
 
     inv_file = getattr(params, "file_inv", None)
 
-    for target in targets:
-        for iteration in range(params.niter):
-            for depth in params.depths:
+    assert len(targets) > 0
+    two_qubit = isinstance(targets[0], tuple)
+    nqubits = len(targets) * (2 if two_qubit else 1)
+
+    qubit_map = {
+        target: (idx * 2, idx * 2 + 1) if two_qubit else (idx,)
+        # Reverse assignment for big-endianess
+        for idx, target in enumerate(reversed(targets))
+    }
+
+    for iteration in range(params.niter):
+        for depth in params.depths:
+            full_circuit = Circuit(nqubits)
+            for target in targets:
                 circuit = layer_circuit(rb_gen, depth, target, interleave)
                 if inverse_layer:
                     add_inverse_layer(circuit, rb_gen, inv_file)
-                add_measurement_layer(circuit)
+                full_circuit.add(circuit.on_qubits(*qubit_map[target]))
 
-                index = CircuitIndex(qubit=target, depth=depth, iteration=iteration)
-                indexed_circuits.append(IndexedCircuit(circuit=circuit, index=index))
+            add_measurement_layer(full_circuit)
+            index = CircuitIndex(
+                targets=list(chain.from_iterable(targets)) if two_qubit else targets,
+                depth=depth,
+                iteration=iteration,
+            )
+            indexed_circuits.append(IndexedCircuit(circuit=full_circuit, index=index))
 
     return indexed_circuits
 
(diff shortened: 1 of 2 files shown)


src/qibocal/protocols/randomized_benchmarking/standard_rb_2q.py
```diff
@@ -34,7 +40,10 @@ def _acquisition(
 ) -> RB2QData:
     """Data acquisition for two qubit Standard Randomized Benchmarking."""
 
-    return twoq_rb_acquisition(params, platform, targets)
+    if len(targets) == 1:
+        return twoq_rb_acquisition(params, platform, targets)
+    else:
+        return twoq_rb_acquisition_parallel(params, platform, targets)
```
RoyStegeman: Why are there two options? If `targets` is a single pair, does `twoq_rb_acquisition_parallel` do something different from `twoq_rb_acquisition`, or can `twoq_rb_acquisition_parallel` not be used if not all qubits participate?

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -574,6 +598,93 @@ def twoq_rb_acquisition(
...
+
+    Args:
+        params (RB2QParameters): The parameters for the randomized benchmarking experiment.
+        targets (list[QubitPairId]): The list of qubit pair IDs on which to perform the benchmarking.
+        inverse_layer (bool, optional): Whether to add an inverse layer to the circuits. Defaults to True.
+        interleave (str, optional): The type of interleaving to apply. Defaults to None.
+
+    Returns:
+        RB2QData: The acquired data for two qubit randomized benchmarking.
+    """
```
RoyStegeman: This docstring is an exact copy of the one in `twoq_rb_acquisition`, so from here it is not possible to understand the difference.

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -574,6 +598,93 @@ def twoq_rb_acquisition(
...
+
+    for (qubit0, qubit1, depth), results in grouped.items():
+        data.register_qubit(
+            dtype=RBType,
+            data_keys=(qubit0, qubit1, depth),
+            data_dict={"survival_probs": results},
+        )
+
+    assert isinstance(data, Union[RB2QData, RB2QInterData])
+    return data
```
RoyStegeman: This logic and code is almost entirely duplicated from `twoq_rb_acquisition`, with the relevant change being the merging of the circuits for the different `target` in the list of `targets` and the corresponding `_process_executed_results`. This should be deduplicated as much as possible. 

Depending on the answer to https://github.com/qiboteam/qibocal/pull/1490/changes#r3348829081, we may not need both functions to begin with.

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -574,6 +598,93 @@ def twoq_rb_acquisition(
     return data
 
 
+def twoq_rb_acquisition_parallel(
```
RoyStegeman: I believe this function does quite a lot of things at once, so it may improve maintainability if it were to consist of calls to several sub-functions such that the logic is split into for example setup, circuit generation and execution, and data processing.

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -443,6 +445,28 @@ def _execute_indexed_circuits(
     return indexed_results
 
 
+def _process_executed_results(counter: Counter) -> list[Counter]:
```
RoyStegeman: Ensuring the correctness of this function is tricky because it requires to check the mapping of the qubit pairs to the indices of the bitstring which depends on both the physical to logical qubit ordering mapping as well as big-endian vs small-endian convention used for the `Counter`. Can you clarify these things in the docstring or through how it is implemented, to ensure that the implementation can more easily be confirmed to be correct? 

Also this seems similar to `marginalize_qubit_counts`, perhaps that can be used here.

RoyStegeman (commented):
Thanks a lot @sorewachigauyo! This is an improvement that we want even ourselves since whenever a list of targets is passed, they should be executed in parallel and not in series. For executing in series the protocol can already be called inside a for loop and there is no need to have two interfaces for series execution. 

Having said that, I did have a few comments and questions. My main concern is that most of the logic you implemented duplicates logic in `twoq_rb_acquisition` (and even a bit `marginalize_qubit_counts`). To make qibocal more maintainable, we should try to avoid repeated implementations of the same logic since if we want to apply a change at a later stage this will also have to be duplicated and if forgotten in one location may lead to bugs or unexpected behaviors.

src/qibocal/protocols/randomized_benchmarking/standard_rb_2q.py
sorewachigauyo: Oh, this was just a misinterpretation of your previous comment
https://github.com/qiboteam/qibocal/pull/1416#issuecomment-4169554385. Sure, I can just modify the original function to have the same functionality as parallel.

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -574,6 +598,93 @@ def twoq_rb_acquisition(
...
+            full_circuit = Circuit(nqubits=platform.nqubits)
+            for target in targets:
+                # Per pair, we generate a standard RB circuit
+                circuit = layer_circuit(rb_gen, depth, target, interleave)
+                if inverse_layer:
+                    add_inverse_layer(circuit, rb_gen, file_inv)
+                add_measurement_layer(circuit)
+
+                # Transpile the intermediate circuit and add it to the full circuit
+                transpiled_circ, _ = transpiler(circuit)
```
alecandido: Why not transpiling just the full thing at the end?

src/qibocal/protocols/randomized_benchmarking/standard_rb_2q.py
RoyStegeman: Okay, just to be sure we don't misunderstand each other: my original comment was that in principle having only a single pair is just specific case of parallel and does not require a separate implementation. So if `twoq_rb_acquisition(params, platform, targets)` does the same as `twoq_rb_acquisition_parallel(params, platform, targets)` for the case where `targets` only contains a single pair (can you confirm this is true?), there is no need to have both. In that case we can rename `twoq_rb_acquisition_parallel` to `twoq_rb_acquisition`, and delete what is now `twoq_rb_acquisition`. Or indeed modify the original function to have the same functionality as parallel. 

This way we will lose the behaviour of `twoq_rb_acquisition` in the case where `targets` consists of multiple pairs, but this is what we want since now it does a loop over `targets` and that is inconsistent with the rest of qibocal, namely that `targets` consisting of multiple pairs means it should be executed in parallel.

sorewachigauyo:
As discussed, the general modification is to the circuit generation to fuse the individual Clifford circuits into a larger `nqubit` circuit and the result handling for this larger circuit. Because of that, I had to modify the single qubit RB to make it parallel as well.

Results on QPU
[sq-rb-test.tar.gz](https://github.com/user-attachments/files/28624321/sq-rb-test.tar.gz)
[rb-2q-test.tar.gz](https://github.com/user-attachments/files/28624319/rb-2q-test.tar.gz)

src/qibocal/protocols/randomized_benchmarking/utils.py
sorewachigauyo: Implemented in https://github.com/qiboteam/qibocal/pull/1490/changes/ad776d74ece1b527227e8433965294c7b6df6878

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -359,7 +360,7 @@ def setup_data(
 def _generate_indexed_circuits(
     params: Parameters,
     rb_gen: RBGenerator,
-    targets,  # list[QubitId] or list[QubitPairId]
+    targets: list[QubitId | QubitPairId],
```
RoyStegeman:
```suggestion
    targets: list[QubitId] | list[QubitPairId],
```

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -27,23 +28,23 @@
...
 
 
 class CircuitIndex(BaseModel):
-    """Tracks the (qubit, depth, iteration) CircuitIndex of a circuit."""
+    """Tracks the (qubits, depth, iteration) CircuitIndex of a circuit."""
 
     model_config = ConfigDict(frozen=True, extra="forbid")
 
-    qubit: QubitId | QubitPairId
+    targets: list[QubitId]
```
RoyStegeman: I implemented `CircuitIndex` as an explicit way to keep track of the relation between a circuit and the settings that were used to generate it (they are stored together in `IndexedCircuit`). Since before your work in this PR, there was no parallel execution but instead passing a list of qubits meant that a new circuit was generated for each qubit in the list, the qubitID was also a piece of information needed to identify a circuit. 

Now that you have parallelized the execution there is no longer a different circuit for each qubit, so there would be no more need to store the qubitID in `CircuitIndex`. Then, the `CircuitIndex.targets` are consumed downstream to build the qubit map, but since they are the same for all instances of `CircuitIndex`, and only build from the input `targets` to the Executor, instead of constructing the `qubit_map` from `CircuitIndex.targets` it may make more sense to pass `targets` directly rather than through the class.

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -380,16 +381,32 @@ def _generate_indexed_circuits(
...
     inv_file = getattr(params, "file_inv", None)
 
-    for target in targets:
-        for iteration in range(params.niter):
-            for depth in params.depths:
+    assert len(targets) > 0
+    two_qubit = isinstance(targets[0], tuple)
+    nqubits = len(targets) * (2 if two_qubit else 1)
+
+    qubit_map = {
```
RoyStegeman: Since elsewhere `qubit_map` is the name used for a list of physical qubit IDs ordered such that the first qubit in the list corresponds to logical qubit 0, the second to logical qubit 1, etc. I think we should avoid using the same name for a different type of object. At least to me, the logic of handling qubits is already fragile/complex enough

RoyStegeman (commented):
Thanks, the implementation is much more compact now. I just have a few minor comments with the main one being about the need to store `targets` in `CircuitIndex`.

I tested for both individual single qubit RB and simultaneous single qubit RB and the results are as expected although the parallel I tried with low enough statistics that it is only with moderate confidence. 

Could you please add some tests to check the handling of the simultaneous targets?

src/qibocal/protocols/randomized_benchmarking/utils.py
sorewachigauyo: Implemented in https://github.com/qiboteam/qibocal/pull/1490/changes/15c8068cac7a627b2902a7bdb540137f940d1138

src/qibocal/protocols/randomized_benchmarking/utils.py
sorewachigauyo: Implemented in https://github.com/qiboteam/qibocal/pull/1490/changes/15c8068cac7a627b2902a7bdb540137f940d1138

src/qibocal/protocols/randomized_benchmarking/utils.py
sorewachigauyo: Implemented in https://github.com/qiboteam/qibocal/pull/1490/commits/64b1a1810e7b6f754c3a4fd052e292c38bf6e5b7

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -27,23 +28,22 @@
...
 )
-from qibocal.protocols.utils import significant_digit
+from qibocal.protocols.utils import marginalize_qubit_counts, significant_digit
 
 from .fitting import fit_exp1B_func
 
 
 class CircuitIndex(BaseModel):
-    """Tracks the (qubit, depth, iteration) CircuitIndex of a circuit."""
+    """Tracks the (qubits, depth, iteration) CircuitIndex of a circuit."""
```
RoyStegeman:
```suggestion
    """Tracks the (depth, iteration) CircuitIndex of a circuit."""
```

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -398,28 +413,24 @@ def _execute_indexed_circuits(
...
         params: Experiment parameters.
         platform: CalibrationPlatform to execute on.
+        targets: List of target qubit IDs.
 
     Returns:
         List of IndexedResult objects with execution results paired with their indices.
     """
 
-    qubit_maps = []
+    qubit_maps = [targets] * len(indexed_circuits)
```
RoyStegeman: Do we still need multiple qubit_maps if they are all the same?

RoyStegeman (commented):
Thanks for the changes. At this point I think it is almost okay to be merged, though I still do not like the way the physical qubit to logical qubit mapping is handled. Are you confident that for example the qubit(pairs) are not assigned to the wrong result? 

This problem is not introduced in this PR, so I will not ask you to handle it here, but if it's something you want to address (even in a different PR) I think that would be a great improvement since I'm not convinced it's without bugs. I think it would be much better if there was a single source of truth for this mapping and/or tests that actually probe this aspect of the RB directly.

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -546,22 +560,23 @@ def twoq_rb_acquisition(
         indexed_circuits=indexed_circuits,
         params=params,
         platform=platform,
+        targets=list(chain.from_iterable(targets)),
```
RoyStegeman: Since we consistenly use `targets` to be the list of input qubit IDs or qubitID pairs, and `qubit_map` to denote the list of phsyical qubit IDs ordered as they are encoded in the logical qubits of the circuit, I think it would be better to rename this to `qubit_map`.

src/qibocal/protocols/randomized_benchmarking/utils.py
sorewachigauyo: Not really, but I'd have to change the behaviour further downstream in `execute_circuits` and check the other consumers. I can do that in another PR, a quick check shows the other consumers are just running a single circuit for this API.

sorewachigauyo:
> Thanks for the changes. At this point I think it is almost okay to be merged, though I still do not like the way the physical qubit to logical qubit mapping is handled. Are you confident that for example the qubit(pairs) are not assigned to the wrong result?
> 
> This problem is not introduced in this PR, so I will not ask you to handle it here, but if it's something you want to address (even in a different PR) I think that would be a great improvement since I'm not convinced it's without bugs. I think it would be much better if there was a single source of truth for this mapping and/or tests that actually probe this aspect of the RB directly.

Yeah, this is something I am still quite tentative about with the opaqueness of `execute_circuits`. On my system, I checked this was correct by comparing the results of `standard_rb` to `standard_rb_sweeper`.

In the previous PR, I handled this by assigning individual measurement register per pair and reading from those at the end, but this was through the [Qibo API](https://github.com/qiboteam/qibocal/pull/1416/changes#diff-c0f56125ac5983ae6588a339a4f13c6c8364c51772138c6d8e262590fd4bff95R618-R619). We could probably attach something like that to the Qibocal circuit execution pipeline (in another PR).

src/qibocal/protocols/randomized_benchmarking/utils.py
sorewachigauyo: Implemented in https://github.com/qiboteam/qibocal/pull/1490/changes/92c4b626182756eb535f97e1bc96ff5fdb6600cc

src/qibocal/protocols/randomized_benchmarking/utils.py
sorewachigauyo: Implemented in https://github.com/qiboteam/qibocal/pull/1490/changes/92c4b626182756eb535f97e1bc96ff5fdb6600cc

src/qibocal/protocols/randomized_benchmarking/utils.py
RoyStegeman: Indeed, there will not longer be a loop over circuits so it should be possible to reflect it downstream. I agree this can be done in a new PR. I opened an issue for it #1524.

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -380,16 +380,31 @@ def _generate_indexed_circuits(
...
-    for target in targets:
-        for iteration in range(params.niter):
-            for depth in params.depths:
+    assert len(targets) > 0
+    two_qubit = isinstance(targets[0], tuple)
+    nqubits = len(targets) * (2 if two_qubit else 1)
+
+    target_id_map = {
+        target: (idx * 2, idx * 2 + 1) if two_qubit else (idx,)
+        # Reverse assignment for big-endianess
```
RoyStegeman: Thinking about this again, each qubit pair encodes a different sub-circuit, so I would say endianess is a concept that only applies within the pairs and does not extend beyond it, while here you are explicitly assigning the concept of endianness to the order of the pairs (if there are pairs). 

If we go up one level from `_generate_indexed_circuits` into `rb_acquisition` or `twoq_rb_acquisition`, in both function calls to `_generate_indexed_circuits` and `_execute_indexed_circuits` are made. In the call of `_execute_indexed_circuits` inside `twoq_rb_acquisition` you now pass 
```
targets=list(chain.from_iterable(targets))
```
which is really a `qubit_map`. But then inside `_generate_indexed_circuits` you do 
```
    target_id_map = {
        target: (idx * 2, idx * 2 + 1) if two_qubit else (idx,)
        # Reverse assignment for big-endianess
        for idx, target in enumerate(reversed(targets))
    }
```
which is then used in 
```
circuit.on_qubits(*target_id_map[target])
```

To me it seems that the reverse assignment makes that the physical-to-logical qubit mapping in `_generate_indexed_circuits` and `_execute_indexed_circuits` differs. Perhaps it makes sense at this point to define the `qubit_map` at the level of `twoq_rb_acquisition` and `rb_acquisition` and create a function that wraps `_generate_indexed_circuits` and `_execute_indexed_circuits` that takes as input `qubit_map` and also `targets` (since for the two-qubit case one is just a list and the other a list of lists), though perhaps there is a cleaner solution than passing twice the same thing.

sorewachigauyo: Well, down the pipeline in https://github.com/qiboteam/qibocal/blob/main/src/qibocal/auto/transpile.py#L163, it gets reversed there, so I would have to reverse it somewhere for reconcilation. (I confused my comment on the order of the endianness due to a misunderstanding of https://github.com/qiboteam/qibocal/pull/1490#discussion_r3349249200)

Anyway, I plan to handle this on a separate PR to modify the circuit execution and transpilation by assigning exact registers.

RoyStegeman: Okay, if you could also open an issue before or together with the PR that would be great.

RoyStegeman (approved):
Thanks a lot for this PR @sorewachigauyo, it is something that was necessary and certainly a good improvement. 

As we discussed, the qubit ordering is still quite messy. We have to follow the endian notation of qibo itself and on top of that we have to handle the physical to logical qubit mapping and at the moment this implementation is very fragile. As agreed this can be dealt with in a follow up PR https://github.com/qiboteam/qibocal/pull/1490#discussion_r3380335457

scarrazza (approved):

PR #1492, opened 2026-05-11
improve intial parameters guess Lorentzian fit

### LINKED ISSUE #1408: resonator spectroscopy failed fit examples
http://login.qrccluster.com:9000/Xrj2xt6wR-Gviisvw6dafw==

RoyStegeman:
closes #1408

The fits in #1408 that before failed now pass.

The background was made linear in an earlier PR. In this PR we heavily rely on the assumption of a linear background to make the guesses for the initial parameters so it is stable only up to a point where the background is linear enough. 

Specifically what changed are the following things:
- the linear background consisting of offset and slope is now estimated from the endpoints of the frequency domain. This linear function is subtracted before peak detection, so argmax/argmin operate on the residual signal.
- sigma is now derived from the half-maximum width of the peak after subtracting the linear background from the data. Before the distance between argmax and argmin of the raw data was used, which is much less stable.
- uncertainties are no longer passed to the fit function since uncertainties in a spectroscopy are largely unnecessary overhead. Bounds have been added to constrain the fit parameters.

### DIFF
--- a/src/qibocal/protocols/utils.py
+++ b/src/qibocal/protocols/utils.py
@@ -133,7 +133,8 @@ def lorentzian_fit(data, resonator_type=None, fit=None):
             frequencies[indices_beyond_half[-1]] - frequencies[indices_beyond_half[0]]
         ) / 2
     else:
-        guess_sigma = frequencies[1] - frequencies[0]
+        # if there is no clear peak, we give a high flexibility
+        guess_sigma = frequencies[-1] - frequencies[0]
 
     guess_amp = guess_peak_height * guess_sigma * np.pi
 
--- a/src/qibocal/protocols/utils.py
+++ b/src/qibocal/protocols/utils.py
@@ -108,34 +108,45 @@ def lorentzian_fit(data, resonator_type=None, fit=None):
     voltages = data.signal
 
     # Guess parameters for Lorentzian max or min
-    # TODO: probably this is not working on HW
-    guess_offset = np.mean(
-        voltages[np.abs(voltages - np.mean(voltages)) < np.std(voltages)]
-    )
-    guess_slope = 0.0
+    guess_slope = (voltages[-1] - voltages[0]) / (frequencies[-1] - frequencies[0])
+    guess_offset = voltages[0] - guess_slope * frequencies[0]
+    voltages_no_background = voltages - (guess_slope * frequencies + guess_offset)
+
     if (resonator_type == "3D" and fit == "resonator") or (
         resonator_type == "2D" and fit == "qubit"
     ):
-        guess_center = frequencies[
-            np.argmax(voltages)
-        ]  # Argmax = Returns the indices of the maximum values along an axis.
-        guess_sigma = abs(frequencies[np.argmin(voltages)] - guess_center)
-        guess_amp = (np.max(voltages) - guess_offset) * guess_sigma * np.pi
-
+        guess_center = frequencies[np.argmax(voltages_no_background)]
+        peak_amp = voltages_no_background.max()
+        indices_beyond_half = np.where(voltages_no_background > peak_amp / 2)[0]
     else:
-        guess_center = frequencies[
-            np.argmin(voltages)
-        ]  # Argmin = Returns the indices of the minimum values along an axis.
-        guess_sigma = abs(frequencies[np.argmax(voltages)] - guess_center)
-        guess_amp = (np.min(voltages) - guess_offset) * guess_sigma * np.pi
+        guess_center = frequencies[np.argmin(voltages_no_background)]
+        peak_amp = voltages_no_background.min()
+        indices_beyond_half = np.where(voltages_no_background < peak_amp / 2)[0]
+
+    if len(indices_beyond_half) >= 1:
+        guess_sigma = (
+            frequencies[indices_beyond_half[-1]] - frequencies[indices_beyond_half[0]]
+        ) / 2
+    else:
+        guess_sigma = frequencies[1] - frequencies[0]
+
+    guess_amp = peak_amp * guess_sigma * np.pi
+    guess_offset_at_center = guess_slope * guess_center + guess_offset
 
     initial_parameters = [
         guess_amp,
         guess_center,
         guess_sigma,
-        guess_offset,
+        guess_offset_at_center,
         guess_slope,
     ]
+
+    freq_domain_size = frequencies[-1] - frequencies[0]
+    bounds = (
+        [-np.inf, frequencies[0], 0.0, -np.inf, -np.inf],
+        [np.inf, frequencies[-1], freq_domain_size, np.inf, np.inf],
+    )
+
     # fit the model with the data and guessed parameters
     try:
         sigma = (
@@ -150,6 +161,7 @@ def lorentzian_fit(data, resonator_type=None, fit=None):
             voltages,
             p0=initial_parameters,
             sigma=sigma,
+            bounds=bounds,
         )
         # The output results are stored in a json, but ndarray is not JSON serializable,
         # so the parameters are converted to list.
--- a/src/qibocal/protocols/utils.py
+++ b/src/qibocal/protocols/utils.py
@@ -140,7 +140,6 @@ def lorentzian_fit(data, resonator_type=None, fit=None):
         guess_offset_at_center,
         guess_slope,
     ]
-
     freq_domain_size = frequencies[-1] - frequencies[0]
     bounds = (
         [-np.inf, frequencies[0], 0.0, -np.inf, -np.inf],
@@ -149,27 +148,16 @@ def lorentzian_fit(data, resonator_type=None, fit=None):
 
     # fit the model with the data and guessed parameters
     try:
-        sigma = (
-            data.error_signal
-            if (hasattr(data, "error_signal") and not np.isnan(data.error_signal).any())
-            else None
-        )
-
         fit_parameters, perr = curve_fit(
             lorentzian,
             frequencies,
             voltages,
             p0=initial_parameters,
-            sigma=sigma,
             bounds=bounds,
         )
         # The output results are stored in a json, but ndarray is not JSON serializable,
         # so the parameters are converted to list.
-        perr = (
-            np.sqrt(np.diag(perr)).tolist()
-            if sigma is not None
-            else [0] * len(initial_parameters)
-        )
+        perr = np.sqrt(np.diag(perr)).tolist()
         model_parameters = fit_parameters.tolist()
         return model_parameters[1] * GHZ_TO_HZ, model_parameters, perr
     except RuntimeError as e:
--- a/src/qibocal/protocols/utils.py
+++ b/src/qibocal/protocols/utils.py
@@ -110,18 +110,23 @@ def lorentzian_fit(data, resonator_type=None, fit=None):
     # Guess parameters for Lorentzian max or min
     guess_slope = (voltages[-1] - voltages[0]) / (frequencies[-1] - frequencies[0])
     guess_offset = voltages[0] - guess_slope * frequencies[0]
-    voltages_no_background = voltages - (guess_slope * frequencies + guess_offset)
+    guess_background = guess_offset + guess_slope * frequencies
+    voltages_no_background = voltages - guess_background
 
     if (resonator_type == "3D" and fit == "resonator") or (
         resonator_type == "2D" and fit == "qubit"
     ):
         guess_center = frequencies[np.argmax(voltages_no_background)]
-        peak_amp = voltages_no_background.max()
-        indices_beyond_half = np.where(voltages_no_background > peak_amp / 2)[0]
+        guess_peak_height = voltages_no_background.max()
+        indices_beyond_half = np.where(voltages_no_background > guess_peak_height / 2)[
+            0
+        ]
     else:
         guess_center = frequencies[np.argmin(voltages_no_background)]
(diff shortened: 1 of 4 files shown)


src/qibocal/protocols/utils.py
```diff
@@ -108,34 +108,45 @@ def lorentzian_fit(data, resonator_type=None, fit=None):
     voltages = data.signal
 
     # Guess parameters for Lorentzian max or min
-    # TODO: probably this is not working on HW
-    guess_offset = np.mean(
-        voltages[np.abs(voltages - np.mean(voltages)) < np.std(voltages)]
-    )
-    guess_slope = 0.0
+    guess_slope = (voltages[-1] - voltages[0]) / (frequencies[-1] - frequencies[0])
```
alecandido, line 111: Maybe it could be worth doing it on a range of values, and taking the median. Something like:
```suggestion
    guess_slope = (voltages[-10:] - voltages[:10]) / (frequencies[-10:] - frequencies[:10])
```

Or using a given fraction of the data points

RoyStegeman: While it should make the initial gradient estimate slightly more stable, I don't recall having seen a case where I believe this would have been beneficial to the fit (where it would make a non-convergent fit converge). The shot noise is quite low even for relatively small number of shots, so I don't believe there is much benefit by effectively integrating out that. Other sources of background noise are on much larger scales so there averaging over the last and first few points doesn't integrate that out.

alecandido: I believe your statement is quite biased by the way you run fits: sometimes it is just easier to simulate a "VNA mode", and go for a *dense* scan (which possibly corresponds to a regular step size on a large window).

To make it more interactive (and that's the main reason of comparison to the VNA) you may balance the resources invested on the scan with shots. Essentially, to keep the runtime still small.
If you do this aggressively enough, you can end up with a quite large shot noise, but the peaks are still quite deep, so there is quite some room to play this game before making them disappearing below the noise floor.

Moreover, take into account that you are taking a difference, which is also doubling the noise just because of that.
If this has little to no chance of adding new failure channels, I would be inclined to introduce some further noise resilience.

alecandido: It's not that your observation is wrong, but I'm just thinking more and more in the direction of automation. The more resilient are our fits, the more we are approaching a good starting point. And we have to build this one bit a time, necessarily

RoyStegeman: The current fit assumes that the Lorentzian peak is the feature that dominates the residues, which is a reasonable assumption for the way we have been doing the resonator fits so far. To support more VNA-like scans where the window is much larger and possibly contains multiple peaks, we should revise the fitting algorithm in general while this PR aims just to improve the stability of the fits in the way we have been doing them so far. Possibly a VNA-style scan would receive a separate flag in which we don't try to do a Lorentzian fit, but just do a peak detections, such that a user can follow up by zooming in on the different peaks and trying a Lorentzian fit. 

Then, I believe main thing that will break this fit is when the residues from the background become large when fitted with just a linear function, but not so much the exact initial guess for the gradient. To demonstrate, for this plot I did `nshots=10`. http://login.qrccluster.com:9000/HG_6MkuZQxaJrs0n5lEmkg==


image

alecandido: You can certainly use the VNA for the many peaks, but that's not what I had in mind here. And, as you wrote yourself, that's also quite easily addressed by a peak finder (though the problem of the "sharpness" of peaks may survive - but it is for a different PR). I agree that, despite sharing the acquisition, it is a different protocol (we could even turn the flag into a completely separate name, since I expect many more options).

In the end, what you plotted is exactly what I was thinking: we tune the window manually but not exactly, and then we take many more points than shots.
Indeed, to speed up our experiments I would start trying with aggressively few shots, trading them for swept points, which are quickly bringing more information.

Would you try to rerun the same, but in a (reasonably) larger window? E.g. 4 times as large, assuming there is no other further peak in scope.
Btw, upload plots with references to data (i.e. links to the report server, or zipped folders). In this case, mostly for the sake of the etiquette

RoyStegeman: Yes instead of a flag it probably make sense to make it a separate protocol.

Here is a report with the same settings except for a window three times as large. http://login.qrccluster.com:9000/ScyXVqJNQmWlsvh02b_fSw==
image

alecandido: Ok, I know it is weird to say, but I find the correction still quite reasonable. It is fair that it has not been able to fit, just because of the background (which is clearly non-linear).

In the hindsight, I even made the wrong call: on a larger window it is not going to be necessarily more noise prone, because you have some stability of your background. It is more sensitive to individual points fluctuations if they are closer together in the y-axis, which in this case they are not, since they are even more separated.

But the former window was already quite a small scan (just 5 MHz), so there is no need to keep exploring the window size. And there would have been no need at all, if I realized this in advance (sorry).

I guess that, all in all, being just an initial estimate, the iterative fit is already introducing "enough" noise-resilience. At least enough for these data - and I have no way to extrapolate a global "enough"

src/qibocal/protocols/utils.py
```diff
@@ -108,56 +108,56 @@ def lorentzian_fit(data, resonator_type=None, fit=None):
...
+
+    if len(indices_beyond_half) >= 1:
+        guess_sigma = (
+            frequencies[indices_beyond_half[-1]] - frequencies[indices_beyond_half[0]]
+        ) / 2
+    else:
+        guess_sigma = frequencies[1] - frequencies[0]
+
+    guess_amp = peak_amp * guess_sigma * np.pi
+    guess_offset_at_center = guess_slope * guess_center + guess_offset
```
alecandido: I know it is just equivalent, but, possibly, a simpler parametrization of the background is around the center. I.e.

```py
def lorentzian(frequency, amplitude, center, sigma, offset, slope):
    f = (frequency - center)
    peak = (amplitude / np.pi) * (sigma / (f ** 2 + sigma**2))
    background = offset + f * slope
    return peak + background
```

src/qibocal/protocols/utils.py
```diff
@@ -108,56 +108,56 @@ def lorentzian_fit(data, resonator_type=None, fit=None):
...
+            bounds=bounds,
         )
         # The output results are stored in a json, but ndarray is not JSON serializable,
         # so the parameters are converted to list.
-        perr = (
-            np.sqrt(np.diag(perr)).tolist()
-            if sigma is not None
-            else [0] * len(initial_parameters)
-        )
+        perr = np.sqrt(np.diag(perr)).tolist()
```
alecandido: To avoid shadowing, and since they are different objects, we may rename the original one
```suggestion
        perr = np.sqrt(np.diag(cov)).tolist()
```

src/qibocal/protocols/utils.py
```diff
@@ -108,56 +108,56 @@ def lorentzian_fit(data, resonator_type=None, fit=None):
...
         resonator_type == "2D" and fit == "qubit"
     ):
-        guess_center = frequencies[
-            np.argmax(voltages)
-        ]  # Argmax = Returns the indices of the maximum values along an axis.
-        guess_sigma = abs(frequencies[np.argmin(voltages)] - guess_center)
-        guess_amp = (np.max(voltages) - guess_offset) * guess_sigma * np.pi
-
+        guess_center = frequencies[np.argmax(voltages_no_background)]
+        peak_amp = voltages_no_background.max()
```
alecandido: Maybe not in this PR, but this is quite a terrible guess: as you showed yourself in the comment before, background fluctuations can greatly exceed the resonance peak.

Better to rely on a `find_peak()` than on `argmax()`

src/qibocal/protocols/utils.py
```diff
@@ -108,56 +108,56 @@ def lorentzian_fit(data, resonator_type=None, fit=None):
...
+        indices_beyond_half = np.where(voltages_no_background > peak_amp / 2)[0]
     else:
-        guess_center = frequencies[
-            np.argmin(voltages)
-        ]  # Argmin = Returns the indices of the minimum values along an axis.
-        guess_sigma = abs(frequencies[np.argmax(voltages)] - guess_center)
-        guess_amp = (np.min(voltages) - guess_offset) * guess_sigma * np.pi
+        guess_center = frequencies[np.argmin(voltages_no_background)]
+        peak_amp = voltages_no_background.min()
+        indices_beyond_half = np.where(voltages_no_background < peak_amp / 2)[0]
```
alecandido: Btw, we should abandon the branching based on `resonator_type`, and replace with the following:

- for the resonator spectroscopy, we know the side of the peak, which is determined by the measurement being performed in transmission or reflection - we can easily consume this information upstream, by applying the same overall fitting strategy (including initial parameters estimation) to either the `signal` or `-signal`
- for the qubit spectroscopy, we do not know the side of the peak, since it depends on the readout frequency (whether reading at the dressed or bare, e.g.) - we should attempt detecting which is the correct sign by running a `find_peak()` with both signs, and comparing the statistics of the most relevant peaks (prominence, displacement wrt background)

alecandido (approved):
The first two comments are stylistic.
The other two need to be addressed in a separate PR. But, if you agree with them, better to write them down in suitable issues immediately

src/qibocal/protocols/utils.py
RoyStegeman: Actually what I did was wrong, so thanks for pointing this out! To correct, I should either change `lorentzian` as you suggest, simply use `guess_offset` directly instead of `guess_offset_at_center`. 

Since in `lorentzian_fit` we have to determine a guess for the offset before we determine a guess for the center, the way to minimize the number of transformations and variables floating around, is by keeping `lorentzian` as it is, and removing `guess_offset_at_center`

src/qibocal/protocols/utils.py
RoyStegeman: Yes, though in the case where background fluctuations greatly exceed the resonance peak, they will also dominate the residue and even with an improved guess I would expect the fit to be highly unstable simply as a consequence of the mismatch between parametrization and data. 

Ignoring even the first peak, if we want to be able to fit a spectrum like this one (where actually still the background does not exceed the peak) possibly we have to revise also the treatment of the background rather than just the initial parameter guess. 
http://login.qrccluster.com:9000/L4OXrFCPSluN5a-J5jmADQ==
Image

src/qibocal/protocols/utils.py
RoyStegeman: I agree https://github.com/qiboteam/qibocal/issues/1501

src/qibocal/protocols/utils.py
RoyStegeman: https://github.com/qiboteam/qibocal/issues/1502

alecandido: Regarding your previous comment, now it makes much more sense to me. And it is very readable.

However, in general, readability often matters more than efficiency, since most of the operations we are performing are not bottlenecks for the execution (networking on its own is already quite offsetting most of the fit operations).
So, before applying performance considerations (as reducing the number of transformations), better to think about whether we care about that performance gain.

But, in this case, readability and performance are aligned. Which is best (and not even that rare)

---

Regarding the issue mention, I guess it is for the comment below

alecandido (approved):

src/qibocal/protocols/utils.py
```diff
@@ -108,26 +108,34 @@ def lorentzian_fit(data, resonator_type=None, fit=None):
...
+        indices_beyond_half = np.where(voltages_no_background < guess_peak_height / 2)[
+            0
+        ]
+
+    if len(indices_beyond_half) >= 1:
+        guess_sigma = (
+            frequencies[indices_beyond_half[-1]] - frequencies[indices_beyond_half[0]]
+        ) / 2
+    else:
+        guess_sigma = frequencies[1] - frequencies[0]
```
alecandido: This could be a problem, if the peak is cut on one side.

If this is our current strategy for width estimation, it would be better to either abort the fit, or rather use one extreme of the window (worst case scenario, you may set to the window size, or half of it, and let the least squares do their job).

src/qibocal/protocols/utils.py
alecandido: > Yes, though in the case where background fluctuations greatly exceed the resonance peak, they will also dominate the residue and even with an improved guess I would expect the fit to be highly unstable simply as a consequence of the mismatch between parametrization and data.
> 
> Ignoring even the first peak, if we want to be able to fit a spectrum like this one (where actually still the background does not exceed the peak) possibly we have to revise also the treatment of the background rather than just the initial parameter guess.

This could be solved by segmentation. I could try to see if #1497 is able to isolate this peak

> http://login.qrccluster.com:9000/L4OXrFCPSluN5a-J5jmADQ==

In any case, thanks for the data. I will try to see if the `find_peaks()` is also providing somehow a better estimate, even without segmentation (which still relies on it, even if it is also using complementary information).

src/qibocal/protocols/utils.py
RoyStegeman: Yes when I mentioned reducing the number of transformations as a benefit I was thinking in terms of readibility rather than performance. 

> Regarding the issue mention, I guess it is for the comment below

Yes indeed

src/qibocal/protocols/utils.py
RoyStegeman: Good suggestion

scarrazza (approved):

Issue #1496, opened 2026-05-13
Regularize variance estimate for Bernoulli averages


lballerio:
During $\chi^2$ fits, it is required to estimate the mean and variance of the Gaussian variables involved.
Gaussianity is often achieved as a limit property for the average of repeated measurements (e.g. shots), because of the central limit theorem. And an unbiased estimator for the mean of this variable is the sample mean.

As discussed in #1486, for the average of i.i.d. Bernoulli variables, also the variance can be estimated just out of the sample mean. This is done just replacing the mean in the variance formula with the sample mean, and applying the usual Bessel correction.

However, while the estimator is unbiased (and intuitively consistent), it can be degenerate for finite estimates, resulting in a null variance.

While theoretically there is no problem for a variance estimate to be zero because of fluctuations, this creates a disproportionate weight for some points in some actual extractions, and cause numerical errors.
Partly, this is also due to the central limit theorem approximation (we are assuming some Gaussian variables, when they only become Gaussian in the limit - and this could be a problem in a "small" extraction"), which may be reconsidered with more thorough modelling (cf. #1486).

Nevertheless, a practical (and theoretically sound) solution could be introduced, by means of a regularization consisting of supplementary pseudo-counts.
A motivation for this can be found in the choice of a conjugate prior for our Bernoulli variable parameter (the $p$ parameter of all i.i.d. extractions), for which we exclude extreme values (i.e $0$ and $1$), and we (mildly) favor intermediate values. This is a reasonable assumption a priori, since we often expect our state  to be either a superposition or a mixed state, because of limited fidelities and thermal populations, and because we expect the center of mass of our trajectories to be often around the equator (for a generic experiment).

To implement this, we just add to our shots `n` counts for both ground and excited states. Where a reasonable value for `n` could be just 1.
In practice, we often have access only to the sample average and the number of shots, this means that the regularized average will be
```py
yreg = (y * nshots + n) / (nshots + n)
```

Which correspond to choose a prior $p \sim \text{Beta}(n+1, n+1)$.

> [!IMPORTANT]
>
> Especially if we want to rely on the emulator, we should find a final solution in efficiently estimating errors and 'regularize them' because sometimes in the emulator might happen that qubit state prob is 1 and then we might have problem in the error, which becomes 0 and `curve_fit` fails.

alecandido:
AI generated

---

[...]

Issue #1486, opened 2026-05-08
Fits with limit variance estimates


alecandido:
When measuring probabilities (i.e. discrimination frequencies), we have some corner values in the estimated variances.

Our variables for the individual shots are Bernoulli distributed, and the sum of many Bernoulli variables should approach a Gaussian by central limit theorem. Thus, it is sufficient to obtain an estimate for the mean and variance to use the standard Gaussian-based approaches.
While the mean has a standard estimate, in this case it is pretty simple to build an asymptotically unbiased estimator of the variance

$$
\begin{align*}
E[y_i] = E[y_i^2] = p &\qquad E[y_i y_j] = E[y_i]E[y_j] = p^2\\
\bar{y} &= \frac{\sum_i y_i}{n} \\
E[\bar{y}(1 - \bar{y})] &= \frac{n-1}{n} p(1-p) = \frac{n-1}{n} \sigma^2
\end{align*}
$$

So, the unbiased estimator is coming with the usual correction factor

$$
\frac{\bar{y}(1 - \bar{y})}{n-1}
$$

However, for a finite statistics, we may obtain a degenerate estimate: if measuring a $p$ very close to 0 or 1, all the samples could have the same value just because of finite-size fluctuations. Especially when the sample is small.
The problem is not restricted to exact zero-variances, because a disproportionately small one can still greatly alter any (standard?) analysis.

In those cases, the sample variance becomes mostly unsuitable for any further analysis, being it a fit or a goodness of fit estimate.

## Temporary proxies

For the sake of moving forward towards automation, we can adopt proxies for the time being, before 
Instead of relying on wrongly estimated uncertainties, we could just leverage the least squares fit itself.

Indeed, a good fit should converge to a minimum in the parameters' space, signalled by a positive Hessian.
Moreover, we would expect our curvature to be sharp. Which, in absence of an absolute scale, we could take relative to the parameters value.

**cons**
- we lose a goodness of fit metric, with a probabilistic interpretation
  - which is already unreliable
- it may falsely fail on parameters which are genuinely vanishing
  - this is not supposed to happen frequently, since most of the parameters we want to estimate should not be null (e.g. Rabi frequency, couplings)
  - especially when measured on the correct scale - so, in presence of a large relative uncertainty, claiming a failure is not completely inappropriate

**pros**
- it will not report a fake $\chi^2$ value, with no probabilistic interpretation
[...]

Issue #1485, opened 2026-05-08
χ² fits replacement for Bernoulli variables


alecandido:
A $\chi^2$ fit is often adopted based on a reasonable approximation, i.e. the Gaussianity of the errors.

Fluctuations around a certain value can be Gaussian for a number of reasons, more or less boiling down to the central limit theorem, when there is a number of noise sources involved for a single variable.
The other option is that Gaussianity arises because considering variables for which the main sources of fluctuations just satisfy the requirements of the central limit theorem (i.e. finite variance), and the Gaussian variable considered is the not the original variable itself, but the sample mean, out of an extraction of copies (e.g. shots).

However, in all the cases in which we are fitting discriminated frequencies, we know the distribution of the underlying variable. And it is not a Gaussian, since it is Bernoulli.
This falls in the second case spelled out above, since the sample average is then binomially distributed, which in the large $n$ limit is known to approach a Gaussian.

Nevertheless, for a finite (small) sample, a Bernoulli variable can still be significantly different from a Gaussian.
On top of this, we may have regularization problems, as described in #1496 

This could be enough motivation to look into a replacement for a least squares (or, better, least $\chi^2$) fit.

A convenient framework can still be the computation of some statistics of the Bayesian posterior, e.g. the MAP estimate, subject to some regularizing prior (but possibly even a flat one could be acceptable). But possibly more information can be extracted from the fit (from credible intervals to parameters correlations).

## Implementation

While a MCMC method may be a perfectly suitable (and efficient enough) method, especially given the reduced size of the problem, it is also possible that, given the problem's simplicity, it could also be mapped to some kind of algorithmic minimization in the parameters' space (curve fitting), with some analytically-computed alteration of the target function (as a replacement of weighted least squares).

Going for MCMC could still be a choice, because more information may be extracted from a posterior sample.

alecandido:
AI generated

---

[...]

PR #1504, opened 2026-05-18
Flipping amplitude

jevillegasd:
This is a 2D sweep for the flipping routine in which a amplitude detuning is used in one axis.


sorewachigauyo:
Maybe put some separation between the colorbar and the "Best amp"

image

(Don't mind the data, there's an issue on my end that I'm debugging)

sorewachigauyo:
Thanks for this, the fitting seems to work well

image

[flipping_2d.tar.gz](https://github.com/user-attachments/files/27983285/flipping_2d.tar.gz)

I had to hack an implementation for sweepers because there were too many sequences to run for my hardware
If you'd like, I can do the PR for that

src/qibocal/protocols/flipping.py
```diff
@@ -46,12 +46,13 @@ def flipping_sequence(
...
         )
         sequence.append((qd_channel, qd_detuned))
-        sequence.append((qd_channel, qd_detuned))
+        # sequence.append((qd_channel, qd_detuned))
 
         if rx90:
             sequence.append((qd_channel, qd_detuned))
             sequence.append((qd_channel, qd_detuned))
 
+    sequence |= natives.R(theta=np.pi / 2, phi=0.0 if flips % 2 == 0 else np.pi)
```
lballerio, line 55: I think we should go back to the previous configuration

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+    nflips_max: int = 21
+    """Maximum number of flips ([RX(pi) - RX(pi)] sequences)."""
+    nflips_step: int = 1
+    """Step size for the number of consecutive flips."""
+    delta_amplitude_min: float = -0.05
+    """Minimum amplitude delta relative to the native pulse amplitude."""
+    delta_amplitude_max: float = 0.05
+    """Maximum amplitude delta relative to the native pulse amplitude."""
+    delta_amplitude_step: float = 0.001
+    """Amplitude delta step."""
```
lballerio, line 40: why this coice for the default values?

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+        if not isinstance(self.nflips_max, int):
+            raise TypeError(
+                f"nflips_max must be int, got {type(self.nflips_max).__name__}"
+            )
+        if not isinstance(self.nflips_step, int):
+            raise TypeError(
+                f"nflips_step must be int, got {type(self.nflips_step).__name__}"
+            )
+        if not isinstance(self.rx90, bool):
+            raise TypeError(f"rx90 must be boolean, got {type(self.rx90).__name__}")
```
lballerio, line 54: I think this checks can be takes as granted or this params can be recasted

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+)
+"""Custom dtype for flipping amplitude sweep."""
+
+
+@dataclass
+class FlippingAmplitudeData(Data):
+    """FlippingAmplitude acquisition outputs."""
+
+    resonator_type: str
+    """Resonator type."""
```
lballerio, line 89:
```suggestion
```

in this protocol is not used at all, can be deleted everywhere

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+
+    For each combination of (flips, delta_amplitude) a sequence is built and
+    executed.  The amplitude values are stored as absolute amplitudes
+    (native + delta).  The resulting 2D map allows identifying the correct
+    drive amplitude: at the true pi-pulse amplitude the excited-state
+    probability should remain flat regardless of the number of flips.
+    """
+
+    data = FlippingAmplitudeData(
+        resonator_type=platform.resonator_type,
```
lballerio, line 115:
```suggestion
```

as said before

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+
+    nflips_max: int = 21
+    """Maximum number of flips ([RX(pi) - RX(pi)] sequences)."""
+    nflips_step: int = 1
+    """Step size for the number of consecutive flips."""
+    delta_amplitude_min: float = -0.05
+    """Minimum amplitude delta relative to the native pulse amplitude."""
+    delta_amplitude_max: float = 0.05
+    """Maximum amplitude delta relative to the native pulse amplitude."""
+    delta_amplitude_step: float = 0.001
```
lballerio, line 39: we want to gradually switch from a notation like this to a range notation, where you put as input directly the (start, end, step) tuple; from now we are still accepting both, so I think is bettter to put these params as optional, then add `@property` or methods that from the input they define a unique `self.delta_amplitude_range` and `self.delta_amplitude_values` to use in the `_acquisition` function:


```suggestion
    delta_amplitude_range: tuple[float, float, float] | None = None
    """Amplitude delta range relative to the native pulse amplitude."""
    delta_amplitude_min: float | None = None
    """Minimum amplitude delta relative to the native pulse amplitude."""
    delta_amplitude_max: float | None = None
    """Maximum amplitude delta relative to the native pulse amplitude."""
    delta_amplitude_step: float | None = None
```

subsequently:

```python
@property
def amplitude_range(self):
   if self.delta_amplitude_range is None:
         return (
                     self.delta_amplitude_start,
                     self.delta_amplitude_stop, 
                     self.delta_amplitude_step
         )
    return self.delta_amplitude_range
```
(BE CAREFUL, RANDOM INDENTATION)

then in the `__post_init__` method you can raise an Error if `self.amplitude_range is None`.

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+@dataclass
+class FlippingAmplitudeParameters(Parameters):
+    """FlippingAmplitude runcard inputs."""
+
+    nflips_max: int = 21
+    """Maximum number of flips ([RX(pi) - RX(pi)] sequences)."""
+    nflips_step: int = 1
+    """Step size for the number of consecutive flips."""
```
lballerio, line 34: same as the comment below for `delta_amplitude_*` params.

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+            raise TypeError(
+                f"nflips_step must be int, got {type(self.nflips_step).__name__}"
+            )
+        if not isinstance(self.rx90, bool):
+            raise TypeError(f"rx90 must be boolean, got {type(self.rx90).__name__}")
+        if self.nflips_max <= 0:
+            raise ValueError("nflips_max must be greater than 0.")
+        if self.nflips_step <= 0:
+            raise ValueError("nflips_step must be greater than 0.")
+
```
lballerio, line 59:
```python
def delta_amplitude_values(self) -> npt.NDArray:
       return np.arange(*self.delta_amplitude_range)
```

same for `nflips` (here we can also force the cast into `int`)

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+        },
+        rx90=params.rx90,
+    )
+
+    flips_range = range(0, params.nflips_max, params.nflips_step)
+    delta_amplitude_range = np.arange(
+        params.delta_amplitude_min,
+        params.delta_amplitude_max,
+        params.delta_amplitude_step,
+    )
```
lballerio, line 130: with the new format no more need, simply call class property or method.

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+            for qubit in targets:
+                sequence += flipping_sequence(
+                    platform=platform,
+                    qubit=qubit,
+                    delta_amplitude=float(delta_amp),
+                    flips=flips,
+                    rx90=params.rx90,
+                )
+            sequences.append(sequence)
+            sweep_params.append((flips, float(delta_amp)))
```
lballerio: use sweeper.

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+            assert isinstance(ro_pulse, Readout)
+            prob = results[ro_pulse.id]
+            error = np.sqrt(prob * (1 - prob) / params.nshots)
+            native_amp = data.pulse_amplitudes[qubit]
+            data.register_qubit(
+                FlippingAmplitudeType,
+                qubit,
+                {
+                    "flips": np.array([flips]),
+                    "amplitude": np.array([native_amp + delta_amp]),
```
lballerio: if you decide to apply the suggestions above: 
```suggestion
                    "flips": params.flip_values(),
                    "amplitude": params.delta_amplitude_values() + native_amp,
```

still I am not sure whether is better to simply save the `delta_amplitude` or the absolute value.

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+    )
+
+    for (flips, delta_amp), sequence in zip(sweep_params, sequences):
+        for qubit in targets:
+            acq_channel = platform.qubits[qubit].acquisition
+            assert acq_channel is not None
+            ro_pulse = list(sequence.channel(acq_channel))[-1]
+            assert isinstance(ro_pulse, Readout)
+            prob = results[ro_pulse.id]
+            error = np.sqrt(prob * (1 - prob) / params.nshots)
```
lballerio: be aware this error estimation is critical since sometimes might explode for `prob={0,1}` (mostly you end up with this error in the emulator).
We are currently working on some regulator to insert, basically fake counts to avoid divergence of the errors.
see #1485, #1486 and #1496 for some suggestions.

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+
+    for qubit in data.qubits:
+        qubit_data = data[qubit]
+        amplitudes = np.unique(qubit_data["amplitude"])
+        variances = []
+
+        for amp in amplitudes:
+            mask = qubit_data["amplitude"] == amp
+            probs = qubit_data["prob"][mask]
+            variances.append(float(np.var(probs)))
```
lballerio, line 214: don't know if it is worth to investigate over more sophisticated methods such as spectral analysis or not.

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+
+        best_idx = int(np.argmin(variances))
+        best_amp = float(amplitudes[best_idx])
+        native_amp = data.pulse_amplitudes[qubit]
+
+        best_amplitudes[qubit] = [best_amp, 0.0]
+        delta_amplitudes[qubit] = [best_amp - native_amp, 0.0]
```
lballerio, line 221: computing both `best_amplitudes` and `delta_amplitudes` is redundant, I would only use one variable, so we also delete a useless dictionary

even for the amplitude uncertainty I don't know how we can estimate it.

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+
+    return FlippingAmplitudeResults(
+        amplitude=best_amplitudes,
+        delta_amplitude=delta_amplitudes,
```
lballerio, line 225: deduplicate

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+    """Plotting function for FlippingAmplitude.
+
+    Produces a heatmap of excited-state probability as a function of flips
+    (x-axis) and pulse amplitude (y-axis).  When fit results are available a
+    dashed horizontal line marks the best amplitude.
+    """
+
+    def ev(prob: np.ndarray) -> np.ndarray:
+        """Helper function to calculate the expectation value."""
+        return 2 * prob - 1
```
lballerio, line 244: why computing the expectation value of Z?

also I am pretty sure it should be: `1 - 2*prob`, since `prob` is the probability of the qubit being in state 1 and for 1 `<Z>=-1`.

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+
+    # Build 2D probability matrix: rows = amplitude, cols = flips
+    z = np.full((len(amplitudes), len(flips_vals)), np.nan)
+    amp_index = {amp: i for i, amp in enumerate(amplitudes)}
+    flip_index = {fl: j for j, fl in enumerate(flips_vals)}
+
+    for row in qubit_data:
+        i = amp_index[row["amplitude"]]
+        j = flip_index[row["flips"]]
+        z[i, j] = ev(row["prob"])
```
lballerio, line 258: I think a simple reshape should be enough:


```suggestion
        z =  qubit_data["prob"].reshape((len(amplitudes), len(flip_vals)))
```

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -0,0 +1,286 @@
...
+    )
+
+    fitting_report = ""
+    if fit is not None and target in fit.amplitude:
+        fitting_report = table_html(
+            table_dict(
+                target,
+                ["Best amplitude [a.u.]", "Delta amplitude [a.u.]"],
+                [fit.amplitude[target], fit.delta_amplitude[target]],
+                display_error=True,
```
lballerio, line 292: here you are displaying a `0.0` error, maybe if we don't know how to estimate the uncertainty `at the moment` we can neglect displaying error.

```suggestion
                display_error=False,
```

lballerio (changes requested):
my PR for this new interesting protocol

lballerio:
PS maybe we should also think about new tests.

lballerio:
> I had to hack an implementation for sweepers because there were too many sequences to run for my hardware
If you'd like, I can do the PR for that

@sorewachigauyo you mean in a separate PR?
I would avoid opening another PR for such a small implementation, I think without sweepers there is no sense in merging it, so if you want to work on it please coordinate with @jevillegasd and try to work on the same PR.
Also @jevillegasd have you tested it? On which chip?

jevillegasd:
> Thanks for this, the fitting seems to work well
> 
> image
> [flipping_2d.tar.gz](https://github.com/user-attachments/files/27983285/flipping_2d.tar.gz)
> 
> I had to hack an implementation for sweepers because there were too many sequences to run for my hardware If you'd like, I can do the PR for that

Yes please do the PR pointing to this one and we can combine it

src/qibocal/protocols/flipping_amplitude.py
jevillegasd: mhhh. 🥇  This is just 10% of the total range in amplitude (assuming a rabi pulse is ~ 0.3, it gives some good variability) and nflips is just a number.

src/qibocal/protocols/flipping_amplitude.py
jevillegasd: true, this I just copied from `flipping` but then I can make these pareameters be a subclass of `FlippingParameters` so that it inherits these checks.

src/qibocal/protocols/flipping_amplitude.py
jevillegasd: true this was just me copying old code.

src/qibocal/protocols/flipping_amplitude.py
jevillegasd: same

lballerio:
Hi @jevillegasd, I've just merged the sweeper implementation of the `flipping_amplitude` experiment.
Now we can do the detailed review and if you want to test it more.

PR #1508, opened 2026-05-19
Add sweeper support for `flipping_amplitude`

sorewachigauyo:
As discussed in PR https://github.com/qiboteam/qibocal/pull/1504, an implementation of the routine with amplitude sweeper

This probably has to be modified based on how https://github.com/qiboteam/qibocal/pull/1504/commits/42aea449a92f8fb89de0019990727bbf66139d1e goes and if https://github.com/qiboteam/qibocal/pull/1507 gets merged due to
```python
            # First two pulses are the pi/2 pulse and unused alignment and the last pulse is a readout alignment
            # So we add every pulse in-between
            # Technically we can reduce the number of pulses by filtering by unique UUID
            pulses_to_sweep[qubit].extend(
                list(seq.channel(platform.qubits[qubit].drive))[2:-1]
            )
```
which I'm not super happy about, but it works for now


src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -123,50 +132,66 @@ def _acquisition(
     )
 
     sequences: list[PulseSequence] = []
-    sweep_params: list[tuple[int, float]] = []
+    pulses_to_sweep: defaultdict[QubitId, list[Pulse]] = defaultdict(list)
+    parallel_sweepers: list[Sweeper] = []
```
lballerio: for parallel sweepers you can import from qibolab `ParallelSweepers` typehint, which exaclty defined as a list of `Sweeper` objects:

```python
ParallelSweepers = list[Sweeper]
```

I know is not a proper class but just a type, however we want to introduce its usage in qibocal, to keep the notation clearer.

```suggestion
    parallel_sweepers: ParallelSweepers = []
```

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -123,50 +132,66 @@ def _acquisition(
...
+                delta_amplitude=0,
+                flips=flips,
+                rx90=params.rx90,
+            )
+            # First two pulses are the pi/2 pulse and unused alignment and the last pulse is a readout alignment
+            # So we add every pulse in-between
+            # Technically we can reduce the number of pulses by filtering by unique UUID
+            pulses_to_sweep[qubit].extend(
+                list(seq.channel(platform.qubits[qubit].drive))[2:-1]
+            )
```
lballerio: surely if we recycle the same pulse we need just one to store.
can be another reason why to keep reusing the same pulse

src/qibocal/protocols/flipping_amplitude.py
```diff
@@ -123,50 +132,66 @@ def _acquisition(
...
+                data.register_qubit(
+                    FlippingAmplitudeType,
+                    qubit,
+                    {
+                        "flips": np.array([flips]),
+                        "amplitude": np.array([amp]),
+                        "prob": np.array([prob]),
+                        "error": np.array([error]),
+                    },
+                )
```
lballerio, line 191: maybe can be simplified, but we can do it in PR #1504.

src/qibocal/protocols/flipping_amplitude.py
lballerio: I am not an expert on this experiment, but should we also sweep on the initial RX90 pulse amplitude?

lballerio (commented):
Thanks @sorewachigauyo for this implementation.
As I said, I think there is still something we can refactor and simplify, but we can postpone this discussion after we merge this PR on PR #1504.
As you mentioned here and in PR #1507, in your implementation we are recycling the same pulse, which I am getting convinced might be the more practical solution since we are sweeping the same way all the pulses in the experiment, so we might actually recycle the pulse so we don't have a huge list of pulses which differ only for the UUID we sweep in the same way...
Btw did you test it?
I am simply waiting for @jevillegasd review.

sorewachigauyo:
> Thanks @sorewachigauyo for this implementation. As I said, I think there is still something we can refactor and simplify, but we can postpone this discussion after we merge this PR on PR #1504. As you mentioned here and in PR #1507, in your implementation we are recycling the same pulse, which I am getting convinced might be the more practical solution since we are sweeping the same way all the pulses in the experiment, so we might actually recycle the pulse so we don't have a huge list of pulses which differ only for the UUID we sweep in the same way... 

On my driver implementation, sweeping/playing multiple pulses with the same UUID is not an issue. My concern from the other PR is if that was an issue with QM/Qblox, which I don't have in my lab to test. If this isn't a problem, then yeah, we should go ahead with the recycling, because in the current implementation (prior to #1507), each call of `flipping_sequence` generates `flips` number of UUIDs due to `replace` making a new pulse and `replace` being used in a loop.

>Btw did you test it? 

Yeah, it was actually in the other PR, but here is a new test
[flipping_2d.tar.gz](https://github.com/user-attachments/files/28047448/flipping_2d.tar.gz)

src/qibocal/protocols/flipping_amplitude.py
sorewachigauyo: Fixed in https://github.com/qiboteam/qibocal/pull/1508/changes/001c80c990c77761491aff673f59aa14a5b46bc5

src/qibocal/protocols/flipping_amplitude.py
sorewachigauyo: Personally, I don't think we have to.

This particular experiment is looking for the minimum variance between adding multiple $2\pi$ rotations from the initial state, so the initial state doesn't have to be the exact middle, so long as we return to it. (And I guess in that sense, we don't have to start from the middle either)

I think the initial state affects the original experiment more under certain conditions, as that decides under/over-rotation fitting.

lballerio: I see your point, yeah probably your right, then probably won't make such a difference.

lballerio:
Hi @sorewachigauyo, I just tried to run this implementation on qblox and actually the sweeper is too big and does not fit, it raises an error.
I am trying to modify a bit the code.

sorewachigauyo:
Hi @lballerio 
Could you try this branch https://github.com/qiboteam/qibocal/tree/flipping_amplitude_sweeper_minpulses?

Its a bit more extreme rewrite that doesn't reuse the existing sequence generation and should make qibolab happier
QPU data [flipping_2d.tar.gz](https://github.com/user-attachments/files/28060888/flipping_2d.tar.gz)

If this doesn't work, then I'll probably change this into a separate PR for sweepers

lballerio:
This is the memory issue on the sweepers, as you can see from the error message:

image

I also tried to run by repeating always the same pulse, trying to see if inserting only once the pulse to sweep it effectively sweeps over all the pulses in the sequence, but at the time being it seems it is not working as it should, but I need to see in more detail.

Ok let me see your branch, thanks!

lballerio:
Hi @sorewachigauyo I tried your new branch and the results I think a re very satifying!
In this way you are simply using the same UUID throughout the entire experiment, right?
Maybe this is better.

image

I also worked on this sweeper implementation and ended up with https://github.com/qiboteam/qibocal/tree/flipping_amplitude_sweeper_lballerio.
In my case I was calling `flipping_sequence` for each `nflips` values, hence I am sweeping over `N` pulses, with `N` total number of `nflips`, since within the same sequence the UUID is the same. However these modifications didn't measure good signal, as you can see from the picture below:

image

if you want to have a look, but I think you're implementation is more than fine...maybe we should move the PR on that branch.
I think in my implementation there is probably some error in the sweeper usage, better to use your which is just the same pulse

sorewachigauyo:
Yeah, its the same UUID/pulse object for the entire experiment

Okay, good to see that it works with Qblox
I'll just push the commit here

src/qibocal/protocols/flipping_amplitude.py
sorewachigauyo: Implemented in https://github.com/qiboteam/qibocal/pull/1508/changes/776a94df770ec5ae3cee500e56524c1a0cab87ce

lballerio:
perfect, yeah I like your implementation!

lballerio (approved):
tested with Qblox on TII qw5q_platinum QPU, I think is good

scarrazza (approved):

PR #1516, opened 2026-06-02
remove `DcFilteredConfig` since duplicate of `DcConfig` in qibolab

RoyStegeman:
`DcFilteredConfig` is a qibocal duplicate of qibolab's `DcConfig` so there is no need for it to exist (at least not any longer). Note the attribute name `filter` was used in `DcFilteredConfig` while `filters` is used in `DcConfig`, this was a problem in the cyoscope protocol. This PR does not fully fix the cyroscope protocol though and further work will be done in separate PRs.


lballerio (approved):
for me is fine, thanks for fixing this.

scarrazza (approved):

PR #1531, opened 2026-06-10
fix dispersive shift plot legend

### LINKED ISSUE #1431: dispersive shift plot "best frequency" line cannot be toggled off
In the dispersive shift plot there is a "best frequency" line that has two issues:
1) there are two toggles while it is a unique frequency for both, and I believe it should indeed be a single line/toggle
2) even if both are toggled off, there is still a vertical line

http://login.qrccluster.com:9000/Vc4AvG5mSHumGB82G1EjkA==/

Image

RoyStegeman:
closes #1431 

Everything for plotting was inside a loop over states 0 and 1, while not everything depended on the state, leading to useless or even duplicate operations (such as two "best frequency" legends). On top of duplicate "best frequency" legends, there was a third "best frequency" line in the plot generated with `vline`. Finally, the legends for the phase plots are now shown which they were not before. 

An example of what it looks like now: http://login.qrccluster.com:9000/lRZ4RucoS7-XKMuiEUS6Mg==

### DIFF
--- a/src/qibocal/protocols/dispersive_shift/dispersive_shift.py
+++ b/src/qibocal/protocols/dispersive_shift/dispersive_shift.py
@@ -149,7 +149,6 @@ def _fit(data: DispersiveShiftData) -> DispersiveShiftResults:
     res_frequencies = {}
     best_freqs = {}
     fitted_parameters = {}
-
     for qubit in qubits:
         freq = []
         fit_params = []
@@ -202,28 +201,18 @@ def _plot(data: DispersiveShiftData, target: QubitId, fit: DispersiveShiftResult
         ),
     )
 
-    fitting_report = ""
+    for state in (0, 1):
+        label = f"State {state}"
+        q_data = data[target, state]
 
-    data_0 = data[target, 0]
-    data_1 = data[target, 1]
-
-    for i, label, q_data in list(
-        zip(
-            (0, 1),
-            ("State 0", "State 1"),
-            (data_0, data_1),
-        )
-    ):
-        opacity = 1
         frequencies = q_data.freq * HZ_TO_GHZ
         fig.add_trace(
             go.Scatter(
                 x=frequencies,
                 y=q_data.signal,
-                opacity=opacity,
-                name=f"{label}",
+                opacity=1,
+                name=f"{label} data",
                 showlegend=True,
-                legendgroup=f"{label}",
             ),
             row=1,
             col=1,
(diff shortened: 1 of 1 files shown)


scarrazza (approved):

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
```diff
@@ -202,28 +201,18 @@ def _plot(data: DispersiveShiftData, target: QubitId, fit: DispersiveShiftResult
...
-        opacity = 1
         frequencies = q_data.freq * HZ_TO_GHZ
         fig.add_trace(
             go.Scatter(
                 x=frequencies,
                 y=q_data.signal,
-                opacity=opacity,
-                name=f"{label}",
+                opacity=1,
+                name=f"{label} data",
```
alecandido: Both signal and phase are *data*.

So, if the other is `State x phase`, this one should be `State x signal` (which we could have called *magnitude*, *amplitude*, *radius*, ... but that's the present terminology).

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
```diff
@@ -202,28 +201,18 @@ def _plot(data: DispersiveShiftData, target: QubitId, fit: DispersiveShiftResult
...
-    ):
-        opacity = 1
         frequencies = q_data.freq * HZ_TO_GHZ
         fig.add_trace(
             go.Scatter(
                 x=frequencies,
                 y=q_data.signal,
-                opacity=opacity,
-                name=f"{label}",
+                opacity=1,
```
alecandido: Isn't it just the default?

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
```diff
@@ -232,80 +221,71 @@ def _plot(data: DispersiveShiftData, target: QubitId, fit: DispersiveShiftResult
...
+                    np.max(all_signals),
+                ],
+                mode="lines",
+                line=go.scatter.Line(color="orange", width=3, dash="dash"),
+                name="Best frequency",
+                showlegend=True,
+            ),
+            row=1,
+            col=1,
+        )
```
alecandido, line 265: Why not just adding a vertical line?

https://plotly.com/python/horizontal-vertical-shapes/

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
```diff
@@ -232,80 +221,71 @@ def _plot(data: DispersiveShiftData, target: QubitId, fit: DispersiveShiftResult
...
+        if fit.frequencies[target] is not None:
             table_entries = [
+                "Ground state Frequency [Hz]",
+                "Excited state Frequency [Hz]",
+                "Chi [Hz]",
```
alecandido, line 270: This has not to be done in a single place, but we definitely need to avoid representing frequencies in Hz, since they just become a bunch of meaningless numbers...

Image

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
RoyStegeman: Yes you're right

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
RoyStegeman: I believe `add_vline` doesn't support legends

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
RoyStegeman: For table generation qibocal has a dedicated function so I think we should address that in a dedicated PR

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
alecandido: True. But if toggling is not strictly required, we may just replace with an annotation. Which is natively supported by the interface.

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
alecandido: Also true. But does the function distinguish between frequencies and everything else?
Here we definitely need more sig figs than in the default representation. So, we should pass the information that is a frequency somehow (or that we want a certain absolute/relative accuracy).

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
RoyStegeman: I believe there is some value to put the "best frequency" line on the same footing as the fit and data lines, rather than as a fixed vertical line with an annotation somewhere inside the plot, which may even be tricky because the plot reshapes based on window size so having it annotated may make it harder to read depending on the window size. Even the code is not that problematic to me.

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
RoyStegeman: > But does the function distinguish between frequencies and everything else?
 
 I don't believe so

Actually I think there is some merit in leaving the value in Hz since that is also how it is presented in the parameters.json. Either way, I think the discussion about how we want to present the numbers and the corresponding changes to the protocols and probably also table plotting function is something that takes some consideration and should not be part of this PR.

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
alecandido: The main advantage I see is being able to toggle it. But my opinion was that is not that relevant.
The label is not that massive, and in practice what matters the most is the size relative to the window. And it quite works for very different window sizes (I tested the Plotly example also on a smartphone size).

Then, my proposal was actually to distinguish it from the fit and data lines, to reduce the size of the legend and the amount of elements that could be toggled. Since having a long list is just more confusing to me.

However, UI is strongly a matter of opinions. So, I'm motivating my owns, but I'm not asserting they are "right" in any way.
I care more not wasting too much time on it. We're always in time to change.

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
alecandido: Fine, I agree

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
RoyStegeman: I don't believe writing "best frequency" as annotation is less confusing than in the legend, and the annotation would be needed to clarify what the vertical line corresponds to, but I acknowledge that it's a matter of opinion.

alecandido (approved):

PR #1532, opened 2026-06-10
Rework circuit measurement handling

### LINKED ISSUE #1524: Replace repeated `qubit_maps` list with a single `qubit_map` in `execute_circuits`
Deferred from https://github.com/qiboteam/qibocal/pull/1490#discussion_r3371400321. In the RB circuit execution, `qubit_maps`, is constructed as a list repeating the same `qubit_map` for every circuit:

```python
qubit_maps = [qubit_map] * len(indexed_circuits)
```

This is a leftover from when a list of targets would generate one circuit per target and execute the circuits in a loop. Since #1490, all targets are executed in parallel on a single circuit, so `qubit_map` can be used directly in favor of `qubit_maps`.

### LINKED ISSUE #1530: Circuit execution pipeline
Currently for `execute_circuits`, passing `qubit_maps` of `[[q0, q2, q3, q7, ..., qn]]`  for a single circuit returns a `Counter` object where the bitstrings are mapped to `(qn, ..., q7, q3, q2, q0)`. Furthermore, if the consumer is only looking at certain combinations of qubits (e.g. pairs), they would have to break the bitstring counter to obtain their relevant bits, which adds to the confusion of result mapping.

I propose an approach where we relax the constraint of https://github.com/qiboteam/qibocal/blob/main/src/qibocal/auto/transpile.py#L89-L97 where only a single qubit can be measured per measurement gate and simply resolve the single/multi-qubit measurement registers with `measurement_maps`. This would place it much closer to the existing circuit execution pipeline of `QibolabBackend`.  Then, we then resolve the mapping of logical qubits onto the physical qubits and return `Counter` objects like in the existing behaviour, but mapped to the requested physical qubits per measurement.

sorewachigauyo:
As described in https://github.com/qiboteam/qibocal/issues/1530, we relax the condition that assigns only one qubit per measurement register and provide handling of multiple qubits per measurement register.

From the normal flow of circuit execution, we get a map of measurement registers/gates to pulse sequences. The qubits in these registers are logical indices, so we can map them back onto the physical qubit IDs thanks to the platform mapping being available. We then reuse the original result handling but assign the state counts to the physical qubit IDs instead of the logical IDs.

TODO

- [x] Further clean and test implementation
- [x] Align and test [RB](https://github.com/user-attachments/files/29075819/rb-test.tar.gz) / [2Q RB](https://github.com/user-attachments/files/29075827/rb-2q-test.tar.gz)
- [x] Align and test [readout mitigation](https://github.com/user-attachments/files/28789426/readout-mitigation.tar.gz)
- [x] Align and test [ttomography ](https://github.com/user-attachments/files/28832223/tomo.tar.gz)
- [x] Align and test [CHSH](https://github.com/user-attachments/files/28831390/chsh.tar.gz)
- [x] Fix tests


src/qibocal/auto/transpile.py
```diff
@@ -17,19 +19,23 @@
...
     """QubitId can be integers or strings. ``pad_circuit`` only works with integer qubit
     IDs, so if the qubit maps contain string IDs, we convert them to integer indices
     based on the platform's qubit order.
     """
     qubits = list(platform.qubits)
-    return [
-        [q if isinstance(q, int) else qubits.index(q) for q in qubit_map]
-        for qubit_map in qubit_maps
-    ]
+    return [q if isinstance(q, int) else qubits.index(q) for q in qubit_map]
```
RoyStegeman: Since this is just needed to accommodate `pad_circuit`, I think it makes sense to call this function inside `pad_circuit`.

src/qibocal/auto/transpile.py
```diff
@@ -52,87 +59,106 @@ def _pad_circuit(nqubits: int, circuit: Circuit, qubit_map: list[int]) -> Circui
...
-    if len(gate.qubits) != 1:
-        raise ValueError(
-            "Measurement gate must measure a single qubit. "
-            f"Got gate with {len(gate.qubits)} qubits."
-        )
-    if gate.qubits[0] not in qubit_map:
-        raise KeyError(f"Qubit {gate.qubits[0]} not found in qubit map: {qubit_map}.")
 
+    new_circuit = _pad_circuit(platform.nqubits, circuit, qubit_map)
+    transpiled_circ, _ = transpiler(new_circuit)
```
RoyStegeman: This way the transpilation is done for each circuit. Can it be done only once on new circuit after constructing it from the sub-circuits? 

Since the `_transpile_circuit` function is now much simpler than the `_transpile_circuits` before there may be no more reason for it to exist in this form. Also because the transpilation is only one of the two steps so the function name does not provide that much information on what it actually does (it doesn't mention anything about the padding).

src/qibocal/auto/transpile.py
```diff
@@ -52,87 +59,106 @@ def _pad_circuit(nqubits: int, circuit: Circuit, qubit_map: list[int]) -> Circui
...
 
-def _transpile_circuits(
-    circuits: list[Circuit],
-    qubit_maps: list[list[QubitId]],
+def _transpile_circuit(
+    circuit: Circuit,
+    qubit_map: list[int],
     platform: Platform,
     transpiler: Passes,
 ) -> list[Circuit]:
```
RoyStegeman:
```suggestion
) -> Circuit:
```

If you use a type checker such as [pyright](https://microsoft.github.io/pyright/#/), it would have pointed this out.

src/qibocal/auto/transpile.py
```diff
@@ -17,19 +19,23 @@
     "MZ": "M",
 }
 
+QubitMap = list[QubitId]
+"""An array where the elements are physical qubit IDs (str/int) and the indices are logical qubit IDs
+on the corresponding circuit
```
RoyStegeman:
```suggestion
```

Since we're talking about qubit IDs in transpilation it's clear that there is a circuit involved. The `QubitMap` object does not contain information about what the corresponding circuit is (e.g. it's not a dict from circuit to a list of `QubitID`), so mentioning a corresponding circuit I think is only confusing.

src/qibocal/auto/transpile.py
```diff
@@ -17,19 +19,23 @@
...
+"""
+ResultMap = dict[QubitId | tuple[QubitId, ...], list[Counter[str]]]
+"""A dictionary mapping the physical qubit ID(s) measured to an array of state counts per requested measurement
+on the corresponding circuit
```
RoyStegeman:
```suggestion
```

Same as above

src/qibocal/protocols/two_qubit_interaction/chsh/protocol.py
```diff
@@ -101,16 +102,14 @@ class CHSHData(Data):
...
+    ):
+        """Store output."""
 
-        # Add zero is state do not appear in state
-        # could be removed by using high number of shots
-        for i in COMPUTATIONAL_BASIS:
-            if i not in frequencies:
-                frequencies[i] = 0
+        for state in COMPUTATIONAL_BASIS:
+            freq = frequencies.get(state, 0)
```
RoyStegeman: > Counter objects have a dictionary interface except that they return a zero count for missing items instead of raising a [KeyError](https://docs.python.org/3/library/exceptions.html#KeyError):

see https://docs.python.org/3/library/collections.html#collections.Counter

src/qibocal/auto/transpile.py
```diff
@@ -197,88 +195,122 @@ def _execute_circuits(
...
-    qubit_maps: list[list[QubitId]],
     platform: Platform,
     transpiler: Passes,
     compiler: Compiler,
     nshots: int,
+    qubit_map: QubitMap | None = None,
+    qubit_maps: list[QubitMap] | None = None,
     averaging_mode: AveragingMode = AveragingMode.SINGLESHOT,
-) -> list[Counter[str]]:
+) -> list[ResultMap]:
```
RoyStegeman, line 159: Given the function name, I think it makes sense to allow only `qubit_maps` as input. Then if the function is called elsewhere, we can pass a list with only a single item if needed. That also allows you to get rid of the `if-elif` block (which arguably should have been `if-else` or `if-elif-else`).

src/qibocal/auto/transpile.py
```diff
@@ -197,88 +195,122 @@ def _execute_circuits(
...
     """Execute multiple quantum circuits.
 
-    Combines :func:`transpile_circuits` and :func:`execute_circuits` into a single call.
+    The circuits are first padded to fit the number of qubits in the platform
+    and rearranged according to the qubit mapping before being transpiled and executed.
```
RoyStegeman:
```suggestion
    The circuits are merged into a single circuit with the number of qubits as defined in the platform, using the provided physical-to-logical mapping from `qubit_maps`.  Then the large circuit is transpiled, compiled, and finally executed in a single call. 
```

This is just intended as an example so feel free to make your own changes, but I think it can be explained a clearer.

src/qibocal/protocols/tomographies/two_qubit_state_tomography.py
```diff
@@ -119,57 +118,46 @@ def _acquisition(
...
-            for i in range(len(targets))
-        )
-        basis_circuit.add(meas for pair in pairs for meas in pair)
+        for pair in logical_pairs:
+            if basis1 != "Z":
+                basis_circuit.add(getattr(gates, basis1)(pair[0]).basis_rotation())
+            if basis2 != "Z":
+                basis_circuit.add(getattr(gates, basis2)(pair[1]).basis_rotation())
 
+        basis_circuit.add([gates.M(*logical_pair) for logical_pair in logical_pairs])
```
RoyStegeman: According to the typehints `Circuit.add` takes as input a `Gate` not a list of gates. If the typehint is correct, the for-loop should be around the .add call

https://github.com/qiboteam/qibo/blob/c0521ba548d70b598e5c44c81850bbacc43ebfcb/src/qibo/models/circuit.py#L608

src/qibocal/protocols/tomographies/two_qubit_state_tomography.py
```diff
@@ -119,57 +118,46 @@ def _acquisition(
...
+            qubit_map=qubits,
         )
 
-        for i, pair in enumerate(targets):
-            frequencies = marginalize_qubit_counts(result, (2 * i, 2 * i + 1))
+        for logical_pair, pair in zip(logical_pairs, targets):
+            [frequencies] = result[pair]
             simulation_probabilities = simulation_result.probabilities(
-                qubits=(2 * i, 2 * i + 1)
+                qubits=logical_pair
```
RoyStegeman: `logical_pair` is a tuple, but according to the typehints of `probabilities` only `list` or `set` is accepted. I know this was even mismatched before, but since we're touchting it again, let's fix it here (or even better, if the function can also handle tuples, in qibo).

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -53,12 +53,13 @@ class IndexedCircuit(BaseModel):
...
 
 class IndexedResult(BaseModel):
-    """An execution result paired with its (depth, iteration) CircuitIndex."""
+    """An execution result paired with its (depth, iteration) CircuitIndex and qubit."""
 
     model_config = ConfigDict(frozen=True, extra="forbid")
 
     result: Counter
     index: CircuitIndex
+    qubit: QubitId | QubitPairId
```
RoyStegeman:
```suggestion
    target: QubitId | QubitPairId
```

For consistency

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -384,10 +385,10 @@ def _generate_indexed_circuits(
     two_qubit = isinstance(targets[0], tuple)
     nqubits = len(targets) * (2 if two_qubit else 1)
 
-    target_id_map = {
+    logical_qubit_map = {
         target: (idx * 2, idx * 2 + 1) if two_qubit else (idx,)
         # Reverse assignment for little-endianess resolution
```
RoyStegeman:
```suggestion
```

src/qibocal/auto/transpile.py
```diff
@@ -17,19 +19,23 @@
     "MZ": "M",
 }
 
+QubitMap = list[QubitId]
+"""An array where the elements are physical qubit IDs (str/int) and the indices are logical qubit IDs
+on the corresponding circuit
+"""
+ResultMap = dict[QubitId | tuple[QubitId, ...], list[Counter[str]]]
```
RoyStegeman:
```suggestion
ResultMap = dict[QubitId | QubitPairId, list[Counter[str]]]
```

This is more general than allowed by for example `IndexedResult`, which expects `QubitId | QubitPairId`. Either downstream should be made more flexible (or `assert` placed), or `ResultMap` more restrictive. Since for as far as I know we only do experiments for single or two qubit, I believe it is fine to make `ResultMap` more restrictive.

tests/test_transpile.py
```diff
@@ -90,24 +97,102 @@ def mock_transpiler(circuit):
     assert np.all(expected.unitary() == transpiled_circuit.unitary())
 
 
+def test_execute_circuits_qubit_mapping():
+    platform = create_platform("dummy")
+    compiler = build_native_gate_compiler(platform)
+    transpiler = build_native_gate_transpiler(platform)
+
+    circuit = Circuit(2)
+    circuit.add(gates.M(*range(2)))
```
RoyStegeman, line 42: Maybe we can also test a circuit contains multiple pairs and multiple individual qubits

src/qibocal/auto/transpile.py
```diff
@@ -41,7 +47,8 @@ def _pad_circuit(nqubits: int, circuit: Circuit, qubit_map: list[int]) -> Circui
     Args:
         nqubits: The total number of qubits in the new circuit.
         circuit: The original quantum circuit to be padded.
-        qubit_map: A list mapping physical qubits to logical qubits in the new circuit.
+        qubit_map: A list mapping the logical qubits in the original circuit to the
+            logical qubits in the padded circuit.
```
RoyStegeman: This does not agree with the description above, which still mentions phsyical qubits. Also, I'm not sure this change is correct since `_pad_circuit` is only called in `_transpile_circuit`, which in turn is only called in `execute_circuits`. `qubit_map(s)` is simply propagated from there and in `execute_circuit` the description is _An array of physical qubit to logical qubit mapping per circuit_.

src/qibocal/auto/transpile.py
RoyStegeman: To keep things uniform, please wrap docstrings at column 88. This is for all docstrings, not just here.

src/qibocal/auto/transpile.py
```diff
@@ -52,87 +59,106 @@ def _pad_circuit(nqubits: int, circuit: Circuit, qubit_map: list[int]) -> Circui
...
+    Iterates across the requested measurements and fetches the corresponding results as a count of states.
+
+    Args:
+        platform_qubit_map: A list where the indices are logical qubit indices and the values are physical qubit IDs.
+        readout: Results from circuit execution.
+        measurement_map: Map of measurement registers to measurement pulse sequences for the current circuit.
+        nshots: Number of shots requested.
+
+    Returns:
+        A dictionary mapping the physical qubit IDs measured to an array of state counters for the current circuit.
```
RoyStegeman: There is not a need to specify all arguments for internal functions, just enough to make clear what the purpose of the function is. If that is best achieved by explicitly writing out all the arguments, then do that, but I don't believe it to be the case here.

src/qibocal/auto/transpile.py
```diff
@@ -52,87 +59,106 @@ def _pad_circuit(nqubits: int, circuit: Circuit, qubit_map: list[int]) -> Circui
...
+    for measure, sequence in measurement_map.items():
+        _validate_measurement(measure, sequence, readout)
+        if len(measure.qubits) == 1:
+            qid = platform_qubit_map[measure.qubits[0]]
+            result = readout[sequence.acquisitions[0][1].id]
+            measurements[qid].append(
+                Counter({str(key): val for key, val in Counter(result).items()})
+            )
+        else:
+            result = {}
```
RoyStegeman:
```suggestion
```

src/qibocal/auto/transpile.py
```diff
@@ -52,87 +59,106 @@ def _pad_circuit(nqubits: int, circuit: Circuit, qubit_map: list[int]) -> Circui
...
+            qid = tuple([platform_qubit_map[qubit_id] for qubit_id in measure.qubits])
+            arr = np.stack(
+                [readout[pulse.id] for (_, pulse) in sequence.acquisitions]
+            ).astype(int)
+            measurements[qid].append(Counter("".join(map(str, col)) for col in arr.T))
```
RoyStegeman: Except for defining `qid`, the `if`-branch can be handled by the logic in the `else`-branch. It probably makes sense to limit the `if-else` to only handling of `qid`.

RoyStegeman (commented):
Thanks again @sorewachigauyo, this is nice improvement. Since I don't have a platform available to check myself, did you run the protocols to make sure the results are still sensible?

src/qibocal/protocols/tomographies/two_qubit_state_tomography.py
sorewachigauyo: The comment of https://github.com/qiboteam/qibo/blob/c0521ba548d70b598e5c44c81850bbacc43ebfcb/src/qibo/models/circuit.py#L614 and the logic of https://github.com/qiboteam/qibo/blob/c0521ba548d70b598e5c44c81850bbacc43ebfcb/src/qibo/models/circuit.py#L622-L624 allows it to accept an iterable of gates.

Should we respect the typehint here or open a PR in Qibo to address that?

sorewachigauyo:
> Thanks again @sorewachigauyo, this is nice improvement. Since I don't have a platform available to check myself, did you run the protocols to make sure the results are still sensible?

Yes, I ran all the protocols on my QPU and attached the results in original comment https://github.com/qiboteam/qibocal/pull/1532#issue-4628855550

src/qibocal/auto/transpile.py
sorewachigauyo: > This does not agree with the description above, which still mentions physical qubits.

Sure, we can change the name to something like `pad_map` or `full_circuit_map`.

> Also, I'm not sure this change is correct since _pad_circuit is only called in _transpile_circuit, which in turn is only called in execute_circuits.

I mean, what the function is doing is mapping qubit `0, 1, 2 ...` of the original circuit to `qubit_map[0], qubit_map[1], qubit_map[2], ...` of the `platform.nqubits` circuit without knowledge of the physical qubit IDs. In that sense, we are purely working with logical qubit indices. 

> qubit_map(s) is simply propagated from there 

`qubit_map(s)` is not directly propagated from `execute_circuits`. It is first passed to `_string_to_integer_qubit_map(s)`, which produces the described logical mapping (in the case of string `QubitId`s).

src/qibocal/protocols/tomographies/two_qubit_state_tomography.py
RoyStegeman: > Should we respect the typehint here or open a PR in Qibo to address that?

In Qibo would be the more suitable place to address it since that's where the signature is not correct, but for this PR also just satisfying the typehints here is fine, so I'll leave the choice to you.

src/qibocal/auto/transpile.py
sorewachigauyo: > This way the transpilation is done for each circuit. Can it be done only once on new circuit after constructing it from the sub-circuits?

Currently it is done only once per circuit and during the `execute_circuits` call.

If you want to provide more agency, we could add a `verbatim` flag similar to `qibo-client` and allow the consumer to handle their own transpilation. I am planning to implement duplicate subcircuit fusion for RB to speedup the protocol and this could be one use, since the transpilation is iterating across all the circuit gates (and we can reduce that time from `ntargets x ngates` to `ngates`).

src/qibocal/auto/transpile.py
sorewachigauyo: Implemented in https://github.com/qiboteam/qibocal/pull/1532/commits/53424875c71538079903aa20d524ef0388375c8e

src/qibocal/auto/transpile.py
sorewachigauyo: Fixed in https://github.com/qiboteam/qibocal/pull/1532/commits/53424875c71538079903aa20d524ef0388375c8e

src/qibocal/auto/transpile.py
sorewachigauyo: Fixed in https://github.com/qiboteam/qibocal/pull/1532/commits/7fc3291f77789acdb821a4800af67d2f6d00e3ed

src/qibocal/auto/transpile.py
sorewachigauyo: Fixed in https://github.com/qiboteam/qibocal/pull/1532/commits/7fc3291f77789acdb821a4800af67d2f6d00e3ed

src/qibocal/auto/transpile.py
sorewachigauyo: Moved transpilation into `execute_circuits` directly in https://github.com/qiboteam/qibocal/pull/1532/changes/7fc3291f77789acdb821a4800af67d2f6d00e3ed

src/qibocal/auto/transpile.py
sorewachigauyo: Implemented in https://github.com/qiboteam/qibocal/pull/1532/changes/7fc3291f77789acdb821a4800af67d2f6d00e3ed

src/qibocal/auto/transpile.py
sorewachigauyo: Anyway, the padding is moved to `execute_circuits` in https://github.com/qiboteam/qibocal/pull/1532/changes/7fc3291f77789acdb821a4800af67d2f6d00e3ed

src/qibocal/auto/transpile.py
sorewachigauyo: This function is removed in https://github.com/qiboteam/qibocal/pull/1532/changes/7fc3291f77789acdb821a4800af67d2f6d00e3ed

src/qibocal/auto/transpile.py
sorewachigauyo: Fixed in https://github.com/qiboteam/qibocal/pull/1532/changes/809df25ae25778975cf958636eb066abfdaa7372

src/qibocal/auto/transpile.py
sorewachigauyo: Fixed in https://github.com/qiboteam/qibocal/pull/1532/changes/809df25ae25778975cf958636eb066abfdaa7372

src/qibocal/auto/transpile.py
sorewachigauyo: Fixed in https://github.com/qiboteam/qibocal/pull/1532/changes/809df25ae25778975cf958636eb066abfdaa7372

src/qibocal/auto/transpile.py
sorewachigauyo: Implemented in https://github.com/qiboteam/qibocal/pull/1532/changes/7fc995132d68efc4d42bb0c4151cb2e4fc5fd59c

src/qibocal/protocols/randomized_benchmarking/utils.py
sorewachigauyo: Fixed in https://github.com/qiboteam/qibocal/pull/1532/commits/f73c53be69d29dc8d33d75dab593aaef1dc494be

src/qibocal/protocols/randomized_benchmarking/utils.py
sorewachigauyo: Fixed in https://github.com/qiboteam/qibocal/pull/1532/commits/f73c53be69d29dc8d33d75dab593aaef1dc494be

src/qibocal/protocols/tomographies/two_qubit_state_tomography.py
sorewachigauyo: Fixed in https://github.com/qiboteam/qibocal/pull/1532/commits/915551972fce21b49b92ad207f66a76c9216efdc

src/qibocal/protocols/tomographies/two_qubit_state_tomography.py
sorewachigauyo: Fixed in https://github.com/qiboteam/qibocal/pull/1532/commits/915551972fce21b49b92ad207f66a76c9216efdc

RoyStegeman:
Thanks @sorewachigauyo I'll have a look. Did you notice the test failing? Nvm, seems it was a fluke

sorewachigauyo:
It seems to be on an unrelated test on windows py312, the other tests were passing by then.

I checked on my end before I pushed.

tests/test_transpile.py
```diff
@@ -21,93 +26,86 @@ def test_natives():
...
 
-    true_circuit = Circuit(5)
-    true_circuit.add(gates.GPI2(1, np.pi / 2))
-    true_circuit.add(gates.GPI2(1, np.pi / 2))
-    true_circuit.add(gates.GPI2(2, np.pi / 2))
-    true_circuit.add(gates.GPI2(2, np.pi / 2))
-    true_circuit.add(gates.Z(1))
-    true_circuit.add(gates.Z(2))
-    assert np.all(true_circuit.unitary() == transpiled_circuit.unitary())
+    qubit_map = [0, 1]
```
RoyStegeman, line 44:
```suggestion
    qubit_map: QubitMap = [0, 1]
```

src/qibocal/auto/transpile.py
```diff
@@ -1,139 +1,92 @@
...
-    `qubit_maps` contain string ids and in the positive case it
-    remap them in integers, following the ids order provided by the
-    platform.
+    measurements: ResultMap = defaultdict(list)
+    for measure, sequence in measurement_map.items():
+        _validate_measurement(measure, sequence, readout)
+        if len(measure.qubits) == 1:
+            qid = platform_qubit_map[measure.qubits[0]]
+        else:
+            qid = tuple([platform_qubit_map[qubit_id] for qubit_id in measure.qubits])
```
RoyStegeman, line 61:
```suggestion
            len(measure.qubits) == 2
            qid = (
                platform_qubit_map[measure.qubits[0]],
                platform_qubit_map[measure.qubits[1]],
            )
```

src/qibocal/protocols/randomized_benchmarking/utils.py
```diff
@@ -567,16 +561,12 @@ def twoq_rb_acquisition(
...
-                indexed_result.result, [pair_id * 2, pair_id * 2 + 1]
-            )
-            key = (qubit_pair[0], qubit_pair[1], indexed_result.index.depth)
-            survival_counts = (
-                partial_result["00"] if inverse_layer else partial_result["11"]
-            )
-            survival_prob = survival_counts / params.nshots
-            grouped[key].append(survival_prob)
+        result = indexed_result.result
+        qubit_pair: QubitPairId = indexed_result.target
```
RoyStegeman, line 565:
```suggestion
        result = indexed_result.result
        assert isinstance(indexed_result.target, tuple)
        qubit_pair: QubitPairId = indexed_result.target
```

src/qibocal/auto/transpile.py
```diff
@@ -1,139 +1,92 @@
...
-    qubits = list(platform.qubits)
-    return [
-        [q if isinstance(q, int) else qubits.index(q) for q in qubit_map]
-        for qubit_map in qubit_maps
-    ]
+QubitMap = list[QubitId]
+"""An array where the elements are physical qubit IDs (str/int) and the indices are
+logical qubit IDs
+"""
+ResultMap = dict[QubitId | QubitPairId, list[Counter[str]]]
```
RoyStegeman, line 27:
```suggestion
ResultMap = dict[QubitId | QubitPairId | tuple[QubitId,...], list[Counter[str]]]
```

Can't this also be `tuple[QubitId,...]` for the readout mitigation matrix protocol?

src/qibocal/protocols/two_qubit_interaction/chsh/protocol.py
```diff
@@ -176,22 +175,22 @@ def _acquisition(
 
         results = execute_circuits(
             circuits,
-            [list(pair)] * len(circuits),
+            [pair],
```
RoyStegeman, line 178: `execute_circuits` now expects a `list[QubitMap]`, so which is not what `[pair]` is. in principle this should be fine since you want to use the pair as a list of `QubitMap`, but you may want to explicitly cast it to a list `list(pair)`.

RoyStegeman (commented):
Thanks, I think it looks good up to a few issues with inconsistent use of types.

src/qibocal/auto/transpile.py
alecandido: assert?

RoyStegeman: Ah yes indeed

src/qibocal/auto/transpile.py
sorewachigauyo: It was, but we can revert
https://github.com/qiboteam/qibocal/pull/1532/changes/BASE..5a6f955a8e3177c244a50489bad59f52f1fad1e8#r3451643784

PR #1533, opened 2026-06-10
qubit flux: use arbitrary units and require drive amplitude

### LINKED ISSUE #1450: qubit flux shouldn't use pi-pulse amplitude
qubit-flux by default uses the amplitude of the drive pulse. This is not a sensible choice since someone working with a platform/chip, may go though the following steps 1) start with determining the sweetspot using long drive pulses with low amplitude 2) calibrating the pi-pulse and possibly more before 3) repeating the qubit flux. During 3) the drive pulse is a mandatory input to the `qubit_flux` protocol, and this duration is generally set much longer than the final drive pulse, so this means that the area of the pulse is much larger than it was while the data was taken during 1). This can lead to effects such as seen in the data in the attached figure. 

http://login.qrccluster.com:9000/Jzw7JBBjRdWELgLNzlRPtA==

Image

### LINKED ISSUE #1470: flux bias is in arbitrary units, not volts
The flux bias is given in arbitrary units, so the plot legend should display `[a.u.]` instead of `[V]`.

RoyStegeman:
This PR changes two things for the qubit flux arc experiment: 
- The flux bias is given in arbitrary units, so the plot legend displays `[a.u.]` instead of `[V]`.
- The default drive pulse amplitude and duration are removed. Previously, the drive duration had a default value of 2000ns while the amplitude defaulted to the calibrated pi-pulse amplitude. This was inconsistent and could become problematic after calibration, as calibrated pulses typically have much shorter durations and larger amplitudes than those used for flux arc spectroscopy.

### DIFF
--- a/src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
+++ b/src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
@@ -41,9 +41,9 @@ __all__ = [
 class QubitFluxParameters(utils.FluxFrequencySweepParameters):
     """QubitFlux runcard inputs."""
 
-    drive_amplitude: float
+    drive_amplitude: float = 0.01
     """Amplitude of the drive pulse."""
-    drive_duration: int
+    drive_duration: int = 2000
     """Duration of the drive pulse."""
 
 
--- a/src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
+++ b/src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
@@ -41,9 +41,8 @@ __all__ = [
 class QubitFluxParameters(utils.FluxFrequencySweepParameters):
     """QubitFlux runcard inputs."""
 
-    drive_amplitude: float | None = None
-    """Drive amplitude (optional). If defined, same amplitude will be used in all qubits.
-    Otherwise the default amplitude defined on the platform runcard will be used"""
+    drive_amplitude: float = 0.01
+    """Amplitude of the drive pulse."""
     drive_duration: int = 2000
     """Duration of the drive pulse."""
 
@@ -127,8 +126,7 @@ def _acquisition(
         ro_channel, ro_pulse = natives.MZ()[0]
 
         qd_pulse = replace(qd_pulse, duration=params.drive_duration)
-        if params.drive_amplitude is not None:
-            qd_pulse = replace(qd_pulse, amplitude=params.drive_amplitude)
+        qd_pulse = replace(qd_pulse, amplitude=params.drive_amplitude)
 
         qd_pulses[q] = qd_pulse
         ro_pulses[q] = ro_pulse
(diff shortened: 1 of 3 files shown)


scarrazza (approved):

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
```diff
@@ -41,10 +41,9 @@
 class QubitFluxParameters(utils.FluxFrequencySweepParameters):
     """QubitFlux runcard inputs."""
 
-    drive_amplitude: float | None = None
-    """Drive amplitude (optional). If defined, same amplitude will be used in all qubits.
-    Otherwise the default amplitude defined on the platform runcard will be used"""
-    drive_duration: int = 2000
+    drive_amplitude: float
+    """Amplitude of the drive pulse."""
+    drive_duration: int
```
alecandido: Any reason for this removal?

It's true that it is an arbitrary value, but it is in a decent range: long enough to cover multiple pi-pulses, and short enough to fit on any electronics.

Removing it is an option, but I wonder if it becomes more verbose without a practical advantage.
But it's also marginally more verbose... more a matter of UI...

RoyStegeman: For consistency, I think `drive_amplitude` and `drive_duration` should behave the same, so either both a default or neither a default. Since I think there is some value in forcing users to explicitly write out their input choices and don't believe the verbosity to be such a problem I went with the latter.

alecandido: I do not really see why they should behave the same. The `drive_amplitude` is subject to a tradeoff, since it should be large enough to excite the qubit, but not too large to excite other transitions. Instead, the `drive_duration` is just bound to be much larger of the typical pi-pulse duration (for that amplitude).

While it is true that the pi-pulse scale is related to the product of the two, the tradeoff for the amplitude is such that a few microseconds is a good value for the duration. With a broad enough range of sensible values.

To be more precise, I believe we could have a natural scale for both amplitude and duration at the qubit level. The exact optimal value for the amplitude would depend on the drive line coupling (which sets the Rabi frequency), but in practice that is often not changing by orders of magnitude. And the lower bound for the duration is just a consequence of it.

In any case, as I said, removing this value we are just transferring this knowledge of the "few microseconds" from Qibocal to its user. With the advantage of having a possibly more accurate estimate (since chip-adapted), even before any measurement. But with the disadvantage of a more verbose interface.

RoyStegeman: I put back default values [Update qubit_flux_dependence.py](https://github.com/qiboteam/qibocal/pull/1533/commits/5954b8dbdf089ffa60c3ca896904851f8caee0eb)

alecandido (approved):

PR #1536, opened 2026-06-11
Rename the `Routine` class to `Protocol`

alecandido:
Just doing that and nothing else, trying to establish consistent naming for this object.

### DIFF
--- a/doc/source/protocols/dispersive_shift.rst
+++ b/doc/source/protocols/dispersive_shift.rst
@@ -31,7 +31,7 @@ where we introduced the dispersive shift
 Equation :eq:`eq_1` shows that the resonator frequency is :math:`\omega_{r,0} = \omega_r - \chi` (:math:`\omega_{r,1} = \omega_r + \chi`) when the
 qubit is in the ground (excited) state. The separation between the two freqiencies is :math:`\lvert 2 \chi \rvert`.
 
-Routine description
+Protocol description
 ^^^^^^^^^^^^^^^^^^^
 After collecting the data from the two spectroscopies, for each readout frequency the distance of the centers of the blobs for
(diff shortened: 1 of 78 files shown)


jevillegasd:
I dont mind ``Routine``, but ``Completed`` has  a loooong due renaming.

Can we still a add a class ``Routine`` that inherits ``Protocol`` for compatibility (I'm not sure how to do that with ``Generic`` objects)., and maybe intercept ``Routine`` instantiation and log a deprecation warning.

This change will certainly break some stuff in my dummy dashbaord :(

alecandido:
The reason to make this fix now is that we have a single unambiguous name, and we are going to work more in the interface and the docs.
Thus, we make the change immediately, and we start settling on a more stable interface.

> This change will certainly break some stuff in my dummy dashbaord :(

This can be fixed quite easily: use the LSP to rename the variable in your entire project.
However, if for any reason does not work, just do
```py
from qibocal import Protocol
Routine = Protocol
```

I could even add it here, and export both `Routine` and `Protocol`. But at some point we want to remove `Routine`, and I do not see any reason why later is better than now.
But if you have, we can schedule the removal and provide the alias.

> but `Completed` has a loooong due renaming.

For this, I do not have immediately a replacement. Most likely, it should be something like `Results`, but it would clash internally with the output of the fit.

At the same time, `Completed` is not expected to be used directly outside Qibocal. You will receive `Completed` objects from Qibocal's protocols. But you will likely not create them yourself, outside Qibocal.
So, I do not see an urgent need to replace it. We can think about it when we will work closer to that.

lballerio (approved):

scarrazza (approved):

PR #1540, opened 2026-06-15
spin lock sequence

jevillegasd:
This is a first trial to address #1525 

The experiment runs, but I run out of memory surprisingly fast, even though I use sweepers for both amplitude and time iterators. Maybe this works better if I use of of the iterations to be a for loop outside (at the cost of the experiment being longer than necessary).


jevillegasd:
Im kind of stuck trying to make the protocol run for long durations of the spin lock pulse. I see that the driver supports [batching](https://github.com/qiboteam/qibolab/blob/main/src/qibolab/_core/instruments/qblox/batching.py) , would this be a good way to allow longer sequences.

alecandido:
> Im kind of stuck trying to make the protocol run for long durations of the spin lock pulse. I see that the driver supports [batching](https://github.com/qiboteam/qibolab/blob/main/src/qibolab/_core/instruments/qblox/batching.py?rgh-link-date=2026-06-15T15%3A52%3A40Z) , would this be a good way to allow longer sequences.

Just for the sake of future reference: batching is quite the opposite. It is a way to accelerate the execution of short sequences, but quite ineffective on long ones.

For very long sequences, we just need to identify the limiting factor (instruction memory, waveform memory, acquisition memory) and check whether there is a specific implementation lifting the bottleneck.

In this case, the sequence is in a sense small, since it is made by few operations. The reason it is failing is because of waveform memory limitations, which can only be lifted by not realizing the spin-locking pulse as a single arbitrary waveform. Which can be realized by either decomposing it in parts (e.g. repeating many back-to-back identical rectangular units) or with a compressed pulse implementation (which is possible for the flux pulse).

src/qibocal/protocols/coherence/spin_lock.py
```diff
@@ -0,0 +1,408 @@
...
+@dataclass
+class SpinLockParameters(Parameters):
+    """SpinLock runcard inputs."""
+
+    duration_min: int
+    """Minimum spin-lock pulse duration [ns]."""
+    duration_max: int
+    """Maximum spin-lock pulse duration [ns]."""
+    duration_step: int
+    """Step spin-lock pulse duration [ns]."""
```
lballerio, line 60: we decided to move into ranges as input, i.e. tuples of the form `(init_value, final_value, step)`.
```suggestion
    duration_range: tuple[float, float, float]
    """Spin-lock pulse duration range [ns]."""
```
Also in theory we can admit times to be float for sub-ns sampling rate; then drivers will convert it accordingly.

src/qibocal/protocols/coherence/spin_lock.py
```diff
@@ -0,0 +1,408 @@
...
+    """Maximum spin-lock pulse amplitude [a.u.]."""
+    n_amplitudes: int
+    """Number of amplitude points, log-spaced between amplitude_min and amplitude_max."""
+    sequence: Literal["SL3", "SL5a", "SL5b"] = "SL3"
+    """Spin-lock sequence type."""
+
+    @property
+    def duration_range(self) -> tuple[int, int, int]:
+        """Return a tuple with the spin-lock pulse duration range."""
+        return self.duration_min, self.duration_max, self.duration_step
```
lballerio, line 73: noy necessary anymore since we are taking range as input.
```suggestion
```

src/qibocal/protocols/coherence/spin_lock.py
```diff
@@ -0,0 +1,408 @@
...
+    duration_max: int
+    """Maximum spin-lock pulse duration [ns]."""
+    duration_step: int
+    """Step spin-lock pulse duration [ns]."""
+    amplitude_min: float
+    """Minimum spin-lock pulse amplitude [a.u.]."""
+    amplitude_max: float
+    """Maximum spin-lock pulse amplitude [a.u.]."""
+    n_amplitudes: int
+    """Number of amplitude points, log-spaced between amplitude_min and amplitude_max."""
```
lballerio, line 66: Honestly here I don't know whether is better to ask for a tuple or not. However I will write `amplitude_values` as a `@property` of the `SpinLockParameters` class.

src/qibocal/protocols/coherence/spin_lock.py
```diff
@@ -0,0 +1,408 @@
...
+@dataclass
+class SpinLockResults(Results):
+    """SpinLock outputs."""
+
+    t1rho: dict[QubitId, list[list[float]]] = field(default_factory=dict)
+    """:math:`T_{1\\rho}` and its error for each spin-lock amplitude [ns]."""
+    gamma: dict[QubitId, list[float]] = field(default_factory=dict)
+    """Relaxation rate :math:`\\Gamma_{1\\rho} = S(\\nu_R)` for each amplitude [Hz]."""
+    gamma_error: dict[QubitId, list[float]] = field(default_factory=dict)
+    """Error on :math:`\\Gamma_{1\\rho}` [Hz]."""
```
lballerio, line 85: Maybe better to save them together?
Something like:

```suggestion
    gamma: dict[QubitId, list[float, float]] = field(default_factory=dict)
    """Relaxation rate :math:`\\Gamma_{1\\rho} = S(\\nu_R)` and its error for each amplitude [Hz]."""
```

src/qibocal/protocols/coherence/spin_lock.py
```diff
@@ -0,0 +1,408 @@
...
+
+
+def _acquisition(
+    params: SpinLockParameters,
+    platform: CalibrationPlatform,
+    targets: list[QubitId],
+) -> SpinLockData:
+    """Data acquisition for the spin-lock (T1rho) experiment."""
+
+    duration_range = np.arange(*params.duration_range)
```
lballerio, line 143: I would call in a different way since it is not a range and then might be confused with `params.duration_range`.

src/qibocal/protocols/coherence/spin_lock.py
```diff
@@ -0,0 +1,408 @@
...
+    pi_pulse_amplitude: dict[QubitId, float] = {}
+    _infra: dict[
+        QubitId, tuple
+    ] = {}  # pulse infrastructure: (y90, echo, qd_channel, ro_channel, ro_pulse)
+
+    for q in targets:
+        natives = platform.natives.single_qubit[q]
+        qd_channel = platform.qubits[q].drive
+        ro_channel, ro_pulse = natives.MZ()[0]
+        rx_pulse = natives.RX()[0][1]
```
lballerio, line 164: Is there a reason why not:

```python
ro_channel, ro_pulse = natives.MZ()[0]
rx_pulse, qd_channel = natives.RX()[0][1]
```
that I cannot see?

src/qibocal/protocols/coherence/spin_lock.py
```diff
@@ -0,0 +1,408 @@
...
+            for q in targets
+        ]
+
+        results = platform.execute(
+            [sequence],
+            [duration_sweeper],
+            nshots=params.nshots,
+            relaxation_time=params.relaxation_time,
+            acquisition_type=AcquisitionType.DISCRIMINATION,
+            averaging_mode=AveragingMode.SINGLESHOT,
```
lballerio, line 237: Is there a reason why you set  `SINGLESHOT` as acquisition mode and not `CYCLIC`?

src/qibocal/protocols/coherence/spin_lock.py
```diff
@@ -0,0 +1,408 @@
...
+        )
+
+        for q in targets:
+            prob = probability(results[ro_pulses[q].id], state=1)
```
lballerio, line 241: if you use `CYCLIC`:

```suggestion
            prob = results[ro_pulses[q].id]
```

src/qibocal/protocols/coherence/spin_lock.py
```diff
@@ -0,0 +1,408 @@
...
+
+    return data
+
+
+def _fit(data: SpinLockData) -> SpinLockResults:
+    """Extract :math:`\\Gamma_{1\\rho}(\\nu_R)` from per-amplitude exponential decays."""
+
+    t1rho = {}
+    gamma = {}
+    gamma_error = {}
```
lballerio, line 262: as before, maybe error and measurement can be incorporated

src/qibocal/protocols/coherence/spin_lock.py
```diff
@@ -0,0 +1,408 @@
...
+                gamma[qubit].append(np.nan)
+                gamma_error[qubit].append(np.nan)
+                fitted_parameters[qubit].append([np.nan, np.nan, np.nan])
+
+        rabi_frequency[qubit] = data.rabi_frequency(qubit).tolist()
+
+    return SpinLockResults(t1rho, gamma, gamma_error, rabi_frequency, fitted_parameters)
+
+
+def _plot(data: SpinLockData, target: QubitId, fit: SpinLockResults = None):
```
lballerio, line 299:
```suggestion
def _plot(data: SpinLockData, target: QubitId, fit: SpinLockResults | None = None):
```

lballerio (changes requested):
Overall I think is good, just small suggestions.
Btw @jevillegasd can you also paste on this PR some data plot you generated while running this experiment?

lballerio:
Btw also rebase (in this case is straightforward) and try to solve the failing tests

Issue #1527, opened 2026-06-09
Spin Lock Noise Spectroscopy


alecandido:
<div type='discussions-op-text'>

<sup>Originally posted by **@jevillegasd** June  8, 2026 in https://github.com/qiboteam/qibocal/discussions/1525</sup>
The idea would be to implement a noise spectroscopy routine based on rotating-frame relaxation ($T_{1\rho}$) to characterize high-frequency environmental noise acting on the qubits during driven evolution, complementing existing free-evolution routines (Ramsey, Hahn-Echo). A version of this was presented by Tatiana some time ago and used to measure noise in the 20Q QPU.

### Background

Some references are [Yan & Oliver (2013)](https://www.nature.com/articles/ncomms3337)  and [Yan & Oliver (2016)](https://www.nature.com/articles/ncomms12964)

Standard continuous-driving Rabi experiments are vulnerable to low-frequency 1/f drifts, causing artifacts in noise reconstruction. By employing a spin-locking sequence, the qubit state is locked parallel to the driving field in the rotating frame. This shifts the decoherence mechanism to longitudinal relaxation ($T_{1\rho}$), filtering out low-frequency phase noise and allowing clean mapping of the noise power spectral density (PSD) at the Rabi frequency $\nu_R$. This is critical for detecting coherent microscopic two-level system (TLS) defects.

### Proposed Pulse Sequence Implementation

The basic implementation folows:
- Preparation: Apply a resonant $\pi/2$ pulse along the $Y$-axis to bring the qubit state to the equator (aligned with the $X$-axis).
- Spin-Lock: Apply a continuous driving field along the $X$-axis (phase-shifted by $90^\circ$ relative to the first pulse) for a variable duration $\tau$. The drive amplitude dictates the target spectroscopy frequency $\nu_R$. The Rabi frequency can be calculate scaling from the pi pulse value.
- Back-Projection: Apply an additional $\pi/2$ pulse along the $Y$ axis to project for  $Z$-axis readout.

To prevent state oscillations caused by phase diffusion during unavoidable inter-pulse gaps, the paper suggests variants (SL-5 a/b) 

### Expected Parameters and Outputs
- Sweep Parameters: Spin-lock duration ($\tau$) and driving amplitude (Rabi frequency $\nu_R$).
- Fits & Plots:Exponential decay fits of qubit population vs. $\tau$ to extract the relaxation rate $\Gamma_{1\rho}(\nu_R)$.
[...]

PR #1543, opened 2026-06-16
Use LongPulse instead of rectangular pulses

jevillegasd:
In this version, I use a LongPulse. This long pulse could be implemented using a wait instead of waveforms and using the offsets of the AWG instead of the waveforms. This approach applicable for rectangular pulses only (and the time resolution is 1 ns (the wait duration) but the precision is 4 ns (the time needed to apply the change in the offsets of the AWG).

A possible path to this is discussed in this [issue](https://github.com/qiboteam/qibocal/issues/1527#issuecomment-4716560580)

### DIFF
--- a/src/qibocal/protocols/coherence/spin_lock.py
+++ b/src/qibocal/protocols/coherence/spin_lock.py
@@ -29,11 +29,11 @@ from qibolab import (
     AcquisitionType,
     AveragingMode,
     Delay,
+    LongPulse,
     ParallelSweepers,
     Parameter,
     Pulse,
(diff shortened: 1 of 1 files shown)


alecandido:
As discussed in private, but to keep the log.

While I hope the solution is working at low-level, the interface should be refined. In particular, `LongPulse` does not contain more information than a regular `Pulse(..., envelope=Rectangular()`.
So, we do not need a new interface for this object, and the distinction can be kept internal to the driver (discriminating the implementation, if the optimized one is available).

PR #1547, opened 2026-06-18
mixer cal: use `_configure_sequencers` instead of `configure` 

RoyStegeman:
This reflects a change in qibolab (https://github.com/qiboteam/qibolab/pull/1494). While the mixer calibration now explicitly depends on a piece of non-public API, I don't believe it is a problem to make this change. The reason is that mixer calibration in qibocal is already hardware dependent while qibocal should be hardware-agnostic.


scarrazza (approved):

RoyStegeman:
We would also need `_configure_modules` to set the LO, so for now we'll keep using the full `cluster.configure`. This is overkill as described in this issue https://github.com/qiboteam/qibocal/issues/1550

PR #1548, opened 2026-06-18
Add postselection to readout characterization

sorewachigauyo:
We have been getting poor readout on our 5Q device for a while. On checking further, I noticed that there is a bit of ground state shots in the excited state clusteroid.

image

By adding a pre-measurement and filtering based on that, the readout fidelity improves by quite a bit. (Results not shown because plotting of the postselected data hasn't been implemented yet) 

image

TODO:

- [ ] Tidy initial dirty implementation
- [ ] Streamline characterization protocol
- [ ] Plotting of pre-and-postselected data
- [ ] Tests?

### DIFF
--- a/src/qibocal/protocols/utils.py
+++ b/src/qibocal/protocols/utils.py
@@ -213,8 +213,10 @@ def compute_qnd(
 
     Returns the QND and the two measurement matrices."""
 
-    p_m1 = np.mean([zeros_first_measure, ones_first_measure], axis=1)
-    p_m2 = np.mean([zeros_second_measure, ones_second_measure], axis=1)
+    p_m1 = np.array([np.mean(arr) for arr in [zeros_first_measure, ones_first_measure]])
+    p_m2 = np.array(
+        [np.mean(arr) for arr in [zeros_second_measure, ones_second_measure]]
+    )
 
     lambda_m = np.stack([1 - p_m1, p_m1])
     lambda_m2 = np.stack([1 - p_m2, p_m2])
--- a/src/qibocal/protocols/readout/readout_characterization.py
+++ b/src/qibocal/protocols/readout/readout_characterization.py
@@ -28,6 +28,8 @@ class ReadoutCharacterizationParameters(Parameters):
 
     delay: float = 0
     """Delay between readouts, could account for resonator deplation or not [ns]."""
+    post_selection: bool = False
+    """If true, a pre-measurement is applied and results will be filtered based on said measurement."""
 
 
 @dataclass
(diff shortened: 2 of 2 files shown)


src/qibocal/protocols/utils.py
```diff
@@ -213,8 +213,10 @@ def compute_qnd(
 
     Returns the QND and the two measurement matrices."""
 
-    p_m1 = np.mean([zeros_first_measure, ones_first_measure], axis=1)
-    p_m2 = np.mean([zeros_second_measure, ones_second_measure], axis=1)
+    p_m1 = np.array([np.mean(arr) for arr in [zeros_first_measure, ones_first_measure]])
+    p_m2 = np.array(
+        [np.mean(arr) for arr in [zeros_second_measure, ones_second_measure]]
+    )
```
alecandido, line 219: Why the `for` comprehension? Wasn't it any longer fine with NumPy?

sorewachigauyo: The two arrays have different length because of the postselection, so the original code was complaining

e.g. `zeros_first_measure` is filtered by `zero_premeasurement` but `ones_first_measure` is filtered by a different premeasurement.

alecandido:
But then, if you are dropping states which are excited at the beginning of the experiment, are those states generated by a too-short relaxation? Or is it a residual thermal population? It seems quite a large difference...

In any case, I'm fine with the concept of this PR, but I was considering that we may want to implement the initialization measurement at a more general level, in Qibocal or Qibolab (preferentially the first one), such that it can be toggled for most (or all) experiments, in a similar fashion to the reset (relaxation time).

sorewachigauyo:
> But then, if you are dropping states which are excited at the beginning of the experiment, are those states generated by a too-short relaxation? Or is it a residual thermal population? It seems quite a large difference...
 
For this chip, the relaxation time is 200us when the T1 is about 20-30us. Currently our suspicion is on the thermalization.

> In any case, I'm fine with the concept of this PR, but I was considering that we may want to implement the initialization measurement at a more general level, in Qibocal or Qibolab (preferentially the first one), such that it can be toggled for most (or all) experiments, in a similar fashion to the reset (relaxation time).

Sure, I can work on that in Qibocal.

alecandido:
> For this chip, the relaxation time is 200us when the T1 is about 20-30us. Currently our suspicion is on the thermalization.

A bit unexpected, but I understand (I do not really see any other option).

> Sure, I can work on that in Qibocal.

No worries: right now there is no way to do that without touching each and every protocol. I will write an issue for that, and it may align with a proposed simplification for the protocols' acquisition definition.

PR #1556, opened 2026-06-19
Capture `ZeroDivisionError` in `drag_simple`

RoyStegeman:
If the single shot classification is not done properly it may be that all values in `drag_simple` return either 0 or 1. In this case the denominator when evaluating the optimal beta may be 0 resulting in a `ZeroDivisionError`. Since we still want to plot the data even if the extraction of the optimal beta parameter fails, we capture the `ZeroDivisionError` in a try-except. I tested that it works, even in the case where one qubit has this error while another does not.

http://login.qrccluster.com:9000/YFvfWN4JR4S4det-R-WXGQ==

### DIFF
--- a/src/qibocal/protocols/drag/drag_simple.py
+++ b/src/qibocal/protocols/drag/drag_simple.py
@@ -186,9 +186,16 @@ def _fit(data: DragTuningSimpleData) -> DragTuningSimpleResults:
                 sigma=qubit_data["error"],
             )
             fitted_parameters[qubit, setup] = popt.tolist()
-        betas_optimal[qubit] = -(
-            fitted_parameters[qubit, "YpX9"][1] - fitted_parameters[qubit, "XpY9"][1]
-        ) / (fitted_parameters[qubit, "YpX9"][0] - fitted_parameters[qubit, "XpY9"][0])
+        try:
+            betas_optimal[qubit] = -(
+                fitted_parameters[qubit, "YpX9"][1]
+                - fitted_parameters[qubit, "XpY9"][1]
+            ) / (
+                fitted_parameters[qubit, "YpX9"][0]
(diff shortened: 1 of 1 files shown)


src/qibocal/protocols/drag/drag_simple.py
```diff
@@ -186,9 +186,16 @@ def _fit(data: DragTuningSimpleData) -> DragTuningSimpleResults:
...
+        try:
+            betas_optimal[qubit] = -(
+                fitted_parameters[qubit, "YpX9"][1]
+                - fitted_parameters[qubit, "XpY9"][1]
+            ) / (
+                fitted_parameters[qubit, "YpX9"][0]
+                - fitted_parameters[qubit, "XpY9"][0]
+            )
+        except ZeroDivisionError:
+            pass
```
scarrazza, line 198: What about adding a warning message?

scarrazza (approved):

src/qibocal/protocols/drag/drag_simple.py
RoyStegeman: If there is no optimal beta than the update will fail and a warning will be printed there. 

https://github.com/qiboteam/qibocal/blob/14ed4a26bb35a76009d4fb46d0fd306a1ef1a910/src/qibocal/auto/task.py#L254-L261

scarrazza: Thanks, looks good to me.

alecandido (approved):
Thanks!

RoyStegeman:
@jevillegasd thanks for pointing out this issue to @alecandido. Now it prints the data even if the excited state probability is always 0 or 1 and thus there is no optimal beta value can be determined.

PR #1559, opened 2026-06-22
CPMG  spectroscopy

jevillegasd:
Loop over CPMG sequences with variable number of flips, to apply a filter function that effectively maps the noise spectra density ($$S_{||}(\omega) \propto \\Gamma_2(\omega)$$)


jevillegasd:
Instead of fitting each T2 with a simple exponmential decay we can use

image

from https://arxiv.org/pdf/1101.4707 eq 23

PR #1561, opened 2026-06-23
Rabi Refactor

### LINKED ISSUE #1521: Use `CYCLIC` instead of `SINGLESHOT` for Rabi
https://github.com/qiboteam/qibocal/blob/62361be9f813f8cf623db4a4b9472cd913066396/src/qibocal/protocols/rabi/amplitude.py#L74-L81

Possibly together with the usual batch of updates (plots, code reuse, fits error handling, ...).

### LINKED ISSUE #1697: Change Rabi length error bars on data from continuous bands to point error bars
Image
http://login.qrccluster.com:9000/3WIZvKNkRmqaeARAf4cdMA==

lballerio:
This PR makes the following changes:

1. fix the Rabi documentation after the refactor/renaming
2. overall refactor of `rabi` folder:
            a. `utils.py` has now been splitted in 3 modules: `acquisition.py`, which contains all rabi sequence functions, `processing.py`, which contains functions for plotting and fitting and `parent_classes.py`, which defines the common `Data`, `Parameters` and `Results` classes.
           b. all rabi experiments refactored accordingly
3. introduced a `range-like` convention for input parameters.
4. In all Rabi experiments was inserted `drive_lines` optional input, a list of qubits which indicates the list of physical lines to use for each qubit; also some quality checks of the input are performed:
        a. if defined, `drive_lines` and `targets` must have the same length
 		b. the must not be repeated elements in the union of `drive_lines` and `targets`.
5. Addition of the `microwave_crosstalk_matrix` field in `calibration.py` with also a validator for `Calibration` class and `get` ad `set` function for this matrix.
6. Every `rabi_amplitude` experiment now also saves in `microwave_crosstalk_matrix` the fitted amplitude from the signal. 
7. In `Calibration` class now is also saved the fitted amplitude of non-crosstalk Rabi experiment for each qubit; this is used later to update `microwave_crosstalk_matrix`.
8. deleted flux crosstalk matrix update in `update.py` and propagated this change in `flux_dependence` folder.
9. introduction of `microwave_cancellation_phase.py` module for calibrating mw crosstalk phase.
10. Propagated the changes in `tests` folder.


lballerio:
check #1108 for crosstalk calibration references.

src/qibocal/protocols/rabi/utils.py
```diff
@@ -259,84 +288,134 @@ def period_correction_factor(phase: float):
...
-    platform: Platform,
+    drive_lines: list[QubitId],
+    platform: CalibrationPlatform,
+    pulse_duration: float | None,
+    pulse_ampl: float | None,
     rx90: bool,
-    use_align: bool = False,
-) -> tuple[PulseSequence, dict, dict, dict]:
-    """Return sequence for rabi length."""
+) -> tuple[PulseSequence, dict[QubitId, PulseLike], dict[QubitId, float], dict]:
```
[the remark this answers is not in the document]
lballerio: fixed in 5a1ec995f104f3b1e728eac573cd71fe17c20047.

src/qibocal/calibration/calibration.py
```diff
@@ -161,14 +199,31 @@ class Calibration(Model):
...
+
+        if self.microwave_crosstalk_matrix.matrix.size == 0:
+            self.microwave_crosstalk_matrix = Matrix(
+                matrix=ndarray_serialize(np.eye(self.nqubits))
+            )
+        if self.microwave_crosstalk_matrix.matrix.shape[0] != self.nqubits:
+            raise ValueError("Drive crosstalk matrix must have as many rows as qubits.")
+        self.microwave_crosstalk_matrix.qubits = self.qubits
+
+        return self
```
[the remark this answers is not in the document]
lballerio: addressed in c4aa1333bcfdc5aa9ff2486596b718695dfe1fb4.

src/qibocal/protocols/rabi/parent_classes.py
```diff
@@ -0,0 +1,183 @@
...
+    """Results container for outputs produced by Rabi protocols."""
+
+    drive_lines: dict[QubitId, QubitId]
+    """List of drive line used for each qubit."""
+    length: dict[QubitId, int | list[float]]
+    """Pi pulse duration for each qubit."""
+    amplitude: dict[QubitId, float | list[float]]
+    """Pi pulse amplitude. Same for all qubits."""
+    fitted_parameters: dict[QubitId, dict[str, float]]
+    """Raw fitting output."""
```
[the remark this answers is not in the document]
lballerio: addressed in 5a1ec995f104f3b1e728eac573cd71fe17c20047.

src/qibocal/protocols/rabi/ef.py
```diff
@@ -94,20 +103,19 @@ def _acquisition(
 
     assert not params.rx90, "Rabi ef available only for RX pulses."
 
-    data = RabiAmplitudeEFData(durations=durations, rx90=False)
+    data = RabiEFSignalData(drive_lines=drive_lines, durations=durations, rx90=False)
+
```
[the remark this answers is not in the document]
lballerio: addressed in 5a1ec995f104f3b1e728eac573cd71fe17c20047.

src/qibocal/protocols/rabi/ef.py
```diff
@@ -57,17 +54,29 @@ def _acquisition(
...
 
+        if q != d:
+            # used when q is being driven with another line (cross rabi)
+            cross_channel = platform.qubits[d].drive
+            qubit12_freq = platform.calibration.configs[qd12_channel].frequency
+            updates |= {cross_channel: {"frequency": qubit12_freq}}
+            qd12_channel = cross_channel
+
         if params.pulse_length is not None:
             qd12_pulse = replace(qd_pulse, duration=params.pulse_length)
```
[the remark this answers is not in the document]
lballerio: addressed in 5a1ec995f104f3b1e728eac573cd71fe17c20047.

src/qibocal/protocols/rabi/ef.py
```diff
@@ -57,17 +54,29 @@ def _acquisition(
...
         qd_channel, qd_pulse = natives.RX()[0]
         qd12_channel, qd12_pulse = natives.RX12()[0]
         ro_channel, ro_pulse = natives.MZ()[0]
 
+        if q != d:
+            # used when q is being driven with another line (cross rabi)
+            cross_channel = platform.qubits[d].drive
+            qubit12_freq = platform.calibration.configs[qd12_channel].frequency
+            updates |= {cross_channel: {"frequency": qubit12_freq}}
+            qd12_channel = cross_channel
```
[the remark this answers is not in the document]
lballerio: addressed in 5a1ec995f104f3b1e728eac573cd71fe17c20047.

src/qibocal/protocols/utils.py
```diff
@@ -1086,13 +1086,13 @@ def fallback_frequency(frequency: NDArray) -> NDArray:
...
-    signal: Any,
-    x: Any,
+    signal: ArrayLike,
+    x: ArrayLike,
     axis: int = -1,
     speedup_flag: bool = False,
     iterations: int = 100,
     tol: int = 1e-8,
-) -> NDArray:
+) -> np.floating[Any]:
```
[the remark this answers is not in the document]
lballerio: addressed in 5a1ec995f104f3b1e728eac573cd71fe17c20047.

src/qibocal/protocols/rabi/parent_classes.py
```diff
@@ -0,0 +1,183 @@
+from dataclasses import dataclass, field
+from typing import Any
+
+from qibocal.auto.operation import Data, Parameters, QubitId, Results
+
+
+class InputError(Exception):
+    """Raised when Rabi length protocol input validation fails."""
+
+    pass
```
RoyStegeman: This is more general than Rabi so it would make sense to put it qibocal-level rather than rabi-level. However, from the name it sounds exactly the same as a [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError)

lballerio: yeah I was also thinking it might be good to define some specific errors in a dedicated script, but if we can use already existing classes better to recycle.

src/qibocal/calibration/calibration.py
```diff
@@ -152,6 +159,37 @@ class TwoQubitCalibration(Model):
     """Qubit-qubit coupling."""
 
 
+class Matrix(Model):
+    """Generic matrix indexed by qubits."""
```
alecandido: Instead of this, I would just use a plain array (dense or sparse) for storage, and resort to a local convention to access them - which has to be already the case, since you are not storing the index (i.e. `.qubits` is excluded from serialization). I would avoid wrapping the array - since then you may need to repeatedly unwrap for any kind of array operation

src/qibocal/calibration/calibration.py
```diff
@@ -161,14 +199,59 @@ class Calibration(Model):
     """Dict with qubit pairs calibration."""
     readout_mitigation_matrix: SparseArray | None = None
     """Readout mitigation matrix."""
-    flux_crosstalk_matrix: NdArray | None = None
+    flux_crosstalk_matrix: Matrix | None = Field(default_factory=Matrix)
     """Crosstalk flux matrix."""
+    microwave_crosstalk_matrix: Matrix | None = Field(default_factory=Matrix)
+    """Microwave crosstalk matrix wiht amplitude information."""
+    mw_cancellation_phases_matrix: Matrix | None = Field(default_factory=Matrix)
+    """Microwave crosstalk matrix."""
```
alecandido: Why do you have `None` if the default is a `Matrix`?

In any case, better to keep `None` as the default

alecandido: Do not store separately amplitudes and phases, but just store a single matrix of complex numbers.

src/qibocal/calibration/calibration.py
```diff
@@ -161,14 +199,59 @@ class Calibration(Model):
...
+        self.mw_cancellation_phases_matrix.qubits = self.qubits
+        if self.mw_cancellation_phases_matrix.matrix.shape != (
+            self.nqubits,
+            self.nqubits,
+        ):
+            raise ValueError(
+                "Microwave crosstalk phase matrix must have shape (nqubits, nqubits)."
+            )
+
+        return self
```
alecandido, line 216: Try to avoid as much as possible complex validation. Just simplify the logic of what you save and load

src/qibocal/calibration/serialize.py
```diff
@@ -49,6 +52,9 @@ def ndarray_serialize(ar: npt.NDArray) -> str:
 
 def ndarray_deserialize(x: str | npt.NDArray) -> npt.NDArray:
     """Deserialize array."""
+    if isinstance(x, np.ndarray):
+        return x
+
```
alecandido: Which is the reason for these functions to be idempotent?

alecandido (commented):
For the time being, just about serialization

src/qibocal/calibration/calibration.py
lballerio: Hi @alecandido, yeah actually I was doubting whether to follow this convention or not for quite a while.
I saw it implemented in #1108 and decided to bring it.
This choice was made exclusively to avoid having `get` and `set` functions for both microwave and flux matrices, which would lead to duplicated code for essentially the same operation.

src/qibocal/calibration/calibration.py
lballerio: yeah you are right, the `| None` is unnecessary.
Why do you think is better `None`?
So then we create the array in the `@model_validator` later?

src/qibocal/calibration/serialize.py
lballerio: Ah idk, I saw from the type hint it also takes an `npt.NDArray`, in which case we need to simply return it without `np.load()`...
Then if we don't want this function to be idempotent we can.

src/qibocal/calibration/calibration.py
lballerio: I don't know, we are simply creating the `np.ndarray` if it is None or empty list and check if the matrix is squared, I don't know how to simplify even more than that.

src/qibocal/calibration/serialize.py
alecandido: It seems like Pydantic is using `Callable[[Any], Any]` as signature. Not very helpful...

https://pydantic.dev/docs/validation/latest/concepts/validators/#field-plain-validator

Let's try to keep `Callable[[str], NDArray]`, which makes sense, and test whether it is causing any problem.

src/qibocal/protocols/rabi/parent_classes.py
alecandido, line 1: Ok, I understand the point, but `parent_classes.py` sounds a bit inaccurate.

On the generic side, `structures.py` could be a candidate. But `data.py` is also an option - though it can be slightly misleading, since we have `*Data` classes.
Or even just `classes.py` would be more accurate on its own (some of the classes are directly used).

doc/source/protocols/rabi/rabi.rst
```diff
@@ -44,13 +44,16 @@ Parameters
...
+	qibocal.protocols.rabi.parent_classes.RabiAmplitudeParameters
 	:noindex:
 
 .. autoclass::
-	qibocal.protocols.rabi.length.RabiLengthParameters
+	qibocal.protocols.rabi.parent_classes.RabiLengthParameters
 	:noindex:
 
+We must note that all Rabi protocols takes as an optional input also `drive_lines`, which indicates the lines to use for each qubit.
+If this parameter is set it has to be of the same length of `targets` list and there must not be repeated elements in these two lists, otherwise an error is raised.
```
alecandido, line 55: Not so clear. I would recommend adding an example of a good parameters' choice, and a bad one.

src/qibocal/calibration/calibration.py
```diff
@@ -139,6 +146,10 @@ class QubitCalibration(Model):
     """T2 hanh echo [ns]."""
     rb_fidelity: Measure | None = None
     """Standard rb pulse fidelity."""
+    rabi_ampl_oscillation: dict[Literal["signal", "classification"], float] = Field(
```
alecandido, line 149: You could give a name to this `Literal[...]` type, since we may use in multiple places.

An alternative option could be using a [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum). But, until we want to use the `Enum` to simplify some kind of consumption (iterating it, dynamically checking if it is part of the set, and similar) a type alias of the `Literal` is more than enough.
Let's avoid overcomplicating things, until we have a use case (which lifts the "over" part).

src/qibocal/calibration/calibration.py
```diff
@@ -163,12 +174,46 @@ class Calibration(Model):
...
-        )
+    @model_validator(mode="after")
+    def validate_flux_matrix(self):
+        """Validate or initialize the flux crosstalk matrix.
+
+        If the matrix is missing or empty, initialize it as an identity matrix of
+        shape (nqubits, nqubits). If provided, verify the matrix shape is correct.
+        """
+        if self.flux_crosstalk_matrix is None or self.flux_crosstalk_matrix.size == 0:
+            self.flux_crosstalk_matrix = np.eye(self.nqubits)
```
alecandido, line 188: I'm not sure whether it is a good idea, but in principle you do not need this to be part of a custom validator, and you do not need `None` as a placeholder default.
It looks like default-setting callables in Pydantic are allowed to depend on previous fields https://pydantic.dev/docs/validation/latest/concepts/fields/#default-values

Thus, you could replace this with:
```py
    flux_crosstalk_matrix: NdArray = Field(default=lambda data: np.eye(len(data["single_qubits"])))
```

src/qibocal/calibration/calibration.py
```diff
@@ -163,12 +174,46 @@ class Calibration(Model):
...
+        matrix of shape (nqubits, nqubits) filled with infinity. If provided,
+        verify the matrix shape is correct and convert its dtype to complex.
+        """
+        if (
+            self.microwave_crosstalk_matrix is None
+            or self.microwave_crosstalk_matrix.size == 0
+        ):
+            self.microwave_crosstalk_matrix = np.full(
+                (self.nqubits, self.nqubits), np.inf, dtype=complex
+            )
```
alecandido, line 209: I'm not sure I understand why you want to initialize it with infinity.

We are not even measuring it in decibels - which is fine, since floats can already do a great job at representing exponents.
Thus, why not `np.zeros()`?

src/qibocal/calibration/calibration.py
```diff
@@ -223,3 +265,61 @@ def get_readout_mitigation_matrix_element(
...
         ids = self._readout_mitigation_matrix_indices(target)
         return self.readout_mitigation_matrix[ids]
+
+    def _get_element(
+        self, matrix: NdArray, qubit: QubitId, line: QubitId
+    ) -> float | complex:
+        """Return the matrix element for a target qubit and a physical line."""
+        return matrix[self.qubits.index(qubit), self.qubits.index(line)]
+
+    def get_flux_crosstalk(self, qubit: QubitId, flux_line: QubitId) -> float:
```
alecandido, line 275: Here and below, the name is ambiguous, and `get_` is not adding anything. I'd rather call it `flux_crosstalk_element`.

However, rather than `Calibration.get_flux_crosstalk(qubit, flux_line)` I'd rework it to make it a bit more modular, such that we can isolate crosstalk, and the information for a given association.
I would propose an API like `Calibration.crosstalk[qubit, line].flux`. In practice, the implementation would be very similar to the one you already made, but it would require two more classes to virtualize a "view" (not much different from what NumPy is doing with its slices).
- A `Crosstalk` container: could even be an actual data structure, or just being a view, holding a pointer back to `Calibration`. Here we can implement a `.__getitem__()` to return the next object.
- A `CrosstalkElement`, which at this point is just a view, and it has to properties (in the literal Python sense) which once accessed perform the look up

In this way, even the setter will be just `Calibration.crosstalk[qubit, line].flux = value`, which is more Pythonic.

src/qibocal/calibration/serialize.py
```diff
@@ -21,8 +21,11 @@ def sparse_serialize(matrix: lil_matrix) -> str:
     return base64.standard_b64encode(buffer.read()).decode()
 
 
-def sparse_deserialize(data: str) -> lil_matrix | None:
+def sparse_deserialize(data: str | lil_matrix) -> lil_matrix | None:
```
alecandido, line 24: Even here: since we are using as `PlainValidator`, I believe we do not need it to be idempotent. It will always receive a `str` as input.

src/qibocal/calibration/serialize.py
```diff
@@ -61,3 +64,24 @@ def ndarray_deserialize(x: str | npt.NDArray) -> npt.NDArray:
     PlainSerializer(ndarray_serialize, return_type=str),
 ]
 """Pydantic-compatible array representation."""
+
+
+def eq(obj1: BaseModel, obj2: BaseModel) -> bool:
```
alecandido, line 69: As most of this file, this is duplicated from Qibolab.

When things are unstable, it is certainly better to keep them internal, and keep control. But, by now, these thin layer over Pydantic is quite stably used since quite long. I wonder whether we should deduplicate it...

However, it's true that I wouldn't like that much to make it public in Qibolab, since it is not related to its core business, and we should try to keep the public "surface" as tiny as possible.

Any consideration? @RoyStegeman

src/qibocal/protocols/flux_dependence/flux_amplitude_frequency.py
```diff
@@ -402,9 +401,11 @@ def _update(
...
 
     # TODO: needs to be inverted
-    flux_qubit = results.crosstalk if results.crosstalk is not None else target
-    update.crosstalk_matrix(
-        results.fitted_parameters_flux[target][0], platform, target, flux_qubit
+    flux_line = results.crosstalk if results.crosstalk is not None else target
+    platform.calibration.set_flux_crosstalk(
+        qubit=target,
+        flux_line=flux_line,
+        value=results.fitted_parameters_flux[target][0],
```
alecandido, line 408: I know it is not part of this PR, but do you understand the meaning of the TODO?

src/qibocal/protocols/rabi/acquisition.py
```diff
@@ -0,0 +1,163 @@
...
+
+from qibocal.auto.operation import QubitId
+from qibocal.calibration import CalibrationPlatform
+from qibocal.protocols.two_qubit_interaction.cross_resonance.cr_parent_classes import (
+    check_qubit_overlap,
+)
+from qibocal.update import replace
+
+
+def check_correct_drive_lines_setup(
```
alecandido, line 12: `correct` is a bit vague.

```suggestion
def check_drives_assignment(
```

src/qibocal/protocols/rabi/acquisition.py
```diff
@@ -0,0 +1,163 @@
...
+        return targets
+
+    if len(input_drivelines) != len(targets):
+        raise ValueError(
+            "Each qubit has to be assigned to a drive line; "
+            "If inserted, drive lines must have the same length of targets list."
+        )
+
+    # check if the pair overlap: not admitted if we want to execute in parallel
+    check_qubit_overlap([(q, d) for q, d in zip(targets, input_drivelines)])
```
alecandido, line 27:
```suggestion
    check_qubit_overlap(list(zip(targets, input_drivelines)))
```
and, if you just need an iterable, you can avoid even `list()`.

src/qibocal/protocols/rabi/acquisition.py
```diff
@@ -0,0 +1,163 @@
...
+
+def sequence_length(
+    targets: list[QubitId],
+    drive_lines: list[QubitId],
+    platform: CalibrationPlatform,
+    pulse_duration: float | None,
+    pulse_ampl: float | None,
+    rx90: bool,
+    use_align: bool = False,
+) -> tuple[PulseSequence, list[PulseLike], list[Delay], dict[QubitId, float], dict]:
```
alecandido, line 129: This function is almost identical to the previous one, `sequence_amplitude()`.

The only differences are:
- the readout placement, which is slightly more complicate here, but just more general - since it is already doing the job in all cases, we do not need to add another one
- which includes `delays` return - which could be return anyhow, also no problem
- and returning the `amplitudes` dictionary, instead of the `durations` one - no one of the two is needed as a return value, since they can be easily inferred outside this function

I would propose to deduplicate the two in a single function. Unless I missed something

src/qibocal/protocols/rabi/acquisition.py
```diff
@@ -0,0 +1,163 @@
...
+    if pulse_ampl is not None:
+        qd_pulse = replace(qd_pulse, amplitude=pulse_ampl)
+
+    if pulse_duration is not None:
+        qd_pulse = replace(qd_pulse, duration=pulse_duration)
+
+    if rx90:
+        single_q_sequence.append((qd_channel, qd_pulse))
+
+    single_q_sequence.append((qd_channel, qd_pulse))
```
alecandido, line 79: No need to initialize this object at the very top, if it is used just now. And the `rx90` conditional can be handled with less duplication:

```suggestion
    single_q_sequence = PulseSequence([(qd_channel, qd_pulse) for i in range(1 if not rx90 else 2])
```

src/qibocal/protocols/rabi/acquisition.py
```diff
@@ -0,0 +1,163 @@
...
+        cross_channel_obj = platform.channels[cross_channel]
+        qubit_channel_obj = platform.channels[qd_channel]
+        update |= {
+            cross_channel: {
+                "frequency": platform.parameters.configs[qd_channel].frequency
+            }
+        }
+        if all(
+            [isinstance(ch, IqChannel) for ch in [qubit_channel_obj, cross_channel_obj]]
+        ):
```
alecandido, line 59: This should always be true, since you can not drive with anything else than an `IqChannel`. So, you could even assert it.

src/qibocal/protocols/rabi/acquisition.py
```diff
@@ -0,0 +1,163 @@
...
+            q_lo_params = platform.parameters.configs[qubit_channel_obj.lo]
+            update |= {
+                cross_channel_obj.lo: {
+                    "frequency": q_lo_params.frequency,
+                    "power": q_lo_params.power,
+                }
+            }
```
alecandido, line 66: Not all drive channels have an LO. So, the `.lo` attribute could be `None`.

So, possibly, above you want to check that the `.lo` attribute is not `None`, instead of checking whether it is an `IqChannel`.

src/qibocal/protocols/rabi/acquisition.py
```diff
@@ -0,0 +1,163 @@
...
+    if target != drive_line:
+        # used when q is being driven with another line (cross rabi)
+        cross_channel = platform.qubits[drive_line].drive
+        cross_channel_obj = platform.channels[cross_channel]
+        qubit_channel_obj = platform.channels[qd_channel]
+        update |= {
+            cross_channel: {
+                "frequency": platform.parameters.configs[qd_channel].frequency
+            }
+        }
```
alecandido, line 56:
```suggestion
        update |= {
            cross_channel: {
                "frequency": platform.parameters.configs[qd_channel].frequency
            }
        }
        
        qubit_channel_obj = platform.channels[qd_channel]
```

Just because this is only needed for the LO update. So, better to keep it close to it.

src/qibocal/protocols/rabi/amplitude_frequency.py
```diff
@@ -59,67 +41,91 @@ class RabiAmplitudeFrequencyResults(RabiAmplitudeFrequencySignalResults):
...
         self.data[qubit] = np.rec.array(data)
 
+    def amplitudes_arr(self, qubit):
+        """Unique qubit amplitudes."""
+        return np.unique(self[qubit].amp)
+
+    def frequencies_arr(self, qubit):
+        """Unique qubit frequency."""
+        return np.unique(self[qubit].freq)
+
```
alecandido, line 68: Instead of making this ugly meshgrid and deduplicating, can't we just save the raw data and the coordinates' arrays?

src/qibocal/protocols/rabi/parent_classes.py
```diff
@@ -0,0 +1,174 @@
...
+
+from qibocal.auto.operation import Data, Parameters, QubitId, Results
+
+
+@dataclass
+class RabiLengthParameters(Parameters):
+    """RabiLength experiments runcard inputs."""
+
+    pulse_duration_range: tuple[float, float, float] | None = None
+    """Pulse duration [ns] range."""
```
alecandido, line 11:
```suggestion
    duration: tuple[float, float, float] | None = None
    """Pulse duration [ns] range."""
```

This is already unambiguous.

src/qibocal/protocols/rabi/parent_classes.py
```diff
@@ -0,0 +1,174 @@
...
+        if any([d is None for d in self.duration_range]):
+            raise ValueError("Valid pulse duration range not inserted.")
+
+
+@dataclass
+class RabiAmplitudeParameters(Parameters):
+    """RabiAmplitude experiments runcard inputs."""
+
+    ampl_range: tuple[float, float, float] | None = None
+    """Pulse minimum amplitude [a.u] range."""
```
alecandido, line 51: As above:

```suggestion
    amplitude: tuple[float, float, float] | None = None
    """Pulse minimum amplitude [a.u] range."""
```

src/qibocal/protocols/rabi/parent_classes.py
```diff
@@ -0,0 +1,174 @@
...
+        if any([d is None for d in self.amplitude_range]):
+            raise ValueError("Valid pulse amplitude range not inserted.")
+
+
+@dataclass
+class RabiLengthFrequencyParameters(RabiLengthParameters):
+    """RabiChevronLength runcard inputs."""
+
+    freq_range: tuple[int, int, int] | None = None
+    """Frequency range as an offset."""
```
alecandido, line 85:
```suggestion
    frequency: tuple[int, int, int] | None = None
    """Frequency range as an offset."""
```

src/qibocal/protocols/rabi/parent_classes.py
```diff
@@ -0,0 +1,174 @@
...
+        if any([f is None for f in self.frequency_range]):
+            raise ValueError("Valid frequency offset range not inserted.")
+
+
+@dataclass
+class RabiAmplitudeFrequencyParameters(RabiAmplitudeParameters):
+    """RabiChevronAmplitude runcard inputs."""
+
+    freq_range: tuple[int, int, int] | None = None
+    """Frequency range as an offset."""
```
alecandido, line 114:
```suggestion
   frequency: tuple[int, int, int] | None = None
    """Frequency range as an offset."""
```

src/qibocal/protocols/rabi/parent_classes.py
```diff
@@ -0,0 +1,174 @@
...
+    def __post_init__(self):
+        super().__post_init__()
+
+        if any([f is None for f in self.frequency_range]):
+            raise ValueError("Valid frequency offset range not inserted.")
+
+
+@dataclass
+class RabiAmplitudeFrequencyParameters(RabiAmplitudeParameters):
+    """RabiChevronAmplitude runcard inputs."""
```
alecandido, line 111: Not sure how well does it work with `@dataclass`, but usually, to reduce this kind of duplication, one uses mixins:

```py
@dataclass
class FrequencyMixin:
    freq_range: tuple[int, int, int] | None = None
    """Frequency range as an offset."""
    min_freq: int | None = None
    """Minimum frequency as an offset."""
    max_freq: int | None = None
    """Maximum frequency as an offset."""
    step_freq: int | None = None
    """Frequency to use as step for the scan."""

    @property
    def frequency_range(self) -> tuple[float, float, float]:
        """
        Return a tuple with the duration times of the pulses.
        """
        if self.freq_range is None:
            return (self.min_freq, self.max_freq, self.step_freq)
        return self.freq_range

    def __post_init__(self):
        super().__post_init__()

        if any([f is None for f in self.frequency_range]):
            raise ValueError("Valid frequency offset range not inserted.")

class RabiLengthFrequencyParameters(RabiLengthParameters, FrequencyMixin):
    """RabiChevronLength runcard inputs."""

class RabiAmplitudeFrequencyParameters(RabiAmplitudeParameters, FrequencyMixin):
    """RabiChevronAmplitude runcard inputs."""
```

alecandido (commented):
Ok, I'm kind at half of it. But this PR is definitely way too big, let me cut the first review here.

PR #1564, opened 2026-06-25
Improve conditional oscillations

alecandido:
- [x] since we are experiencing problems with sweeping the phase for this experiment, I will provide an option for unrolling it instead
    - multiple sequences will need to be generated, for all the values of the phase
    - each will have its own measurements, and they should be eventually regrouped, in the same shape of the sweeper

### DIFF
--- a/src/qibocal/protocols/two_qubit_interaction/virtual_z_phases.py
+++ b/src/qibocal/protocols/two_qubit_interaction/virtual_z_phases.py
@@ -1,6 +1,6 @@
 """CZ virtual correction experiment for two qubit gates, tune landscape."""
 
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from typing import Literal
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


src/qibocal/protocols/two_qubit_interaction/virtual_z_phases.py
```diff
@@ -56,13 +58,30 @@ class VirtualZPhasesParameters(Parameters):
...
+       are now providing the double implementation.
+       The difference it is just an internal detail, and ideally the result is supposed
+       to be the exact same when sweeping phases is properly supported. Sweeping is
+       often a bit faster.
+
+       The recommended usage is just to toggle this to ``False`` when the user is
+       confident the sweeping is not working properly. Or just to check whether that is
+       actually the problem or not.
+
+    """
```
alecandido, line 77: I wonder whether I should now remove this. It was introduced mostly for Qblox.

And *a bit faster* is of course an euphemism. Especially, taking into account the per-job overhead.

PR #1565, opened 2026-06-26
Fix Cryoscope to work with qibolab filters interface

### LINKED ISSUE #1566: Cryoscope is not propagating parameters in the format expected by the qibolab interface
As the title suggests, the parameters propagated by the Cryoscope protocol do not fit correspond to the qibolab interface.

RoyStegeman:
The Cryoscope protocol was implemented specifically for the QM interface and is not compatible with the qibolab interface for filters (at least not Qblox). This PR extends support to the qibolab filters interface, including the Qblox backend.

Changes on top the fix:
- The CMA algorithm has been replaced by a `least_squares` fit. This is faster, doesn't come with an additional dependence just for this protocol (though still through qibo), and the cma was even creating directories (I imagine optionally, but did not check).
- The hardcoded QM bounds on the taps has been removed, but now there are no bounds at all. This means that an unstable fit will be a problem, although this is largely mitigated by raising an error if there are more degrees of freedom than datapoints. 
- No longer assume that the sampling rate is 1 GSps.

The implementation is not supporting multiple IIR filter, and for that there is an issue to track https://github.com/qiboteam/qibocal/issues/1646.

Example: http://login.qrccluster.com:9000/cmxg0T2oTmSFQENq9UhdVQ==


pyproject.toml
```diff
@@ -17,7 +17,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = ">=3.10,<3.14"
-qibolab = "^0.2.8"
+qibolab = "^0.2.12"
```
alecandido, line 20: Is this bump required? Do you remember for which reason?

src/qibocal/protocols/flux_dependence/cryoscope.py
```diff
@@ -48,8 +65,38 @@ class CryoscopeParameters(Parameters):
...
+            raise ValueError(
+                f"Cannot fit {self.fir} FIR taps and {iir_free_parameters} exponential "
+                f"parameters with only {n_points} duration points: the fit would be "
+                "underdetermined."
+            )
+
+        # TODO: padding is in samples while duration in ns. This protocol mixes ns and
+        # samples, so only works for 1 GHz sampling.
+        if self.duration_min < self.padding:
+            raise ValueError("Ensure that duration_min >= padding.")
```
alecandido: In the acquisition function you have access to the `platform`. You could move this validation there, such that you can perform the conversion with the sampling rate.

src/qibocal/protocols/flux_dependence/cryoscope.py
```diff
@@ -66,17 +113,28 @@ class CryoscopeResults(Results):
...
+    # TODO: this is here for the plotting for now, but needs to go
     feedforward_taps: dict[QubitId, list[float]] = field(default_factory=dict)
     """feedforward taps"""
     feedforward_taps_iir: dict[QubitId, list[float]] = field(default_factory=dict)
     """feedforward taps for IIR"""
     feedback_taps: dict[QubitId, list[float]] = field(default_factory=dict)
     """feedback taps"""
 
+    # TODO: we only need this because params is not passed to the fit function, probably
+    # it makes sense to make the input parameters also accessible for the fit function.
```
alecandido: The first line we can keep, but the second is a matter for an issue/a discussion. Not very useful to reproduce it in each and every protocol.

I would also like to make the parameters accessible to the fitting and plotting functions. At least we could avoid passing the coordinates over and over.
But, as usual, I would make it optional, such that we can transition gradually from the current state.

src/qibocal/protocols/flux_dependence/cryoscope.py
```diff
@@ -88,21 +146,26 @@ def __contains__(self, key):
 def generate_sequences(
     platform: Platform,
     qubit: QubitId,
-    duration: float,
     params: CryoscopeParameters,
-) -> tuple[PulseSequence, PulseSequence]:
-    """Compute sequences at fixed duration of flux pulse for <X> and <Y>"""
+) -> tuple[PulseSequence, PulseSequence, PulseId, PulseId, Pulse]:
```
alecandido: A tuple with five elements is still acceptable. But, beyond this, and possibly including, it would be more explicit to give it some structure. Even just making a structure to represent the return type of this function. Or splitting it in subgroups.

src/qibocal/protocols/flux_dependence/cryoscope.py
```diff
@@ -154,18 +217,34 @@ class CryoscopeData(Data):
...
     flux_coefficients: dict[QubitId, list[float]] = field(default_factory=dict)
     """Flux - amplitude relation coefficients obtained from flux_amplitude_frequency routine"""
-    filters: dict[QubitId, float] = field(default_factory=dict)
+    has_filters: dict[QubitId, bool] = field(default_factory=dict)
     """Check if there are filters already."""
     data: dict[tuple[QubitId, str], npt.NDArray[CryoscopeType]] = field(
         default_factory=dict
     )
+    # TODO: we only need this because params is not passed to the fit function, probably
+    # it makes sense to make the input parameters also accessible for the fit function.
```
alecandido: Same as https://github.com/qiboteam/qibocal/pull/1565/changes#r3853020533

src/qibocal/protocols/flux_dependence/cryoscope.py
```diff
@@ -154,18 +217,34 @@ class CryoscopeData(Data):
...
 
-    def has_filters(self, qubit: QubitId) -> bool:
-        """Checking if for qubit there are already filters."""
-        try:
-            return len(self.filters[qubit]) > 0
-        except AttributeError:
-            return False
+
+def _check_phase_can_be_unwrapped(
+    data: CryoscopeData, qubit: QubitId, duration_step: float
```
alecandido: Since you only need the `Data.flux_coefficients` and `Data.flux_pulse_amplitude`, I would recommend to only require those two floats, instead of `CryoscopeData` and `QubitId`. In this way, the function will be more isolated and more modular - which makes it easier to reuse, also in testing (which, in lack of any other consumer, it's always a potentially good candidate).

src/qibocal/protocols/flux_dependence/cryoscope.py
```diff
@@ -154,18 +217,34 @@ class CryoscopeData(Data):
...
+) -> None:
+    """Check if the sampling rate is above the Nyquist rate."""
+    f = np.poly1d(data.flux_coefficients[qubit])
+    detuning = abs(f(data.flux_pulse_amplitude) - f(0.0))
+    cycles_per_step = detuning * duration_step
+    nyquist_cycles_per_sample = 0.5
```
alecandido: Maybe this constant could be lifted top-level, with a brief docstring commenting it?

src/qibocal/protocols/flux_dependence/cryoscope.py
```diff
@@ -189,117 +268,93 @@ def _acquisition(
...
+            qubit_sequence_y,
+            ro_pulse_id_x,
+            ro_pulse_id_y,
+            flux_pulse,
+        ) = generate_sequences(platform, qubit, params)
+        sequence_x += qubit_sequence_x
+        sequence_y += qubit_sequence_y
+        flux_pulses.append(flux_pulse)
+        ro_ids_x[qubit] = ro_pulse_id_x
+        ro_ids_y[qubit] = ro_pulse_id_y
```
alecandido: Just a matter of style, and not that relevant.

However, just to point out, this operation is more or less a transposition: from a list of tuples (one `generate_sequence()` output per qubit) to a tuple of lists. Plus some concatenation.
Another possible way of writing it is just something along the line of
```py
(
    sequence_x,
    sequence_y,
    ro_ids_x,
    ro_ids_y,
    flux_pulses,
) = zip(*(generate_sequences(platform, qubit, params) for qubit in targets))

# for what you actually need to concatenate, or collect differently
sequence_x_ = sum(sequence_x)
sequence_y_ = sum(sequence_x)
ro_ids_x_ = dict(zip(targets, ro_ids_x))
ro_ids_y_ = dict(zip(targets, ro_ids_y))
```

Just because `zip()` is effectively a transposition. And, despite being less explicit than NumPy's `.T`, it is still more explicit than a `for` loop.

src/qibocal/protocols/flux_dependence/cryoscope.py
```diff
@@ -189,117 +268,93 @@ def _acquisition(
...
+    for measure, ro_ids in zip(["MX", "MY"], [ro_ids_x, ro_ids_y]):
+        for qubit in targets:
+            data.register_qubit(
+                CryoscopeType,
+                (qubit, measure),
+                {
+                    "duration": duration_range,
+                    "prob_1": results[ro_ids[qubit]],
+                },
+            )
```
alecandido: Since you are anyhow reviewing quite massively this protocol, I would recommend to drop the `CryoscopeType`, and just store in `data` the raw measurements, i.e. `results`, on the blueprint of what we did for the qubit spectroscopy
https://github.com/qiboteam/qibocal/blob/0a5078ad49645511d13f0864e5df7474f6672278/src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py#L236-L246
*Of course, without the extra complication of concatenating the batches. Which is specific for that protocol.*

src/qibocal/protocols/flux_dependence/cryoscope.py
```diff
@@ -189,117 +268,93 @@ def _acquisition(
...
-    alpha = 1 - np.exp(-1 / (sampling_rate * tau * (1 + exp_amplitude)))
-    k = (
-        exp_amplitude / ((1 + exp_amplitude) * (1 - alpha))
-        if exp_amplitude < 0
-        else exp_amplitude / (1 + exp_amplitude - alpha)
+    popt, _ = scipy.optimize.curve_fit(
+        _expmodel,
+        durations,
+        step_response,
+        p0=[10.0, 1.0, 1.0],
```
alecandido: Any reason to have fixed initial parameters? Or simply because we are not very interested in it right now?

The second option is also perfectly fine. But, in that case, I would leave an issue for the initial parameters estimate.

src/qibocal/protocols/flux_dependence/cryoscope.py
```diff
@@ -329,34 +384,40 @@ def _fit(data: CryoscopeData) -> CryoscopeResults:
...
         Y_exp = 1 - 2 * data[(qubit, "MY")].prob_1
 
         norm_data = X_exp + 1j * Y_exp
 
         # demodulation frequency found by fitting sinusoidal
+        sampling_rate = 1 / (duration_step)
         demod_freq = -fitted_parameters[qubit, "MX"][2] / 2 / np.pi * sampling_rate
         # to be used in savgol_filter
+        # TODO: expose this 7 since it's a hyperparameter that, according to PycQED
+        # "Needs some playing around sometimes"
```
alecandido: Issue?

Not sure, if in your experience is not that relevant, we can even leave as it is. Whenever we will find problems, we will read the source again, and the `TODO` will be there.
But issues are usually easier to point the rest of the world at.

In any case, I would add a URL to support the citation, if possible.

src/qibocal/protocols/flux_dependence/cryoscope.py
```diff
@@ -382,36 +443,39 @@ def _fit(data: CryoscopeData) -> CryoscopeResults:
...
+            else:
+                exp_params = [0.0, 0.0, 1.0]
+                feedback_taps[qubit] = [1.0]
+                feedforward_taps_iir[qubit] = [1.0]
+
             time_decay[qubit], alpha[qubit], g[qubit] = exp_params
-            iir_correction = lfilter(
+            iir_correction = scipy.signal.lfilter(
                 feedforward_taps_iir[qubit], feedback_taps[qubit], step_response[qubit]
             )
```
alecandido, line 501: I know that you want to make the two branches of the conditional symmetric at this level. But isn't it better to branch after, and just skip the `iir_correction`, in case IIR is excluded?

One branching is anyhow needed, and in this way you keep the `data.iir == False` branch simpler
```suggestion
            if data.iir:
                exp_params = exponential_params(step_response[qubit], durations)
                tau, exp_amplitude, _ = exp_params
                iir_filter = ExponentialFilter(amplitude=exp_amplitude, tau=tau)
                feedback_taps[qubit] = iir_filter.feedback
                feedforward_taps_iir[qubit] = iir_filter.feedforward
                time_decay[qubit], alpha[qubit], g[qubit] = exp_params
                iir_correction = scipy.signal.lfilter(
                    feedforward_taps_iir[qubit], feedback_taps[qubit], step_response[qubit]
                )
            else:
                iir_correction = step_response[qubit]
```

Not exactly this with these names, but I hope you got the idea.

src/qibocal/protocols/flux_dependence/cryoscope.py
```diff
@@ -382,36 +443,39 @@ def _fit(data: CryoscopeData) -> CryoscopeResults:
...
             # FIR corrections
 
             taps = data.fir
             baseline = g[qubit]
-            x0 = [1] + (taps - 1) * [0]
-
-            fir = cma.fmin2(_fir_cost_function(iir_correction, baseline), x0, 0.5)[0]
 
+            # toeplitz_matrix @ taps == lfilter(taps, [1], iir_correction)
+            toeplitz_matrix = scipy.linalg.toeplitz(iir_correction, np.zeros(taps))
```
alecandido: Maybe expand a comment with one more line, reminding in words that the product by a Toeplitz matrix is essentially implementing the convolution, since the row is sliding.

src/qibocal/protocols/flux_dependence/cryoscope.py
```diff
@@ -536,8 +603,17 @@ def _plot(data: CryoscopeData, fit: CryoscopeResults, target: QubitId):
 
 def _update(results: CryoscopeResults, platform: Platform, target: QubitId):
     try:
-        update.feedforward(results.feedforward_taps[target], platform, target)
-        update.feedback(results.feedback_taps[target], platform, target)
+        filters = [{"kind": "fir", "coefficients": results.fir_taps[target]}]
+        # TODO: multiple iir filters?
```
alecandido: Not for the time being. But it would be useful to update the docs to explain that an update will wipe out the existing filters.

alecandido (commented):
I know they are a bunch, but they are mostly minor comments. The PR seems pretty good, and almost ready to be merged.

pyproject.toml
RoyStegeman: iirc it's where the current filters interface was added

alecandido: This is one of the last commit modifying it https://github.com/qiboteam/qibolab/commit/32466a05f0a429f857d182ef306cca6f1bc106f3, other than a Ruff v0.16 pass by yourself. According to GitHub, it is included since v0.2.10.

In any case, we have no need to be super-picky. v0.2.12 can also be fine. It is more about the methodology, and attempting to be consistent in it. And that's why I was asking.

RoyStegeman: You're right. I probably was targeting specifically support for qblox pre-distortion filters https://github.com/qiboteam/qibolab/pull/1331

I think it makes sense to make that the minimum version, but it does not change the public API, so if you prefer to follow the rule that we only care about the public API then I can do up to 0.2.10 as well.

alecandido: I would do that, but only if we could test it. Given that it makes little difference, with the current situation I would stick to the present one. Unless it takes much longer to merge (hope not).

Even though we are not really testing the combinations consistently, and most often just sticking to the last one. But it seems a bit risky to now bounce the version back. Maybe it's just psychological

RoyStegeman: I don't really understand what you're trying to say. 

>  would stick to the present one

Is this 0.2.12 or do you mean 0.2.8?

With regards to testing: any environment solver I used, including poetry, resolves to the latest qibolab release and during development I always use main, so combinations of current qibocal with older versions of qibolab are untested. In that sense, making the range more restrictive reduces any risk. At the same time, since it is always resolved to the latest versions, any change of the lower bound has no effect in practice.

src/qibocal/protocols/flux_dependence/cryoscope.py
RoyStegeman: They were already fixed before this PR. The only thing that changed is that previously `exp_amplitude` was fixed to 10 and `tau` to 1, while I think it makes more sense for them to be swapped (and I assume this was also originally intended). Thinking about it now, the expected amplitude is even negative since the initial few samples undershoot the desired square pulse. 

Having said that, since the default value for the initial paramters is 1, I doubt it makes much of a difference.

src/qibocal/protocols/flux_dependence/cryoscope.py
RoyStegeman: I'll add a link. 

Since I wasn't able to get a chevron, I admittedly don't have much feeling for the cryoscope either. This PR fixes some things, and I believe/hope it should allow us to build other 2q protocols on top, but it is not tested by me to the same level as previous protocols (another example of this is that multiple iir is not supported).

The point is that beyond just this one possible issue, once we have a better platform to test it is worth revisiting and properly finalizing the cryosope protocol in general. 

That doesn't mean we shouldn't merge now, because for the time being this PR still represents an important fix and doing things in steps makes it more manageable.

src/qibocal/protocols/flux_dependence/cryoscope.py
RoyStegeman: Actually, if there are already filters, it doesn't overwrite with the new ones. This is to allow to determine and subsequently validate the cryscope by running it twice.

I'll indeed mention this behaviour in the docs.

src/qibocal/protocols/flux_dependence/cryoscope.py
RoyStegeman: [f66a4b5](https://github.com/qiboteam/qibocal/pull/1565/commits/f66a4b54008a76f6389f67918df622380ad0cc90)

src/qibocal/protocols/flux_dependence/cryoscope.py
RoyStegeman: [f66a4b5](https://github.com/qiboteam/qibocal/pull/1565/commits/f66a4b54008a76f6389f67918df622380ad0cc90)

src/qibocal/protocols/flux_dependence/cryoscope.py
RoyStegeman: [f66a4b5](https://github.com/qiboteam/qibocal/pull/1565/commits/f66a4b54008a76f6389f67918df622380ad0cc90)

src/qibocal/protocols/flux_dependence/cryoscope.py
RoyStegeman: I removed them since I don't believe them to be needed. If this becomes a problem, we can think of improved initial guesses. 

[f66a4b5](https://github.com/qiboteam/qibocal/pull/1565/commits/f66a4b54008a76f6389f67918df622380ad0cc90)

src/qibocal/protocols/flux_dependence/cryoscope.py
RoyStegeman: Added the link here [f66a4b5](https://github.com/qiboteam/qibocal/pull/1565/commits/f66a4b54008a76f6389f67918df622380ad0cc90)

pyproject.toml
alecandido: > Is this 0.2.12 or do you mean 0.2.8?

v0.2.12, with the alternative being v0.2.10 - the bare minimum we need.

> With regards to testing: any environment solver I used, including poetry, resolves to the latest qibolab release and during development I always use main, so combinations of current qibocal with older versions of qibolab are untested. In that sense, making the range more restrictive reduces any risk. At the same time, since it is always resolved to the latest versions, any change of the lower bound has no effect in practice.

That's what I was saying: being more restrictive is safer. But being exceedingly restrictive is just forcing unneeded updates, and reducing wiggle room for dependency resolution.

Testing the matrix of all deps is just unfeasible. Qibolab is of course a bit special to us, but still remains quite demanding to test for multiple versions.
Ideally, we should rely on semver, and new versions should not be breaking. And they are breaking very little - but still not nothing.

So, I would rely on semver and changelog, and just aim for reasonable choices. Not necessarily perfect ones.

src/qibocal/protocols/flux_dependence/cryoscope.py
alecandido: > I'll indeed mention this behaviour in the docs.

Great!


> Actually, if there are already filters, it doesn't overwrite with the new ones.

Wait, the `platform.update()` below is not conditional on anything. So, it is always dropping the existing ones, isn't it?
To avoid updating, you should manually disable it with `update=False`. Which is perfectly fine, and the best solution (to me). But I believe it is overwriting. Isn't it?

src/qibocal/protocols/flux_dependence/cryoscope.py
RoyStegeman: [ae9025c](https://github.com/qiboteam/qibocal/pull/1565/commits/ae9025c2abe80474d765701721cf381410b34e9c)

This ended up being quite a large commit since I decided to tackle also the conistent handling of number of samples vs duration. Before this was not done correctly and so the protocol only worked for a sampling rate of 1 GSps.

src/qibocal/protocols/flux_dependence/cryoscope.py
RoyStegeman: [c47af64](https://github.com/qiboteam/qibocal/pull/1565/commits/c47af644c5d6bb38351c6ade599df34aa995df18)

src/qibocal/protocols/flux_dependence/cryoscope.py
RoyStegeman: [006422a](https://github.com/qiboteam/qibocal/pull/1565/commits/006422aa2a77ca021a8c9edf18fdd37f9cbac066)

src/qibocal/protocols/flux_dependence/cryoscope.py
RoyStegeman: [48ccc66](https://github.com/qiboteam/qibocal/pull/1565/commits/48ccc66298a710d414594ffeecd25c37fd669cda)

src/qibocal/protocols/flux_dependence/cryoscope.py
RoyStegeman: [48ccc66](https://github.com/qiboteam/qibocal/pull/1565/commits/48ccc66298a710d414594ffeecd25c37fd669cda)

src/qibocal/protocols/flux_dependence/cryoscope.py
RoyStegeman: [0f33dff](https://github.com/qiboteam/qibocal/pull/1565/commits/0f33dffcb2f346908e1f24e56837911799f62f92)

src/qibocal/protocols/flux_dependence/cryoscope.py
RoyStegeman: The issue is that `CryoscopeResults` expects `feedback_taps` and `feedforward_taps_iir` regardless, for plotting the `all_corrections`. That could be refactored, but I think its probably best done when adding support for multiple iir filters. Then we can also decide if we want to support only IIR and no FIR.

src/qibocal/protocols/flux_dependence/cryoscope.py
RoyStegeman: It was implicitly handled through the `except KeyError`. Now it has been made explicit in [9c78f4a](https://github.com/qiboteam/qibocal/pull/1565/commits/9c78f4a8bacb9977e94ecb0f29279b16edbbf9e4)
and documented in [4162b45](https://github.com/qiboteam/qibocal/pull/1565/commits/4162b4502a71799fcc3db2b242afdc839e04bb81). 

I imagine this behaviour may change at some point, but we should choose between iteratively adding more filters (which iirc is what is done in the paper) or simply determining the best solution for the requested number of filters after collecting a single dataset. Perhaps the former is unnecessary, and the latter could be achieved by setting the filters to 0 before the acquisition, but I'd like to postpone the discussion for when it becomes relevant.

alecandido: > but I'd like to postpone the discussion for when it becomes relevant.

Agreed.

> I imagine this behaviour may change at some point, but we should choose between iteratively adding more filters (which iirc is what is done in the paper)

I'm not sure if it is strictly implied there that you should acquire new data every time. But we're not necessarily bound to follow what they write.

> Perhaps the former is unnecessary, and the latter could be achieved by setting the filters to 0 before the acquisition

I'm inclined to keep doing the determination with unfiltered data, just to avoid the mess of the many combinations you can go through (n filters determined in one shot, one at a time - or even determine some, delete part of them, and redetermine some other). To handle all of this, we need to make very clear which are the options to the user, and provide straightforward tools for manipulation.

However, we will always need two acquisition modes, regardless of what we do:
- from scratch, since it is certainly needed for new calibrations
- validation, which may not determine new filters, but it should be able to run with existing filters, for validation

alecandido (approved):

scarrazza (approved):

PR #1581, opened 2026-07-02
Restrict mixer calibration to target qubits' channels

jevillegasd:
Previously calibrate_mixers configured and calibrated every channel in the platform regardless of `targets`. Now only the drive/probe channels of the requested qubits are touched (all targets together, to preserve intermodulation products), the report table is split per qubit, and offset_i/offset_q updates propagate to sibling channels sharing a frequency-multiplexed readout port, since those get recalibrated in hardware too.

To do this I patch the `_channels_by_module` method from `qibolab.instruments.qblox.cluster` to keep only the channels associated to the target qubit(s).

### DIFF
--- a/src/qibocal/protocols/calibrate_mixers.py
+++ b/src/qibocal/protocols/calibrate_mixers.py
@@ -107,6 +107,9 @@ class CalibrateMixersData(Data):
     final_calibration: dict[str, ModuleCalibrationData] = field(default_factory=dict)
     """Final calibration values after running calibration."""
 
+    target_modules: dict[str, list[str]] = field(default_factory=dict)
+    """Module short names associated to each target's channels."""
+
 
 def _cluster(platform: CalibrationPlatform) -> Cluster:
     """Return the Qblox cluster available in the platform."""
@@ -208,6 +211,63 @@ def _perform_calibration(cluster: Cluster, seq_map: SequencerMap):
             sequencer.sideband_cal()
 
 
+def _target_channels(platform: CalibrationPlatform, targets: list[str]) -> set[str]:
+    """Drive and readout channels for the given target qubits.
+
+    All targets are combined into one set so channels sharing a module are configured
+    together, preserving cross-channel intermodulation products.
+    """
+    channels: set[str] = set()
+    for target in targets:
+        qubit = platform.qubits[target]
+        channels.update(
+            ch for ch in (qubit.drive, qubit.probe, qubit.acquisition) if ch is not None
+        )
+        channels.update(qubit.drive_extra.values())
+    return channels
+
+
+def _configure(cluster: Cluster, configs: dict, channels: set[str]) -> SequencerMap:
+    """Preconfigure sequencers, restricted to the given channels.
+
+    ``Cluster.configure`` always configures every channel in
+    ``Cluster._channels_by_module`` (the whole platform, by default), so the map is
+    narrowed temporarily and restored right after, to avoid touching unrelated modules
+    or leaking the restriction into later experiments.
+    """
+    cluster._channels_by_module = {
+        slot: filtered
+        for slot, chs in cluster._channels_by_module.items()
+        if (filtered := [(ch, address) for ch, address in chs if ch in channels])
+    }
+    try:
+        seq_map, _ = cluster.configure(configs=configs)
+    finally:
+        del cluster._channels_by_module
+    return seq_map
+
+
+def _target_modules(
+    cluster: Cluster,
+    platform: CalibrationPlatform,
+    seq_map: SequencerMap,
+    targets: list[str],
+) -> dict[str, list[str]]:
+    """Module short names touched by each individual target's channels."""
+    mapping = {}
+    for target in targets:
+        target_channels = _target_channels(platform, [target])
+        mapping[target] = sorted(
+            {
+                cluster._modules[slot].short_name
+                for slot, channels in seq_map.items()
+                for ch_name in channels
+                if ch_name in target_channels
+            }
+        )
+    return mapping
+
+
 def _acquisition(
     params: CalibrateMixersParameters,
     platform: CalibrationPlatform,
@@ -216,13 +276,16 @@ def _acquisition(
     """Data acquisition for mixer calibration.
 
     This routine calibrates the IQ mixer offsets, gain ratios, and phase offsets
-    for all Qblox RF modules in the platform. This is currently done specifically for
-    Qblox electronics but could be included as part of qibolab to make it more general.
+    for the Qblox RF modules associated to the drive and readout channels of
+    ``targets``. This is currently done specifically for Qblox electronics but could
+    be included as part of qibolab to make it more general.
 
     Args:
         params: Input parameters (currently empty)
         platform: Qibolab's calibration platform
-        targets: List of target qubits (not used, calibrates all mixers)
+        targets: List of target qubits. Only the drive and readout channels of these
+            qubits are calibrated; all of them are configured together so that
+            intermodulation products across their channels are accounted for.
 
     Returns:
         CalibrateMixersData with initial and final calibration values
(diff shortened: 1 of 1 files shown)


jevillegasd:
How I handle the update is a bit dummy running first in a reduced version of the hardware (only the target channels) and later in the full cluster to update offsets (and make sure all mixers that share an LO get the same offset update).

RoyStegeman:
Hi @jevillegasd, to for me to understand, what is the main motivation for this? Is it the time saved or something else?

jevillegasd:
> Hi @jevillegasd, to for me to understand, what is the main motivation for this? Is it the time saved or something else?

Hey @RoyStegeman , you may recall we talked about needing this when you have more pulses than sequencers. This may happen for example if in a platform we defined the ge, ef, and cr pulses, for two qubits, this already exhausts the available sequencers, even if they are not used simultaneously in a practical application. The mixer calibration would fail in such cases because it loads by default all existing sequencers.

Additionally, when you re-calibrate the mixer in practice you need to re-calibrate your pulse amplitudes and drag coefficients, so so it does make sense to have some level of control of which ports are being re calibrated.

src/qibocal/protocols/calibrate_mixers.py
```diff
@@ -208,6 +211,63 @@ def _perform_calibration(cluster: Cluster, seq_map: SequencerMap):
             sequencer.sideband_cal()
 
 
+def _target_channels(platform: CalibrationPlatform, targets: list[str]) -> set[str]:
```
RoyStegeman, line 217:
```suggestion
def _target_channels(platform: CalibrationPlatform, targets: list[QubitId]) -> set[ChannelId]:
```

Using these type aliases improves readability. I point out just here, but of course it should be done for all instances where there is now `str` that can be a more specific type alias.

src/qibocal/protocols/calibrate_mixers.py
```diff
@@ -208,6 +211,63 @@ def _perform_calibration(cluster: Cluster, seq_map: SequencerMap):
...
+                for slot, channels in seq_map.items()
+                for ch_name in channels
+                if ch_name in target_channels
+            }
+        )
+    return mapping
+
+
 def _acquisition(
     params: CalibrateMixersParameters,
```
RoyStegeman, line 277:
```suggestion
    _params: CalibrateMixersParameters,
```

to indicate it's not consumed

src/qibocal/protocols/calibrate_mixers.py
```diff
@@ -208,6 +211,63 @@ def _perform_calibration(cluster: Cluster, seq_map: SequencerMap):
...
+    """
+    cluster._channels_by_module = {
+        slot: filtered
+        for slot, chs in cluster._channels_by_module.items()
+        if (filtered := [(ch, address) for ch, address in chs if ch in channels])
+    }
+    try:
+        seq_map, _ = cluster.configure(configs=configs)
+    finally:
+        del cluster._channels_by_module
```
RoyStegeman:
```suggestion
    original_channels_by_module = cluster._channels_by_module
    cluster._channels_by_module = {
        slot: filtered
        for slot, chs in cluster._channels_by_module.items()
        if (filtered := [(ch, address) for ch, address in chs if ch in channels])
    }
    try:
        seq_map, _ = cluster.configure(configs=configs)
    finally:
        cluster._channels_by_module = original_channels_by_module
```

You may as well directly restore the previous value. It's more explicit.

src/qibocal/protocols/calibrate_mixers.py
```diff
@@ -208,6 +211,63 @@ def _perform_calibration(cluster: Cluster, seq_map: SequencerMap):
...
+    return channels
+
+
+def _configure(cluster: Cluster, configs: dict, channels: set[str]) -> SequencerMap:
+    """Preconfigure sequencers, restricted to the given channels.
+
+    ``Cluster.configure`` always configures every channel in
+    ``Cluster._channels_by_module`` (the whole platform, by default), so the map is
+    narrowed temporarily and restored right after, to avoid touching unrelated modules
+    or leaking the restriction into later experiments.
```
RoyStegeman, line 239: Please mention that this handling is needed because `Cluster._channels_by_module` is a cached property. 

Also mention that `try` is there not because you expect it to fail, but just to be absolutely certain that if it does then the previous `cluster._channels_by_module` is still restored. Or if there is some other reason, explain that one.

src/qibocal/protocols/calibrate_mixers.py
```diff
@@ -208,6 +211,63 @@ def _perform_calibration(cluster: Cluster, seq_map: SequencerMap):
...
+    seq_map: SequencerMap,
+    targets: list[str],
+) -> dict[str, list[str]]:
+    """Module short names touched by each individual target's channels."""
+    mapping = {}
+    for target in targets:
+        target_channels = _target_channels(platform, [target])
+        mapping[target] = sorted(
+            {
+                cluster._modules[slot].short_name
```
RoyStegeman, line 266: `cluster._modules` is not cached so you're generating the same dict it for each `target`, it can be moved out of the loop.

src/qibocal/protocols/calibrate_mixers.py
```diff
@@ -208,6 +211,63 @@ def _perform_calibration(cluster: Cluster, seq_map: SequencerMap):
...
+                for slot, channels in seq_map.items()
```
RoyStegeman: The loop over `seq_map` is done for each target but unchanged. It can be moved out. 
```
module_by_channel = {ch_name:  cluster._modules[slot].short_name
                for slot, channels in seq_map.items()
                for ch_name in channels
            }
```

src/qibocal/protocols/calibrate_mixers.py
RoyStegeman, line 266: Only the rf modules should be selected

src/qibocal/protocols/calibrate_mixers.py
```diff
@@ -452,39 +515,57 @@ def _update(
...
+            ch = channel_mixer(ch_id)
+            port = PortAddress.from_path(cluster.channels[ch_id].path).ports[0] - 1
+            updates[f"configs.{ch.mixer}.scale_q"] = cal.gain_ratio[port][seq_id]
+            updates[f"configs.{ch.mixer}.phase_q"] = cal.phase_offset[port][seq_id]
+            calibrated_ports.add((slot, port))
+
+    # round2
+    # offset_i/offset_q are per physical port, shared by every qubit multiplexed on it
+    # Calibrating one recalibrates all of them in hardware, so propagate the update to
+    # update to every channel on a sahred calibrated port, not just the targets.
```
RoyStegeman:
```suggestion
    # update to every channel on a shared calibrated port, not just the targets.
```

src/qibocal/protocols/calibrate_mixers.py
```diff
@@ -452,39 +515,57 @@ def _update(
...
+    # channels have matching `final_cal` entries.
+    calibrated_ports: set[tuple[int, int]] = set()
+    for slot, channels in results.sequencer_map.items():
+        mod_name = cluster._modules[slot].short_name
+        if mod_name not in final_cal:
+            continue  # Skip if no calibration data for this module
+        cal = final_cal[mod_name]
+        for ch_id, seq_id in channels.items():
+            ch = channel_mixer(ch_id)
+            port = PortAddress.from_path(cluster.channels[ch_id].path).ports[0] - 1
```
RoyStegeman, line 551:
```suggestion
            port = PortAddress.from_path(ch.path).ports[0] - 1
```

src/qibocal/protocols/calibrate_mixers.py
```diff
@@ -452,39 +515,57 @@ def _update(
...
+
+    # round1
+    # Per-sequencer values, for channels actually calibrated. Use `results.sequencer_map`
+    # rather than the platform-wide `cluster._channels_by_module`, since only calibrated
+    # channels have matching `final_cal` entries.
+    calibrated_ports: set[tuple[int, int]] = set()
+    for slot, channels in results.sequencer_map.items():
+        mod_name = cluster._modules[slot].short_name
+        if mod_name not in final_cal:
+            continue  # Skip if no calibration data for this module
```
RoyStegeman, line 547: This is duplicated in the second round. The `mod_name` handling can be done only once

src/qibocal/protocols/calibrate_mixers.py
RoyStegeman, line 336: These are not empty

RoyStegeman, line 98: The report always says "An error occurred when performing the fit.", even if that's not the case.

Image

The reason is that `Results.__contains__` checks if `QubitId` is contained within all of the keys of the attributes of the `Results` (child) class. In this case that's not there, but you could consider something like this

```suggestion
    def __contains__(self, target: QubitId) -> bool:
        return target in self.target_modules

```

src/qibocal/protocols/calibrate_mixers.py
jevillegasd: True this would be necessary if we pass more than one target to the routine.

PR #1594, opened 2026-07-06
Bootstrap calibrations

alecandido:
I'm trying to make it possible to avoid any manual initialization of the `calibration.json` file, and ideally to populate it from protocols' updates - whenever the fits are allowing to do so.

### DIFF
--- a/src/qibocal/calibration/platform.py
+++ b/src/qibocal/calibration/platform.py
@@ -55,7 +55,12 @@ class CalibrationPlatform(Platform):
     def from_platform(cls, platform: Platform):
         name = platform.name
         path = locate_platform(name)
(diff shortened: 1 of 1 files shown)


alecandido:
Maybe just better to merge this single commit, for the time being, and postpone the review of the protocols. When there will be a more stable platform, we can play at will the game of attempting a calibration touching as little as possible (or not at all) `parameters.json` and `calibration.json`.

In any case, the ranges introduced in #1595, and expanded in #1605, should quite help in making a large part of the content of `calibration.json` optional to run the protocols. Such they can be run, and, if successful, use their updates to populate `calibration.json` automatically.

scarrazza (approved):

RoyStegeman (approved):

PR #1595, opened 2026-07-06
Introduce range syntax in resonator spectroscopy

### LINKED ISSUE #1300: Move to ranges for parameters

alecandido:
E.g.:
```py
e.resonator_spectroscopy(
    frequency=("linwindow", 7.2e9, 1e8, 100),
    power_level="high",
    amplitude=0.2,
    nshots=100,
    relaxation_time=500,
)
```

### DIFF
--- a/src/qibocal/protocols/resonator_spectroscopies/resonator_spectroscopy.py
+++ b/src/qibocal/protocols/resonator_spectroscopies/resonator_spectroscopy.py
@@ -11,9 +11,12 @@ from qibocal.auto.operation import Data, Parameters, Protocol, QubitId, Results
 from qibocal.calibration import CalibrationPlatform
 from qibocal.protocols.utils import (
     PowerLevel,
+    Range,
+    RangeLike,
     lorentzian_fit,
     lorentzian_with_linear_background,
     readout_frequency,
+    to_range,
 )
 from qibocal.result import magnitude, phase
 from qibocal.update import replace
@@ -67,11 +70,21 @@ FITS = {
 class ResonatorSpectroscopyParameters(Parameters):
     """ResonatorSpectroscopy runcard inputs."""
 
-    freq_width: int
-    """Width for frequency sweep relative  to the readout frequency [Hz]."""
-    freq_step: int
-    """Frequency step for sweep [Hz]."""
-    power_level: PowerLevel | str
+    frequency: RangeLike | None = None
+    """Frequency range for sweep [Hz]."""
+    freq_width: int | None = None
+    """Width for frequency sweep relative  to the readout frequency [Hz].
+
+    .. deprecated:: 0.2.6
+        Use :attr:`frequency` instead.
+    """
+    freq_step: int | None = None
+    """Frequency step for sweep [Hz].
+
+    .. deprecated:: 0.2.6
+        Use :attr:`frequency` instead.
+    """
+    power_level: PowerLevel | str = PowerLevel.low
     """Power regime (low or high). If low the readout frequency will be updated.
     If high both the readout frequency and the bare resonator frequency will be updated."""
     fit_function: str = "lorentzian"
@@ -86,6 +99,21 @@ class ResonatorSpectroscopyParameters(Parameters):
     def __post_init__(self):
         if isinstance(self.power_level, str):
             self.power_level = PowerLevel(self.power_level)
+
+    def frequency_range(self, q: QubitId, platform: CalibrationPlatform) -> Range:
+        try:
+            center = readout_frequency(q, platform, self.power_level)
+        except KeyError:
+            center = 0.0
+        return (
+            to_range(self.frequency, center=center)
+            if self.frequency is not None
+            else (
+                center - self.freq_width / 2,
+                center + self.freq_width / 2,
+                self.freq_step,
+            )
+        )
 
 
 @dataclass
@@ -160,16 +188,10 @@ def _acquisition(
         ro_pulses[q] = pulse
         sequence.append((channel, pulse))
 
-    # define the parameter to sweep and its range:
-    delta_frequency_range = np.arange(
-        -params.freq_width / 2, params.freq_width / 2, params.freq_step
-    )
-
     sweepers = [
         Sweeper(
             parameter=Parameter.frequency,
-            values=readout_frequency(q, platform, params.power_level)
-            + delta_frequency_range,
+            range=params.frequency_range(q, platform),
             channels=[platform.qubits[q].probe],
         )
         for q in targets
@@ -196,16 +218,15 @@ def _acquisition(
     for q in targets:
         result = results[ro_pulses[q].id]
         # store the results
-        ro_frequency = readout_frequency(q, platform, params.power_level)
         signal = magnitude(result)
         phase_ = phase(result)
         data.register_qubit(
             ResSpecType,
-            (q),
+            q,
             dict(
                 signal=signal,
                 phase=phase_,
-                freq=delta_frequency_range + ro_frequency,
+                freq=np.arange(*params.frequency_range(q, platform)).tolist(),
             ),
         )
     return data
--- a/src/qibocal/protocols/utils.py
+++ b/src/qibocal/protocols/utils.py
@@ -1187,6 +1187,9 @@ def fallback_period(period: NDArray) -> NDArray:
     return np.where(np.isnan(period), 4, period)
 
 
+Range = tuple[float, float, float]
+"""Value range, corresponding to ``(start, stop, step)``."""
+
 RangeLike = (
     tuple[float, float, float]
     | tuple[Literal["linspace"], float, float, int]
@@ -1224,9 +1227,7 @@ The other variants are the following:
 """
 
 
-def to_range(
-    spec: RangeLike, center: float | None = None
-) -> tuple[float, float, float]:
+def to_range(spec: RangeLike, center: float | None = None) -> Range:
     """Convert any range specification into the default representation."""
 
     if not isinstance(spec[0], str):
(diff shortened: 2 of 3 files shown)


scarrazza (approved):

src/qibocal/protocols/utils.py
```diff
@@ -1185,3 +1185,73 @@ def fallback_period(period: NDArray) -> NDArray:
...
+- ``window``, which allows to specify the center and width of the window, instead of its
+  extremes, i.e. ``("window", center, width, step)``
+- ``center``, similar to the former, but assuming a default for the center, which
+  depends the chosen protocol, i.e. ``("center", width, step)``
+- ``asym``, which is also a way to build upon an implicit center, but with specifying an
+  asymmetric interval around it, i.e. ``("asim", (left-shift, right-shift), step)``
+"""
+
+
+def to_range(spec: RangeLike, center: float | None = None) -> Range:
```
RoyStegeman, line 1233: Maybe here we should raise an explicit error if the user passes an unsupported input. One way to achieve this is by explicitly checking the full string below instead of separating `endswith` and `startswith`. It will be at the expense of making the function less concise, but I think that's justifiable. This may also be a good use-case for match/case instead of if/else, which I believe will also remove the warning I see from pyright.

src/qibocal/protocols/utils.py
RoyStegeman, line 1: I think the range feature could have it's own module

tests/test_utils.py
```diff
@@ -12,3 +13,20 @@ def test_eval_magnitude():
...
+    assert pytest.approx(to_range(("linwindow", 50, 10, int(5e2)))[:2]) == (45, 55)
+    assert pytest.approx(to_range(("center", 1e5, 1e-1), center=1e6)[0]) == 1e6 - 5e4
+    assert pytest.approx(to_range(("lincenter", 1e5, 1234), center=1e6)[1]) == 1e6 + 5e4
+    assert pytest.approx(to_range(("asym", (1e5, -2e4), 1e-1), center=1e6)[0]) == 0.9e6
+    assert (
+        pytest.approx(to_range(("linasym", (1e5, -2e4), 987), center=1e6)[1])
+        == 1e6 - 2e4
+    )
+
+    with pytest.raises(ValueError, match="center"):
```
RoyStegeman, line 32: Perhaps here we can extend the test with a few more incorrect syntaxes.

src/qibocal/protocols/resonator_spectroscopies/resonator_spectroscopy.py
```diff
@@ -67,11 +70,21 @@ class ResonatorSpectroscopyFit:
...
+
+    .. deprecated:: 0.2.6
+        Use :attr:`frequency` instead.
+    """
+    freq_step: int | None = None
+    """Frequency step for sweep [Hz].
+
+    .. deprecated:: 0.2.6
+        Use :attr:`frequency` instead.
+    """
```
RoyStegeman, line 86: Since they all default to None, pydantic will no longer raise an error for missing inputs. The error will occur later in the protocol, but we can also use `@model_validator(mode='after')` to ensure that either `freq_step and `freq_width` are not None or `frequency` is not None.

src/qibocal/protocols/utils.py
```diff
@@ -1185,3 +1185,73 @@ def fallback_period(period: NDArray) -> NDArray:
...
+    """Convert any range specification into the default representation."""
+
+    if not isinstance(spec[0], str):
+        # Default case: assume it's a tuple of (start, stop, step)
+        return spec
+
+    if any(lab in spec[0] for lab in {"center", "asym"}) and center is None:
+        raise ValueError(
+            f"Center must be provided for '{spec[0]}' range specification."
+        )
```
RoyStegeman: We should also validate that width is positive and the number of steps in linspace are more than 1.

RoyStegeman (commented):
Thanks for implementing this. I think it is a good proposal and will make in particular the frequency sweep experiments much more user-friendly. My comment is just about putting some guardrails since now there are very few checks which may lead to errors that are not as helpful as they could be, or no error but unexpected behaviour. 

As an example of what I assume is unintended behaviour. Here the 10 used used as both the width and the stepsize. 
```
In [2]: to_range(("window", 50, 10))
Out[2]: (45.0, 55.0, 10)
```

src/qibocal/protocols/utils.py
```diff
@@ -1185,3 +1185,73 @@ def fallback_period(period: NDArray) -> NDArray:
...
+    | tuple[Literal["linwindow"], float, float, int]
+    | tuple[Literal["center"], float, float]
+    | tuple[Literal["lincenter"], float, int]
+    | tuple[Literal["asym"], tuple[float, float], float]
+    | tuple[Literal["linasym"], tuple[float, float], int]
+)
+"""Range specification.
+
+The alternative options allow for multiple representations of a range, which could all
+be reconducted to a sequence of evenly spaced values.
```
RoyStegeman:
```suggestion
be mapped to a sequence of evenly spaced values.
```

src/qibocal/protocols/utils.py
```diff
@@ -1185,3 +1185,73 @@ def fallback_period(period: NDArray) -> NDArray:
...
+The semantics of the default one is equivalent to that of the built-in :func:`range`, in
+which the three values correspond to ``(start, stop, step)``. Unlike :func:`range`, all
+values are mandatory, since the usual defaults are often unsuitable.
+
+The other variants are unambiguously discriminated by a starting label, e.g.
+``("linspace", 1.2, 1.5, 70)``.
+
+In the ``linspace`` variant, the ``step`` element is replaced with the number of steps.
+Since the step specification is always mandatory, for each of the other variants a
+further ``lin<...>`` version is also avaialble, making the same substitution.
```
RoyStegeman:
```suggestion
further ``lin<...>`` version is also available, making the same substitution.
```

src/qibocal/protocols/utils.py
alecandido: The negative width is intentionally allowed: it has a smooth interpretation as reversing the interval, which is just fine.

For the number of points you're right.

src/qibocal/protocols/utils.py
alecandido: I will do validate it with Pydantic, and keep the function concise. I will think something for Pyright later on...

src/qibocal/protocols/utils.py
alecandido: It could, as many things in this file. I just avoided solving this problem in this PR.

You're right that we could avoid making it worse, but I'd rather procrastinate it a bit more, and then take some proper time to reorganize the package.
It is a bit boring, but every now and then I'm also taking care of these things. I simply just did for Qibolab platforms...

src/qibocal/protocols/resonator_spectroscopies/resonator_spectroscopy.py
alecandido: True, but I would let it be, with the argument that the alternative input is deprecated, and it will be removed - removing also the `None` default. In any case, it is quite easy to figure out that you forgot to specify the frequencies. I would just avoid doing the thing and reverting soon, given it is not that critical (in terms of failure and user base).

Regarding `@model_validator` you can not, since this class is not Pydantic. You should validate in `__post_init__`. But let's avoid bloating further, until possible

src/qibocal/protocols/utils.py
alecandido: Done in 4e88039

tests/test_utils.py
alecandido: Done in 6c72a22

src/qibocal/protocols/utils.py
alecandido: Done in 173e5cd

RoyStegeman (approved):

PR #1596, opened 2026-07-06
Disentangle qubit spectroscopy from resonator spectroscopy

### LINKED ISSUE #1500: qubit spectroscopy fit sometimes fails on good data
The two sets of data below were acquired minutes after eachother and look very similar, but while the first failed the second (and many afterwards) have been fitted successfully. 

http://login.qrccluster.com:9000/e_YhqVkdQP2T6Sk2muVAHw==
Image

http://login.qrccluster.com:9000/lC0emD-lSJKG5HWSHZX0VQ==
Image

RoyStegeman:
The qubit spectroscopy protocol was using objects from the resonator spectroscopy protocol such as the data class. This meant that it would have attribute such as `resonator_type`, which is not needed for qubit spectroscopy and in fact complicates things. 

In this PR the qubit and resonator spectroscopies have been disentangled such that in a follow-up we can deduplicate in a cleaner way.

Besides a refactor, it changes the fitting function by removing the linear term in the background and leaving only the constant offset. As a consequence this fixes https://github.com/qiboteam/qibocal/issues/1500.


src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
```diff
@@ -62,9 +72,17 @@ class QubitSpectroscopyResults(Results):
...
 
-class QubitSpectroscopyData(ResonatorSpectroscopyData):
+@dataclass
+class QubitSpectroscopyData(Data):
     """QubitSpectroscopy acquisition outputs."""
 
+    amplitudes: dict[QubitId, float]
+    """Amplitudes provided by the user."""
+    fit_function: str = "lorentzian"
+    """Fit function (optional) used for the resonance."""
```
alecandido: Do we need this at all?

We do not provide alternatives for fitting the qubit spectroscopy. Also because the model itself is actually Lorentzian, for a two-level system.

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
```diff
@@ -216,15 +232,107 @@ def _acquisition(
...
+    return [
+        guess_amp,
+        guess_center,
+        guess_sigma,
+        guess_offset,
+    ]
+
+
+def _lorentzian_fit(data):
+    frequencies = data.freq * scipy.constants.nano
```
alecandido: Why do you rescale for fitting?

In general, I would do as little manipulations as possible, and postpone them at the end, for UI/standardization purposes.

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
```diff
@@ -216,15 +232,107 @@ def _acquisition(
...
+    voltages = data.signal
```
alecandido: Since, in general, it is in arbitrary units, I would keep calling it `signal`, rather then `voltages`.

Most often it will be a voltage - but we are truly not making any effort to guarantee the units. And, until now, we did not even care that much...

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
```diff
@@ -216,15 +232,107 @@ def _acquisition(
...
+
+    freq_domain_size = frequencies[-1] - frequencies[0]
+    bounds = (
+        [-np.inf, frequencies[0], 0.0, -np.inf],
```
alecandido: Are you sure to leave the `amplitude` the option to be negative, since you will be looking separately for peaks and dips?

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
```diff
@@ -216,15 +232,107 @@ def _acquisition(
     return data
 
 
+def _lorentzian_with_offset(frequency, amplitude, center, sigma, offset):
+    return lorentzian(frequency, amplitude, center, sigma) + offset
+
+
+def _guess_initial_parameters(voltages, frequencies, is_peak):
+
+    guess_offset = (voltages[0] + voltages[-1]) / 2
```
alecandido: Maybe you could get a 10% of the points in the window per-side?

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
```diff
@@ -216,15 +232,107 @@ def _acquisition(
...
+        guess_peak_height = voltages_no_background.max()
+        indices_beyond_half = np.where(voltages_no_background > guess_peak_height / 2)[
+            0
+        ]
+    else:
+        guess_center = frequencies[np.argmin(voltages_no_background)]
+        guess_peak_height = voltages_no_background.min()
+        indices_beyond_half = np.where(voltages_no_background < guess_peak_height / 2)[
+            0
+        ]
```
alecandido: Why not a `find_peak()`?

It is quite reliable and pretty fast.

alecandido: Instead of repeating the three lines, you could just flip the sign if `voltages`.

I.e.
```py
voltages_no_background = (voltages - guess_offset) * (1 if is_peak else -1)
```

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
RoyStegeman: Indeed it's not used anywhere so can go

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
RoyStegeman: I did not change much about the fitting function, other than trying both peak and dip. The purpose of this PR was to separate the qubit and resonator, and the individual protocols I will improve in a follow up. The reason that I changed to trying both peak and dip is because previously the choice depended on `resonator_type`.

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
RoyStegeman: Yes that makes sense

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
RoyStegeman: We don't want to get stuck in a local minimum. The peak and dip attempts only change the initial guesses, so while we know that in some cases the result is sensitive to the initial condition,  I imagine it will not be uncommon for both to get a good fit, which is also fine.

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
RoyStegeman: I could, but that introduces a hyperparameter and I don't believe the fit will be that sensitive to the initial guess for the offset. I'd prefer to not introduce a hyperparameter until we have some proof that it improves the fit stability.

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
RoyStegeman: Not a bad suggestion, but as said before, in this PR I wanted to stop at separating qubit spectroscopy from the resonator spectroscopy and in a next PR I'll deal with further changes.

RoyStegeman: Yes good suggestion

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
alecandido: Well, take 1 point or 10% of the points is introducing the same amount of hyperparameters. Even the 1 point could be 2, 3, or more.

I understand that with your comment you meant that 10% is not working for all possible windows. But it should be good enough - if you have at least 10 points (which for a spectroscopy...). And it sensible that if you fit a Lorentzian you should see a bit of the tails.

I'm much more worried about the 1 point to be a bad hyperparameter fine-tuning, since 1 is very sensitive to the shot noise. And this would prevent to make a spectroscopy with few shots, if you want to have a chance to fit it.

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
alecandido: Ok, then we can just open an issue for this :)

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
alecandido: Agreed: it is a small change, but it should be tested.

If you have data to fit and spare time, it can be done in this PR. But, on a general ground, I agree that PRs should be scoped - so, if it is not the best time, just leave a `TODO:`.

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
alecandido: True, but you are assuming that not putting a boundary can just give a chance to the wrong side to converge to the correct minimum.

What I was thinking was more in the direction that it could instead spoil the one with the correct sign. In principle, there is also the chance of getting both bad fits, because of some fluctuations, which could be interpreted as peak-like. Instead, one of the two may have a bunch of them filtered out, if they are on the wrong side.

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
RoyStegeman: Done

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
RoyStegeman: In the end we're doing a gradient descent without any stochastic features to avoid local minima, so I imagine it more likely that if we place a bound at 0, and if the minimum wrt the starting point is on the other side of 0, we just get stuck near 0. In principle it may be possible that putting a bound redirects to a different local minimum, but if that happens, we're in a region where fits are very fragile since the wrong local minimum could also have been on the same side of 0 as the local minimum we are interested in.

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
RoyStegeman: > Well, take 1 point or 10% of the points is introducing the same amount of hyperparameters. Even the 1 point could be 2, 3, or more.

Okay fair enough, it doesn't really increase complexity of the algorithm. Done

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
RoyStegeman: https://github.com/qiboteam/qibocal/issues/1599

src/qibocal/protocols/qubit_spectroscopies/qubit_spectroscopy.py
alecandido: In general, I do not see the advantage of having two fits converging to the same solution, while one has clearly an advantage over the other. I believe it is reasonable to assume that, if the fit is stable enough, the one with the guess on the correct side will always converge to the correct minimum when the other is doing it.

If there is a different local minimum on the same side of the correct one, I would not even trust the one starting from the wrong side to have found the global one, regardless of the loss. It is more likely to be just a different local minimum.

I believe we are not very interested in retaining the comparison between two initializations, when one of the two is meaningful (as much as we can make it) and the other is pretty random. The only reason why we keep both is that we do not know which one is the meaningful one.

> so I imagine it more likely that if we place a bound at 0, and if the minimum wrt the starting point is on the other side of 0, we just get stuck near 0.

Well, that would be a good thing, since it will clearly allow to discriminate the good initialization from the random one.

> In the end we're doing a gradient descent without any stochastic features to avoid local minima

At a technical level, we (i.e. `curve_fit) are actually not using gradient descent for minimization, but something which is more in the direction of a quadratic method, first-order approximating the least squares (though it is somehow alternating with gradient descent).

RoyStegeman: > Well, that would be a good thing, since it will clearly allow to discriminate the good initialization from the random one.

We don't want to discriminate the two initial conditions, the goal is simply to obtain at least one good fit by trying two different initializations. I added the bound because I also don't think there is a reason to expect it should make a significant difference.

alecandido: If that is the only goal, then we should consider using an array of randomly initialized parameters, like in a genetic algorithm.

I believe we are trying to locate the global minimum, just by guessing the properties of our curve in advance, in order to start close to the optimum.
And, in this case, we know that only one of the two can be right.

RoyStegeman: Honestly, I'm not sure why we are not trying random different initializations. 

> I believe we are trying to locate the global minimum

If we are looking for the global minimum and not the local minimum near the initial parameters (which is a statement I agree with), I believe we can apply different strategies than only trying to optimize as best we can the initial parameters. Such as indeed trying more randomly initialized initial parameters. The fits are very fast so the time it takes to try fits starting from different initial points I don't expect to be a problem.

alecandido: > Honestly, I'm not sure why we are not trying random different initializations.

Because it is expensive and it gives less guarantees of success than an informed approach.

Random initialization suffers from the curse of dimensionality, and it makes no use of the available information. We know that it is supposed to be a Lorentzian, and we are going to fit just that - why disregarding this information?

> The fits are very fast so the time it takes to try fits starting from different initial points I don't expect to be a problem.

The speed could be not much of a concern. But if you need to go for orders of magnitude more points, at some point it will become more and more concerning.

---

Do you have a reason to think than throwing ~10 points at random in a 3-dimensional (often larger) parameters space we will land closer to the global minimum, rather than trying to optimize noise-resilient estimates?

src/qibocal/protocols/resonator_spectroscopies/resonator_spectroscopy.py
```diff
@@ -236,13 +236,13 @@ def _fit(
         )
         if fit_result is not None:
             frequency[qubit], fitted_parameters[qubit], _ = fit_result
-            if data.power_level is PowerLevel.high:
+            if data.power_level == PowerLevel.high:
```
alecandido, line 239: Why not `is`? That's the recommended way to compare variants of enumerations, and all objects which are unique, as `None`.

That's why you write `is None` and not `== None`

src/qibocal/protocols/resonator_spectroscopies/resonator_spectroscopy.py
```diff
@@ -236,13 +236,13 @@ def _fit(
...
                 bare_frequency[qubit] = frequency[qubit]
 
     return ResonatorSpectroscopyResults(
         frequency=frequency,
         fitted_parameters=fitted_parameters,
-        bare_frequency=bare_frequency if data.power_level is PowerLevel.high else {},
+        bare_frequency=bare_frequency if data.power_level == PowerLevel.high else {},
```
alecandido, line 245: Same

tests/test_chsh.py
alecandido, line 1: How did this end up in this PR?

src/qibocal/protocols/resonator_spectroscopies/resonator_utils.py
```diff
@@ -177,24 +176,21 @@ def spectroscopy_plot(data, qubit, fit: Results = None):
...
                 name="Fit",
                 line=go.scatter.Line(dash="dot"),
             ),
             row=1,
             col=1,
         )
 
-        if data.power_level is PowerLevel.low:
-            label = "Readout Frequency [Hz]"
+        if data.power_level == PowerLevel.low:
```
alecandido, line 188: Same

alecandido (approved):

scarrazza (approved):

Issue #1590, opened 2026-07-03
Fast SNZ landscape


alecandido:
While #1558 should provide the most stable experiment to fit the full conditional phase and leakage landscape for the SNZ pulse, the problem is that it is also going to be quite expensive.

### Estimate

E.g. for a limited number of pixels, we estimate that the protocol may take \~1h on Qblox, because of the many executions required due to the 3 nested sweepers.
This back-of-the-envelope calculation could be based on experiments dominated by a $300~\mu s$ reset through relaxation, considering 50 amplitude points, 20 values for the shape ($B/A$ ratio), and 10 phases for the conditional phase fit. If we add \~7s of reset time between different executions for the shape, this would be:
```py
(
  3e-4 *  # relaxation
  1e3 *   # shots
  50 *    # amplitudes
  10 +    # phases
  7       # experiment overhead (cluster reset)
) * 20    # shapes
```

This amounts to 50 minutes of experiments, and couple of minutes just going into the reset.

### Proposal

Instead, we could compute the relative phase with just two measurements, which will be the X and Y coordinates (expectation values) of the point rotated by the gate.
We will measure the same phase, with the backdraw of being more noise-prone, resulting in a less accurate measurement per-pixel.

However, this would be a speed-up of a factor 5 over the 10 phases considered above. And still a boost of 2-3, even if we assumed we could fit a sine with 4-6 points strategically distributed (which, most likely, it means evenly spaced).

PR #1601, opened 2026-07-08
Standard Error Amplification circuit

ElStabilini:
I am trying to implement the Standard Error Amplification (SEA) circuit to estimate the conditional phase error in CZ gate implementation.

The intended circuit is the following*:

SEA_circuit

The idea was to implement a protocol to correct the pulse amplitude based on the conditional phase error, it doesn't include information on leakage or single-qubit rotation imperfection.

Right now I'm only estimating the conditional phase error without implementing any update as I haven't implemented any protocol to estimate the phase amplitude slope**.

*I didn't find the original reference for the protocol, but I took the idea for the implementation from what is described in [https://arxiv.org/pdf/2607.01422](https://arxiv.org/pdf/2607.01422)

** Maybe there's already something similar but I didn't look for that.


alecandido:
Thanks @ElStabilini!

The experiment is certainly interesting, since it is measuring the conditional phase with a single sequence.
In particular, for gate candidates, we still need to measure the individual virtual phases, to compensate them. But it is extremely interesting for tuneup landscapes, since it prevents the need of a phase sweep (which can be reduced at most to 2 points, as described in #1590) and the double experiment. I.e. at least a factor 4 speedup, and no need for pixel fits. Super nice :D

In principle, the amplification could be useful for landscape fine-tuning. But I don't have a feeling for how much precise we can get with the usual CZ scheme, since there is a tiny tension between the phase condition and leakage.

However, for sure we will try it. As soon as the PR is ready (and we have a suitable platform available)

ElStabilini:
> In principle, the amplification could be useful for landscape fine-tuning. But I don't have a feeling for how much precise we can get with the usual CZ scheme, since there is a tiny tension between the phase condition and leakage.

Yes, this is the reason why in [[https://arxiv.org/pdf/2607.01422](https://github.com/qiboteam/qibocal/pull/url)](url) they use also the Echoed Leakage Error Amplification protocol. I think that the idea could be to write a cost function that takes the result from both to fine tune the calibration.

> However, for sure we will try it. As soon as the PR is ready (and we have a suitable platform available)

This might take some time on my side :)

alecandido:
> This might take some time on my side :)

No worries, just take the time you need ^^

If needed on our side, we will just reimplement the idea in a landscape - which will be anyhow a different experiment (because of the further sweeps).

> Yes, this is the reason why in [[https://arxiv.org/pdf/2607.01422](https://github.com/qiboteam/qibocal/pull/url)](url) they use also the Echoed Leakage Error Amplification protocol. I think that the idea could be to write a cost function that takes the result from both to fine tune the calibration.

Indeed! Thanks again, also for the reference

sorewachigauyo:
There is also section III of this paper https://arxiv.org/pdf/2508.16437, though it is tailored more towards couplers to reduce |02> state leakage

ElStabilini:
> There is also section III of this paper https://arxiv.org/pdf/2508.16437, though it is tailored more towards couplers to reduce |02> state leakage

As far as I know the main difference between the PALEA protocol described in section III of [https://arxiv.org/pdf/2508.16437](url) and the ELEA described [https://arxiv.org/pdf/2607.01422](url) is the phase sweep. I guess that choice depends on the kind of optimization loop one wants to perform.

> If needed on our side, we will just reimplement the idea in a landscape - which will be anyhow a different experiment (because of the further sweeps).

As you prefer, the only reason why I implemented the protocol as it is was to understand how to write such a circuit in the context of pulse calibration. Eventually I can also leave the protocol simply as a way to estomate the conditional phase error without updating any of the pulses parameters. The update might be performed in an external optimization loop where the phase error is also combined with other FOMs.

alecandido:
> As you prefer, the only reason why I implemented the protocol as it is was to understand how to write such a circuit in the context of pulse calibration. Eventually I can also leave the protocol simply as a way to estomate the conditional phase error without updating any of the pulses parameters. 

This is perfectly fine. We also have a virtual phases experiment, which is used for measuring the phases for the chosen gate. It could be useful to have the individual point, since when you settled on the landscape, you may want to have a measurement (possibly more precise) of that point, without performing any scan on the other parameters.

> The update might be performed in an external optimization loop where the phase error is also combined with other FOMs.

In this case, I would make a separate experiment, instead of an external loop. The reason is that I would implement in with hardware sweeps.
But once the experiment is known, it costs us little to upgrade it to use sweepers. And if we have a different experiment, we can burn in its fit also the figure of merits, and suitable plots.

ElStabilini:
> In this case, I would make a separate experiment, instead of an external loop. The reason is that I would implement in with hardware sweeps.
But once the experiment is known, it costs us little to upgrade it to use sweepers. And if we have a different experiment, we can burn in its fit also the figure of merits, and suitable plots.

It definitely makes sense.

> This is perfectly fine. We also have a virtual phases experiment, which is used for measuring the phases for the chosen gate. It could be useful to have the individual point, since when you settled on the landscape, you may want to have a measurement (possibly more precise) of that point, without performing any scan on the other parameters.

Then I'll proceed to keep only the conditional phase error estimate :).

ElStabilini:
I should have kept only the conditional phase estimation. Some things are still missing (real tests + circuit image) but overall I think it's ready for review.

RoyStegeman:
@ElStabilini do you perhaps have a report generated from this experiment? I'm not sure if you have access to the qrc cluster to upload the reports, but if not you can also just zip the folder and share it in a comment.

src/qibocal/protocols/two_qubit_interaction/sea.py
```diff
@@ -0,0 +1,311 @@
...
+                    "repetitions": np.array([repetitions]),
+                    "prob": np.array([prob]),
+                    "error": np.array([error]),
+                },
+            )
+    return data
+
+
+def sea_fit(x, offset, amplitude, omega, phase, gamma):
+    return np.sin(x * omega + phase) * amplitude * np.exp(-x * gamma) + offset
```
RoyStegeman: The model in the docs is `sin^2`. Here the space of solutions is the same and you only care about `omega`, but a comment explaining why omega is the same as in the documentation would be helpful. Even better would be to match the functional form to that in the documentation.

doc/source/protocols/standard_error_amplification.rst
```diff
@@ -0,0 +1,62 @@
...
+  missing svg image of the circuit
+
+The net phase accumulated by :math:`Q_a` after the full sequence is, up to known and calculable offsets,
+
+.. math:: \phi_n = n\,\delta .
+    :label: sea_phase
+
+The final :math:`X(\pi/2)` pulse on :math:`Q_a` is the second pulse of the Ramsey pair: it converts the phase :math:`\phi_n` of :eq:`sea_phase` into a measurable excited-state population
+
+.. math:: P(\ket{1}_{Q_a}) \approx \sin^2\left(\frac{n\,\delta}{2}\right)
```
RoyStegeman: Also since for `n=0` the excited state probability is 1, should this be `cos^2` instead?

src/qibocal/protocols/two_qubit_interaction/sea.py
```diff
@@ -0,0 +1,311 @@
...
+        try:
+            popt, perr = curve_fit(
+                sea_fit,
+                x,
+                y,
+                p0=pguess,
+                maxfev=2000000,
+                bounds=(
+                    [0.4, 0.4, -np.inf, -np.pi / 4, 0],
+                    [0.6, 0.6, np.inf, np.pi / 4, np.inf],
```
RoyStegeman: There are the same as in flipping, but should they be? My understanding is that for example `offset` is expected to be 1 (as long as the function is sin instead of cos).

src/qibocal/protocols/two_qubit_interaction/sea.py
```diff
@@ -0,0 +1,311 @@
...
+    [("repetitions", np.float64), ("prob", np.float64), ("error", np.float64)]
+)
+
+
+@dataclass
+class StandardErrorAmplificationData(Data):
+    """CZ conditional-phase SEA acquisition outputs."""
+
+    resonator_type: str
+    """Resonator type."""
```
RoyStegeman: I believe this is not used.

src/qibocal/protocols/two_qubit_interaction/sea.py
```diff
@@ -0,0 +1,311 @@
...
+
+            chi2[pair] = [
+                chi2_reduced(
+                    y,
+                    sea_fit(x, *popt),
+                    pair_data["error"],
+                ),
+                np.sqrt(2 / len(x)),
+            ]
+        except Exception as e:
```
RoyStegeman: Try being more specific, e.g `except RunTimeError`

doc/source/protocols/standard_error_amplification.rst
RoyStegeman, line 1: To be accessible this page should be indexed.

src/qibocal/protocols/two_qubit_interaction/sea.py
```diff
@@ -0,0 +1,311 @@
...
+
+    qubit_a, qubit_b = pair
+
+    natives_a = platform.natives.single_qubit[qubit_a]
+    natives_b = platform.natives.single_qubit[qubit_b]
+    natives_pair = platform.natives.two_qubit[pair]
+
+    sequence = natives_a.R(theta=np.pi / 2)
+
+    cz_channel, cz_pulse = natives_pair.CZ()[0]
```
RoyStegeman: `CZ` generally consists of multiple pulses on different channels. Here you're taking only the first one but presumably you want the full set.

RoyStegeman (commented):
A few quick observations. I don't have a platform to test with a CZ now, but if you can provide a report that would make my review a lot easier.

src/qibocal/protocols/two_qubit_interaction/sea.py
ElStabilini: I can confirm, I tested and it's not used.

doc/source/protocols/standard_error_amplification.rst
ElStabilini: True, thank you for noticing!

src/qibocal/protocols/two_qubit_interaction/sea.py
ElStabilini: On this I think I must have a look again. At the beginning I had implemented the protocol in a way that was very similar to a flipping protocol and that's the reason why I had chosen those parameters. They probably come from there, I will check it.

src/qibocal/protocols/two_qubit_interaction/sea.py
ElStabilini: That's true thank you. I hadn't inspected the output of `natives_pair.CZ()`, I had assumed it was similar to the one of a single qubit gate which is not the case as it is a sequence of pulses.

ElStabilini:
Thank you @RoyStegeman for the review. I should have addressed most of the issues.

Regarding the full functional form, I am not sure about keeping the exponential decay, I would probably test against some data but we are still trying to calibrate single qubit gates so I have no way to that. 
Eventually can convert the PR back to draft and wait until I have a report :) .

alecandido:
> Eventually can convert the PR back to draft and wait until I have a report :) .

We may be able to test on our side, but it could happen by the end of September or a bit later.

But you could also run it on the simulator :)

ElStabilini:
I reviewed the protocol implementation and documentation that I had proposed previously, by focusing in particular on the fitting procedure and the analytical form of the fitting.

My understanding is that the shape of the curve is determined by the number of active CZ which is the `n` set by the user. This happens because qubit Qb is in state 0 for CZs 1, 3, 5, … and in state 1 for CZs 2, 4, 6, … . This second case is when the conditional phase gets applied to Qa, so at the end of the circuit I have `n` active CZs, each contributing π + δ.

A consequence of this is that the probability of finding the qubit in state 1 is 0*5[1-(-1)^n cos(n * delta)], to avoid the (-1)^n I added  a conditional phase on the final gate.

```
    final_phi = np.pi if repetitions % 2 else 0.0
    final_channel, final_pulse = natives_a.R(theta=np.pi / 2, phi=final_phi)[0]
```

I am not sure this is correct and also I am starting to question wether the protocol is useful at all (we do not know the sign of the error, we only know the magnitude). I will still leave the (draft) PR open for now.

PR #1603, opened 2026-07-10
fixing bugs in `zz_coupling` protocol

lballerio:
@alecandido pointed out two small bugs in `zz_coupling` module.
In particular the code crashes when qubit pair is a `list` object, this since it is used as a jey in the data dictionary, resulting in an error.
Also it uses some initial parameters that are optional, hence when those values are not set the error raises.

### DIFF
--- a/src/qibocal/auto/execute.py
+++ b/src/qibocal/auto/execute.py
@@ -137,8 +137,9 @@ class Executor(BaseModel):
             targets: Targets | None = None,
             **kwargs: Any,
         ):
-            # casting targest to be of type Targets
-            targets = TypeAdapter(Targets).validate_python(targets)
+            # casting targest to be of type Targets if not None
+            if targets is not None:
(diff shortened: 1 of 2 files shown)


alecandido (approved):

scarrazza (approved):

src/qibocal/auto/execute.py
```diff
@@ -137,6 +137,10 @@ def wrapper(
             targets: Targets | None = None,
             **kwargs: Any,
         ):
+            # casting targest to be of type Targets if not None
+            if targets is not None:
+                targets = TypeAdapter(Targets | None).validate_python(targets)
```
alecandido: If it is not `None`, no need for it in the type hint
```suggestion
                targets = TypeAdapter(Targets).validate_python(targets)
```

alecandido (approved):

PR #1616, opened 2026-07-16
adding input tuple overlap at executor level

lballerio:
I am adding in `_wrapped_protocol` in `Executor` class an additional check (`check_qubit_tuples_overlap`) for the input targets in case this is a list of qubit tuples: `list[QubitPairId | QubitTuplesId]`.
This check asserts whether there in no overlap between different qubit tuples in the list and was first implemented in cross resonance protocols, but I think it can be generalized for all two-qubit protocols.

### DIFF
--- a/src/qibocal/auto/execute.py
+++ b/src/qibocal/auto/execute.py
@@ -9,6 +9,7 @@ from contextlib import contextmanager
 from copy import deepcopy
 from dataclasses import fields
 from functools import cached_property, reduce
+from itertools import combinations
 from pathlib import Path
 from typing import Any
 
@@ -22,12 +23,18 @@ from qibocal.config import log
 from ..calibration import CalibrationPlatform, create_calibration_platform
 from .history import History
 from .mode import AUTOCALIBRATION, ExecutionMode
-from .operation import Protocol, ProtocolsCollection
+from .operation import Protocol, ProtocolsCollection, QubitPairId, QubitTupleId
 from .output import Metadata, Output
 from .task import Action, Completed, Targets, Task
 
 PLATFORM_DIR = "platform"
 """Folder where platform will be dumped."""
+
+
+def check_qubit_tuples_overlap(targets: list[QubitPairId] | list[QubitTupleId]) -> None:
+    """This function checks if the input qubit tuples are independent, i.e. they don't share any qubit."""
+    if any(set(t1) & set(t2) for t1, t2 in combinations(targets, 2)):
+        raise ValueError("Target tuples must be independent, but are overlapping.")
 
 
 class Executor(BaseModel):
@@ -140,6 +147,10 @@ class Executor(BaseModel):
             # casting targest to be of type Targets if not None
             if targets is not None:
                 targets = TypeAdapter(Targets).validate_python(targets)
+
+            if all(isinstance(t, tuple) for t in targets):
+                # check if input was given correctly
+                check_qubit_tuples_overlap(targets)
 
             positional = dict(
                 zip((f.name for f in fields(protocol.parameters_type)), args)
--- a/src/qibocal/auto/execute.py
+++ b/src/qibocal/auto/execute.py
@@ -31,6 +31,12 @@ PLATFORM_DIR = "platform"
 """Folder where platform will be dumped."""
 
 
+def check_overlap_in_input_qubits(targets: np.typing.ArrayLike):
+    # check if input was given correctly
+    targ = np.asarray(targets)
+    assert np.unique(targ).size == targ.size, "One or more target qubits were repeated"
+
+
 class Executor(BaseModel):
     """Execute a tasks' graph and tracks its history."""
 
@@ -58,6 +64,8 @@ class Executor(BaseModel):
         for name, protocol in self.protocols.items():
             object.__setattr__(self, name, self._wrapped_protocol(protocol, name))
 
+        check_overlap_in_input_qubits(self.targets)
+
     @cached_property
     def protocols(self) -> ProtocolsCollection:
         return reduce(operator.or_, [protocols.PROTOCOLS] + self.sources)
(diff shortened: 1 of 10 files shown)


src/qibocal/auto/execute.py
```diff
@@ -30,6 +32,12 @@
 """Folder where platform will be dumped."""
 
 
+def check_qubit_tuples_overlap(targets: list[QubitPairId] | list[QubitTupleId]) -> None:
+    """This function checks if the input qubit tuples are independent, i.e. they don't share any qubit."""
+    if any(set(t1) & set(t2) for t1, t2 in combinations(targets, 2)):
```
RoyStegeman: This doesn't raise an error if two qubits within a tuple are the same? Is that intentional? Perhaps it is already checked somewhere else and not needed here. 

Instead of trying for all possible pairs you could flatten the targets and check if all entries are unique.

src/qibocal/calibration/calibration.py
```diff
@@ -18,6 +18,9 @@
 ]
 """Qubit pair name."""
 
+QubitTupleId = tuple[QubitId, ...]
+"""Qubit tuple name."""
```
RoyStegeman: Why did you introduce this? Is QubitPairId not sufficient?

src/qibocal/auto/execute.py
```diff
@@ -141,6 +149,10 @@ def wrapper(
             if targets is not None:
                 targets = TypeAdapter(Targets).validate_python(targets)
 
+                if all(isinstance(t, tuple) for t in targets):
+                    # check if input was given correctly
+                    check_qubit_tuples_overlap(targets)
```
RoyStegeman: I think in none of the protocols we would allow for the same qubit to appear twice, weather it is in a list of `QubitPairId` or just `QubitId`. Does this check already exist elsewhere for `QubitId`? Otherwise you could do the check for both types.

src/qibocal/auto/execute.py
```diff
@@ -141,6 +149,10 @@ def wrapper(
             if targets is not None:
                 targets = TypeAdapter(Targets).validate_python(targets)
 
+                if all(isinstance(t, tuple) for t in targets):
```
RoyStegeman: By placing the check inside the wrapper you only catch the error in case the call to the protocol is made through the wrapper. The check will not be done if `run_protocol` is used directly. Since the wrapper also calls `run_protocol`, I think all ways of running a protocol go through there, so that would be a more suitable place for the validation.

RoyStegeman (commented):
This need some tests, to ensure the error is captured when it should be.

src/qibocal/auto/execute.py
lballerio: no it does not., for example a list `[(1,1), (2,3)]` won't raise any error.
I think in general it might be OK this behavior, you are mantaining some flexibility and charging the user to be careful of inserting the right couple. For example I think if you do CR and you insert qubit pair `(1,1)` either the check is done protocol level or it might run.

src/qibocal/calibration/calibration.py
lballerio: So actually `tuple[QubitId, ...]` was already used in `task.py` when defining `Targets` type; what I did is explicitly define its associated type, then idk if this is needed or not.

src/qibocal/auto/execute.py
lballerio: yeah actually this is a good point.

src/qibocal/auto/execute.py
lballerio: yeah this is another good point.

src/qibocal/auto/execute.py
RoyStegeman: > you are mantaining some flexibility and charging the user to be careful of inserting the right couple

Sure, but you would also maintain flexibility by allowing for different pairs to contain the same qubit. The reason you implement this check, is that if a user provides such an input, then it is certainly a typo and we may as well fail early with a clear error message. 

> For example I think if you do CR and you insert qubit pair `(1,1)` 

Is this something a user may want to do?

src/qibocal/calibration/calibration.py
RoyStegeman: What is the benefit of the explicit type? Generally it is used to make things more readable but I don't think that it is the case here.   

From what I can tell it is only relevant for the readout mitigation matrix.

src/qibocal/auto/execute.py
```diff
@@ -30,6 +31,12 @@
 """Folder where platform will be dumped."""
 
 
+def check_overlap_in_input_qubits(targets: np.typing.ArrayLike):
+    # check if input was given correctly
+    targ = np.asarray(targets)
+    assert np.unique(targ).size == targ.size, "One or more target qubits were repeated"
```
RoyStegeman: While not strictly a rule, I believe asserts are particularly useful for consistency checks during development but in principle they should never be triggered by a user since it would indicate a bug. If a user provides an input that does not correspond to a sensible protocol execution, we can raise an actual error.

RoyStegeman (approved):
Thanks, other than replacing the assert with an error, this seems fine to me, although I think it would still be good to have a CI test that ensure the error is actually raised (assuming you agree to make it an error). After removing the type,  the PR description in the first message is now slightly outdated. 

Please also answer the questions I asked. Even if for example in the one about CR the answer is that there is no reason to pass (1,1) and the question seems rethorical, it is not. Since you are much more familiar with that protocol than I am, it would be good if you could confirm that there is no scenario in which a user would provide that input.

src/qibocal/auto/execute.py
lballerio: ok perfect, I'll replace with an error

lballerio:
> Please also answer the questions I asked. Even if for example in the one about CR the answer is that there is no reason to pass (1,1) and the question seems rethorical, it is not. Since you are much more familiar with that protocol than I am, it would be good if you could confirm that there is no scenario in which a user would provide that input.

No I confirm such a scenario does not have to happen in CR.

lballerio:
Also, adding tests for this check.

RoyStegeman (approved):

scarrazza (approved):

PR #1618, opened 2026-07-20
Ramsey: don't duplicate `target in fit.fitted_parameters` check 

lballerio:
after rebasing on main, this PR is essentially modifying the checks in ramsey fitting functions.
from the following if-condition:
`if fit is not None and target in fit.fitted_parameters`
 the second condition is already checked in `executor`, as pointed out below by @RoyStegeman.

### DIFF
--- a/src/qibocal/protocols/ramsey/processing.py
+++ b/src/qibocal/protocols/ramsey/processing.py
@@ -231,7 +231,7 @@ def signal_plot(
         ]
     )
 
-    if fit is not None and target in fit.fitted_parameters:
+    if fit is not None:
         fitting_report = fit_plot(
             target=target,
             fit=fit,
--- a/src/qibocal/protocols/ramsey/classification.py
+++ b/src/qibocal/protocols/ramsey/classification.py
@@ -105,6 +105,7 @@ def _fit(data: RamseyData) -> RamseyResults:
         except RuntimeError as e:
             log.warning(f"Ramsey fitting failed for qubit {qubit} due to {e}.")
             continue
+
         (
             freq_measure[qubit],
             t2_measure[qubit],
@@ -112,6 +113,7 @@ def _fit(data: RamseyData) -> RamseyResults:
             delta_fitting_measure[qubit],
             popts[qubit],
         ) = process_fit(popt, perr, qubit_freq, data.detuning)
+
     return RamseyResults(
         detuning=data.detuning,
         frequency=freq_measure,


scarrazza (approved):

scarrazza (approved):

src/qibocal/protocols/ramsey/processing.py
```diff
@@ -216,7 +216,7 @@ def signal_plot(
         ]
     )
 
-    if fit is not None:
+    if fit is not None and target in fit.fitted_parameters:
```
RoyStegeman: https://github.com/qiboteam/qibocal/blob/45b843da43e35f1fb57edd85b050c16a5c884588/src/qibocal/cli/report.py#L25-L41

I didn't check exhaustively, but I suspect that the plot function is always called through here, which would mean that if the fit is not None, then `target in fit.fitted_parameters` is always True. Either way, not a bad thing to make it explicit here.

src/qibocal/protocols/ramsey/zz_coupling.py
```diff
@@ -318,35 +318,33 @@ def _fit(data: RamseyZZData) -> RamseyZZResults:
...
+                coupling[pair] = coupling_strength(
+                    omega1=data.qubit_freqs[target],
+                    omega2=data.qubit_freqs[spectator],
+                    anharmonicity1=data.anharmonicity[target],
+                    anharmonicity2=data.anharmonicity[spectator],
+                    zz=zz[pair],
+                )
+
             except Exception as e:
                 log.warning(f"Ramsey fitting failed for qubit pair {pair} due to {e}.")
```
RoyStegeman: This in general is not a good idea. You want to capture maybe a specific error, but not blindly capture all.

src/qibocal/protocols/ramsey/processing.py
lballerio: you're right, it is not necessary.
If we want we can remove for ease

src/qibocal/protocols/ramsey/zz_coupling.py
lballerio: modified in 660c0a7

src/qibocal/protocols/ramsey/processing.py
lballerio: modified in 660c0a7dd681c2dcf7641c052d55876a56321625

scarrazza (approved):

src/qibocal/protocols/ramsey/processing.py
RoyStegeman: No, for me it's fine to do it here as well. That way if the upstream logic for this function changes, this one is not broken. Ah I see you already reverted it. Also fine, but in this case I really just meant to point out that within the current worfklow of the code, this check is degenerate.

src/qibocal/protocols/ramsey/zz_coupling.py
```diff
@@ -318,35 +318,33 @@ def _fit(data: RamseyZZData) -> RamseyZZResults:
...
-                omega2=data.qubit_freqs[spectator],
-                anharmonicity1=data.anharmonicity[target],
-                anharmonicity2=data.anharmonicity[spectator],
-                zz=zz[pair],
-            )
-        except AnharmonicityError:
-            # e-f transitions are not calibrated, cannot compute coupling strength
-            pass
+
+            except RuntimeError as e:
```
RoyStegeman: Should the RuntimeError be around evertything or is it just for the call to `fitting`?

src/qibocal/protocols/ramsey/zz_coupling.py
```diff
@@ -297,13 +297,13 @@ def _fit(data: RamseyZZData) -> RamseyZZResults:
...
-    delta_fitting_measure: dict[QubitId, list[float]] = {}
-    zz: dict[QubitId, list[float]] = {}
-    coupling: dict[QubitId, list[float]] = {}
+    popts: dict[tuple[QubitPairId, str], list[float]] = {}
+    freq_measure: dict[tuple[QubitPairId, str], list[float]] = {}
+    t2_measure: dict[tuple[QubitPairId, str], list[float]] = {}
+    delta_phys_measure: dict[tuple[QubitPairId, str], list[float]] = {}
+    delta_fitting_measure: dict[tuple[QubitPairId, str], list[float]] = {}
+    zz: dict[QubitPairId, list[float]] = {}
+    coupling: dict[QubitPairId, list[float]] = {}
```
RoyStegeman: You're changing it here, but the data class still expects dicts with `QubitId` as key. Within qibocal I'm mostly ignoring typehints and only fixing when I'm actually touching something, but since you're working on this protocol in another PR, if you want to update the typehints please do it here rather than in the other PR. If you want to leave it as is, that's also fine for me.

RoyStegeman (approved):

PR #1621, opened 2026-07-21
use RANSAC for qubit and resonator flux arc fits

### LINKED ISSUE #1382: examples of failed flux arc fits
http://login.qrccluster.com:9000/YwaOc48OQ2q7kRnxjymwEg==
http://login.qrccluster.com:9000/nBEP1tuoSk-PlfXfXCh_0g==
http://login.qrccluster.com:9000/ZKpGLQtSQGmyYqOaTuP-ig==
http://login.qrccluster.com:9000/gvu1WMgoRJ6fBoz4xvNAgw==

RoyStegeman:
Improve the stability of the qubit flux arc and resonator flux arc fits by implementing a RANSAC-based algorithm (see e.g. [the wikipedia page](https://en.wikipedia.org/wiki/Random_sample_consensus)). 

The algorithm consists of three main steps: 
1. identify peaks in the landscape
2. identify which of those peaks are on the flux arc
3. perform a fit to only those points on the flux arc

Note that in terms of implementation both steps 2 and 3 are inside the `ransac_fit` function, where step 3 is just a single call to scipy's `curve_fit`.

Step 1 was also done before and mainly relies [scipy's `find_peaks` ](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.find_peaks.html), along with some smoothing depending on the specific features of the background signal and noise. However, since the features of the data are different, it is also desirable that they do not share a single function for the pre-processing. 

Step 2 is where the main idea of this PR comes in. Namely, after step 1 we find peaks that may look something like the image below, although the features can also be very different. For example, depending on the binning, the signal of the arc itself may be much more sparse. The idea of RANSAC is that we don't have to apply a completely general "pattern finding" algorithm to identify which points constitute the feature of interest, but we can make use of the fact that we know the functional form of the feature of interest (the arc). 

image

The RANSAC algorithm samples random subsets of all peaks and fits the functional form to this random subset. If all points in the subset are part of the arc, then most of the non-fitted peaks should still be close (according to some threshold) to the fitted arc. Any point close to the fitted function we call an "inlier". However, if one or more of the points in the random subset are part of the background noise, then a fit to this subset will provide a poor description of the non-fitted points. In step 2 many different random subsets are fitted and the result with most inliers is selected. 

Step 3 takes the largest set of inliers found in step 2 as input, and fits this dataset. As initial parameters also the fitted parameters found in step 2 are used. This step does not change the final result by much, but reduces some noise by fitting a larger number of datapoints. 

image

---
Besides the fitting, this PR also changes the extraction of the sweetspot to be numerical rather than analytical. I'm not sure to what extend this is justifiable.  The motivation for this is that extracting it analytically was not consistently giving the correct result, possibly because there are some degeneracies in the fit parameters, although I did not investigate this further. 

---
qubit example: http://login.qrccluster.com:9000/kn3AtSEHTpCkHUrdf_EL8A==
image

resonator example: http://login.qrccluster.com:9000/1kaQVn9WSEqtm6R0ugOD3g==
image


scarrazza (approved):

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
```diff
@@ -209,54 +249,57 @@ def _fit(data: QubitFluxData) -> QubitFluxResults:
     for qubit in qubits:
         qubit_data = data[qubit]
 
-        # extract signal from 2D plot based on SNR mask
-        frequencies, biases = data.filtered_data(qubit)
+        freq, freq_idx = np.unique(qubit_data.freq, return_inverse=True)
+        bias, bias_idx = np.unique(qubit_data.bias, return_inverse=True)
+        signal = np.full((len(bias), len(freq)), np.nan)
+        signal[bias_idx, freq_idx] = qubit_data.signal
```
lballerio: is not the same as 
```python
signal = qubit_data.signal.copy()
```
or am I missing something why we need this?

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
```diff
@@ -209,54 +249,57 @@ def _fit(data: QubitFluxData) -> QubitFluxResults:
...
+                "charging_energy": data.charging_energy[qubit] * HZ_TO_GHZ,
+            }
+            frequency[qubit] = popt[0] * GHZ_TO_HZ
+            middle_bias = (np.max(qubit_data.bias) + np.min(qubit_data.bias)) / 2
+            sweetspot[qubit] = (
+                np.round(popt[1] * middle_bias + popt[2]) - popt[2]
+            ) / popt[1]
+            matrix_element[qubit] = popt[1]
+            successful_fit[qubit] = True
+        except (ValueError, RuntimeError) as e:
```
lballerio, line 326: what is `ValueError` accounting for?

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
```diff
@@ -168,14 +166,112 @@ def _acquisition(
...
+        residual = row - baseline
+
+        # Estimate the std from median absolute deviation because a naive std is
+        # inflated by the arc we're trying to detect
+        row_mad = np.median(np.abs(residual - np.median(residual)))
+        row_std = 1.0 / (np.sqrt(2) * erfinv(0.5)) * row_mad
+
+        # Detect both peaks and dips by finding prominent extrema in the absolute
+        # residual
+        peaks, props = find_peaks(np.abs(residual), prominence=row_std)
```
lballerio: why in this case we do not discriminate dips and peaks like in the qubit protocol?

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
```diff
@@ -189,84 +285,51 @@ def _fit(data: ResonatorFluxData) -> ResonatorFluxResults:
...
+            bias=bias,
+            signal=signal,
+        )
+
+        try:
+            w_max = data.qubit_frequency[qubit] * HZ_TO_GHZ
+            fit_function = partial(
+                _fit_function,
+                w_max=w_max,
+            )
```
lballerio: I saw this was already implemented in `main`, but now I was wondering why we are fixing the maximum frequency of the qubit; in principle this protocol does not require qubit spectroscopy, right? Do you have any explanation?

src/qibocal/protocols/flux_dependence/utils.py
```diff
@@ -398,3 +389,123 @@ def flux_extract_feature(
...
+    (points with residual below ``residual_threshold``), and keeps the best-performing
+    model. The number of trials adapts dynamically based on the current inlier ratio,
+    following the standard RANSAC stopping criterion, and is bounded by ``min_trials``
+    and ``max_trials``. A final least-squares refit is performed on the best inlier set.
+
+    Returns:
+        Optimal fit parameters from the least-squares refit on the best inlier set.
+
+    """
+    rng = np.random.RandomState(random_state)
```
lballerio, line 538: do we want to use always the same seed?

tests/runcards/protocols.yml
```diff
@@ -77,16 +77,6 @@ actions:
...
       nshots: 10
 
-  - id: resonator flux dependence bias
-    operation: resonator_flux
-    parameters:
-      freq_width: 10_000_000
-      freq_step: 500_000
-      bias_width: 0.8
-      bias_step:  0.1
-      nshots: 10
```
lballerio, line 87: I know is not a sensitive question because the test itself is meaningless, but what's the reason behind that? was RANSAC having problems with this test?

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
RoyStegeman: It is for reshaping. I thought it was more convenient to have a `(n_bias,n_freq)` sized matrix for signal and then the correspnding `n_bias` sized array for bias and `n_freq` sized array for frequencies. However, `qubit_data.freq`, `qubit_data.bias`, and `qubit_data.signal` are all flat arrays of the same size .

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
RoyStegeman: When the functional form has more degrees of freedom than there are bins along the flux-axis. In that case `rng.choice` will raise a ValueError. I'll explcitly capture it and make the message related to the RANSAC.

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
RoyStegeman: Because if both resonator frequency and readout are at the same point, then the dispersive shift will move the resonator away from the readout point. However, if the resonator is away from the readout point, then the dispersive shift can move the resonator frequency closer towards the readout frequency. This impacts the shape of the peak/dip and even change it's sign. I have one example of seeing that, though the data is also so poor that it would not be terrible if we did not fit it:
image
[qubit_data_11.npz.zip](https://github.com/user-attachments/files/30412609/qubit_data_11.npz.zip)

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
RoyStegeman: No I don't know. Possibly just because it's a parameter in the functional form, so some choice had to be made. Possibly we could reconsider how bad that assumption makes the parameter extraction from the fit (i.e. how unreliable our fit of the other parameters becomes), and if the extraction becomes bad then maybe there is no point in using a physics inspired functional form if we can do with something simpler. 

But the answer is just that I did not think much about the functional form.

src/qibocal/protocols/flux_dependence/utils.py
RoyStegeman: For regular use the seed doesn't matter, but for debugging/testing it can be useful to remove a source of noise.

tests/runcards/protocols.yml
RoyStegeman: Yes ransac did not find a large enough subset of inliers. Which makes sense because the dof of the flux arc functional form is almost the same as the number of bias points sampled. And then the test data is even random. 

https://github.com/qiboteam/qibocal/actions/runs/29926270115/job/88944115559?pr=1621

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
lballerio: Ok understood

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
lballerio: Ok that makes sense and it is handling a degenerate use case.
And this is due to the flag `replace=False` (which of course we want to be `False`) right?

RoyStegeman: Yes it needs to be False. If it is not, we can still have fewer points than the degrees of freedom.

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
lballerio: yeah honestly here I see no signal, but maybe is my screen's colors ahaah.
Anyway I see your point and I agree on that.

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
lballerio: No indeed, I am in the same situation as you, just out of curiosity.

src/qibocal/protocols/flux_dependence/utils.py
lballerio: no indeed, for testing I understand, was more a production related doubt for me, but ok I guess this is just for testing and later on it will be removed

tests/runcards/protocols.yml
lballerio: yeah ok, no point in trying to fix it

lballerio:
For me the code makes sense and at the moment I don't have issues to point out, the code is readable and well documented, so also for a user who does not know RANSAC alg still I think can understand what is going on.
@RoyStegeman  what is the status of the tests?

src/qibocal/protocols/flux_dependence/utils.py
RoyStegeman: No I was not planning to remove. Is there an advantage to doing so?

RoyStegeman:
You tagged a different person, but I will add the tests now.

src/qibocal/protocols/flux_dependence/utils.py
RoyStegeman: I added an inline comment with the reason for leaving it.

RoyStegeman:
I included bounds which makes it slower, but at least now the parameters are physically sensible and the sweespot can be computed analytically rather than numerically. Possibly we could provide the option of doing a fast fit without bounds, or one with a simpler functional form and determining the sweetspot numerically. In that case the parameters will be physically meaningless/unreliable, but if a user is only interested in getting the sweetspot that is not a problem. 

Or maybe there is some other way to speed it up with bounds.

lballerio:
Hi @RoyStegeman, in the last days you mentioned me a data which was not fitting with RANSAC, did you manage to fit also that one?
Can you please share the problematic signal?

RoyStegeman:
Yes, fitting any one parabola isn’t really the problem. The issue is that making a small change to improve the fit for one can cause another one to no longer fit. So, addressing them all simultaneously would probably require a more complex fitting algorithm, which at this point is likely getting into diminishing returns.

I made a few small changes this morning, but for now I’ll focus on finding something that is relatively stable and saving a large number of example datasets, including some that fail. Then, if we want to, we can always come back later and try to improve it further to account for more of those examples.

RoyStegeman:
It's not perfect (see some of the plots generated that can be generated with `test_qubit_flux_fit.py`), but at this point changing something about the algorithm to accommodate one dataset means breaking the fit to another dataset. The fits did become a lot more stable overall, and at this point continuing is getting into diminishing returns, but I just want to mention that it will be possible to come back to it later and improve on the speed and stability of the fits. 

One thing I am a not sure about is storing so many examples in the `tests_data` directory. They are useful because making the fit more stable for a wide variety of possible features in the data, requires a database containing a wide variety of example datasets. Of course they are not part of the src/ folder, so should not be shipped as part of qibocal, but they do take up space in the repo (~12MB, but if we do this for more protocols it will grow).

src/qibocal/protocols/flux_dependence/utils.py
```diff
@@ -397,3 +374,219 @@ def flux_extract_feature(
...
+        if low <= candidate <= high:
+            candidates.append(candidate)
+
+    if candidates:
+        # If there is at least one sweetspot in the bias window, take the bias with abs
+        # closest to 0.0
+        return min(candidates, key=abs)
+
+    # No sweetspot lies inside the acquired window. Find the closest one to the window.
+    n = int(np.floor(offset + 0.5))
```
lballerio: why `0.5`?

src/qibocal/protocols/flux_dependence/utils.py
lballerio: ok perfect

src/qibocal/protocols/flux_dependence/utils.py
```diff
@@ -397,3 +374,219 @@ def flux_extract_feature(
...
+        # TODO: would be nice to have this as a callback
+        unique_y_count = len(np.unique(yvals[inliers]))
+        score = (
+            _continuity_score(xvals, yvals, inliers),
+            unique_y_count,
+        )
+
+        # Accept as a new best if it has a better continuity score and at least matches
+        # the minimal sample size.
+        if inliers.sum() >= len(subset) and score > best_score:
```
lballerio: I think it would be great if you could comment why you want to maximize the continuity score, I have some intuition but idk if this is correct.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
lballerio: perfect, resolved for me

src/qibocal/protocols/flux_dependence/utils.py
lballerio: > No I was not planning to remove. Is there an advantage to doing so?

No no, I thought usually for production line we don't want to set a seed, but I think in this case we can keep since we generate numbers for extracting data to fit.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
lballerio, line 86: if I remember correctly this field is pretty useless, so not here but we should also consider to delete it.

lballerio (commented):
I see for test we need `parameters.json`, `data.npz` and I strongly believe also `data.json`, but do we need also `action.yml` to upload?
If we can avoid uploading `action.yml` I would avoid.
Other than these comments, I think this PR is good to merge.

RoyStegeman:
@alecandido @scarrazza I'd like your input on what to do with the data examples I stored in the `tests_data` directory. The idea behind having a large variety of examples is that at this point the fit is quite stable, so improvements have to be made carefully. If we encounter a failed fit, we can no longer just tweak the algorithm a bit to capture also that case while assuming that in general it was an improvement. Instead, I would like to have a relatively large database with a variety of examples to run against, to confirm that making any change to accommodate a failed fit, did not deteriorate other fits.

However, I'm concerned about their size, and the fact they are .npz files. Of course they are not part of the src/ folder, so should not be shipped as part of qibocal, but they do take up space in the repo (~12MB, but if we do this for more protocols it will grow). 

Alternatively we can avoid uploading the datasets here and make a separate repo, or have some place where we store links to reports instead of the data itself. They are both slightly less convenient, but will prevent us uploading many binary files. While I uploaded them here, I think I'm actually leaning towards keeping them separate instead, not even because I'm very worried about 12MB, but since it is unclear to me at this point how much it will grow if we do this for other experiments (although there are not many 2d experiments) or when we change something about the protocol and the format of the .npz changes.

scarrazza:
@RoyStegeman, indeed, we could move this data to a git lfs repository, so the CI just run git clone. Another option is to store this in zenodo and modify the conftest.py to download before running tests.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
RoyStegeman: Yes good point, but this is still expected by the flux crosstalk experiment and I do not want to touch that experiment here.

src/qibocal/protocols/flux_dependence/utils.py
RoyStegeman: This is to symmetrize the `np.floor` around 0. It could also have been `np.round`. Having said that, I wanted to try the points on either side of the acquisition window, so I'll change it.

src/qibocal/protocols/flux_dependence/utils.py
RoyStegeman: Okay wrote an explanation. I hope it is clear now.

RoyStegeman:
Let's take some time to think about the best solution. We could also use the qrc cluster as storage and download from there. [Free tier git lfs has a 10GiB/month bandwidth limit](https://docs.github.com/en/billing/concepts/product-billing/git-lfs#free-use-of-git-lfs), which would not be enough for us.

For the time being I removed the .npz files from the commit history and moved them here https://github.com/RoyStegeman/QRC/tree/5e4a6ab9bd96b3a5987f6223d6c6de43b92cd85f/test_data.

> I see for test we need `parameters.json`, `data.npz` and I strongly believe also `data.json`, but do we need also `action.yml` to upload?
> If we can avoid uploading `action.yml` I would avoid.

They provide an easily parseable definition of the experiment that generated the data. I think it's worth keeping.  

With that, I would like to merge this.

.gitignore
```diff
@@ -145,3 +145,6 @@ dmypy.json
 
 # Marimo
 __marimo__
+
+# Plots to visually confirm regression fit is good, they are not for CI testing
+regression_fit_plots/
```
alecandido: Isn't it going to
```suggestion
tests/regression_fit_plots/
```
?

This
https://github.com/qiboteam/qibocal/blob/fa48fab275377e195a5380f4b0a61e1ab6f3b8a9/tests/test_resonator_flux_fit.py#L51-L52
is the same folder of
https://github.com/qiboteam/qibocal/blob/fa48fab275377e195a5380f4b0a61e1ab6f3b8a9/tests/test_resonator_flux_fit.py#L16

Which also raises a few further points. I will comment in there

tests/test_resonator_flux_fit.py
```diff
@@ -0,0 +1,135 @@
...
+    """Run all fits and generate plots to help assess whether a fit is good or not.
+
+    This is intended for use while working on the resonator flux protocol. The
+    PATH_TESTING_DATA directory contains many examples with very different features, but
+    for all it is clear where the sweetspot should be, and thus a good algorithm should
+    be able to find the sweetspot for all of them.
+    """
+    from matplotlib import pyplot as plt  # not a test dependency
+
+    output_base = Path(__file__).parent / "regression_fit_plots/resonator_flux"
```
alecandido: Why isn't it using the defined top-level constant?

And why the constant is not just defined in `conftest.py`? To be consistently reused across the different tests

tests/test_resonator_flux_fit.py
```diff
@@ -0,0 +1,135 @@
...
+        assert (
+            pytest.approx(expected.frequency[qubit], abs=1e3) == fitted.frequency[qubit]
+        )
+        assert (
+            pytest.approx(expected.sweetspot[qubit], abs=1e-3)
+            == fitted.sweetspot[qubit]
+        )
+
+
+if __name__ == "__main__":
```
alecandido: Actually, I'm inclined to believe that all the code that follows would be better stored somewhere else.

We could have a `dev/` folder top-level, or anything else on that line. But putting it below tests seems a bit counterintuitive to me, since this is not part of the test suite (i.e. not supposed to be run by Pytest).

tests/test_qubit_flux_fit.py
```diff
@@ -0,0 +1,136 @@
...
+
+        # overwrite the existing results.json files, such that if we are happy with
+        # the new fit, the regression test can easily be updated.
+        fitted.save(results_path)
+
+        for qubit in expected.frequency:
+            freq, freq_idx = np.unique(data.data[qubit].freq, return_inverse=True)
+            bias, bias_idx = np.unique(data.data[qubit].bias, return_inverse=True)
+            signal = np.full((len(bias), len(freq)), np.nan)
+            signal[bias_idx, freq_idx] = data.data[qubit].signal
```
alecandido: If you load the parameters, you can save this ugly processing...

In general, with @lballerio we started shifting towards raw data storage. As in the [punchout](https://github.com/qiboteam/qibocal/blob/66de4b9a7bc3faf4ee3d9c7cdadb687d486ac920/src/qibocal/protocols/resonator_spectroscopies/resonator_punchout.py#L144-L146), where I dropped the usage of `.register_qubit()`.

I'm not suggesting here to propagate it to the protocol, where you still need to propagate something (since there is no `Parameters` access in the `Protocol.fit()` function).
But, if feasible, we could start avoiding this verbose loading.

However, I may still be losing something, since the initial point of storing NumPy arrays was preserving the shape. While here it seems for some reason to be [flattened](https://github.com/qiboteam/qibocal/blob/66de4b9a7bc3faf4ee3d9c7cdadb687d486ac920/src/qibocal/protocols/flux_dependence/utils.py#L45-L53).
But since you know the dimensions from the parameters, and the order in which the sweepers are executed, you can deterministically reconstruct the shape, and just do `data.data[qubit].signal.reshape(...)`, using the coordinates determined from the parameters. No opaque `np.unique()` call is needed

tests/test_qubit_flux_fit.py
```diff
@@ -0,0 +1,136 @@
...
+from qibocal.protocols.utils import GHZ_TO_HZ
+
+TEST_FILE_DIR = Path(__file__).resolve().parent
+PATH_TESTING_DATA = TEST_FILE_DIR / "tests_data"
+
+RESULT_FOLDERS = sorted(PATH_TESTING_DATA.glob("qubit_flux-*"))
+
+
+@pytest.mark.parametrize("results_folder", [p for p in RESULT_FOLDERS])
+def test_qubit_flux_fit(results_folder):
```
alecandido, line 16: Maybe add an extremely brief docstring to clarify it is a regression test. It may be nice to see it at first sight. We could even create a Pytest marker to label them.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
```diff
@@ -226,41 +265,61 @@ def _fit(data: QubitFluxData) -> QubitFluxResults:
     for qubit in qubits:
         qubit_data = data[qubit]
 
-        # extract signal from 2D plot based on SNR mask
-        frequencies, biases = data.filtered_data(qubit)
+        freq, freq_idx = np.unique(qubit_data.freq, return_inverse=True)
+        bias, bias_idx = np.unique(qubit_data.bias, return_inverse=True)
+        signal = np.full((len(bias), len(freq)), np.nan)
+        signal[bias_idx, freq_idx] = qubit_data.signal
```
alecandido: This is connected to the comment https://github.com/qiboteam/qibocal/pull/1621/changes#r3714239407, but here you do not have access to the parameters, unfortunately.

The alternative option is a bit more elaborate here: save as data just the raw data (even the I/Q pairs, rather than the signal itself) preserving its shape, and store the frequencies and biases arrays generated for the sweepers (or their range representations) in the `QubitFluxData` class.

The reason why it makes this more convoluted than what it is suggested in the other comment is that you need to modify the `*Data` class. However, you could still go halfway, by preserving it and just extend redundantly, saving the array of swept parameters.
But the idea is still to preserve the shape, and avoid reshaping - which will require anyhow to propagate the change. It is actually very frustrating that this has been flattened in the first place...

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
```diff
@@ -226,41 +265,61 @@ def _fit(data: QubitFluxData) -> QubitFluxResults:
...
+        )
+
+        bounds = (
+            [
+                data.qubit_frequency[qubit] * HZ_TO_GHZ - 1,
+                0,
+                -1,
+            ],
+            [
+                data.qubit_frequency[qubit] * HZ_TO_GHZ + 1,
```
alecandido: Replace `HZ_TO_GHZ` with `nano` from SciPy

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
```diff
@@ -226,41 +265,61 @@ def _fit(data: QubitFluxData) -> QubitFluxResults:
...
+                np.inf,
+                1,
+            ],
+        )
```
alecandido, line 290: Are these bounds actually worth it?

They seem extremely lenient, since they are:
- searching the sweetspot in a 2 GHz band
- imposing a positive curvature, and
- restricting the offset in the largest possible domain

The only requirement which seems to carry a significant amount of information is the second. While the other two are just providing boundaries for fits that most likely are already badly failing, if the data can not constrain the sweetspot close to the data region.

A different option could be to lift these bounds here (retaining just the curvature one) in the hope to speed up the process, and just impose them in the fitting/plotting, considering the fit as "failed" if the results is exceeding these bounds. And thus not reporting it in the final plot (and possibly excluding from the update as well).

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
```diff
@@ -187,6 +175,57 @@ def _acquisition(
...
+
+    # Sometimes we observe bright lines for certain bias values that are constant in
+    # frequency
+    centred_signal = signal - np.median(signal, axis=1, keepdims=True)
+
+    peak_biases, peak_frequencies = [], []
+    for bias, signal_row in zip(biases, centred_signal):
+        # The Gaussian filter reduces noise in the background and helps make a noisy
+        # peak into a stronger signal.
+        smoothed_row = gaussian_filter1d(signal_row, sigma=2)
```
alecandido: I'm not sure I would apply this processing per row.

It's a matter of intuition, and I should try to formulate some examples to clarify - but I'm hesitant to do it, because they may not generalize.

However, I would rather try to keep the radius as small as possible, such that any arbitrarily small "bulk" of the signal (if the thickness of our signal exceeds the single pixel) is strongly preserved.
Instead, by doing it in 2D, we gain the other dimension "for free". Our signal is still 1D, so it will lose weight by averaging only for the orthogonal dimension. Instead, the noise is assumed to be uncorrelated, which makes it "0D", and it will lose intensity by averaging in both directions.

In a sense, if you imagine our signal to be a vertical line, by applying the horizontal filters you are diluting signal and noise by the same dimension. While if you use 2D masks, the additional neighbors in the signal region (the vertical ones) will be coherent. And in the noise region they will be incoherent, gaining the extra averaging dimension - and thus extra contrast.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
```diff
@@ -187,6 +175,57 @@ def _acquisition(
...
+
+        # The standard deviation is computed from the median absolute deviation instead
+        # of the standard deviation itself to avoid the peaks in the arc from affecting
+        # the estimate of the background noise. While this prominence threshold is
+        # somewhat motivated, it is still a choice and it has been observed that the
+        # result is not very sensitive to it and probably it is even fine to set the
+        # threshold to 0.
+        row_mad = np.median(np.abs(smoothed_row - np.median(smoothed_row)))
+        row_std = 1.0 / (np.sqrt(2) * erfinv(0.5)) * row_mad
```
alecandido: I understand the reasoning behind the usage of the median shifts and its clever,  because it exploits the different dimensionality, and thus the hierarchy in the signal vs background region - by treating the signal as an outlier.

However, I'm not sure I understand the logic for the prefactor in `row_std`.
It seems like you're normalizing the median by the coordinate of the 0.75 quantile of a normal distribution. But since the quantile itself is arbitrary in the first place, not sure what's the purpose of going through `erfinv()` to eventually rescale that value by an equally arbitrary 50% more...

alecandido: Btw, did you compute which is the expected value of `row_mad` for a white noise 1D-Gaussian-filtered normally-distributed row?

This can be done even empirically, by extracting a bunch of samples.

Even though you're still missing a scale for it, since rows are not normalized, so the noise can be arbitrarily scaled (i.e. $\sigma$ can not be assumed to be just 1 as a rough approximation, since we are completely missing a scale).

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
```diff
@@ -168,28 +161,77 @@ def _acquisition(
...
+    bias: npt.NDArray[np.float64],
+    signal: npt.NDArray[np.float64],
+) -> tuple[npt.NDArray[np.float64], npt.NDArray[np.float64]]:
+    """Extract the most prominent peaks in the resonator (flux,frequency) landscape. At
+    most one peak per flux bin.
+    """
+    # Sometimes there are bright spots for a given bias. Not sure what causes them,
+    # but this should get rid of them.
+    median_per_bias = np.median(signal, axis=1, keepdims=True)
+    centered_signal = signal - median_per_bias
```
alecandido, line 180: Do you have an example at hand? Just to develop some intuition

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
```diff
@@ -168,28 +161,77 @@ def _acquisition(
...
+    is_peak = []
+    for bias_val, row in zip(bias, centered_signal):
+        # There may be fluctuations along the frequency axis caused by elements such
+        # cables or amplifiers. In principle this is bias independent, so we do the same
+        # as we did before, and subtract the median per frequency. However, the arc may
+        # be very flat and take up the majority of the window (perhaps together with
+        # another background feature of the same extremum), in which case we end up
+        # subtracting the arc rather than background. To avoid this, we use
+        # median_filter.
+        samples_per_peak = np.ceil(APPROXIMATE_RESONATOR_PEAK_WIDTH / np.diff(freq)[0])
```
alecandido:
```suggestion
        samples_per_peak = np.ceil(APPROXIMATE_RESONATOR_PEAK_WIDTH / (freq[1] - freq[0]))
```

In this way, to me, its even clearer that you are using the first one because they are all the same. And it is irrelevantly more efficient (at least asymptotically)

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
```diff
@@ -168,28 +161,77 @@ def _acquisition(
...
+        baseline = median_filter(row, size=int(samples_per_peak), mode="mirror")
+        residual = baseline - np.median(baseline)
```
alecandido: I'm not sure I understood how the median filter is avoiding the problem of the median falling in the signal region.

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
```diff
@@ -168,28 +161,77 @@ def _acquisition(
...
+    for bias_val, row in zip(bias, centered_signal):
+        # There may be fluctuations along the frequency axis caused by elements such
+        # cables or amplifiers. In principle this is bias independent, so we do the same
+        # as we did before, and subtract the median per frequency. However, the arc may
+        # be very flat and take up the majority of the window (perhaps together with
+        # another background feature of the same extremum), in which case we end up
+        # subtracting the arc rather than background. To avoid this, we use
+        # median_filter.
+        samples_per_peak = np.ceil(APPROXIMATE_RESONATOR_PEAK_WIDTH / np.diff(freq)[0])
+        baseline = median_filter(row, size=int(samples_per_peak), mode="mirror")
```
alecandido: Why not using the *nearest* mode?

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
```diff
@@ -168,28 +161,77 @@ def _acquisition(
...
+            continue
+
+        # Keep the most prominent extremum, along with its prominence, and whether it is
+        # a peak or dip
+        best = peaks[np.argmax(props["prominences"])]
+        bias_pts.append(bias_val)
+        freq_pts.append(freq[best])
+        is_peak.append(row[best] > 0)
+
+    # Keep only the dominant extremum type and ignore extrema of the opposite feature
```
alecandido: Maybe worth to comment that you expect the dispersive shift to be just in one direction, and the rest would be something different.

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
```diff
@@ -217,62 +258,68 @@ def _fit(data: ResonatorFluxData) -> ResonatorFluxResults:
...
+                "g": popt[0],
+            }
+            matrix_element[qubit] = popt[3]
+            sweetspot[qubit] = utils.select_sweetspot(
+                popt[2],
+                popt[3],
+                (np.min(data[qubit].bias), np.max(data[qubit].bias)),
+                max_distance=0.3,
+            )
+            resonator_freq[qubit] = fit_function(sweetspot[qubit], *popt) * GHZ_TO_HZ
```
alecandido: Other instances of custom conversion factors. I would recommend using SciPy ones.

alecandido (commented):
Still missing the RANSAC implementation itself (i.e. the entire `utils.py` diff), but better to start releasing something.

I did not think very carefully about the bounds for the resonator flux arc. But being a bit more convoluted, better to postpone it to when we will discuss again the role of the parameters, and what we expect to be able to constrain.

In general, I agree with some of the choices in the preprocessing, but I would reconsider using more often the 2D Gaussian (as described in detail in one of the comments), but also the ZCA, which was working quite well for the gradient removal.
It's true that fitting both peaks and dips with a better reference is making less critical the problem of the gradient, since the analysis is done per-row anyhow. But my intuition is still that it is propagating some 2D information in the rows, so we may gain a bit in contrast. But I'm not at all confident about this statement, and it could be irrelevant

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
RoyStegeman: Because it may cause a peak/dip near a boundary if the large value happened to be very large/small

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
RoyStegeman: I think `HZ_TO_GHZ` and `GHZ_TO_HZ` are more readable than multiplying by `nano` or dividing by `giga`. What is the reason for getting rid of them?

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
RoyStegeman: If the bounds are so loose they don't significantly impact the result of the fit, then removing them by setting the limit to `np.inf` or `-np.inf` will not gain us a performance improvement. 

One thing I did consider was to use symmetries in the equations to reduce the constraints. E.g. `offset` is just about periodicity so it can be left unconstrained but take the mod afterward. This didn't allow me to fully remove get rid of all constraints. I tried removing all constraints and then applying this symmetrization where possible, but I found that in many cases none of the RANSAC instances fell within the bounds even after symmetrization. Perhaps an idea like this can still be used to speed up the fit though.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
RoyStegeman: > the additional neighbors in the signal region (the vertical ones) will be coherent

This is not universally true - in the example in the description of this PR also the signal would be diluted if we used a 2d filter. Because of this my intuition was that features along the vertical axis vary less between different datasets, and thus will allow to build a more stable filter. 

Having said that, I'm not saying that a 2d filter cannot be an improvement of the current preprocessing, it may well be, but as you say, it requires testing on examples to conclude. At this point I already iterated quite a few times and I believe optimizing the preprocessing is really diminishing returns that should not go in this PR.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
RoyStegeman: The prefactor is the relation between the median absolute deviation (MAD) and standard deviation. Another way to think about it is that MAD is the 50% confidence interval and the factor rescales a 50% confidence interval to 68% assuming a Gaussian distribution. https://en.wikipedia.org/wiki/Median_absolute_deviation#Relation_to_standard_deviation

It's just there because I'm used to thinking of signal strength in terms of standard deviations, but tbh the prominence threshold is pretty much arbitrary and I think even setting it to 0 does not change the outcome of the fits in a significant way.

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
RoyStegeman: image

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
RoyStegeman: changed.

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
RoyStegeman: I'm resolving this comment to do the discussion in the other thread.

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
RoyStegeman: It doesn't originally my idea was actually based on the observation that part of the background varies slower than the peak we're interested it. So by using a sliding window that is wider than the expected peak by a factor >2 such that the median is not part of the peak, we could remove the local background, i.e. it should have been something like
```
        baseline = median_filter(row, size=5*int(samples_per_peak), mode="mirror")
        residual = row - np.median(baseline)
```

I tried and the result is effectively the same as the current implementation, so I'll go with this since at least we have some motivation for it.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
alecandido: That they are defined over and over, with inconsistent usage and documentation.

Moreover, I see the apparent advantage, since they are seemingly more explicit. But I would claim it is just weird perspective: the unit is carried by the `data.qubit_frequency[qubit]` quantity, and it should already be unambiguous. Thus, the constant name is just redundancy.
If you were writing it in formulas, it would be $10^9$ or $10^{-9}$, and you would be surprised if someone start qualifying it as conversion between Hz to GHz, or s to ns. They are factors, and they are just all the same.

I would say it becomes more consistent to just use `giga` and `nano`: it is sufficiently explicit, and conveys the message of the unit conversion. Being defined in a third-party library and clearly documented, it saves the diff generated by "ah, but this is already defined there, you could import and reuse" after copy-paste re-definitions.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
alecandido: > If the bounds are so loose they don't significantly impact the result of the fit, then removing them by setting the limit to `np.inf` or `-np.inf` will not gain us a performance improvement.

Wait: is the `trf` algorithm significantly slower even with the bounds extended to infinity? Considering the [dogleg algorithm described on wikipedia](https://en.wikipedia.org/wiki/Powell's_dog_leg_method), if the radius of the trust region is extremely large, it should just become just Gauss-Newton. At high-level, it should not be terribly different from Levenberg-Marquardt, unless the gradient descent part is giving a large boost.

If this is the case, maybe better to resort to just Levenberg-Marquardt with a custom loss. In any case, we hardly have sharp boundaries, and they are always approximate.

> E.g. `offset` is just about periodicity so it can be left unconstrained but take the mod afterward.

True, it's about the periodicity. But with an unknown period...

> One thing I did consider was to use symmetries in the equations to reduce the constraints.

Even for this, possibly the regularization is just simpler. Especially if rather flat (thus agnostic) in the "acceptable" region

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
alecandido: > This is not universally true - in the example in the description of this PR also the signal would be diluted if we used a 2d filter.

While this is certainly true, it should not get worse than the 1D performance, since the signal will be reduced by the same amount of the noise, keeping the relative contrast unchanged.
Btw, that's also quite a degenerate example, with an odd distribution of pixels: if you such large jumps, you often want to increase the density in the offset axis. Though, as said, the algorithm should work as well as the 1D one.

> Having said that, I'm not saying that a 2d filter cannot be an improvement of the current preprocessing, it may well be, but as you say, it requires testing on examples to conclude. At this point I already iterated quite a few times and I believe optimizing the preprocessing is really diminishing returns that should not go in this PR.

Well, here I disagree: the 2D filtering was the previous workhorse, and we had several hints that was working well enough.
Then, the previous algorithm was failing as a whole, while passing on test cases. Now, you performed the study on a wider and more representative set of test cases, and in this sense the proof is superior. But, when we inspected the failing instances for the previous algorithm (which are close to be the sources of your test set), the outcome was almost always blaming the clustering part, not the preprocessing and peak finder.

I agree about being concrete, and avoid wasting resources in a territory of diminishing returns. But since this 1D preprocessing is a separate change itself, and not at all a necessary part of the RANSAC algorithm, the strategy shift should be justified for it in the first place.

Even if the PR is not strictly atomic, it is still acceptable to work on both together. But, as a consequence, it is quite natural that it will also increase the review time, and I'm against to bundle as a single package which "just works". Unless we're in extreme urgency

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
alecandido: Ok, I see. But even in terms of $\sigma$, taking 1 as a threshold is just an arbitrary amount, so I would avoid the conversion. You could use the MAD, or it rescaled by the amount you preferred, but no need to get the equivalent standard deviation.
Especially because there is a reason why you are avoiding the actual standard deviation. So, getting it back is just confusing...

> I think even setting it to 0 does not change the outcome of the fits in a significant way.

This I'm not sure I understand: it means that, effectively, you're not being able to threshold anything, and always end up with one peak per row.
If that's the case, yes, we could save all this business about the MAD, and keep it simpler.

But that seems to point out that the chosen rescaling is really not optimal: in principle, the reason of cutting with the MAD is that the prominence of the signal is a much stronger outlier than the prominence of a peak in the background, when normalized with the MAD. If this assumption is as reasonable as it seems, there should be a threshold which is mostly separating the two regions.
Then, we may be conservative, and accept a large fraction of noise to retain signal, since the RANSAC is noise-resilient itself. But the prominence cut could be a more efficient way of performing this task. Giving up in full does not seem fully motivated...

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
alecandido: I'm a bit lost: that's what I thought about the `mirror`.

If you have a slope, possibly even because of a fluctuation, and you mirror, that becomes a local maximum/minimum.

But if you saturate to that value, it would be just less prominent, isn't it?

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
RoyStegeman: That is for the qubit, where we actually don't make the same assumption. Here it is jus wether we measure reflection or transition. I'll indeed clarify the comment. [69102d2](https://github.com/qiboteam/qibocal/pull/1621/commits/69102d21fff1339f97d28f92816344c07125c049)

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
alecandido: I see your point now.

However, I guess you'd need a stable hierarchy between the background features size and the resonators width. Which usually is there, and it is the way we distinguish them - but not the only one.

Often, they are clearly visible also because their transmission is much lower than the background. Which is a vertical criterion, as opposed to the horizontal one.
But, in the end, it is always a combination of the two, and a local one.

<p align="center">
image
</p>

> I think that as it is right now, the median filter only results in some smoothening of noise and sharpening of the peaks. I'll update the comment to reflect this.

If the window is just smaller than the expected width, this is what it should happen. Not much different from the Gaussian blurring we are using for the qubit arc.
For the Gaussian blurring in `main` we were using 1 pixel as radius. And in practice it was quite reasonable. The justification was that:
- making it significantly smaller makes the filter almost ineffective (i.e. an identity)
- it was already quite enough to remove a significant amount of incoherent noise

In this case, measuring in pixel was quite effective, because it was the "unit of information", and strongly related to the nature of our noise - assumed to be incoherent per bin.
And if the signal is so small to be contained in a single pixel per row, and the pixels are far apart in different rows (an extreme version of the data in the OP), than no local filtering can be effective, since the data are locally indistinguishable from incoherent fluctuations. You need a global model to "connect the dots" (what the RANSAC is doing), and a lot of contrast as well - which, by definition, would survive the filtering.

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
RoyStegeman: With `nearest` your window appends the final value in the row, for all entries in the window that go out of the data range. So if it is an outlier, than at some point the median will be heavily affected. By doing a mirror, the values that get appended for the points in the window that fall outside the data region, is just walking back from the edge, so the median is not affected by an outlier in the same way. 

Having said that, I did not simulate nor can I reproduce it now, so perhaps my memory is incorrect that the mirror instead of nearest solved the issue.

alecandido: Ok, you're right. Now I understood

RoyStegeman:
I removed the function that updates the regression test data and generates the corresponding plots. While I believe it to be useful, it was implemented in a very-proof-of-concept style and if we want to propagate the idea to all protocols we should do it with some more care. 

Since we have already started the discussion about the database of examples (which the function is a part of), I believe it   is better to invest the effort there and avoid adding noise by having this remnant in main.

tests/test_qubit_flux_fit.py
RoyStegeman: added the docstring

tests/test_qubit_flux_fit.py
RoyStegeman: Yes that's true, the idea is that now we don't make the assumption and thereby decouple the generation of the data from the processing. But it is also true that there is no reason to assume the signal will be stored with frequency-first and varying all bias since frequency can be swept and bias cannot so the order will not change. 

changed it.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
RoyStegeman: > That they are defined over and over, with inconsistent usage and documentation.

Doing a search only returns a single place where they are defined for both. 

I agree that Scipy's constants are a good way to signal that the multiplication factor is just about unit conversion, but ot is not so obvious that frequency is in Hz since `transmon_frequency` for example is in GHz, while naively I would expect that to be Hz, so I think in that case it is a helpful indicaiton. We could improve this handling such that the conversion to GHz is more localized and therefore one can assume that all frequencies are Hz unless basically changed in the line before, but at the moment it's not so clear.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
RoyStegeman: > Wait: is the `trf` algorithm significantly slower even with the bounds extended to infinity? 

I don't believe so. You pointed out that the bounds are already quite loose so we could completely remove them. I wanted to point out that if they truly loose already, then going to infinity doesn't change performance, and if it does change performance then the bounds are also genuinely constraining the fit so we cannot just loosen them without affecting the fit result. 

> Even for this, possibly the regularization is just simpler. 

Bu regularization you mean a custom loss?

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
RoyStegeman: I tried putting back your filtering, and for qubit it improves 
image
image

 
However, for the resonator I did not have an example of a failed peak detection (I have a failed fit, but the peak detection is perfect), and now one of the examples failed. It seems to be due to the failed automatic sign detection related to the peak and dip for a fixed bias, where in your filter the sign was assumed. However, since I did not have an example of a failed detection to begin with, I'm not convinced this is worth chasing.
image
image

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
RoyStegeman: This is replaced in favor of the old pre-processing.

src/qibocal/protocols/flux_dependence/resonator_flux_dependence.py
alecandido: Ok, I see the problem. And I don't know where is it coming from, as if the magnetic was affecting the output of the transmission line...

After some experiments, it seems that the median subtraction works best for this type of gradients, where you know to be columns-aligned (and you remove the gradient per row).
However, even the ZCA is able to provide some results in this case - though not as good as the median subtraction.

image

But with the benefit that works even for different backgrounds
image

[background_removal.py](https://github.com/user-attachments/files/30792924/gradient_removal.py)

alecandido: For a comparison to the previous state: this is what it happens in the first case, when you add an `ndimage.gaussian_filter(_, 1)` downstream to the ZCA

image

Still not as clean, but it seems almost equally good for a peak finder.

RoyStegeman: Yes indeed, if you wish I can implement it like that. Here you can see an example of just zca and minmax: https://github.com/qiboteam/qibocal/pull/1621#discussion_r3729359818. 

But as I also mention, the fits succeed for all the datasets that I have (except for one, but the peak finding is perfect), so the exercise is mainly academic.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
alecandido: > I wanted to point out that if they truly loose already, then going to infinity doesn't change performance

I actually meant "is it slower **than `lm`** even with bounds extended to infinity?".

I agree that they are not so loose not to affect the fit. My perspective was more than they are sufficiently loose that they convey little physical information. If some steps stop on the boundary, most likely the fit is hard to recover - unless it finds a canyon back...

> Bu regularization you mean a custom loss?

Yes, exactly

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
alecandido: *Ok, now this discussion is mixed with the one below. Nothing bad :)*

However, I agree that the second case is unpleasant: the arc developed a shadow, which is not present in the original data. And this confuses the peak finder.

Somehow, this is due to the coherent noise, which is what the ZCA is acting on.
While it is easy to get some intuition for what the Gaussian blurring is doing, and even in this case it seems to cause no problem, it looks like the ZCA is doing more than expected.

You could actually see ripples and thin shadows even in the other discussion https://github.com/qiboteam/qibocal/pull/1621#discussion_r3729521711. But they are so tiny that the Gaussian filtering is suppressing them.

In this case, the original data already have very little contrast. And that horizontal strip is the main both the main source of contrast, and the first one to introduce a shadow (the yellow spot).

The following is mostly guesswork.
It seems that the median subtraction works well in this case exactly because it encodes the information about the possible presence of these horizontal stripes.
At this point, we could argue that it always enhance or preserve the contrast, since it is just a horizontally-constant subtraction. Which is certainly based on background, under the mild assumption that our 1D curve is not as thick as the whole picture (which would arguably make it not 1D any longer...).
Then, we may decide to always apply the median subtraction *before* the ZCA and Gaussian blurring.

At this point, I'm quite convinced that preserving the Gaussian blurring is harmless, and possibly beneficial.
Keeping the ZCA should also be helpful in some cases. But I'm not convinced at all it will not regenerate the shadow again, even after the median subtraction. In which case, we could ditch the ZCA for the resonator, and keep just the median subtraction + Gaussian blurring. It should be anyhow a sensible candidate

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
alecandido: Ok, but then I would open an issue for the future.

Despite the update strategy, minmax + 0.2 is as arbitrary as a thresholding mechanism, I agree.
And it is well possible that we have a chance of getting a much better separation of the two clusters. And the MAD could be an ingredient of that. Regarding it, also to me it seems pretty solid, and better than `max - min` in a noisy environment.

The part I was most skeptical about was exactly your comment about the possibility to set the threshold to 0 - which triggered the consideration that we need better motivation for a threshold value.

alecandido:
> Since we have already started the discussion about the database of examples (which the function is a part of), I believe it is better to invest the effort there and avoid adding noise by having this remnant in main.

Agreed. I'm opening the discussion https://github.com/qiboteam/qibocal/discussions/1651

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
RoyStegeman: The reason I said we can set it to 0 are the following:
1) The prominence threshold is arbitrary.
2) At some point I found examples where filtering the noise required a threshold that also filtered the peaks of the arc. Of course this may be avoided with a better pre-processing, but that's a separate point. 
3) Because RANSAC is specifically intended for noisy background. That doesn't mean we shouldn't worry about background peaks at all, because they may not be from random noise but some feature such as a straight line, in which case the RANSAC can still get thrown off. Also, if we can filter more without losing the peaks on the arc it will make the fit faster.

alecandido: I perfectly agree on point 3.

Regarding points 1. and 2., they are correct, but I believe we can assume that we have some distributions of prominence for both the intended signal and background peaks.
In this case, the goal is of course not to discriminate them exactly at the time of peak finding. And of course, because of RANSAC resilience, we will lean towards accepting more peaks, to retain more signal at the expense of keeping even more background peaks.
But we can also afford to lose some signal peaks. And, if it is removing a significant amount of background ones, it is certainly worth

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
RoyStegeman: I wrote down some of the things I tried to make it faster without much success https://github.com/qiboteam/qibocal/issues/1652

If you insist I can try a custom loss after the holiday, although I believe the bounds were slower not because each iteration takes much longer but because more iterations are needed. This would also be the case with a custom loss. While I believe at the moment the fit in some cases takes longer than is acceptable, I also believe there is some value in merging it as it is now, and using for a while to see how stable the fit really is on unseen data. Namely, if we want to change something about the fitting algorithm or the filtering because we realize there is still some instability, this may also affect how we would optimize the fitting part.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
RoyStegeman: >  In which case, we could ditch the ZCA for the resonator, and keep just the median subtraction + Gaussian blurring. It should be anyhow a sensible candidate. 

This is what we do now. Also because currently I have no examples of a bad extraction of peaks using this algorithm for the resonator. For the qubit I did put back the ZCA + 2d gaussian filter + minmax scaling, because it performs better than what I had. However, the minmax scaling I'm not so convinced by since there are rows (fixed bias) where there is no arc, so doing minmax means blowing up the noise. 

Actually that last thing might be a good way to filter some noise and reduce the ransac time.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
alecandido: As I said, I'm perfectly willing to open issues and postpone non-blocking improvements.

> If you insist I can try a custom loss after the holiday, although I believe the bounds were slower not because each iteration takes much longer but because more iterations are needed. This would also be the case with a custom loss.

Glad that you checked, I was not convinced at all of the source of the slowdown.

However, if it takes many more iterations than in the unbounded and non-physical case, it could well be that it depends on the shape of the boundary. Possibly, once it crashes on the boundary itself, it is making much smaller steps - which leads to the increased number of iterations. In such a case, a softer penalty term could avoid getting trapped with small gradients - or not.

Anyhow, a solution should start with an analysis of the problem, understanding as much as possible which is the source of the slowdown. Just throwing options and testing them has a high chance of wasting a lot of time. And it could result in poor generalization.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
alecandido: > Doing a search only returns a single place where they are defined for both.

Glad it is the case. It was not in Qibolab, where it was repeated (at least per-driver).

> but ot is not so obvious that frequency is in Hz since `transmon_frequency` for example is in GHz, while naively I would expect that to be Hz

Exactly, this is what I mean. We should reduce choose a single internal unit, and only convert at the interface and external calls, where needed.
Following Qibolab, I would go for Hz and s. For superconducting it would have been just better to use GHz and ns, but it is what it is. And, adopting the standard units, we can remain less opinionated, and more open also to other technologies.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
alecandido: To base the result on performances, at some point we need to be rather careful in the way we define and measure them. Otherwise, we risk "overfitting".

In any case, we are not yet at the point were results become little intelligible. So, we could still work on the kind of considerations you're now making for the minmax.

About the individual items:
- per-bias median subtraction is just fine
- for the ZCA, I'd be inclined to keep trying, since it could remove patterns which would be otherwise ignored - however, applying it only after the median subtraction, just as an attempt - we already saw it failing without
- the 2D pixel-size Gaussian blurring it seems just beneficial, and I see nothing taking its role in the resonator
- the minmax scaling was needed just for prominence thresholding, but the MAD proposal seemed more robust - I'm not convinced that enhancing the rate of background peaks gives any help to reduce the RANSAC time

Another thing I wonder is whether we should do a two-pass peak detection, to first identify the sign of the peaks, and then just look for those. If peak-detection is very cheap even for many pixels, it could be worth to reduce the noise in the RANSAC.

alecandido:
Another thing that we may want to take into account is that, for a (relatively) large amount of power, the flux arc is broadening, and developing shadows:
image

In extreme cases, the broadening can span the full picture, though you can still see an arc:
image

We may want to add some of these to the data set. But we may still decide that we do not want to solve a problem as complicate as the second case.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
RoyStegeman: Since you agree to consider it non-blocking improvement, let's move further discussion to this issue https://github.com/qiboteam/qibocal/issues/1652

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
RoyStegeman: Agreed on the individual items and that's what's done in the state of the code as of [4166031](https://github.com/qiboteam/qibocal/pull/1621/commits/4166031b37b14144476c2dec14494fe7dfe70215)

RoyStegeman:
> Another thing that we may want to take into account is that, for a (relatively) large amount of power, the flux arc is broadening, and developing shadows:

The examples contain one dataset (see image) that has some shadowing similar to your first image but not the second. If you believe them to be useful additions to the development datasets, which I believe they will be since the arc is clear in both, then adding them there would be the best thing to do. 

image

RoyStegeman:
@alecandido please have another look. We had some long discussions, but I believe [4166031](https://github.com/qiboteam/qibocal/pull/1621/commits/4166031b37b14144476c2dec14494fe7dfe70215) improves many relevant things. In particular the filtering based on height expressed in MAD quite reliably removes a lot of the background peaks.

src/qibocal/protocols/flux_dependence/qubit_flux_dependence.py
```diff
@@ -187,6 +180,52 @@ def _acquisition(
...
+        # Use find_peaks instead of argmax because there may be nothing in a row. Try
+        # both peak and dip per row, since this may differ per row due to moving of the
+        # resonator frequency.
+        median_subtracted = row - np.median(row)
+        peaks, peak_props = find_peaks(
+            median_subtracted, height=height_limit, prominence=0
+        )
+        dips, dip_props = find_peaks(
+            -median_subtracted, height=height_limit, prominence=0
+        )
```
alecandido, line 208: In order to put a prominence cut, we may actually normalize by the median, instead of subtracting. Something like `row / np.median(row) - 1`.

But we would be left with the problem of "which threshold". Let's forget, for the time being

src/qibocal/protocols/flux_dependence/utils.py
```diff
@@ -397,3 +414,235 @@ def flux_extract_feature(
...
+        inliers = residuals_all < residual_threshold
+        if inliers.sum() == len(xvals):
+            # all points are inliers, so we cannot do better
+            best_inliers = inliers
+            best_params = popt
+            break
+
+        # Use a continuity score to reduce the sensitivity to a noisy background.
+        # Especially if the arc is only in a small part of the bias range.
+        # TODO: would be nice to have this as a callback
```
alecandido: Do you mean to expose the (loss) function as a parameter of the algorithm?

src/qibocal/protocols/flux_dependence/utils.py
```diff
@@ -397,3 +414,235 @@ def flux_extract_feature(
...
+
+        # Prefer fits with long consecutive inlier runs (high continuity) over many
+        # scattered inliers, making noisy fits less likely to be selected.
+        if inliers.sum() >= len(subset) and score > best_score:
+            best_inliers = inliers
+            best_params = popt
+            best_score = score
+            # Re-estimate how many trials are needed to have `stop_probability`
+            # confidence of drawing an all-inlier minimal sample.
+            denom = np.log(1 - (best_inliers.sum() / len(xvals)) ** len(subset))
```
alecandido, line 623: Not sure what is the most explicit alternative, but `len(subset)` seems something dynamically dependent on `subset`, while in practice it is just `function_dof`.

alecandido, line 623: I'm also not sure whether this `denom`, and the whole `N_needed` truly needs to be estimated dynamically: we are mostly fitting under the expectation that there are at least `function_dof` inliers.

I know that this has the effect of making an earlier termination if we have already found a large enough number of `inliers`. But, in principle, we'd like all of them to be inliers.
Is this better than just relying on `max_trials`? Or estimating `max_trials` with this formula, but replacing `best_inliers` with `function_dof` as well?

I'm not suggesting to change this now. I was wondering whether this may be a reason for slower fits. But it should be checked.

alecandido (approved):
A few more remarks, but I'm fine with the current state of the PR

src/qibocal/protocols/flux_dependence/utils.py
RoyStegeman: Yes that's what I meant, but I no longer agree with this. Since we only use it for flux dependence protocols and in the directory structure it's even placed there, and for flux dependence it was decided to go with this, we should not generalize beyond what we need. It's also easy enough to make the loss function an input if it's something we need. I'll remove the TODO.

src/qibocal/protocols/flux_dependence/utils.py
RoyStegeman: Yes that's true, but evaluating `len` is not costly, and if `len(subset)` were ever decoupled from `function_dof` this would be wrong as well. It also makes it easier to confirm that what we do here is the same as the prescription for the dynamic ransac iteration determination.

RoyStegeman: As mentioned some lines above, this determination for the number of iterations is standard literature and I did not scrutinize it too much. 

If the outlier ratio is higher we'll want to do more trials than with hardly any outliers. In principle if there are no outliers then any sample of Ndof datapoints would work. Setting a large fixed iterations wastes a lot of time that we don't always have to pay, so the reason I implemented this dynamic number of iterations was to avoid an excessive number of iterations. 

We could replace `best_inliers.sum()` with `function_dof` but then we lose the dependence on the ratio of inliers and may be overly conservative in doing many iterations. 

> Is this better than just relying on max_trials?

I would say yes, because there is a large variance in the inlier ratios between different datasets, from datasets where we don't have outliers at all, to ones where 95% are outliers.

alecandido: > We could replace `best_inliers.sum()` with `function_dof` but then we lose the dependence on the ratio of inliers and may be overly conservative in doing many iterations.

No wait, `function_dof` is strictly more than `best_inliers.sum()`, and this appears in the denominator.
By approximating $\log\(1-x\) \sim -x$ we would obtain
```py
N_needed ~ stop_probability * len(xvals) ** len(subset) / best_inliers.sum() ** len(subset)
```

So, always saturating `best_inliers.sum()` to `function_dof`, we should reduce `N_needed`, to a non-adaptive value which should already be the "asymptotic" one for the adaptive formula. Which, being lower, it could be passed by the number of iterations, before being adopted as a limit.

image

> As mentioned some lines above, this determination for the number of iterations is standard literature and I did not scrutinize it too much.

I just read the Wikipedia page you linked. Essentially, there it is explaining that the value is *adaptive* in the sense that you set a certain probability which is acceptable to you for failing the search, and then, according to the parameters of the problem, the formula estimates how many iterations are approximately needed in order to obtain that probability.

But it is not adaptive during the *evolution*, i.e. it is not changing during the iterations, it is a fixed target.

What you are doing is to say "not enough inliers, try harder". But this effectively invalidates the meaning of `stop_probability`: that's not any longer the target failure rate, because if it appears to be failing, you are just trying harder.
But, artificially, you are just lowering the `stop_probability` at the expense of the number of iterations. As part of that failure rate, there would be cases in which the inliers are there, but you are repeatedly failing to identify them in a fixed number of iterations.

> I would say yes, because there is a large variance in the inlier ratios between different datasets, from datasets where we don't have outliers at all, to ones where 95% are outliers.

Then, indeed, there is no trivial way to estimate $w$ (the inliers ratio), and what you do is to dynamically estimate a lower bound, looking at the data themselves.

But, my point was that we already have a more restrictive lower bound. If we are attempting to fit a function with `function_dof` degrees of freedom, we assume that, in the whole dataset, there are at least `function_dof` inliers - otherwise our parameters would be under-determined. And `function_dof` is already more restrictive than `best_inliers.sum()`, despite being the bare minimum we would allow.

---

I would be absolutely in favor of supplementing your approach in case we were going to extract a sample larger than `function_dof`. In that case, we could start `N_needed` from the one determined through `function_dof`, but we could make it even smaller, if we already passed the `function_dof` inliers in our `best_inliers` selection.

Than, this is on paper. But I would argue that your algorithm can be just replace with a smaller `stop_probability`. And, eventually, if it's small enough, it will be cut by `max_trials`. And that's why I was asking

alecandido: Fine by me.

In general, I would not guide a choice anticipating what may come in the future. Though I'm prone to do it myself.

In this case, I agree that `len()` is inexpensive. It is truly fine both ways.

RoyStegeman: > No wait, `function_dof` is strictly more than `best_inliers.sum()`

No, `best_inliers.sum() >= function_dof`. The if-branch that writes `best_inliers` is conditional on `inliers.sum() >= len(subset)`.

Ignoring the continuity scoring, we're optimizing for the largest number of inliers defined as `inliers = residuals_all < residual_threshold`, where `residuals_all` is evaluated for peaks, not just those in the random subset of size `function_dof`. So `N_needed` is updated during the fit.

This is an example of how `N_needed` may develop for a qubit flux fit.
```
N_needed: 13167.037821395525
N_needed: 25720.163157272193
N_needed: 1642.8552020558734
N_needed: 42.263121242434885
```

alecandido: > No, `best_inliers.sum() >= function_dof`. The if-branch that writes `best_inliers` is conditional on `inliers.sum() >= len(subset)`.

Ok, then I missed this part, and I misunderstood as the number of inliers within the sample.

In this case, it is coinciding with what I wrote I'd be in favor of. And I am :)

RoyStegeman: > I misunderstood as the number of inliers within the sample.

This is what I suspected. I'll add a comment to emphasize it more clearly.

PR #1622, opened 2026-07-21
FIT mode now uses platform with no hardware memory.

lballerio:
We noticed that when executing only FIT operation (e.g. though `qq fit` command) Qibocal was building the platform using the state of `platform_folder` (e.g. the ones in `qibolab_platforms_qrc`) and not the state saved in the `experiment_data_folder`.
We fix this issue in the PR, also avoiding to construct the `Hardware` sensitive fields of the `Platform` for operations that do not include data acquisition.

Also, `Completed` object now always requires a `path:Path` input an is no more optional, as I think it should be.

> [!NOTE]
> Discuss about eventual tests to add.

### DIFF
--- a/src/qibocal/auto/execute.py
+++ b/src/qibocal/auto/execute.py
@@ -31,7 +31,7 @@ PLATFORM_DIR = "platform"
 """Folder where platform will be dumped."""
 
 
-def check_overlap_in_input_qubits(targets: np.typing.ArrayLike):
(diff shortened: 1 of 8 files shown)


scarrazza (approved):

scarrazza:
Sorry @lballerio, the app interface got stuck and this was merged. If you prefer we can undo the merge or just open another PR with @RoyStegeman feedback.

Issue #1535, opened 2026-06-11
1q complementary experiments


alecandido:
We also have experiments which are occasionally useful, but not strictly required for the single qubit calibration.

Most of them are in a worse state than those which are more regularly used, and the may benefit from a review.

Issue #1582, opened 2026-07-02
Include parameter deltas alongside the fitted parameters in reports


RoyStegeman:
Currently, reports display the relevant parameters extracted from a protocol. In many cases, however, it is also useful to show how these parameters have changed relative to their previous values. 

The mixer calibration will be an exception since the parameters are not very meaningful, but no other exceptions immediately come to mind. 

In cases where many parameters are shown, the deltas could be included in a collapsible table.

PR #1624, opened 2026-07-21
FIT mode now creates platform with no hardware memory.

### LINKED ISSUE #1413: `qq fit` uses the `QIBOLAB_PLATFORM`, while it should load the platform from the results folder

lballerio:
We noticed that when executing only FIT operation (e.g. though `qq fit` command) Qibocal was building the platform using the state of `platform_folder` (e.g. the ones in `qibolab_platforms_qrc`) and not the state saved in the `experiment_data_folder`.
We fix this issue in the PR, also avoiding to construct the `Hardware` sensitive fields of the `Platform` for operations that do not include data acquisition.

Also, `Completed` object now always requires a `path:Path` input an is no more optional, as I think it should be.

> [!NOTE]
> Discuss about eventual tests to add.

linked to [Qibolab #1535](https://github.com/qiboteam/qibolab/pull/1535).

### DIFF
--- a/src/qibocal/auto/execute.py
+++ b/src/qibocal/auto/execute.py
@@ -31,7 +31,7 @@ PLATFORM_DIR = "platform"
 """Folder where platform will be dumped."""
 
 
-def check_overlap_in_input_qubits(targets: np.typing.ArrayLike):
+def check_overlap_in_input_qubits(targets: list):
     """Check that target qubits do not contain duplicates."""
 
     targ = np.asarray(targets)
@@ -77,14 +77,37 @@ class Executor(BaseModel):
         protocol: Protocol,
         parameters: Action,
         mode: ExecutionMode = AUTOCALIBRATION,
-        output: Path | None = None,
     ) -> Completed:
-        """Run single protocol in ExecutionMode mode."""
+        """Run a calibration protocol and record the completed task.
+
+        The executor preserves the execution history and chooses the platform
+        instance used for the task based on the requested mode. If acquisition is
+        requested, the current live platform is used. If only fitting or analysis
+        is required, the platform is reconstructed from the output folder so that
+        the exact experiment configuration is reused.
+        If the mode contains :class:`ExecutionMode.FIT`, and the action is
+        configured to update the platform, it is updated using the fitted parameters.
+        """
+
+        output = self.path
 
         task = Task(action=parameters, operation=protocol)
         log.info(f"Executing mode {mode} on {task.action.id}.")
         completed = task.run(
-            platform=self.platform,
+            platform=(
+                # if I need to acquire data I need to create from scratch
+                # the platform with all its hardware configurations;
+                # when executor is just fitting I need to create the exact same
+                # platform of the experiment (saved in the experiment folder), and here
+                # the hardware configuration is unnecessary.
+                self.platform
+                if ExecutionMode.ACQUIRE in mode
+                else CalibrationPlatform.from_datafolder(
+                    folder_path=output,
+                    platform_name=self.platform.name,
+                    dummy_hardware=False,
+                )
+            ),
             targets=self.targets,
             mode=mode,
             folder=self.history.task_path(
@@ -168,9 +191,7 @@ class Executor(BaseModel):
                     "parameters": params | positional | kwargs,
                 }
             )
-            return self.run_protocol(
-                protocol, parameters=action, mode=mode, output=self.path
-            )
+            return self.run_protocol(protocol, parameters=action, mode=mode)
 
         return wrapper
 
--- a/src/qibocal/auto/execute.py
+++ b/src/qibocal/auto/execute.py
@@ -31,7 +31,7 @@ PLATFORM_DIR = "platform"
 """Folder where platform will be dumped."""
 
 
-def check_overlap_in_input_qubits(targets: list):
+def check_overlap_in_input_qubits(targets: np.typing.ArrayLike):
     """Check that target qubits do not contain duplicates."""
 
     targ = np.asarray(targets)
@@ -105,7 +105,7 @@ class Executor(BaseModel):
                 else CalibrationPlatform.from_datafolder(
                     folder_path=output,
                     platform_name=self.platform.name,
-                    dummy_hardware=False,
+                    dummy_hardware=True,
                 )
             ),
             targets=self.targets,
--- a/src/qibocal/auto/execute.py
+++ b/src/qibocal/auto/execute.py
@@ -31,7 +31,7 @@ PLATFORM_DIR = "platform"
 """Folder where platform will be dumped."""
 
 
-def check_overlap_in_input_qubits(targets: np.typing.ArrayLike):
+def check_overlap_in_input_qubits(targets: list):
     """Check that target qubits do not contain duplicates."""
 
     targ = np.asarray(targets)
@@ -77,14 +77,37 @@ class Executor(BaseModel):
         protocol: Protocol,
         parameters: Action,
         mode: ExecutionMode = AUTOCALIBRATION,
-        output: Path | None = None,
     ) -> Completed:
-        """Run single protocol in ExecutionMode mode."""
+        """Run a calibration protocol and record the completed task.
+
+        The executor preserves the execution history and chooses the platform
+        instance used for the task based on the requested mode. If acquisition is
+        requested, the current live platform is used. If only fitting or analysis
+        is required, the platform is reconstructed from the output folder so that
+        the exact experiment configuration is reused.
+        If the mode contains :class:`ExecutionMode.FIT`, and the action is
+        configured to update the platform, it is updated using the fitted parameters.
+        """
+
+        output = self.path
 
         task = Task(action=parameters, operation=protocol)
         log.info(f"Executing mode {mode} on {task.action.id}.")
         completed = task.run(
-            platform=self.platform,
+            platform=(
+                # if I need to acquire data I need to create from scratch
+                # the platform with all its hardware configurations;
+                # when executor is just fitting I need to create the exact same
+                # platform of the experiment (saved in the experiment folder), and here
+                # the hardware configuration is unnecessary.
+                self.platform
+                if ExecutionMode.ACQUIRE in mode
+                else CalibrationPlatform.from_datafolder(
+                    folder_path=output,
+                    platform_name=self.platform.name,
+                    dummy_hardware=False,
+                )
+            ),
             targets=self.targets,
             mode=mode,
             folder=self.history.task_path(
@@ -167,9 +190,7 @@ class Executor(BaseModel):
                     "parameters": params | positional | kwargs,
                 }
             )
-            return self.run_protocol(
-                protocol, parameters=action, mode=mode, output=self.path
-            )
+            return self.run_protocol(protocol, parameters=action, mode=mode)
 
         return wrapper
 
--- a/src/qibocal/auto/execute.py
+++ b/src/qibocal/auto/execute.py
@@ -103,7 +103,7 @@ class Executor(BaseModel):
                 self.platform
                 if ExecutionMode.ACQUIRE in mode
                 else CalibrationPlatform.from_datafolder(
-                    folder_path=output,
+                    folder_path=output / PLATFORM_DIR,
                     platform_name=self.platform.name,
                     dummy_hardware=True,
                 )
--- a/src/qibocal/auto/output.py
+++ b/src/qibocal/auto/output.py
@@ -231,12 +231,37 @@ class Output:
         update: bool = True,
         force: bool = False,
     ):
-        """Process existing output."""
-        backend = construct_backend(
-            backend=self.meta.backend, platform=self.meta.platform
-        )
-        assert backend.platform is not None
-        self.platform = CalibrationPlatform.from_platform(backend.platform)
+        """Process an existing output directory.
+
+        Reconstruct the calibration platform from the live backend during
+        acquisition or from the saved datafolder during fitting, then rerun each
+        completed task using the requested execution mode and output folder.
+        If ``update`` is enabled and the task supports platform updates, the
+        platform is refreshed accordingly. When ``force`` is ``False``, tasks
+        that already contain fitting results raise an error to avoid
+        overwriting them.
+        """
+        # NOTE: this function in principle takes also ``ExecutionMode.ACQUIRE`` as ``mode``
+        # value, but if we want this function to be used only for offline post-processing
+        # this input is completely unnecessary (we cannot acquire offline), so we might
+        # think of removing it.
+
+        # during acquisition we need the information of the hardware
+        if ExecutionMode.ACQUIRE in mode:
+            backend = construct_backend(
+                backend=self.meta.backend, platform=self.meta.platform
+            )
+            assert backend.platform is not None
+            self.platform = CalibrationPlatform.from_platform(backend.platform)
+        else:
+            # while performing a fitting task we do not need hardware information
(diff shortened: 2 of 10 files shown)


src/qibocal/calibration/platform.py
```diff
@@ -1,7 +1,9 @@
 from dataclasses import dataclass
 from pathlib import Path
 
-from qibolab import Platform, create_platform, locate_platform
+from qibolab import Parameters, Platform, create_platform, locate_platform
+from qibolab._core.dummy.platform import create_dummy
+from qibolab._core.platform.platform import PARAMETERS
```
RoyStegeman: This is not part of the public API, meaning qibolab can change this without warning, so we should not use it in qibocal

src/qibocal/calibration/platform.py
```diff
@@ -64,11 +66,47 @@ def from_platform(cls, platform: Platform):
...
+    def from_datafolder(
+        cls, folder_path: Path, platform_name: str, dummy_hardware: bool
+    ):
+        """Create a calibration platform from a serialized data folder.
+
+        The platform is rebuilt from the configuration saved in the experiment history,
+        using the ``parameters.json`` and ``calibration.json`` files stored in the data folder
+        rather than the platform definition. If a ``platform_name`` is provided, the hardware
+        configuration is loaded from that platform; otherwise, a dummy hardware
+        configuration is used so that acquisition-related fields are still present
```
RoyStegeman: > If a ``platform_name`` is provided, the hardware
> configuration is loaded from that platform; otherwise, a dummy hardware
> is used 

According to this the boolean `dummy_hardware` is not needed, but depends on whether a `platform_name` is provided. Although `platform_name` can't be `None`, so the docstring is not very clear.

src/qibocal/calibration/platform.py
```diff
@@ -64,11 +66,47 @@ def from_platform(cls, platform: Platform):
...
 
+    @classmethod
+    def from_datafolder(
+        cls, folder_path: Path, platform_name: str, dummy_hardware: bool
+    ):
+        """Create a calibration platform from a serialized data folder.
+
+        The platform is rebuilt from the configuration saved in the experiment history,
+        using the ``parameters.json`` and ``calibration.json`` files stored in the data folder
+        rather than the platform definition. If a ``platform_name`` is provided, the hardware
```
RoyStegeman: I think by "the platform definition" you mean the `QIBOLAB_PLATFORMS` environment variable. If so, please write that. If you meant something else, please clarify.

src/qibocal/calibration/platform.py
```diff
@@ -64,11 +66,47 @@ def from_platform(cls, platform: Platform):
...
+
+        calibration = Calibration.model_validate_json(
+            (folder_path / CALIBRATION).read_text()
+        )
+
+        platform = (
+            create_dummy() if dummy_hardware is None else create_platform(platform_name)
+        )
+        platform.parameters = parameters
+        platform.name = platform_name
```
RoyStegeman, line 95: Why do we need to overwrite the name?

src/qibocal/auto/execute.py
```diff
@@ -31,7 +31,7 @@
 """Folder where platform will be dumped."""
 
 
-def check_overlap_in_input_qubits(targets: np.typing.ArrayLike):
+def check_overlap_in_input_qubits(targets: list):
```
RoyStegeman: The first thing this function does is call `targ = np.asarray(targets)` so the previous type should have been correct.

src/qibocal/calibration/platform.py
```diff
@@ -64,11 +66,47 @@ def from_platform(cls, platform: Platform):
...
+        parameters = Parameters.model_validate_json(
+            (folder_path / PARAMETERS).read_text()
+        )
+
+        calibration = Calibration.model_validate_json(
+            (folder_path / CALIBRATION).read_text()
+        )
+
+        platform = (
+            create_dummy() if dummy_hardware is None else create_platform(platform_name)
```
RoyStegeman:
```suggestion
            create_dummy() if dummy_hardware is True else create_platform(platform_name)
```
Or maybe `if platform_name is None`; see [comment](https://github.com/qiboteam/qibocal/pull/1624/changes#r3637842606).

src/qibocal/auto/execute.py
```diff
@@ -77,14 +77,37 @@ def run_protocol(
...
+                # the platform with all its hardware configurations;
+                # when executor is just fitting I need to create the exact same
+                # platform of the experiment (saved in the experiment folder), and here
+                # the hardware configuration is unnecessary.
+                self.platform
+                if ExecutionMode.ACQUIRE in mode
+                else CalibrationPlatform.from_datafolder(
+                    folder_path=output,
+                    platform_name=self.platform.name,
+                    dummy_hardware=False,
```
RoyStegeman:
```suggestion
                    dummy_hardware=True,
```

I think? Since in this case you're just fitting.

src/qibocal/auto/task.py
```diff
@@ -231,18 +232,15 @@ def results(self, value):
...
         """Dumping data."""
-        if self.path is not None:
-            self._data.save(self.path)
+        self._data.save(self.path)
 
     def dump_results(self):
         """Dumping results."""
-        if self.path is not None:
-            self._results.save(self.path)
+        self._results.save(self.path)
```
RoyStegeman: `self._data` and `self._results` can still be None, in which case calling `None.save` will raise an error.

RoyStegeman (commented):
Some comments. The main one being a bug the line below: while running `qq fit`, `dummy_hardware` will be `True` instead of None.  
```python
platform = create_dummy() if dummy_hardware is None else create_platform(platform_name)
```

src/qibocal/calibration/platform.py
lballerio: Ok I'll find a solution.

src/qibocal/calibration/platform.py
lballerio: That action is only effective when platform is a `DummyPlatform`, I've decided to overwrite to propagate the name even for dummy platforms, maybe you need later in the code and I didn't want to create confusion by printing `dummy` (or whatever name  `DummyPlatform` has).

src/qibocal/auto/execute.py
lballerio: it's true, but also we call this function after casting `self.targets` as `Target` types. which indeed is always a list.
I assumed a more strict typehint since still `list` is an `ArrayLike`.

src/qibocal/calibration/platform.py
lballerio: yes, thanks!

src/qibocal/auto/task.py
lballerio: yes you are right, thanks

src/qibocal/auto/execute.py
lballerio: yes thanks for pointing out

src/qibocal/auto/task.py
lballerio: addressed in e47d647

src/qibocal/auto/execute.py
lballerio: addressed in e47d647

src/qibocal/calibration/platform.py
lballerio: addressed in e47d647

src/qibocal/auto/execute.py
lballerio: addressed in e47d647

src/qibocal/calibration/platform.py
lballerio: addressed in e47d647

src/qibocal/calibration/platform.py
lballerio: addressed in e47d647

src/qibocal/calibration/platform.py
lballerio: in order to address this point, I opened [Qibolab #1535](https://github.com/qiboteam/qibolab/pull/1535) and then corrected the import in e47d647.

alecandido:
The proposal is fine by me. Possibly, we just want to eventually exclude platform access from the `Protocol.fit` invocation, and have `qq fit` to just work without creating any platform, not even a dummy one. But, for the time being, this is a much faster workaround. And seemingly solid.

> Also, `Completed` object now always requires a `path:Path` input an is no more optional, as I think it should be.

This is a more complicated situation. The original design was meant to avoid passing paths everywhere, since disk I/O can be quite annoying in a HPC-like infrastructure. Because of the non-locality of the storage, and also the recommended folders access.

However, we may even forget this part now, and decide to always work with paths. This happened accidentally, because the original design was violated (following the frequent quick & dirty approach), but we could even decide to systematically reconsider the initial assumptions.
In principle, if we need performance (no disk I/O) and using a temporary scratch-like folder, we could also deal with at high-level, with an in-memory filesystem, and copying the results folder out of scratch at the desired time - e.g. at the end of the workload, or with intermediate "backups".

Whatever it is done here it is fine and acceptable. But let's keep this in mind, and redesign together.

src/qibocal/cli/update.py
```diff
@@ -25,15 +25,21 @@ def update(path: pathlib.Path, skip_qubits: list[QubitId] | None):
     platform_name = json.loads((path / META).read_text())["platform"]
 
     platform_path = locate_platform(platform_name)
```
RoyStegeman, line 27:
```
(qpu165) roy.stegeman@dalma:~/calibration$ qq update /home/users/roy.stegeman/calibration/results/tuna5/260804/17-19-51-cryoscope-0
  File "/nfs/users/roy.stegeman/.venvs/qpu165/bin/qq", line 8, in <module>
    sys.exit(command())
  File "/nfs/users/roy.stegeman/.venvs/qpu165/lib/python3.10/site-packages/click/core.py", line 1514, in __call__
  File "/nfs/users/roy.stegeman/.venvs/qpu165/lib/python3.10/site-packages/click/core.py", line 1435, in main
    rv = self.invoke(ctx)
  File "/nfs/users/roy.stegeman/.venvs/qpu165/lib/python3.10/site-packages/click/core.py", line 1902, in invoke
  File "/nfs/users/roy.stegeman/.venvs/qpu165/lib/python3.10/site-packages/click/core.py", line 1298, in invoke
  File "/nfs/users/roy.stegeman/.venvs/qpu165/lib/python3.10/site-packages/click/core.py", line 853, in invoke
    return callback(*args, **kwargs)
  File "/nfs/users/roy.stegeman/github/qibocal/src/qibocal/cli/_base.py", line 145, in update
    updating(folder, skip_qubits)
  File "/nfs/users/roy.stegeman/github/qibocal/src/qibocal/cli/update.py", line 27, in update
    platform_path = locate_platform(platform_name)
  File "/nfs/users/roy.stegeman/github/qibolab/src/qibolab/_core/platform/load.py", line 81, in locate_platform
    return _search(name, paths)
  File "/nfs/users/roy.stegeman/github/qibolab/src/qibolab/_core/platform/load.py", line 56, in _search
    raise ValueError(
ValueError: Platform tuna5 not found. Check $QIBOLAB_PLATFORMS environment variable.
```

Sine no path is passed to `locate_platform` it will `_search` in the default `_platforms_paths()`, which is still 
`PLATFORMS_PATH = "QIBOLAB_PLATFORMS"`

RoyStegeman (changes requested):

src/qibocal/cli/update.py
lballerio: thanks for this comment @RoyStegeman.
Probably you already had a look at the the function where the problem comes in, but I'll paste here for reference:

```python
def update(path: pathlib.Path, skip_qubits: list[QubitId] | None):
    """Perform copy of updated platform in QIBOLAB_PLATFORM

    Arguments:
        - input_path: Qibocal output folder.
    """
    new_platform_path = path / UPDATED_PLATFORM

    if not new_platform_path.exists():
        raise_error(FileNotFoundError, f"No updated runcard platform found in {path}.")

    platform_name = json.loads((path / META).read_text())["platform"]

    platform_path = locate_platform(platform_name)
    old_platform = create_calibration_platform(platform_name)
    for filename in os.listdir(new_platform_path):
        shutil.copy(
            new_platform_path / filename,
            platform_path / filename,
        )

    if skip_qubits is not None:
        new_platform = create_calibration_platform(platform_name)
        updated_platform = merge_with_skipped_qubits(
            old_platform, new_platform, skip_qubits
        )
        updated_platform.dump(platform_path)

    log.info(f"Platform {platform_name} configuration has been updated.")
```

this is where `locate_platform()` is called; however I think it is intended to be called for locating not the datafolder but instead the folder to write the updates on, so it is telling us: "to update a platform it has to be stored in `PLATFORM_PATHS`"; this works in the assumption that the datafolder has been generated using the platform in `PLATFORM_PATHS`, then one can say: "ok but maybe I want to update the 'main' platform in `PLATFORM_PATHS` even if this does not exist", which means write from scratch from the datafolder, but this is another problem I think.

src/qibocal/calibration/platform.py
```diff
@@ -64,11 +69,44 @@ def from_platform(cls, platform: Platform):
...
+
     def dump(self, path: Path):
         super().dump(path)
         self.calibration.dump(path)
 
 
 def create_calibration_platform(name: str) -> CalibrationPlatform:
+    """This function builds a ``CalibrationPlatform`` object which is sentitive of the hardware,
+    so it needs information about the clusters and its connection. Has to be used for acquisition.
+    """
```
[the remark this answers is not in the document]
lballerio: applied in 19e5b5b

src/qibocal/auto/output.py
```diff
@@ -231,12 +231,37 @@ def process(
...
+        if ExecutionMode.ACQUIRE in mode:
+            backend = construct_backend(
+                backend=self.meta.backend, platform=self.meta.platform
+            )
+            assert backend.platform is not None
+            self.platform = CalibrationPlatform.from_platform(backend.platform)
+        else:
+            # while performing a fitting task we do not need hardware information
+            # but also we need the params saved in the datafolder, since in the
+            # platform folder might been changed.
```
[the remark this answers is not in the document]
lballerio: addressed in 19e5b5b

src/qibocal/auto/execute.py
```diff
@@ -77,14 +77,37 @@ def run_protocol(
...
+            platform=(
+                # if I need to acquire data I need to create from scratch
+                # the platform with all its hardware configurations;
+                # when executor is just fitting I need to create the exact same
+                # platform of the experiment (saved in the experiment folder), and here
+                # the hardware configuration is unnecessary.
+                self.platform
+                if ExecutionMode.ACQUIRE in mode
+                else CalibrationPlatform.from_datafolder(
+                    folder_path=output,
```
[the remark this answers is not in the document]
lballerio: addressed in 19e5b5b

lballerio:
Now we also need [Qibolab #1582](https://github.com/qiboteam/qibolab/pull/1582) to be approved and merged before this

alecandido:
> Now we also need [Qibolab #1582](https://github.com/qiboteam/qibolab/pull/1582) to be approved and merged before this

Linking PRs between Qibocal and Qibolab has been the standard approach for too long. For as long as we can, let's stop doing it. If you really need and it is unavoidable - e.g. because you are testing some genuinely new capability that we planned for Qibolab, and they are best tested with a Qibocal experiment - keep the PR in Qibocal draft for as long as Qibolab is not exposing that. At times, this is definitely acceptable. But let's try to avoid pressuring specific PRs by linking them to others.

PR #1633, opened 2026-07-28
Ruff v0.16 default rule changes and consolidate linting tools

RoyStegeman:
See also https://github.com/qiboteam/qibolab/pull/1534

---

Ruff v0.16 was merged for pre-commit https://github.com/qiboteam/qibocal/pull/1632

However, it results in failed tests since it significantly extends the set of default rules: https://astral.sh/blog/ruff-v0.16.0

In this PR we address most of them, though the warning `BLE001` about not blindly doing `except Exception` is still there since replacing it with `except RuntimeError` makes many of the protocols tests fail. The fact that, that make the tests fail is also a good indication that those tests are quite bad. I opened an issue #1634.

NOTE: much of this was done manually, since ruff does not automate everything

---

This PR also consolidates the functionality of pyupgrade and pycln in ruff. 
- pycln removes unused imports, which is also done by ruff rule F401. It seems pycln is slightly more extensive in some casses https://github.com/astral-sh/ruff/issues/8149, but also much slower and the `__init__.py` files which do the import with `*` were skipped anyway. 
- pyupgrade is prominently mentioned to be replaceable by ruff in the ruff documention. Specifically the `UP` set of rules are those from pyupgrade.


scarrazza (approved):

RoyStegeman:
Okay I scrolled through it again and think the changes are all acceptable.

src/qibocal/protocols/two_qubit_interaction/chsh/utils.py
```diff
@@ -7,14 +7,13 @@ def compute_chsh(frequencies, basis, i):
     """Computes the chsh inequality out of the frequencies of the 4 circuits executed."""
     chsh = 0
     aux = 0
```
lballerio: I think this line can be deleted.

lballerio (changes requested):

src/qibocal/protocols/two_qubit_interaction/chsh/utils.py
RoyStegeman: You're right! And I even did this one manually...

lballerio:
Other than that small fix for me is fine, the changes are self contained and are mostly changing the syntax, they do not seem breaking changes.

tests/test_randomized_benchmarking.py
```diff
@@ -83,7 +83,7 @@ def test_exp2_fitting():
         assert np.allclose(fitting.exp2_func(x, A1, A2, f1, f2), y)
         # Distort ``y`` a bit.
         y_dist = y + np.random.uniform(-1, 1, size=len(y)) * 0.001
-        popt, perr = fitting.fit_exp2_func(x, y_dist)
+        popt, _perr = fitting.fit_exp2_func(x, y_dist)
```
lballerio, line 86: Super minor consideration:
I see here `_perr` is never used, so maybe we can change into `_` so we don't have the mismatch between `_perr` and `perr` present in other functions

lballerio (approved):

tests/test_randomized_benchmarking.py
RoyStegeman: The prepended underscore already indicates that the variable is unused. By keeping `_perr` instead of `_` we also provide some information about what is being returned. In this case I would have been fine with either, but I don't think it needs to be changed.

Issue #1293, opened 2025-11-21
Plot 1D experimental data as scatter points


alecandido:
No more than a review of existing plots. But faster performed after the deduplication in #1290 - hopefully it will be sufficient to just follow this (common) convention during the refactoring process.

jevillegasd:
This was taking a while so I put it on [#1641](https://github.com/qiboteam/qibocal/pull/1641).

PR #1641, opened 2026-07-31
Homogenize 1D plots 

jevillegasd:
All 1D plots to have scatter points for data and soliud lines for fittings, as discussed in #1293.  

I am not sure what is the best format for waveform traces (in time of flight or cryoscope  for example), here I put them as markers.


lballerio:
Hi @jevillegasd, thanks for your contribution!
Yeah we are slowly replacing every plot style with scatter-plots.
However I don't know if every plot should be scatter plots (such as you mentioned cryoscope, but I am mainly thinking about experiments where we do not have any fit to plot, such as coherence-vs-flux).
That's why I will leave the decision to @alecandido, then he will decide whether to go on with this PR.

lballerio:
Anyway if you manage I would rebase onto main because some protocols might already been fixed.

jevillegasd:
> Anyway if you manage I would rebase onto main because some protocols might already been fixed.

I think its pretty much up to date with main (only the commit on the iSWAP update is ahead)

alecandido:
In general, I'm not a huge fan of large rewritings, because it is hard to realize the impact of all changes.
If you add that the aesthetics of plots is quite a marginal topic, this is the main reason why we took an incremental approach.

For a PR affecting many protocols, I would approve after confirming that the proposed changes are fitting the intended purpose, and they do not introduce unintended behavior. This could be automated by unit tests - which are especially poor in Qibocal.
For this PR, the best review would consist in producing plots for all the affected protocols, and check we are satisfied with the new appearance.

In practice, we know the surface is minimal, and most plots were "wrong", according to a widely agreed concept.
Since the PR is here, I would be lenient in the review, and accept the new state, assuming on average to be no worse than the previous one. We will anyhow review them one by one later on, also to standardize the rest of the layout.

The only review I would request here is to make sure that plots that have already been reviewed are not affected in a negative way, and then accept the rest with no further comment. If I were completely sure about the first condition, I would have already accepted, since the changes are affecting almost exclusively the `mode` argument of `go.Scatter()` calls (or exactly that, if I did not miss anything).
But, iirc, @RoyStegeman reviewed recently at least the DRAG and allXY (though maybe not the one with resonator depletion). So, I would wait for his approval

For future PRs, strongly favor changes which are scoped by protocol, rather than horizontal. This makes the review process much more effective, and avoids changing things back and forth

RoyStegeman:
> For a PR affecting many protocols, I would approve after confirming that the proposed changes are fitting the intended purpose, and they do not introduce unintended behavior. This could be automated by unit tests - which are especially poor in Qibocal.

This is why I had not approved yet. Looking at it now the protocols that are change here, I did not work on the layout, so I have no reason to believe this PR could be a deterioration of any of them. 

>  @RoyStegeman reviewed recently at least the DRAG and allXY (though maybe not the one with resonator depletion). So, I would wait for his approval

The allXY that is changed in this PR is specifically for the resonator depletion, which I did not touch. The only change in Drag is to explicitly set `mode="lines", which I'm happy with since the default is a bit strange:

> [If there are less than 20 points and the trace is not stacked then the default is “lines+markers”. Otherwise, “lines”](https://plotly.com/python-api-reference/generated/plotly.graph_objects.Scatter.html)

RoyStegeman (approved):
Thanks. This is fine for me

scarrazza (approved):

PR #1649, opened 2026-08-05
fix: move colormap scale

ElStabilini:
A very small fix to avoid the following problem:

image

### DIFF
--- a/src/qibocal/protocols/flux_dependence/utils.py
+++ b/src/qibocal/protocols/flux_dependence/utils.py
@@ -66,6 +66,7 @@ def flux_dependence_plot(data, fit, qubit, fit_function=None):
             z=qubit_data.signal,
             colorbar={"title": "Signal [a.u.]"},
             colorscale="Viridis",
+            colorbar_x=1.01,
         ),
     )
 
--- a/src/qibocal/protocols/qubit_spectroscopies/qubit_power_spectroscopy.py
+++ b/src/qibocal/protocols/qubit_spectroscopies/qubit_power_spectroscopy.py
@@ -142,6 +142,7 @@ def _plot(
             z=data.normalized_signal(target).ravel(),
             colorbar={"title": "Normalized signal"},
             colorscale="Viridis",
+            colorbar_x=1.01,
         )
     )
 


alecandido (dismissed):
I imagine it, but could you upload two reports for the before and after?

If you do not have access to the report server, you can just upload the zipped folders in a comment.

In any case, fine by me :)

RoyStegeman:
Hi @ElStabilini, thanks for opening this PR, that is indeed annoying.. 

Looking at the [plotly documentation](https://plotly.com/python/reference/heatmap/#heatmap-colorbar-x) it seems that the default value for `colorbar_x` is conditional on `orientation` and `xref`, but neither of those is explicitly set and thus default to `xref='paper'` and `orientation='v'`. In that case `colorbar_x` should be 1.02. 

Does this change something for you (other than 1.02 to 1.01), and if it does, do you understand if my reasoning is going wrong somewhere?

alecandido:
Actually, reading the docs

> Sets the x position with respect to `xref` of the color bar (in plot fraction). When `xref` is "paper", defaults to 1.02 when `orientation` is "v" and 0.5 when `orientation` is "h". When `xref` is "container", defaults to "1" when `orientation` is "v" and 0.5 when `orientation` is "h". Must be between "0" and "1" if `xref` is "container" and between "-2" and "3" if `xref` is "paper".

It seems to be as if the orientation was "h", since the bar is in the middle. I wonder whether that can change dynamically according to some specs in the browser.
In any case, "h" just for the purpose of `colorbar_x`, since the actual orientation in the picture is vertical.

I wonder whether this could even be related to the Plotly version. But that one should always be the latest that the CDN can retrieve
https://github.com/qiboteam/qibocal/blob/7e59f55ae8b920bfa6be4f31e6bdf41f4591885e/src/qibocal/web/templates/template.html#L24

ElStabilini:
@RoyStegeman @alecandido thank you for the comments, I am looking into this.
Honestly I hadn't test it, we had only run qubit and resonator experiments and I assumed that in other experiments (eg. chevron) the plot was being displayed correctly, for this reason I had simply copy - pasted from there.

At the moment I also have limited access to the hardware, as soon as I will have a more clear answer and/or reports I will past it here.

RoyStegeman:
Thanks for pointing that out, I did not realize it was also done in other protocols. If my interpretation was correct that it is already the default, then it can also be removed in the other places but it would be good to understand before we do.

ElStabilini:
While looking more into this I have noticed that we are working with qibocal v0.2.4 (apparently because of qua, I didn't investigate more) while the latest is version 0.2.6. 

In qibocal v0.2.4 the colorbar coordinate for most of the plots was set as `colorbar_x = 0.46`  causing the overlap with the plot, changing to `colorbar_x=1.01` fixed the problem.

If you do not have any issue on your reports I think that the problem had already been solved, I apologise for the confusion :) .

RoyStegeman:
Thanks for looking into it! You're right, the colorbar coordinate has indeed been changed since v0.2.4 of qibocal. Before we close the PR though, do you believe we can safely remove all occurrences of `colorbar_x=1.01` from qibocal and rely on the default instead? 

I'm not aware of any regression in qibocal that affects support for qua between v0.2.4 and v0.2.6, and I believe there should not be one. If you or your coworkers observed an issue, then of course we welcome an issue either here or in qibolab.

ElStabilini:
> Thanks for looking into it! You're right, the colorbar coordinate has indeed been changed since v0.2.4 of qibocal. Before we close the PR though, do you believe we can safely remove all occurrences of `colorbar_x=1.01` from qibocal and rely on the default instead?

I would say so but I couldn't directly test it.

RoyStegeman:
Actually, all `colorbar_x=1.01` that are left are paired with a `colorbar_x=0.46`. Presumably because two figures are placed side by side, so being explicit may clarify something even if it's not necessary.

In any case, given that removing it would be very minor, I think we can indeed close the discussion. Thanks again for reporting it

ElStabilini:
Regarding the compatibility question instead, my understanding of the problem is that:

qibocal==0.2.5 requires qibo>=0.3.1 which requires numpy>=2.0.0
qibolab-qm==0.0.1 requires requires qm-qua==1.2.1 which requires numpy>=1.26.0,<2

alecandido:
> qibocal==0.2.5 requires qibo>=0.3.1 which requires numpy>=2.0.0
> qibolab-qm==0.0.1 requires requires qm-qua==1.2.1 which requires numpy>=1.26.0,<2

Yes, that's known. We will update soon `qm-qua`, which has also updated the dependency on NumPy, and that should solve the problem.

For the time being, just install `qibolab-qm` on top with `pip`, which will overwrite the NumPy version. All in all, it should mostly not matter which version of NumPy you are actually using at runtime.

PR #1656, opened 2026-08-09
Introduce range syntax for flux arcs

alecandido:
On top of what the title says, this is also:

- migrating the qubit flux dependence (flux arc spectroscopy) to use rectangular pulses
- fixing a small mess introduced by the range syntax initial support for the qubit spectroscopy

Most likely, the second point should be described better in the documentation of the qubit spectroscopy. But that is possibly better done with a thorough refactor of the qubit spectroscopy implementation, modularizing it a bit more, and making the batching and consequent LO shifts reusable for other protocols. Primarily, for the resonator spectroscopy.

However, the algorithm is "simply" the following:
1. batches, consisting of begin-end and LO positions, are computed, just taking into account the full width of the requested interval - which can be computed independently of its center, which is allowed to be qubit dependent
2. the center is computed for each qubit, just by passing the channel frequency as "candidate center" to the usual range function - boiling down to a call to `to_range()`; this is determining the entire range, which could be specified independently of the default centering (drive channel frequency), and out of it a center is computed - which is delegated to `params.f0()`
3. on top of this center the actual range for the batch is computed, summing the extrema of the batch, computed at point 1. relative to a `0.0` center

### DIFF
--- a/src/qibocal/protocols/flux_dependence/qubit_crosstalk.py
+++ b/src/qibocal/protocols/flux_dependence/qubit_crosstalk.py
@@ -155,7 +155,7 @@ def _acquisition(
         freq_sweepers.append(
             Sweeper(
                 parameter=Parameter.frequency,
-                values=platform.config(qd_channel).frequency + params.frequency_range,
+                range=params.frequency_range(platform.config(qd_channel).frequency),
                 channels=[qd_channel],
             )
         )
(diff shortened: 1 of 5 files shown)


scarrazza (approved):

RoyStegeman (approved):
Looks good, don't see any issues

PR #1662, opened 2026-08-13
Correct numpification of Quinn Fernandes algorithm

lballerio:
As mentioned in the title, I realized that the `quinn_fernandes` implementation in `protocols/utils.py` does not work correctly when the `axis` parameter is different from `-1`, as the axis was not being handled properly.

Additionally, at the end of the function, we were computing the `np.median` of the estimated omegas in the `ndarray`. This was originally introduced for the `cross_resonance` protocols, particularly for simultaneous fitting in Hamiltonian tomography. However, this behavior is not generally desirable, so I moved this operation to the specific module where it is needed.

### DIFF
--- a/src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_processing.py
+++ b/src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_processing.py
@@ -409,13 +409,11 @@ def tomography_cr_fit(
                 ]
             ).reshape(len(Basis), -1)
 
-            total_omega_guess = float(
-                np.median(
+            total_omega_guess = np.median(
                     quinn_fernandes_algorithm(
                         concatenated_signal, vector_x, speedup_flag=True
                     )
-                )
-            )
+                ).tolist()
 
             if fit_with_evolution:
                 popt = dynamic_evolution_optimizer(
--- a/src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_processing.py
+++ b/src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_processing.py
@@ -410,10 +410,10 @@ def tomography_cr_fit(
             ).reshape(len(Basis), -1)
 
             total_omega_guess = np.median(
-                    quinn_fernandes_algorithm(
-                        concatenated_signal, vector_x, speedup_flag=True
-                    )
-                ).tolist()
+                quinn_fernandes_algorithm(
+                    concatenated_signal, vector_x, speedup_flag=True
+                )
+            ).tolist()
 
             if fit_with_evolution:
                 popt = dynamic_evolution_optimizer(
--- a/src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_processing.py
+++ b/src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_processing.py
@@ -409,9 +409,13 @@ def tomography_cr_fit(
                 ]
             ).reshape(len(Basis), -1)
 
-            total_omega_guess = float(np.median(quinn_fernandes_algorithm(
-                concatenated_signal, vector_x, speedup_flag=True
-            )))
+            total_omega_guess = float(
+                np.median(
+                    quinn_fernandes_algorithm(
+                        concatenated_signal, vector_x, speedup_flag=True
+                    )
+                )
+            )
 
             if fit_with_evolution:
                 popt = dynamic_evolution_optimizer(
--- a/src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_processing.py
+++ b/src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_processing.py
@@ -409,9 +409,9 @@ def tomography_cr_fit(
                 ]
             ).reshape(len(Basis), -1)
 
-            total_omega_guess = quinn_fernandes_algorithm(
+            total_omega_guess = float(np.median(quinn_fernandes_algorithm(
                 concatenated_signal, vector_x, speedup_flag=True
-            )
+            )))
 
             if fit_with_evolution:
                 popt = dynamic_evolution_optimizer(
--- a/src/qibocal/protocols/utils.py
+++ b/src/qibocal/protocols/utils.py
@@ -8,7 +8,7 @@ import numpy as np
 import pandas as pd
 import plotly.express as px
 import plotly.graph_objects as go
-from numpy.typing import NDArray
+from numpy.typing import NDArray, ArrayLike
 from plotly.subplots import make_subplots
 from pydantic import TypeAdapter
 from scipy import constants, sparse
@@ -1065,9 +1065,7 @@ def guess_frequency(x: NDArray, y: NDArray, axis: int = -1) -> NDArray:
     mags = np.abs(fft)
     mags[..., 0] = 0
 
-    selected_freqs = fft_freqs[np.argmax(mags, axis=-1)]
-
-    return selected_freqs
+    return fft_freqs[np.argmax(mags, axis=-1)]
 
 
 def fallback_frequency(frequency: NDArray) -> NDArray:
@@ -1084,13 +1082,13 @@ def fallback_frequency(frequency: NDArray) -> NDArray:
 
 
 def quinn_fernandes_algorithm(
-    signal: np.typing.ArrayLike,
-    x: np.typing.ArrayLike,
+    signal: ArrayLike,
+    x: ArrayLike,
     axis: int = -1,
     speedup_flag: bool = False,
     iterations: int = 100,
     tol: float = 1e-8,
-) -> NDArray | float:
+) -> NDArray[float]:
     """This is a custom implementation of the Quinn-Fernandes algorithm.
     We compute the signal sampling rate from :param:x, hence this function assumes x to be
     ordered.
@@ -1109,16 +1107,13 @@ def quinn_fernandes_algorithm(
 
     fs = 1 / abs(x[0] - x[1])
 
-    if signal.ndim == 1:
-        signal = np.asarray(signal[np.newaxis, :])
-
     omegas = 2 * np.pi * fallback_frequency(guess_frequency(x, signal, axis=-1))
     alpha = 2 * np.cos(omegas)
 
     null_mean_signal = signal - np.mean(signal, axis=-1, keepdims=True)
 
     sig_shape = list(null_mean_signal.shape)
-    # adding two elements in the last dimensionto be used in the algorithm below
+    # adding two elements in the last dimension to be used in the algorithm below
     sig_shape[-1] += 2
     buffer_beta = []
 
@@ -1133,7 +1128,8 @@ def quinn_fernandes_algorithm(
         beta = np.sum((xi[..., 2:] + xi[..., :-2]) * xi[..., 1:-1], axis=-1) / np.sum(
             xi[..., :-1] ** 2, axis=-1
         )
-        beta[np.isnan(beta)] = 0
+        # np.where() works with scalars as well, so there is not need to add a dimension in 1D case
+        beta = np.where(np.isfinite(beta), beta, 0)
         if len(buffer_beta) >= 5:
             buffer_beta.pop(0)
         buffer_beta.append(beta)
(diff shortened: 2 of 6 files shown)


scarrazza (approved):

src/qibocal/protocols/utils.py
```diff
@@ -1104,32 +1104,34 @@ def quinn_fernandes_algorithm(
...
     alpha = 2 * np.cos(omegas)
 
-    signal = signal - np.mean(signal, axis=axis, keepdims=True)
+    null_mean_signal = signal - np.mean(signal, axis=-1, keepdims=True)
 
-    sig_shape = list(signal.shape)
-    sig_shape[axis] += 2
+    sig_shape = list(null_mean_signal.shape)
+    # adding two elements in the last dimensionto be used in the algorithm below
+    sig_shape[-1] += 2
```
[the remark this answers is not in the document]
alecandido: Ok, marginal - but it is marking itself as low priority. So, fair

src/qibocal/protocols/utils.py
```diff
@@ -1146,14 +1148,16 @@ def quinn_fernandes_algorithm(
 
     alpha = np.clip(alpha, -2, 2)
     omega_est = np.arccos(alpha / 2)
-    med_omega = np.median(omega_est)
 
-    return med_omega * fs
+    if omega_est.size == 1:
+        return float(omega_est) * fs
+
+    return omega_est * fs
```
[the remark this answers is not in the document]
alecandido: Uhm, this is really achieving a different result than the original code.

In a sense, part of the objection is fair: instead of checking for the size (`omega_est.size == 1`) you should check the rank (`omega_est.ndim == 0`).

Then, if supported on the receiving end, I agree it is better to avoid casting to floats, and instead always keep 0-dim arrays as such.
Generating the exceptional case means that you will have to handle it through the whole processing. Instead, keeping it as a perfectly regular limit case allows you to always relying on a stable interface, and you can always manipulate as a NumPy array.
Eventually, we may cast to Python native object. Which is exactly what is doing the `.tolist()` function, which works well even in the limit case
```py
# notice that is not an "instance" in the sense of being any weird
# subclass - it is exactly the native `float` type
>>> type(np.squeeze(np.arange(1) * 1.0).tolist()) is float
True
```

I hope the downstream analysis is also correct. I guess it is for the `fallback_period` function, but I wonder if it is the only place where it is used. Any comment? @copilot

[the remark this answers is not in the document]
alecandido: Ok, the commit was completely unrequested. I removed it.

But it highlighted another another good point, regarding the `signal` array creation.

src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_processing.py
```diff
@@ -409,8 +409,12 @@ def tomography_cr_fit(
...
             ).reshape(len(Basis), -1)
 
-            total_omega_guess = quinn_fernandes_algorithm(
-                concatenated_signal, vector_x, speedup_flag=True
+            total_omega_guess = float(
+                np.median(
+                    quinn_fernandes_algorithm(
+                        concatenated_signal, vector_x, speedup_flag=True
+                    )
+                )
```
alecandido: Why did you ad this `np.median()`? There is not even any `axis` parameter, so this would be a single number.

Not sure what was before this PR, but:
- if it was a single number, why now do we need a median?
- if it was an array, why now are we reducing to a single number?

src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_processing.py
```diff
@@ -409,8 +409,12 @@ def tomography_cr_fit(
...
             )
```
alecandido: Having discover that `.tolist()` does the correct conversion also for scalars, I would encourage to always use it. At least, we do not have to provide any further information, and delegate to it the full cast.
Usually, when we do this conversion to and from NumPy, we do not care terribly about the dtype (e.g. if it is `float64` or `float32`) or the exact shape. We just want to pass from one domain (Python native) to the other (NumPy), and back again.

So, from now on, I will encourage using `np.array()` and `.tolist()` for these conversions. Unless there is a clear motivation to be more explicit, or generally doing something else.

```suggestion
            total_omega_guess = np.median(
                quinn_fernandes_algorithm(
                    concatenated_signal, vector_x, speedup_flag=True
                )
            ).tolist()
```

src/qibocal/protocols/utils.py
```diff
@@ -1063,11 +1063,11 @@ def guess_frequency(x: NDArray, y: NDArray, axis: int = -1) -> NDArray:
     fft = np.fft.rfft(y, axis=-1)
     fft_freqs = np.fft.rfftfreq(y.shape[-1], d=(x[1] - x[0]))
     mags = np.abs(fft)
-    mags[:, 0] = 0
+    mags[..., 0] = 0
 
     selected_freqs = fft_freqs[np.argmax(mags, axis=-1)]
 
-    return np.moveaxis(selected_freqs, -1, axis)
+    return selected_freqs
```
alecandido: No need for giving it a name if all you are doing is just returning.

Names help making things clear and more explicit. But the return object of a function is clear from the function name and docs - no need for extra annotations.

Just use
```py
    return fft_freqs[np.argmax(mags, axis=-1)]
```

src/qibocal/protocols/utils.py
```diff
@@ -1084,13 +1084,13 @@ def fallback_frequency(frequency: NDArray) -> NDArray:
 
 
 def quinn_fernandes_algorithm(
-    signal: Any,
-    x: Any,
+    signal: np.typing.ArrayLike,
+    x: np.typing.ArrayLike,
```
alecandido: A common pattern is to `import numpy.typing as npt`, in parallel to `import numpy as np`.

src/qibocal/protocols/utils.py
```diff
@@ -1084,13 +1084,13 @@ def fallback_frequency(frequency: NDArray) -> NDArray:
...
     axis: int = -1,
     speedup_flag: bool = False,
     iterations: int = 100,
-    tol: int = 1e-8,
-) -> NDArray:
+    tol: float = 1e-8,
+) -> NDArray | float:
```
alecandido: As [discussed with copilot](https://github.com/qiboteam/qibocal/pull/1662/changes#r3778325151), better to keep this just a single type, and not a union. Unless there is a strong motivation for it.

src/qibocal/protocols/utils.py
```diff
@@ -1104,32 +1104,34 @@ def quinn_fernandes_algorithm(
...
+    x = np.asarray(x)
+    signal = np.moveaxis(np.asarray(signal), axis, -1)
 
-    if not isinstance(signal, np.ndarray):
-        signal = np.array(signal)
+    fs = 1 / abs(x[0] - x[1])
 
     if signal.ndim == 1:
-        signal = signal[np.newaxis, :]
+        signal = np.asarray(signal[np.newaxis, :])
```
alecandido: What is the purpose of this additional `np.asarray()`? Indexing already returns an array

alecandido: A first question is: why the 1D is special? Why do you have to turn it into a 2D?

However, if there is a reason for that, consider that NumPy has a function which is doing the exact same of these two lines. It is called [`np.atleast_2d()`](https://numpy.org/doc/stable/reference/generated/numpy.atleast_2d.html)

src/qibocal/protocols/utils.py
```diff
@@ -1146,14 +1148,16 @@ def quinn_fernandes_algorithm(
...
+    if omega_est.size == 1:
+        return float(omega_est) * fs
+
+    return omega_est * fs
 
 
 def guess_period(
     x: NDArray, y: NDArray, axis: int = -1, speedup_flag: bool = True
-) -> NDArray:
+) -> NDArray | float:
```
alecandido: Same as above, https://github.com/qiboteam/qibocal/pull/1662/changes#r3781806067

src/qibocal/protocols/utils.py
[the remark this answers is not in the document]
lballerio: addressed in 1e8585f.

src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_processing.py
lballerio: so `concatenated_signal` contains the info about X, Y and Z axes we compute from the hamiltonian tomography process. But we know that period has to be the same for all these axes, that's why we compute the median between the `omega` estimated along each component so we choose one.

src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_processing.py
lballerio: addressed in 1e8585f.

src/qibocal/protocols/utils.py
lballerio: yeah, didn't think about it, I simply removed the `np.moveaxis()` call and left the variable, but you are right.
addressed in 1e8585f.

src/qibocal/protocols/utils.py
lballerio: Yeah then at the top of the file they also explicitly import it, idk why I did like this ahahah. Then no harm...
Addressed in 1e8585f.

src/qibocal/protocols/utils.py
lballerio: you are right, I agree.
Addressed in 1e8585f.

src/qibocal/protocols/utils.py
lballerio: never mind, some leftover for type checking.
Removed in 1e8585f.

src/qibocal/protocols/utils.py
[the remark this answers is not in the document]
lballerio: so I saw the code and actually another point is that `np.clip()` returns a Numpy scalar if the input is a 0-dim array. So I decided to wrap it with `np.asarray()` call.
Same for `np.arccos()` call.

src/qibocal/protocols/utils.py
lballerio: addressed in 1e8585f.

src/qibocal/protocols/utils.py
lballerio: this was to always work with numpy arrays, while calls like `np.sum()`, `np.mean()` etc. for a 1D signal recast the result as a scalar.
However, I changed it and now I wrapped all these functions with `np.asarray()` calls, to keep the variable as a numpy array.
Addressed in 1e8585f.

src/qibocal/protocols/utils.py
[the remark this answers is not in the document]
alecandido: No strong need. At most, if you want to faithfully honor the return type, you may wrap with `np.asarray()` the final product. But wrapping the individual factors is not really needed

src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_processing.py
alecandido: And what was happening before this PR? Why `np.median()` was not needed before?

alecandido (approved):
There are the last couple of discussions open, but they are minor anyhow.

For me it is quite good to go. Thank you very much @lballerio!

RoyStegeman:
Is there a reason this has not been merged?

src/qibocal/protocols/two_qubit_interaction/cross_resonance/cross_resonance_processing.py
lballerio: @alecandido  No even in `main` `np.median()` is called, but it is called inside the `quinn_fernandes` function in `protocols/utils.py`, which indeed is even conceptually wrong.

alecandido:
> Is there a reason this has not been merged?

I believe that, when I approved, @lballerio was on holiday. But it should be possible to rebase and merge

src/qibocal/protocols/utils.py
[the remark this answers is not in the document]
lballerio: addressed in 12a51e5.

PR #1670, opened 2026-08-20
Clip arccos argument at 1.0 to avoid NaNs in flux amplitude frequency

### LINKED ISSUE #1669: flux_amplitude_frequency: detuning values clip to 0 for small absolute detuning
This was done intentionally, but probably this threshold is too conservative. 

https://github.com/qiboteam/qibocal/blob/c4ffb0fb722e94c4ce43bd51fb15ae774c4bd4f7/src/qibocal/protocols/flux_dependence/flux_amplitude_frequency.py#L284-L285

Image

http://login.qrccluster.com:9000/qUzlPqRaRmK2wXfkXt3xzQ==

RoyStegeman:
closes #1669

Replace the arbitrary threshold `det[np.abs(det) < 1e-3] = 0`, with `np.minimum` of the arccoss argument that should be equal to or smaller than 1. This still avoids arccos nans as a result of noise, but is not so aggressive to affect many datasets.

### DIFF
--- a/src/qibocal/protocols/flux_dependence/flux_amplitude_frequency.py
+++ b/src/qibocal/protocols/flux_dependence/flux_amplitude_frequency.py
@@ -281,10 +281,10 @@ def _fit(data: FluxAmplitudeFrequencyData) -> FluxAmplitudeFrequencyResults:
         other_det = data.detuning[qubit]
         f = data.qubit_frequency[qubit]
         det = phase / data.flux_pulse_duration / 2 / np.pi + other_det
-        # to make sure that flux is invertible
-        det[np.abs(det) < 1e-3] = 0
+        # clip to make sure that flux is invertible
+        freq_ratio = np.minimum(((f + det) / f) ** 2, 1.0)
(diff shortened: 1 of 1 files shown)


scarrazza (approved):

alecandido (approved):

alecandido:
@RoyStegeman, if the purpose is just to avoid the `nan`, yours is perfectly fine. And certainly better than the alternative in `main`.

Instead I wonder whether we should try to also change branch, to be able to resolve the full arc. In a sense, we could just detect whether the argument of the `arccos` has a local maximum, and it is somewhat around 1. In that way, we could cut the array in two, and choose two different branches of `np.arccos()` (i.e. + and -) for reconstructing the flux shape.

We do not have a compelling need for it, since usually this protocol is needed just for calibrating the curvature of the parabola, needed for the cryoscope experiment.
However, if we could traverse the turning point, the linear shifts of the parabola could be better determined, since they correspond to translations of the tip itself. And it could even be a more refined measurement of the sweetspot - which can even be used for monitoring its fluctuations.

PR #1688, opened 2026-09-04
review and improvement of readout optimization protocols

lballerio:
As per statement.

1. filenames and protocols names were changed
2.  In `ro_amplitude` protocol, I added the `Sweeper` syntax for performance improvement.
3. Now IQ measurements are saved during acquisition, while the classification is being moved during fitting.

here there are data and plots of the new protocols:
1. Readout Amplitude optimization
http://login.qrccluster.com:9000/yMDVRwzATUOO5O4sa237ww==

2. Readout Frequency Optimization:
http://login.qrccluster.com:9000/U4srY2jNTzWctciUXtzk8g==

3. Readout Amplitude-Frequency Optimization: 
http://login.qrccluster.com:9000/teIvgJuqReeKKWQHIMreJg==

Experiments run on QW Tuna5


alecandido:
Quick comment: also the `resonator_optimization` can be renamed to `readout_optimization`, and, for the files
- either you drop the `_optimization` suffix, since they are in the `readout_optimization/` folder
- or, even better, just keep the current names, and move the files into the `readout/` folder, we do not really need infinite many folders

Target a specific issue. If you are addressing multiple ones, let's discuss (even briefly).

lballerio:
> Quick comment: also the `resonator_optimization` can be renamed to `readout_optimization`, and, for the files
>     * either you drop the `_optimization` suffix, since they are in the `readout_optimization/` folder

yes in the following commits I also changed the name of that protocol, I was just doing step by step, first I just worked on 1D protocols.

>     * or, even better, just keep the current names, and move the files into the `readout/` folder, we do not really need infinite many folders

We can do it later on (eventually even in a new PR), after a fist batch of reviews, so we don't mess with file names and so on, do yo agree?

src/qibocal/protocols/readout/amplitude_frequency_optimization.py
```diff
@@ -0,0 +1,403 @@
...
+        pixel_data: dict[tuple[QubitId, int, int], np.ndarray] = {}
+        for state in [0, 1]:
+            for m in range(3):
+                pixel_data[target, state, m] = results[ro_pulses[target, state, m].id]
+
+        data.data |= elaborate_raw_data(
+            qubit=target,
+            pixel_measurements=pixel_data,
+            ampl_sweep=data.amplitudes_swept,
+            freq_sweep=data.frequencies_swept[target],
```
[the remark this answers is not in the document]
lballerio: this is intended to be like this

src/qibocal/protocols/readout/amplitude_optimization.py
```diff
@@ -0,0 +1,102 @@
...
+def _acquisition(
+    params: ReadoutAmplitudeParameters,
+    platform: CalibrationPlatform,
+    targets: list[QubitId],
+) -> ReadoutData:
+    r"""
+    Data acquisition for resoantor amplitude optmization.
+    This protocol sweeps the readout amplitude performing a classification routine
+    and evaluating the error probability at each step. The sweep will be interrupted
+    if the probability error is less than the `error_threshold`.
```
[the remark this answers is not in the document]
lballerio: addressed in 178267c

src/qibocal/protocols/readout/utils.py
```diff
@@ -0,0 +1,219 @@
...
+        state_1 = np.asarray(results[qubit_ros[1].id])
+        for idx, param in enumerate(parameter_dict[qubit]):
+            sweep_state_0 = state_0[:, idx, :]
+            sweep_state_1 = state_1[:, idx, :]
+
+            if save_iq:
+                data[qubit, 0, param] = sweep_state_0
+                data[qubit, 1, param] = sweep_state_1
+
+            fitted_result = fit_classification_model(sweep_state_0, sweep_state_1)
```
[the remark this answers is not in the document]
lballerio: addressed in 178267c.

src/qibocal/protocols/readout/amplitude_frequency_optimization.py
```diff
@@ -0,0 +1,403 @@
...
+    """Amplitude RangeLike object; for further information, see
+    :class:`qibocal.protocols.utils.RangeLike`."""
+    delay: float = 0
+    """Delay between readouts, could account for resonator depletion or not [ns]."""
+    save_iq: bool = False
+    """Whether to save the IQ data during the acquisition."""
+
+    @property
+    def _amplitude_range(self) -> Range:
+        return to_range(self.amplitude_range)
```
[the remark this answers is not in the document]
lballerio: addressed in 4857e4a

src/qibocal/protocols/readout/__init__.py
```diff
@@ -1,4 +1,13 @@
+from .amplitude_frequency_optimization import ro_amplitude_frequency
+from .amplitude_optimization import ro_amplitude
+from .frequency_optimization import ro_frequency
```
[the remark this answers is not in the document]
lballerio: addressed in bc528f5

src/qibocal/protocols/readout/utils.py
```diff
@@ -0,0 +1,228 @@
...
+    # saving flag `save_iq` is set to False.
+    # In that case, the `data` attribute will be empty and the `qubits`
+    # property will return an empty list.
+    @property
+    def qubits(self) -> list[QubitId]:
+        """Return the list of qubits for which data was acquired."""
+        return list(self.swept_parameter.keys())
+
+
+def readout_sequence(
```
alecandido: Either call it just `sequence()`/`base_sequence()`, relying on the scope, or you need a more specific name. `readout_sequence()` is a bit misleading, because it gives the impression that is only the measurement part, while there is also the $\pi$-pulse

src/qibocal/protocols/readout/amplitude_optimization.py
```diff
@@ -0,0 +1,101 @@
...
+__all__ = ["ro_amplitude"]
+
+
+@dataclass
+class ReadoutAmplitudeParameters(Parameters):
+    """ReadoutAmplitude runcard inputs."""
+
+    amplitude_range: RangeLike
+    """Amplitude RangeLike object; for further information, see
+    :class:`qibocal.protocols.utils.RangeLike`."""
```
alecandido: Just keep the first line short:
```suggestion
    """Amplitude RangeLike object.
    
    For further information, see :class:`qibocal.protocols.utils.RangeLike`.
    """
```

src/qibocal/protocols/readout/amplitude_optimization.py
```diff
@@ -0,0 +1,101 @@
...
+def _acquisition(
+    params: ReadoutAmplitudeParameters,
+    platform: CalibrationPlatform,
+    targets: list[QubitId],
+) -> ReadoutData:
+    r"""
+    Data acquisition for resonator amplitude optimization.
+    This protocol sweeps the readout amplitude performing a classification routine
+    and evaluating the assignment fidelity at each step.
+    """
```
alecandido:
```suggestion
    """Data acquisition for resonator amplitude optimization.
    
    This protocol sweeps the readout amplitude performing a classification routine
    and evaluating the assignment fidelity at each step.
    """
```

However, this is internal, and we do not necessarily need to document it for completeness, but just if there is anything relevant to point out.
The explanation about the protocol can go to the object at the bottom.

src/qibocal/protocols/readout/amplitude_optimization.py
```diff
@@ -0,0 +1,101 @@
...
+    sequences, probe_pulses_dict = readout_sequence(platform, targets)
+
+    sweepers: dict[QubitId, Sweeper] = {}
+    amplitude_values: dict[QubitId, list[float]] = {}
+    for qubit in targets:
+        sweepers[qubit] = Sweeper(
+            parameter=Parameter.amplitude,
+            range=to_range(
+                spec=params.amplitude_range,
+                center=probe_pulses_dict[qubit][0].probe.amplitude,
```
alecandido: Maybe I would take it explicitly from the native, instead of doing the round-trip through the `readout_sequence()` function, which is increasing a bit the complexity. Let's cut the link, and go to the source.

src/qibocal/protocols/readout/amplitude_optimization.py
```diff
@@ -0,0 +1,101 @@
...
+            ),
+            pulses=list(probe_pulses_dict[qubit].values()),
+        )
+        amplitude_values[qubit] = sweepers[qubit].values.astype(float).tolist()
```
alecandido: Isn't `.tolist()` already doing the full job? Or, maybe a step back, why are you casting explicitly?

```suggestion
        amplitude_values[qubit] = sweepers[qubit].values.tolist()
```

src/qibocal/protocols/readout/amplitude_optimization.py
```diff
@@ -0,0 +1,101 @@
...
+
+    results = platform.execute(
+        sequences,
+        [list(sweepers.values())],
```
alecandido: You actually do need a full dict for sweepers. Unless I'm missing something.

```suggestion
    sweepers: list[Sweeper] = {}
    amplitude_values: dict[QubitId, list[float]] = {}
    for qubit in targets:
        sweeper = Sweeper(
            parameter=Parameter.amplitude,
            range=to_range(
                spec=params.amplitude_range,
                center=probe_pulses_dict[qubit][0].probe.amplitude,
            ),
            pulses=list(probe_pulses_dict[qubit].values()),
        )
        amplitude_values[qubit] = sweeper.values.astype(float).tolist()
        sweepers.append(sweeper)

    results = platform.execute(
        sequences,
        [sweepers],
```

src/qibocal/protocols/readout/utils.py
```diff
@@ -0,0 +1,228 @@
...
+    save_iq: bool,
+) -> tuple[
+    dict[tuple[QubitId, int, float], npt.NDArray[np.float64]],
+    dict[QubitId, list[list[float]]],
+]:
+    """Extract and optionally classify readout data for each parameter value.
+
+    It returns a mapping keyed by ``(qubit, state, parameter)`` containing either raw
+    IQ samples or classification metrics.
+    """
```
alecandido: The name of this function is quite wrong, the docstring sounds correct. Pass the docstring to AI, and get a new name :P

*To be explicit: what I would expect from a function named `save_data()` is to dump data to disk. What it is doing is mostly collecting them and conditionally running a fit. There is clearly a mismatch.*

src/qibocal/protocols/readout/amplitude_optimization.py
```diff
@@ -0,0 +1,101 @@
...
+        swept_parameter=amplitude_values,
+        save_iq=params.save_iq,
+    )
+    data.data, data.classification_info = save_data(
+        targets=targets,
+        parameter_dict=amplitude_values,
+        pulses_dict=probe_pulses_dict,
+        results=results,
+        save_iq=params.save_iq,
+    )
```
alecandido: At this point, I would recommend reverting the workflow: return an object of type `ReadoutData` from the `save_data()` function (or whatever it will be called, after renaming) and modify the `swept_parameter` attribute of the returned object.

All the other attributes (`.data`, `.classification_info`, `.save_iq`) are known to the `save_data()` function. And returning a single named object (as `ReadoutData`) makes it simpler than anonymous tuples.

src/qibocal/protocols/readout/amplitude_frequency_optimization.py
```diff
@@ -0,0 +1,406 @@
...
+    table_html,
+    to_range,
+)
+
+from .utils import fit_classification_model
+
+__all__ = ["ro_amplitude_frequency"]
+
+
+def elaborate_raw_data(
```
alecandido, line 45: It should be possible to reuse here the same function of the 1D scan, since the inputs are homogeneous (single-shot points) and the outputs as well (assignment fidelity, QND, QNDpi).

However, this becomes especially natural once everything is vectorized, as described in #1693. Just add a reminder in the OP there, to migrate and deduplicate also this function.

src/qibocal/protocols/readout/amplitude_frequency_optimization.py
```diff
@@ -0,0 +1,406 @@
...
+
+
+@dataclass
+class ReadoutAmplitudeFrequencyData(Data):
+    """Data class for readout optimization protocol."""
+
+    frequencies_swept: dict[QubitId, list[float]] = field(default_factory=dict)
+    """Frequency swept for each qubit."""
+    amplitudes_swept: dict[QubitId, list[float]] = field(default_factory=dict)
+    """Amplitude swept for each qubit."""
```
alecandido: Do we need empty defaults here?

src/qibocal/protocols/readout/amplitude_frequency_optimization.py
```diff
@@ -0,0 +1,406 @@
...
+def _acquisition(
+    params: ReadoutAmplitudeFrequencyParameters,
+    platform: CalibrationPlatform,
+    targets: list[QubitId],
+) -> ReadoutAmplitudeFrequencyData:
+    """Protocol to optimize readout frequency and readout amplitude.
+
+    After preparing either state 0 or state 1 we perform two consecutive measurements to
+    evaluate QND. Additionally we apply a pi pulse and we perform a third measurement to
+    evaluate the QND-pi following https://arxiv.org/pdf/2110.04285"""
```
alecandido: Same of the other protocol docstring.

src/qibocal/protocols/readout/frequency_optimization.py
alecandido, line 1: Just a reminder to propagate accepted comments from the other 1D protocol.

src/qibocal/protocols/readout/utils.py
```diff
@@ -0,0 +1,228 @@
...
+    measured_fidelities: dict[QubitId, list]
+    """Measured assignment fidelities."""
+
+
+@dataclass
+class ReadoutData(Data):
+    """Optimization RO frequency acquisition outputs."""
+
+    swept_parameter: dict[QubitId, list[float]] = field(default_factory=dict)
+    """List of parameter swept for each qubit."""
```
alecandido:
```suggestion
    swept_parameter: dict[QubitId, list[float]] = field(default_factory=dict)
    """List of parameter values swept for each qubit."""
```

src/qibocal/protocols/readout/utils.py
```diff
@@ -0,0 +1,228 @@
...
+        sequence_1 += PulseSequence([(qd_channel, qd_pulse)]) | PulseSequence(
+            [(ro_channel, ro_pulse_1)]
+        )
+
+        probe_pulses_dict[qubit] = {
+            0: ro_pulse_0,
+            1: ro_pulse_1,
+        }
+
+    return [sequence_0, sequence_1], probe_pulses_dict
```
alecandido:
```suggestion
    readouts: dict[QubitId, dict[int, Readout]] = {}
    for qubit in targets:
        natives = platform.natives.single_qubit[qubit]

        # measuring the ground state
        sequence_0 += natives.MZ()

        # preparing and measuring the excited state
        sequence_1 += natives.RX() | natives.MZ()

        readouts[qubit] = {
            0: sequence_0[-1][1],
            1: sequence_1[-1][1],
        }

    return [sequence_0, sequence_1], readouts
```

Notes
1. The `probe_pulses_dict` contains the IDs that will be later associated with the results. The rationale of these IDs is that they are associated with input operations, as the `Acquisition`s. `Readout` are also input operations, as they contain an acquisition and *also* a probe pulse, which is the **output** part; so, the name to the dictionary was quite wrong. If anything, the part directly relevant for the results is the other one (acquisition, not probe). But, more correctly, the result is associated to the whole `Readout` operation, in this case.
2. No need to first extract the channels and pulses to then recombine them. The (implicit) assumption here is that the `MZ` is a sequence made by a single `Readout`. With the same implicit assumption, it can be made significantly simpler.
3. Using the `__call__` method of the natives (e.g. `MZ()`) is already calling `.new()` internally. No need to make it explicit.

alecandido (commented):
The PR is quite fine, but I would make a final round of refinements before merging.

---

@lballerio @RoyStegeman this review is consciously coming after the AI one. However, some of the points are things that AI could check for us, e.g. the consistency of functions' names with their docstrings and implementation, or not-so-relevant defaults, because the fields are actually treated as mandatory everywhere they are used.

If we could start making a selection of the types of comments that we agree about, we could start writing an `AGENTS.md` (or the most suitable file) for the purpose, requesting AI to dedicate further attention to these topics.

I would start with a per-repo file, at the cost of duplication. If we consistently do this, we can consider deduplicating some of them per-organization, or some similar strategy.

src/qibocal/protocols/readout/amplitude_frequency_optimization.py
lballerio: Yes I was thinking the same adfter I wrote #1693, but indeed, I think this might be much more feasible once we vectorize the classification fit, so idk whether to tackle this point now or later in another PR.

src/qibocal/protocols/readout/amplitude_frequency_optimization.py
lballerio: Yes if we save the swept freqs and ampls values the way we currently do (if you have other ideas tell me)

src/qibocal/protocols/readout/utils.py
lballerio: ok thanks for the changing, addressed in 5a58609

src/qibocal/protocols/readout/utils.py
lballerio: addressed in 5a58609

src/qibocal/protocols/readout/frequency_optimization.py
lballerio: yes, trying to propagate even in `frequency_optimization.py` in 5a58609

src/qibocal/protocols/readout/amplitude_frequency_optimization.py
lballerio: addressed in 5a58609

src/qibocal/protocols/readout/amplitude_optimization.py
lballerio: we can actually do the entire initializaiton of the `ReadoutData` object inside the `save_data()` function without further modifying the `swept_parameter` field.

Addressed in 5a58609

src/qibocal/protocols/readout/utils.py
lballerio: actually even the docstring is a bit outdated, modified in 5a58609

src/qibocal/protocols/readout/amplitude_optimization.py
lballerio: yes you are right.
addressed in 5a58609

src/qibocal/protocols/readout/amplitude_optimization.py
lballerio: yes, idk why but at some point I think I saw without casting it was turning it into a list of `np.float64`...
Anyway, addressed in 5a58609

src/qibocal/protocols/readout/amplitude_optimization.py
lballerio: ok perfect, done in 5a58609

src/qibocal/protocols/readout/amplitude_optimization.py
lballerio: ok moved the description at the bottom for all 3 protocols, addressed in 5a58609

src/qibocal/protocols/readout/amplitude_optimization.py
lballerio: addressed in 5a58609

src/qibocal/protocols/readout/utils.py
lballerio: ok, addressed in 5a58609

alecandido:
For those comments that you just addressed according to the content of my message, just resolve the conversation.

If I have to go through them again, it is taking just more time. Instead, I would rather focus on those on which the reply needs a further discussion.

src/qibocal/protocols/readout/amplitude_frequency_optimization.py
alecandido: What I mean is that, if the swept values are mandatory, can't we just make the fields mandatory as well?

Instead of doing:
```py
data = ReadoutAmplitudeFrequencyData()
data.frequencies_swept[qubit] = ...
```
we could do:
```py
freqs = {}
freqs[qubit] = ...
data = ReadoutAmplitudeFrequencyData(frequencies_swept=freqs, ...)
```

alecandido (commented):
This PR is already cleaning up significantly the readout optimization protocols. Better to focus on merging, rather than improving it any further.

Even the current pending comment is optional: it was already like this in the existing protocol, better not to postpone this PR much further.

In general, before merging, I would just pass one last time AI, just for a last sanity check. However, in this case I would skip, just because the diff is massive (since the files have been moved) but it is not faithful of the changes in this PR. And we already know that we want to postpone part of the changes.

If it's working, let's merge.

alecandido (approved):

alecandido (approved):

alecandido (approved):

alecandido:
Sorry for the extra noise: I had some connection problem with GitHub 🤦

src/qibocal/protocols/readout/amplitude_frequency_optimization.py
lballerio: no no it was clear

scarrazza (approved):

PR #1701, opened 2026-09-14
fix: crosstalk report failure

ElStabilini:
Recently I was running a crosstalk experiment with one qubit as a target and three qubits that I was biasing.
The acquisition was completed correctly but the report was not generated with the following error message:

```
  File "/Users/stabilini/Desktop/CQC/qibolab_platform_cqc/sandbox/bin/qq", line 10, in <module>
    sys.exit(command())
             ~~~~~~~^^
  File "/Users/stabilini/Desktop/CQC/qibolab_platform_cqc/sandbox/lib/python3.13/site-packages/click/core.py", line 1569, in __call__
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/stabilini/Desktop/CQC/qibolab_platform_cqc/sandbox/lib/python3.13/site-packages/click/core.py", line 1490, in main
    rv = self.invoke(ctx)
  File "/Users/stabilini/Desktop/CQC/qibolab_platform_cqc/sandbox/lib/python3.13/site-packages/click/core.py", line 1970, in invoke
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/Users/stabilini/Desktop/CQC/qibolab_platform_cqc/sandbox/lib/python3.13/site-packages/click/core.py", line 1353, in invoke
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/stabilini/Desktop/CQC/qibolab_platform_cqc/sandbox/lib/python3.13/site-packages/click/core.py", line 907, in invoke
    return callback(*args, **kwargs)
  File "/Users/stabilini/Desktop/CQC/qibolab_platform_cqc/sandbox/lib/python3.13/site-packages/qibocal/cli/_base.py", line 78, in run
    protocols_execution(runcard, folder, force, update)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/stabilini/Desktop/CQC/qibolab_platform_cqc/sandbox/lib/python3.13/site-packages/qibocal/cli/run.py", line 62, in protocols_execution
    report(path, history)
    ~~~~~~^^^^^^^^^^^^^^^
  File "/Users/stabilini/Desktop/CQC/qibolab_platform_cqc/sandbox/lib/python3.13/site-packages/qibocal/cli/report.py", line 82, in report
    html = template.render(
        is_static=True,
    ...<10 lines>...
        ),
    )
  File "/Users/stabilini/Desktop/CQC/qibolab_platform_cqc/sandbox/lib/python3.13/site-packages/jinja2/environment.py", line 1295, in render
    self.environment.handle_exception()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/stabilini/Desktop/CQC/qibolab_platform_cqc/sandbox/lib/python3.13/site-packages/jinja2/environment.py", line 942, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "/Users/stabilini/Desktop/CQC/qibolab_platform_cqc/sandbox/lib/python3.13/site-packages/qibocal/web/templates/template.html", line 169, in top-level template code
    {% set figures, fitting_report = report.plotter(report.history[task_id], qubit) %}
    ^^^^^^^^^^^^^
  File "/Users/stabilini/Desktop/CQC/qibolab_platform_cqc/sandbox/lib/python3.13/site-packages/qibocal/cli/report.py", line 53, in plotter
    figures, fitting_report = generate_figures_and_report(node, target)
                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/Users/stabilini/Desktop/CQC/qibolab_platform_cqc/sandbox/lib/python3.13/site-packages/qibocal/cli/report.py", line 42, in generate_figures_and_report
    return node.task.operation.report(data=node.data, fit=node.results, target=target)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/stabilini/Desktop/CQC/qibolab_platform_cqc/sandbox/lib/python3.13/site-packages/qibocal/protocols/flux_dependence/qubit_crosstalk.py", line 305, in _plot
    figures, fitting_report = utils.flux_crosstalk_plot(
                              ~~~~~~~~~~~~~~~~~~~~~~~~~^
        data, target, fit, fit_function=utils.transmon_frequency
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/stabilini/Desktop/CQC/qibolab_platform_cqc/sandbox/lib/python3.13/site-packages/qibocal/protocols/flux_dependence/utils.py", line 232, in flux_crosstalk_plot
    xj=qubit_data.bias, **fit.fitted_parameters[flux_qubit]
                          ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
KeyError: (0, 1)
```

After a few trials I noticed that the failure happens only for the case of runcard where i'm flux biasing multiple qubits and the fit fails for some of them (but not all of them).
I should have fixed it with the single change implemented in this PR.

I tested it by running `qq report` with the folder containing the data that generated the previous error message after having implemented the single change I added here.


ElStabilini:
I took the chance to try to fix also another bug that I was seeing. When trying to display the fit for the dispersive shift I had the following plot on the report:

image

where if I remove the fit I obtain

image

I would appreciate a double check, but my understanding was that the wrong x values were fed to the `lorentzian_with_linear_background()` function. 

This is an example of the current plot I can get:

image

lballerio:
Hi @ElStabilini, thanks a lot for your contributions; actually you anticipated us in the dispersive shift protocol because we also noticed this problem and was in our to-do-list.
Also for the crosstalk experiment has to be review and validated, so thanks again to spotting and fixing this issue.
I wanted also to communicate you that this week all the Middleware team is out of office for the IEEE Quantum Week, so we probably won't be able to follow up this week.
Thanks a lot!

ElStabilini:
Thanks, @lballerio no problem. I opened the PR simply because I am working on a chip calibration at the moment, I will probably continue to test around and eventually make small changes. Enjoy the conference!

ElStabilini:
Same thing for qutrit dispersive shift

before:
image

after:
image


small note: there's quite some redundant code in the plotting routines for dispersive shifts and resonator spectroscopy maybe in the future would be nice to unify the functions as much as possible. If I have time I'll do that.

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
```diff
@@ -230,11 +230,11 @@ def _plot(data: DispersiveShiftData, target: QubitId, fit: DispersiveShiftResult
         fit_group = f"state{state}-fit"
         q_data = data[target, state]
 
-        frequencies = q_data.freq * scipy.constants.nano
+        frequencies = q_data.freq
         for col, y in enumerate([q_data.signal, q_data.phase], start=1):
             fig.add_trace(
                 go.Scatter(
-                    x=frequencies,
+                    x=frequencies * scipy.constants.nano,
```
lballerio, line 237: I think you change this because the `fit_function` the frequency must be in Hz...however I don't quite like this multiple conversion, isn't possible to do once per all?
Have you also tried to rescale in GHz even for `fit`?

Ps this is a relatively minor problem.

src/qibocal/protocols/dispersive_shift/dispersive_shift_qutrit.py
```diff
@@ -212,10 +212,10 @@ def _plot(
         )
     ):
         opacity = 1
-        frequencies = q_data.freq * HZ_TO_GHZ
+        frequencies = q_data.freq
         fig.add_trace(
             go.Scatter(
-                x=frequencies,
+                x=frequencies * scipy.constants.nano,
```
lballerio, line 218: Same as previous comment.

src/qibocal/protocols/flux_dependence/qubit_crosstalk.py
lballerio, line 322: Do we want to print in the report even unmodified crosstalk terms?

lballerio:
Hi @ElStabilini, I checked your PR, for me is mostly fine, some small comments!

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
ElStabilini: I think that the previous code version was trying to d that by converting all the frequencies (line 223: `frequencies = q_data.freq * scipy.constants.nano`). The problem was that then the plot had `lorentzian_with_linear_background(freqrange, *params)` with `freqrange` in GHz but with parameters in Hz, this caused the curve to "collapse" to the offset value.

What I tried to do do was a minimal fix based on what is already done in the resonator spectroscpy.
Probably I would touch the `fit` only in a second moment, eventually while also reducing the code redundancy between `dispersive_shift`, `dispersive_shift_qutrits` and `resonator_utils`.

I would be happy to work on that but right now to me seems like a secondary problem that might be targeted in the future.

src/qibocal/protocols/flux_dependence/qubit_crosstalk.py
ElStabilini: This was not intended actually, I'll double check and fix it, thank you.

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
RoyStegeman: I think the way currently done by @ElStabilini is preferred. 

To give some context: one of the issues with qibocal at the moment is that units are mixed, and from the name of the variables it's often not clear what the unit of a physical parameter is. This is particularly true for frequencies, where a variable named `freq` may be either Hz or GHz and the distinction can only be made by tracing back far enough. To avoid this confusion, we should use Hz everywhere such that we can trust `freq` to be in Herz, unless very clearly specified otherwise in exceptional cases. 

That doesn't mean we cannot have GHz anywhere, but if the unit deviates from Hz it should be explicit (docstring, variable name, ...), and ideally the translation should be done as late as possible. In this case, I would say doing the transformation in the function call is a good way to reduce the chance of confusion/mistakes due to unit changes, although arguably assigning  a `frequencies_ghz` is also fine.

While we need this transformation to plot in GHz, personally, I would be in favor of plotting in Hz as well and just having something like `+9` on the x-axis. But since it was done this way probably not everyone agrees with it, and I won't delay this PR to have that discussion here.

src/qibocal/protocols/flux_dependence/qubit_crosstalk.py
ElStabilini: I doubled check this and I do not see why unmodified crosstalk terms should get printed. 
I also run a small check using just noise and the report (in case of fit failures for all the `flux_qubits`) will look like this:
image

src/qibocal/protocols/dispersive_shift/dispersive_shift.py
alecandido: More something like `*1e9`, but I perfectly agree with @RoyStegeman

PR #1704, opened 2026-09-16
Additional ef transition protocols

ElStabilini:
Example of `ef_length` output:
image

Example of `ramsey_ef` output:
image


alecandido:
@ElStabilini before you push this too far, I would strongly advise to make small PRs about very specific features.

E.g. the present content in 59578ca should be split in at least 3 different PRs:
- one about introducing `ef_length.py`
- one for `ramsey_ef.py`
- and a further PR for amending the others

Long PRs take infinitely long to review, since we are few people and they may never end to be high on our priority list.
Moreover, if they age it becomes more and more challenging to merge, and the overhead in keeping them updated spikes.

For as long as it is draft there is no problem ever. But make your considerations about when it is best to split them apart, because eventually they are better submitted as separate ones.

ElStabilini:
Hi @alecandido, thank you for the comment and for pointing this out. 

Regarding `ef_length` and  `ramsey_ef` I actually thought about splitting them but then I thought that se the protocols are not completely new maybe opening too many PRs was not a good idea.

Regarding the other amends, separated from these two protocols, they are indeed the content of a different PR ( #1701), for this reason I kept this PR as a draft for now. 

On a side note, and as a small explanation of how I ended up in this situation :sweat_smile:
Ideally I would have liked to open this PR using as base the branch `ElStabilini:corsstalk_fit` (the branch corresponding to  #1701)  but I've not been able to do so. 
The problem is that in my fork of the I need to have a branch with all the changes I am implementing, so that I can give the collaborators an environment that is easy to configure ,and where they can replicate all the experiments I make/they requested. 
If you have any suggestion on how to better manage these two aspects at the same time I would be happy to follow your instructions :grin:

EDIT: I think I have a solution for my problem, I will implement in the next PRs

